PR #3, opened 2016-02-15
Add cxx feature cxx_noexcept

hean01:
Fixes build with gcc

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -311,6 +311,8 @@ add_executable(solvespace WIN32 MACOSX_BUNDLE
     ${solvespace_HEADERS}
     ${solvespace_SOURCES})
 
+target_compile_features(solvespace PRIVATE cxx_noexcept)
+
 target_link_libraries(solvespace
     "${OPENGL_LIBRARIES}"
     "${PNG_LIBRARIES}"


whitequark:
CI builds with gcc-4.6 succeed. What gcc version and what compilation problem do you have?

hean01:
This is on FC23, gcc (GCC) 5.3.1 20151207 (Red Hat 5.3.1-2)

The errors are hundreds of lines but see the top of output down below...

Cmake will add -std=gnu++11, when defining compiler features cxx_noexcept, which is the correct way of doing this.

```
[ 15%] Built target slvs
[ 17%] Building CXX object src/CMakeFiles/solvespace.dir/gtk/gtkmain.cpp.o
In file included from /usr/include/c++/5.3.1/type_traits:35:0,
                 from /usr/include/sigc++-2.0/sigc++/visit_each.h:22,
                 from /usr/include/sigc++-2.0/sigc++/functors/slot.h:6,
                 from /usr/include/sigc++-2.0/sigc++/signal_base.h:27,
                 from /usr/include/sigc++-2.0/sigc++/signal.h:8,
                 from /usr/include/sigc++-2.0/sigc++/sigc++.h:86,
                 from /usr/include/glibmm-2.4/glibmm/iochannel.h:30,
                 from /usr/include/glibmm-2.4/glibmm/main.h:26,
                 from /home/henrik/Development/upstream/solvespace.git/src/gtk/gtkmain.cpp:18:
/usr/include/c++/5.3.1/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
 #error This file requires compiler and library support for the \
  ^
In file included from /usr/include/glibmm-2.4/glibmm/main.h:23:0,
                 from /home/henrik/Development/upstream/solvespace.git/src/gtk/gtkmain.cpp:18:
/usr/include/glibmm-2.4/glibmm/refptr.h:83:70: error: ‘nullptr’ was not declared in this scope
     static const bool value = sizeof(check(static_cast<T_CppObject*>(nullptr))) == sizeof(big);
                                                                      ^
/usr/include/glibmm-2.4/glibmm/refptr.h:86:34: error: ‘Glib::RefPtr::is_allowed_type::value’ is not a type
   static_assert(is_allowed_type::value,
                                  ^
/usr/include/glibmm-2.4/glibmm/refptr.h:87:5: error: expected identifier before string constant
     "Glib::RefPtr must not be used with this class.");
     ^
/usr/include/glibmm-2.4/glibmm/refptr.h:87:5: error: expected ‘,’ or ‘...’ before string constant
/usr/include/glibmm-2.4/glibmm/refptr.h:87:53: error: ISO C++ forbids declaration of ‘static_assert’ with no type [-fpermissive]
     "Glib::RefPtr must not be used with this class.");
                                                     ^
/usr/include/glibmm-2.4/glibmm/refptr.h:95:17: error: expected ‘;’ at end of member declaration
   inline RefPtr() noexcept;
                 ^
/usr/include/glibmm-2.4/glibmm/refptr.h:95:19: error: ‘noexcept’ does not name a type
   inline RefPtr() noexcept;

```

whitequark:
Oh yes, I see, glibmm upgraded to requiring C++11 instead of C++0x.

> Cmake will add -std=gnu++11, when defining compiler features cxx_noexcept, which is the correct way of doing this.

It's not. If I want C++11, I should request C++11, not some random feature that carries no semantic meaning as to why it's requested. It's failing on `static_assert`, not `noexcept`, isn't it? CMake's solution is stupid.

Ok, I will look into bumping the requirement. SolveSpace on non-Linux systems already carried the C++11 requirement for a while so it should be OK; mostly it's C++0x just for Travis CI, which has an old gcc for some reason.

hean01:
> It's not. If I want C++11, I should request C++11, not some random feature that carries no semantic meaning as to why it's requested. It's failing on static_assert, not noexcept, isn't it? CMake's solution is stupid.

Ah, that is true. I got confused and cmake is cmake :). Bumping up the requirement for SolveSpace is the proper way of doing it.

whitequark:
Oh no, you aren't confused, this is indeed what the CMake manual recommends. But it's still stupid.

whitequark:
I've fixed this and updated the for-upstream branch.

Symbian9:
February 2016:
> CI builds with gcc-4.6 succeed. What gcc version and what compilation problem do you have?


https://github.com/solvespace/solvespace/blob/master/CMakeLists.txt
> GCC 4.8/4.9 ship with broken but present <regex>. meh.

@whitequark, @Evil-Spirit, is it a way keep support for GCC 4.8.2 and Trusty (14.04)?

Or how run GCC 5.0 on Trusty if its not officially support?
* https://gist.github.com/beci/2a2091f282042ed20cda

PR #4, opened 2016-05-16
Robust UTF8 conversion

r8zr:
Old UTF8 conversion works wrong with degree symbol that is showing while using angle constraint. 
Calculating of bytes offset was wrong and then we stuck in loop trying to read some thrash in memory.

### DIFF
--- a/src/glhelper.cpp
+++ b/src/glhelper.cpp
@@ -48,11 +48,11 @@ double ssglStrFontSize(double h)
 }
 double ssglStrWidth(const std::string &str, double h)
 {
-    int width = 0;
-    for(char32_t chr : ReadUTF8(str)) {
-        const VectorGlyph &glyph = GetVectorGlyph(chr);
+	auto width = 0;
+    for(auto chr : ReadUTF8(str)) {
+        const auto& glyph = GetVectorGlyph(chr);
         if(glyph.baseCharacter != 0) {
-            const VectorGlyph &baseGlyph = GetVectorGlyph(glyph.baseCharacter);
+            const auto& baseGlyph = GetVectorGlyph(glyph.baseCharacter);
             width += max(glyph.advanceWidth, baseGlyph.advanceWidth);
         } else {
             width += glyph.advanceWidth;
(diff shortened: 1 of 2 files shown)


whitequark:
What exactly was wrong with it? The old code I took from glib (`g_unichar_to_utf`), I doubt there's a bug in it.

r8zr:
i've tried to add "Constraint angle" into my sketch, and after clicking on the button i've got next error (see attachment). 
untitled

I've tried to debug it, and found that a problem is actually with a degree symbol (°) that is using in label of "Constraint angle". Then in char32_t utf8_iterator::operator*() i've found that this line
this->n = (const char*) (it + 1);
return thrash instead of end of line, so we will stuck in a loop in double ssglStrWidth(const std::string &str, double h) method. So looks like calculating of offset to the next character is working wrong here.
My converting routine should work perfect with all UTF8 characters. The problem could be only with JIS encoding, but as i understand SolveSpace doesn't support it at all, so working with JIS we will have a problem in any case.

whitequark:
Well, the existing routine should work perfect with all UTF-8 characters as well. Let me try to reproduce this.

whitequark:
Ok, so I've just created such a constraint using the e129755d build from Appveyor on Windows XP SP3:


What configuration are you using exactly?

r8zr:
Intel Core 2 Quad Q6600, Windows 7 Ultimate SP1 x86, Visual Studio Ultimate 2013 Version 12.0.40629.00 Update 5

whitequark:
Ok, I'll install that. Did you build it in the default configuration, or did you change some CMake parameters?

r8zr:
Default
untitled
untitled

whitequark:
I just rebuilt using your exact VS version and OS. I cannot reproduce your crash. Does the binary from https://ci.appveyor.com/project/whitequark/solvespace/build/2.1.7/artifacts crash as well?

r8zr:
Sorry for delay here.. So, looks like problem is gone after cloning a repo in new location and rebuilding. Still don't know why it was crashing. Sorry for false alarm and wasting your time!

whitequark:
I suspect the UTF-8 in the source got broken somehow; the UTF-8 iterator is optimized for speed, not robustness, and it's all but supposed to crash that way.

This isn't really important because all input data (except that read from savefiles) is already sanitized at boundary, and savefiles will be sanitized soon since I have to rewrite the serialization code anyway.

PR #7, opened 2016-06-12
Mac OS X: Fix Window vanishing if losing focus

xythobuz:
I don't know if this is the intended behavior for other platforms, but on Mac OS X, all SolveSpace windows completely disappear if the app loses focus, for example by clicking on another program. It is then no longer visible in the Mac 'Mission Control' that should show all opened windows. After clicking on the dock icon, the window reappears.

When setting this flag on the browser window, both windows stay visible when losing focus. This resembles the expected behavior on this platform.

### DIFF
--- a/src/cocoa/cocoamain.mm
+++ b/src/cocoa/cocoamain.mm
@@ -1025,6 +1025,7 @@ void InitTextWindow() {
     [TW setFrameAutosaveName:@"TextWindow"];
     [TW setFloatingPanel:YES];
     [TW setBecomesKeyOnlyIfNeeded:YES];
+    [TW setHidesOnDeactivate:NO];
     [GW addChildWindow:TW ordered:NSWindowAbove];
 
     NSScrollView *scrollView = [[NSScrollView alloc] init];


whitequark:
So, this is interesting. As far as I'm aware (OS X is not my primary platform), OS X's builtin behavior is to hide all *tool* windows when the application loses focus, which happens on my OS X install (a bit dated, I think) and this seems to be consistent with what all other applications do (e.g. invoke a file open dialog in anything else).

What your patch does, as far as I see, is cancels this built-in behavior for the (tool) property browser window. Setting aside the question of whether or not it should be hidden (seems like it should to me, but I have no strong opinion here), this seems rather odd, as this doesn't do anything to the graphics window.

What's your OS X version?

xythobuz:
Yeah, it took me quite some time to find this because I did not expect the tool window flags to have any effect on the main window.

Just to be clear: as you say, I'd expect the tool window to vanish and the main window to stay, but on my system, both vanish without my patch!

I'm still on Yosemite 10.10.5.

whitequark:
Can you try `[GW setHidesOnDeactivate:NO];` instead?

xythobuz:
Oh, I should have mentioned, I tried that first, it has no effect!

whitequark:
Ok, I can confirm this behavior on my OS X install as well.

whitequark:
Fixed in both 2.x and master.

PR #10, opened 2016-06-19
Mac Spacemouse support

xythobuz:
After some discussion with @whitequark I think I'm more or less finished implementing 3DconnexionClient Spacemouse support on Mac OS X.

Instead of linking to the Framework, I'm now loading it completely dynamically using CFBundle. This way, users of the official 3DconnexionClient Framework and users of my open re-implementation can both benefit from this.

Also, I wasn't sure what to do with the 3DconnexionClient Headers. I didn't want to just use them, because that would mean every developer would have to have a version of the 3DconnexionClient installed on his machine. Instead, I've simply copy-pasted the (very small) relevant parts into the solvespace source code. I'm not sure if that's the cleanest solution, but it works.


src/platform/cocoamain.mm
```diff
@@ -1218,6 +1218,227 @@ - (IBAction)preferences:(id)sender {
     [NSApp stop:nil];
 }
 
+#ifdef HAVE_SPACEWARE
```
whitequark: Um. First you set SPACEWARE_FOUND to TRUE, then work around it in the build system to fix the linked files, then add this ifdef, which is always included. Why not just drop the ceremony entirely?

src/platform/cocoamain.mm
```diff
@@ -1218,6 +1218,227 @@ - (IBAction)preferences:(id)sender {
...
+- (UInt16) registerConnexionClient:(UInt32)signature Name:(UInt8 *)name
+                              Mode:(UInt16)mode Mask:(UInt32)mask;
+
+- (void) unregisterConnexionClient:(UInt16)client;
+@end
+
+@implementation FrameworkWrapper
+- (id)init {
+    self = [super init];
+    NSString *bundlePath = @"/Library/Frameworks/3DconnexionClient.framework";
```
whitequark: Is hardcoding this the right thing to do on OS X? Seems like it'd have a lookup function or something.

src/platform/cocoamain.mm
```diff
@@ -1218,6 +1218,227 @@ - (IBAction)preferences:(id)sender {
...
+typedef void (*ConnexionAddedHandlerProc)(io_connect_t);
+typedef void (*ConnexionRemovedHandlerProc)(io_connect_t);
+typedef void (*ConnexionMessageHandlerProc)(io_connect_t, natural_t, void *);
+
+typedef OSErr (*InstallConnexionHandlersProc)(ConnexionMessageHandlerProc, ConnexionAddedHandlerProc, ConnexionRemovedHandlerProc);
+typedef void (*CleanupConnexionHandlersProc)(void);
+typedef UInt16 (*RegisterConnexionClientProc)(UInt32, UInt8 *, UInt16, UInt32);
+typedef void (*UnregisterConnexionClientProc)(UInt16);
+
+@interface FrameworkWrapper : NSObject {
```
whitequark: Why bother with the Objective-C wrapper if the only way it's ever used is from some static C functions?

whitequark:
Seems alright overall, just lots of redundant ceremony. I'm OK with pulling in definitions from headers, doubt they'll ever change.

src/platform/cocoamain.mm
xythobuz: Oh, now that you mention it, this is no longer needed. Whoopsie, I'll remove it...

But I'm not linking to anything anymore.

src/platform/cocoamain.mm
xythobuz: Hmm, true, I've got the idea of using this wrapper object from the Apple documentation, but it's not really needed here. I'll remove it.

src/platform/cocoamain.mm
whitequark: Yes, exactly, you have an if in CMakeLists to *not* link to `${SPACEWARE_LIBRARIES}` on Apple platforms.

xythobuz: Yes, because it no longer should be linked to when using CFBundle, as I'm loading it dynamically...?

whitequark: I mean, if you do not set `SPACEWARE_FOUND` to true then that's not needed. (And the library is indeed not found at compile-time.)

xythobuz: Oh, yeah, now I get it :) Will do, thanks!

src/platform/cocoamain.mm
xythobuz: That is a good question, I've spent some time searching but didn't find anything. I suppose it could be in other locations, like `/System/Library/Frameworks`, but as far as I am aware the official 3DconnexionClient is always installed in `/Library/Frameworks`.

NSBundle can be used to get the paths for frameworks inside of the application bundle, but I can't find anything for the system level.

xythobuz:
How about this? :smile:

src/platform/cocoamain.mm
```diff
@@ -1218,6 +1218,182 @@ - (IBAction)preferences:(id)sender {
...
+    space.data0 = (double)device->axis[0] * -0.25;
+    space.data1 = (double)device->axis[1] * -0.25;
+    space.data2 = (double)device->axis[2] * 0.25;
+    space.data3 = (double)device->axis[3] * -0.0005;
+    space.data4 = (double)device->axis[4] * -0.0005;
+    space.data5 = (double)device->axis[5] * -0.0005;
+    [space performSelectorOnMainThread:@selector(invoke)
+                            withObject:nil
+                         waitUntilDone:NO];
+}
```
whitequark: Ok--one last change. Let's use `dispatch_async(dispatch_get_main_queue(), ^(void){...})` instead of this object used once; `dispatch_async` seems to be the recommended API anyway.

xythobuz:
You're right, using dispatch_async seems cleaner. Done.

whitequark:
Thanks!

PR #19, opened 2016-07-19
Fix resource path resolution

hzeller:
Attempting to execute
  build$ src/solvespace
crashes, because resource_dir.erase(resource_dir.rfind('/')); is executed twice, but the second rfind() will return string::npos.

This only goes back one slash to get the directory and then assembles the path as a relative path.

So both:
  build$ src/solvespace
  solvespace$ build/src/solvespace
will work.


whitequark:
Fixed in master by using `/proc/self/exe` instead.

PR #36, opened 2016-08-24
Switch icon placements on toolbar

Symbian9:
1) under "Extrude" must be "Step translate"
2) under "Lathe" must be "Step rotate"


whitequark:
This change makes no difference except for confusing people who are used to the old layout.

Symbian9:
Its add better logic for UI, where linear operation in one column, rotation in another one.
SolveSpace need UI improvements, and for most number of user it would be no any confuses while SolveSpace 3.0 not released

Issue #105, opened 2016-11-17
Add option for deselecting with left-click


lineprinter:

whitequark:
@jwesthues Was there any particular reason you didn't implement deselection this way originally?

whitequark:
Instead of an option I made it possible to deselect entities with Ctrl-LMB in 6352405, because that's the least intrusive way to let laptop users correct misclicks.

PR #106, opened 2016-11-17
Add option for deselecting with left-click.

lineprinter:
See #105


whitequark:
@lineprinter do you still have code for this somewhere? it looks like you've force pushed over master since

Issue #101, opened 2016-11-16
Solvespace crashes when trying to 'File > Import...' not .dxf or .dwg file


lineprinter:
```
File /home/lineprinter/solvespace/src/solvespace.cpp, line 547, function MenuFile:
Assertion 'Unexpected extension of file to import' failed: ((false) == false).
Backtrace:
 0: solvespace(_ZN10SolveSpace14assert_failureEPKcjS1_S1_S1_+0x78) [0x5626adb1ea2d]
 1: solvespace(_ZN10SolveSpace12SolveSpaceUI8MenuFileENS_7CommandE+0xe0e) [0x5626adb02ee2]
 2: solvespace(_ZN10SolveSpace12MainMenuItemIN3Gtk8MenuItemEE11on_activateEv+0x7d) [0x5626adb19019]
 3: /usr/lib/x86_64-linux-gnu/libgtkmm-2.4.so.1(_ZN3Gtk14MenuItem_Class17activate_callbackEP12_GtkMenuItem+0x50) [0x7f42a2aa3240]
 4: /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_closure_invoke+0x145) [0x7f42a0a64f75]
 5: /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0(+0x2195c) [0x7f42a0a7695c]
 6: /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit_valist+0xe3c) [0x7f42a0a7fbcc]
 7: /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_signal_emit+0x8f) [0x7f42a0a7ffaf]
 8: /usr/lib/x86_64-linux-gnu/libgtk-x11-2.0.so.0(gtk_widget_activate+0x6e) [0x7f42a21bc70e]
 9: /usr/lib/x86_64-linux-gnu/libgtk-x11-2.0.so.0(gtk_menu_shell_activate_item+0xfd) [0x7f42a20b759d]
10: /usr/lib/x86_64-linux-gnu/libgtk-x11-2.0.so.0(+0x143906) [0x7f42a20b7906]
11: /usr/lib/x86_64-linux-gnu/libgtk-x11-2.0.so.0(+0x13184c) [0x7f42a20a584c]
12: /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0(g_closure_invoke+0x145) [0x7f42a0a64f75]
[...]

PR #110, opened 2016-11-17
Fix crash when trying to import not .dxf or .dwg files

lineprinter:
See #101


src/solvespace.cpp
```diff
@@ -544,7 +544,7 @@ void SolveSpaceUI::MenuFile(Command id) {
                 ImportDxf(importFile);
             } else if(Extension(importFile) == "dwg") {
                 ImportDwg(importFile);
-            } else ssassert(false, "Unexpected extension of file to import");
+            } else Error("Unexpected extension of file to import");
```
whitequark: A better message would be the one from export.cpp:836.

PR #133, opened 2016-12-13
add missing build dependency on libspnav (spaceware)

SebKuzminsky:


whitequark:
Fixed in master, with additional clarification.

whitequark:
Ah, and 2.x too.

Issue #134, opened 2016-12-13
regression in master when entering fractions for a Distance/Diameter/Length constraint


SebKuzminsky:
I just switched from the 2.x branch (6802a12a) to master (3dd2fc00), and noticed a regression.  When entering a value for a Distance/Diameter/Length constraint, fractions no longer work right.  In 2.x I could enter this:

    3/4

And the constraint would have the expected value of 0.75.  But in master, entering that fraction gives me a value of 1.33 (4/3).  A couple of other fractions also gave me reciprocal values.

Bisecting shows that the bug first appeared in commit 61f2488.

whitequark:
Fixed in master.

PR #135, opened 2016-12-13
Master fix fractions

SebKuzminsky:
This PR adds some test cases to reproduce the problem in #134, then fixes the problem.

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -787,7 +787,7 @@ bool ExprParser::Reduce(std::string *error) {
         case TokenType::BINARY_OP: {
             Token b = PopOperand(error);
             if(b.IsError()) return false;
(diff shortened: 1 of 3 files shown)


whitequark:
Unfortunately until the CLA situation is sorted out we cannot accept further pull requests. Sorry for the trouble; GitHub doesn't have any good way to indicate this in the UI.

I've pushed the expr fix, and I'll implement a more comprehensive test suite the next time I refactor that part of code (which is soon).

whitequark:
There are a few things I would do differently for testing this, such as implementing a helper function to reduce the amount of noise required for the test, and use the coverage information to make sure 100% of the branches in the parser are covered.

whitequark:
Please see commit d497415, which achieves 100% line coverage and almost 100% branch coverage in ExprParser, and fixes two crash bugs.

PR #168, opened 2017-01-24
don't prevent selection of first point of pending cubic

### LINKED ISSUE #167: Can't make periodic cubic
regression in [55ae10b](https://github.com/solvespace/solvespace/commit/55ae10b5b84ce51042ac9957cfbf01e70d48c952). you're supposed to be able to click again on the first point of a cubic to make a periodic cubic ([HitTestMakeSelection allows this](https://github.com/solvespace/solvespace/blob/55ae10b5b84ce51042ac9957cfbf01e70d48c952/src/draw.cpp#L402)), but [this](https://github.com/solvespace/solvespace/commit/55ae10b5b84ce51042ac9957cfbf01e70d48c952#diff-d1e73a0157de46c51742e677b04e8c0c) prevents selecting the point.

mdejean:
closes #167
but don't just take this-

* the logic is duplicated from HitTestMakeSelection which calls this
* shouldn't you be able to draw a closed loop with the line tool too?

### DIFF
--- a/src/draw.cpp
+++ b/src/draw.cpp
@@ -338,7 +338,15 @@ GraphicsWindow::Selection GraphicsWindow::ChooseFromHoverToSelect() {
         if(hov.selection.entity.v != 0) {
             hEntity he = hov.selection.entity;
             if(he.isFromRequest() && IsFromPending(he.request())) {
-                continue;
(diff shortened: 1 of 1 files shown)


mdejean:
obsoleted by e25a163.

Issue #166, opened 2017-01-23
refresh file from the CLI


wpwrak:
Scenario: part A is assembled from part B, and B in turn is assembled from part C, each in its own .slvs file.
After changing c.slvs, b.slvs has to be updated (loaded and saved) before the changes become visible in A.

Supposedly, resolving all such dependencies properly would require all of https://github.com/solvespace/solvespace/issues/80 i.e., a lot of work.

However, it would seem that, as a work-around, one could just express such dependencies in a (manually created and maintained) Makefile and do the updates manually. The Makefile would ensure that no update is missed. To automate this a bit more, it would be useful to have the ability to load a .slvs file, picking up any changes in parts it directly depends on, and then save the updated version.

whitequark:
Done in master, command "regenerate".

PR #169, opened 2017-01-24
New CLI command "dump" to load and then save a .slvs, to refresh assemblies

wpwrak:
This implements the functionality proposed in
https://github.com/solvespace/solvespace/issues/166

A set of test files can be found in
https://github.com/wpwrak/3dstuff/tree/master/slvs-dump

### DIFF
--- a/src/platform/climain.cpp
+++ b/src/platform/climain.cpp
@@ -46,6 +46,9 @@ static void ShowUsage(const std::string &argv0) {
         being triangulated first.
     export-surfaces --output <pattern>
         Exports exact surfaces of solids in the sketch, if any.
+    dump --output <pattern>
+        Load and immediately write again, updating any copies of parts directly
+        referenced by assemblies.
 )");
 
     auto FormatListFromFileFilter = [](const FileFilter *filter) {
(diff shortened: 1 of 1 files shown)


whitequark:
This doesn't actually work properly because it doesn't call SS.ReloadAllImported. In any case I've implemented it correctly in master.

wpwrak:
Regarding the call to SS.ReloadAllImported, I see that there's one in SolveSpaceUI::SaveToFile. Shouldn't that be enough ?

More importantly, there's a problem with the directory: when I do
solvespace-cli regenerate foo.slvs
it saves to /foo.slvs

As a work-around, prefixing the name with `pwd`/ or ./ produces the desired result.

whitequark:
> More importantly, there's a problem with the directory: when I do
> solvespace-cli regenerate foo.slvs
> it saves to /foo.slvs

Fixed in master.

> Regarding the call to SS.ReloadAllImported, I see that there's one in SolveSpaceUI::SaveToFile. Shouldn't that be enough ?

Hmm, I forgot about that.

PR #185, opened 2017-02-02
vft_app/etc are already defined except in very old compilers

mdejean:
I guess whatever template this was based on was ancient?


whitequark:
Fixed in master, thanks.

PR #200, opened 2017-02-10
if DOF > 0, indicate in group list

wpwrak:
This should help to have a better overview of which groups need some DOF cleanup:
When a group's status would be "ok" (green), but it has > 0 DOF, then show "Dn" (yellow), where "n" is the number of DOF.
If the number is greater that nine, "D+" is shown.

Obligatory screenshot:
show-dof


Evil-Spirit:
@wpwrak, may be it is straightforward to introduce new column for this?

wpwrak:
@Evil-Spirit: like this #204 ? Advantages: easier to guess what it means, code a little nicer. Drawbacks: a bit more clutter on the screen, and it eats up space.

whitequark:
Superceded by #290.

PR #201, opened 2017-02-12
"ok" is not always okay

wpwrak:
A sketch may contain errors that cause later groups to fail, but the group list still says that everything is "ok". Here is an example:

[not-okay.zip](https://github.com/solvespace/solvespace/files/769622/not-okay.zip)

g004-extrude is based on the outline in g003-sk-outline, which in turn uses geometry from g002-sk-bad-influence.

- select g002-sk-bad-influence, change c00a-pt-pt-distance to REF, drag the right point to the left of the left point.
- activate g004-extrude. The extrusion is now a mess, because g003 has become self-intersecting. But the group list still shows a reassuringly green "ok".

With this commit, the "ok" of a group containing a sketch with problems will be shown in yellow:
not-ok-example

Note that this commit touches the same area a my previous DOF indication changed. Since each makes the color selection a little more complicated, I disentangled it here.

Addendum: the "ok" may not turn yellow immediate. E.g., in the above example, it remains green until the group is recalculated, which doesn't happen automatically/immediately in this case.

### DIFF
--- a/src/textscreens.cpp
+++ b/src/textscreens.cpp
@@ -110,6 +110,17 @@ void TextWindow::ShowListOfGroups() {
 	bool no_dof = !g->solved.dof;
 	char dof[3] = { 'D', g->solved.dof > 9 ? '+' :
 	    (char) ('0' + g->solved.dof), 0 };
+	char status_color = 's';	// all is well
+
+	if (!ok) {
+		status_color = 'x';	// "serious" error
+	} else if (!no_dof) {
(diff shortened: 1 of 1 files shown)


whitequark:
@wpwrak Change looks good to me. Can you please rebase on master?

whitequark:
@wpwrak ping

whitequark:
I went ahead and merged this manually, with minor changes.

PR #204, opened 2017-02-15
show DOF in group list

wpwrak:
This is an alternative implementation for #200. It introduces a new column to the group list, showing the DOF. For DOF > 99, it just shows 99.

It looks like this:
alt-show-dof


Evil-Spirit:
@wpwrak, looks better than previous implementation for me.

whitequark:
Superceded by #290.

Issue #184, opened 2017-02-02
problem with text input field on Arch Linux


GerhardPaulus:
solvespace on Arch Linux (both current) :

text input field in GraphicsWindow and TextWindow gets black transparent background, as soon as you start to type. Editing is effectively black on black.

The simplest workaround I found is commenting out the line in src/platform/gtkmain.cpp where the frame of Gtk::Entry is removed. 

class EditorOverlay : public Gtk::Fixed {
public:
    EditorOverlay(Gtk::Widget &underlay) : _underlay(underlay) {
        set_size_request(0, 0);
        add(_underlay);
        _entry.set_no_show_all(true);   
//        _entry.set_has_frame(false);
        add(_entry);
        _entry.signal_activate().
            connect(sigc::mem_fun(this, &EditorOverlay::on_activate));
    }

I checked with _entry.get_style_context()->get_background_color().to_string()  before and after _entry.set_has_frame(false) and this instruction currently changes the standard background color of this widget from  opaque white rgb(255,255,255) to transparent black rgba(0,0,0,0) . 

Gerhard

whitequark:
@GerhardPaulus Unfortunately `set_has_frame(true)` screws up positioning.

The correct patch should be somewhere along these lines:
```
diff --git a/src/platform/gtkmain.cpp b/src/platform/gtkmain.cpp
index 079fd395..5bab2dec 100644
--- a/src/platform/gtkmain.cpp
+++ b/src/platform/gtkmain.cpp
@@ -17,6 +17,7 @@
 #include <glibmm/convert.h>
 #include <giomm/file.h>
 #include <gdkmm/cursor.h>
+#include <gtkmm/cssprovider.h>
 #include <gtkmm/drawingarea.h>
 #include <gtkmm/glarea.h>
 #include <gtkmm/scrollbar.h>
@@ -1417,6 +1418,17 @@ int main(int argc, char** argv) {
 
     Gtk::Main main(argc, argv);
 
+    // Add our application-specific styles, to override GTK defaults.
+    Glib::RefPtr<Gtk::CssProvider> style_provider = Gtk::CssProvider::create();
+    style_provider->load_from_data(R"(
+    GtkEntry {
+        background: red;
+    }
+    )");
+    Gtk::StyleContext::add_provider_for_screen(Gdk::Screen::get_default(),
+                                               style_provider,
+                                               600 /*Gtk::STYLE_PROVIDER_PRIORITY_APPLICATION*/);
+
[...]

PR #216, opened 2017-03-19
Decorate GtkEntry fields with CSS to ensure that they are visible

tpimh:
The code is based on [this patch](https://github.com/solvespace/solvespace/issues/184#issuecomment-277509013), I only changed CSS code.

Before:
before
After:
after

### DIFF
--- a/src/platform/gtkmain.cpp
+++ b/src/platform/gtkmain.cpp
@@ -17,6 +17,7 @@
 #include <glibmm/convert.h>
 #include <giomm/file.h>
 #include <gdkmm/cursor.h>
+#include <gtkmm/cssprovider.h>
 #include <gtkmm/drawingarea.h>
 #include <gtkmm/glarea.h>
 #include <gtkmm/scrollbar.h>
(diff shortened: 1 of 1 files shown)


src/platform/gtkmain.cpp
```diff
@@ -1410,6 +1411,19 @@ int main(int argc, char** argv) {
 
     Gtk::Main main(argc, argv);
 
+    // Add our application-specific styles, to override GTK defaults.
+    Glib::RefPtr<Gtk::CssProvider> style_provider = Gtk::CssProvider::create();
+    style_provider->load_from_data(R"(
+      entry {
+        background: black;
+        color: white;
+        font-size: 12px;
```
whitequark: This isn't either correct or necessary, the font size is set explicitly.

whitequark:
I've applied a different CSS in master, thanks for discovering the necessary trick for this.

tpimh:
Now it looks like this:
new
I personally would prefer white text on black background, because the interface scheme is dark and dark textbox fits better. That's just a matter of taste.
I agree, overriding font-size here is not a good idea. I think that 16px font is too large and 12px would be optimal. Does it look different with different DPI or GTK+ theme?
Does `font_height` of `16` look oversized on Win32 and Mac?

whitequark:
> I agree, overriding font-size here is not a good idea. I think that 16px font is too large and 12px would be optimal. Does it look different with different DPI or GTK+ theme?

Looks like this for me:
screenshot_20170319_195157

And, since the font size is overridden and inferred from text window font size, the theme should not matter.

By the way, please don't send pull requests, I cannot accept them as there is no CLA yet.

whitequark:
> Does font_height of 16 look oversized on Win32 and Mac?

Oh, and this is irrelevant because we don't use GTK on any platform except Linux.

tpimh:
`CHAR_HEIGHT` is a constant and it is 16 for all platforms.
On your setup textbox looks well-centered vertically and doesn't seem oversized. Probably it's Adwaita's problem.

whitequark:
> CHAR_HEIGHT is a constant and it is 16 for all platforms.

This doesn't matter. The size of the text entry in the graphics window varies, as it fits the constraint label.

> Probably it's Adwaita's problem.

No. The theme should not matter.

Issue #92, opened 2016-10-22
stack overflow when compiling win32 - solved


jmkinzer:
Compiling under VS2015 I received an immediate crash due to stack overflow initializing SolveSpace::SS (at least under the debug build). I found it necessary to bump the stack to around 32MB to resolve this, i.e.:
/STACK:33554432
Would recommend adding something like this to CMakeLists.txt (untested):
`
if(MSVC)
    set(CMAKE_EXE_LINKER_FLAGS "/STACK:33554432 ${CMAKE_EXE_LINKER_FLAGS}")
endif()
`

whitequark:
Yes, this has been a problem with debug builds in the past with other functions... interestingly, VS2013 doesn't have that. Ideally this should be fixed by redoing the way the text window works.

Evil-Spirit:
@jmkinzer, thank you, I have the same issue but not actually try to understand why it happens.

bjaraujo:
Got the same issue. 
Going to try that fix.

plachenko:
This allowed me to compile in VS2017-- will this be added to master?

ruevs:
@whitequark 
Let me resurrect this ancient one. I think I know why the stack overflow was (is) happening on Windows. Using the code analysis tool in MSVC 2019 I get the following warnings:

```
.\src\solvespace.h(813): warning C6262: Function uses '5614392' bytes of stack:  exceeds /analyze:stacksize '16384'.. This allocation was for a compiler-generated temporary for 'class SolveSpace::TextWindow' at line 813.  Consider moving some data to heap.
.\src\solvespace.h(814): warning C6262: Function uses '21016632' bytes of stack:  exceeds /analyze:stacksize '16384'.. This allocation was for a compiler-generated temporary for 'class SolveSpace::System' at line 814.  Consider moving some data to heap.
```

The lines in question are these:
https://github.com/solvespace/solvespace/blob/master/src/solvespace.h#L813

So the analyser thinks that this constructor:
```
    SolveSpaceUI()
        : pTW(new TextWindow({})), TW(*pTW),
          pSys(new System({})), sys(*pSys) {}
```
somehow temporarily puts a copy of both `TextWindow` and `System` on the stack or about 26MB - not nice. I'll try to debug this on assebler level to confirm that this is indeed the case.

whitequark:
[...]

PR #238, opened 2017-04-17
Use in-tree external libraries and added fix for stack overflow issue #92

bjaraujo:
Help compile/debug on windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,24 +103,31 @@ if(WIN32)
     include(FindVendoredPackage)
     include(AddVendoredSubdirectory)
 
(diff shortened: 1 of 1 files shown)


whitequark:
Sorry, no. The `find_vendored_package` function was written as it is for a reason. And the fix for #92 does not involve enlarging the stacks but rewriting the code in question to not rely on massive arrays.

whitequark:
(Also, your changes don't even work: https://ci.appveyor.com/project/whitequark/solvespace/build/3.0.423).

bjaraujo:
It is better than not compiling at all...

PR #252, opened 2017-05-04
Remove extra setjmp include.

romeojulietthotel:
#include <setjmp.h> appears twice, so remove one of them.


whitequark:
Fixed in f54dabbb5fb049.

PR #257, opened 2017-05-08
French translation

patrovite:
This is the french translation for SolveSpace.
Status of the translation : 100%

Note: The hexadecimal code for the country (0x040C for French) must be in capital character otherwise the program crash. May be a note in the doc would good.

### DIFF
--- a/res/CMakeLists.txt
+++ b/res/CMakeLists.txt
@@ -197,6 +197,7 @@ add_resources(
     locales/en_US.po
     locales/uk_UA.po
     locales/ru_RU.po
+    locales/fr_FR.po
     fonts/unifont.hex.gz
     fonts/private/0-check-false.png
     fonts/private/1-check-true.png
--- a/res/locales.txt
+++ b/res/locales.txt
@@ -3,3 +3,4 @@
 en-US,0409,English (US)
 ru-RU,0419,Русский
 uk-UA,0422,Українська
+fr-FR,040C,Français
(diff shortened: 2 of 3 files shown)


whitequark:
Is there a reason you didn't preserve the orthography of menu items? E.g.:
```
msgid "Save File"
msgstr "Sauver fichier"
```
In the former every word is capitalized, in the latter is not.

Otherwise looks good to me!

patrovite:
I'll answer with a question : Why did you capitalized every word ? :)
It's more natural to have only one capital per sentence at the beginning.
In the translation I made some small change in order to more clear, e.g. : 
Activate one with Sketch -> In Workplane. >> Activez un plan avec "Dessin -> Dans plan de travail". 

Also it wasn't possible to put "&" at the same place and sometime you have 2 same shorcuts letters in the same menu.

Let's see what other user think about it...

whitequark:
> I'll answer with a question : Why did you capitalized every word ? :)

It's a stylistic choice made by @jwesthues. I don't see it as either good or bad, but I think we should be consistent across languages...

> Also it wasn't possible to put "&" at the same place and sometime you have 2 same shorcuts letters in the same menu.

Please feel free to put `&` wherever they feel natural for you as a French user, there is absolutely no need to keep them consistent between locales, it is much more important to have an unique one for every menu item.

patrovite:
I capitalized some start of words of the french translation in order to be consistent with the original.

whitequark:

In addition, could you please update the translation to address changes made in master? I apologize that it took me so long to get around to this pull request.

patrovite:
Hi,
I updated the french translation in my fork.
Forks & PR are a new for me. Something else to do?

PR #284, opened 2017-08-25
Update README.md with caveats building Solvespace on 16.04.

cr1901:
As you probably saw in IRC today/recently, I had some difficulties building Solvespace, even cross-compiling from Linux. It turns out the mingw-w64 package provided by default is too old.

I'm certain many people use Ubuntu 16.04 (or 14.04) LTS, so I've provided some instructions on how to build on LTS distros in the README.md to save others some moderate frustration. All upgraded packages do not depend on any newer libraries than those already provided by 16.04 by default (can't speak for 14.04).

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,21 @@ is built as `build/bin/solvespace-cli.exe`.
 
 Space Navigator support will not be available.
 
+If using Ubuntu 16.04 LTS (Xenial Xerus), the `mingw-w64` package provided by the distribution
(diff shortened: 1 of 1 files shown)


whitequark:
Note for myself: preferably just apply this patch: https://fossies.org/linux/qt-everywhere-opensource-src/qtbase/src/angle/patches/0002-ANGLE-Fix-compilation-with-MinGW.patch

whitequark:
I've added a more concise note myself.

PR #286, opened 2017-08-30
Js fixes

cr1901:
This pull request fixes bugs in the THREE.js viewer, including:

* `mousemove` handler no longer relies on `event.button` having a valid value (per [MDN](https://developer.mozilla.org/en-US/docs/Web/Events/mousemove), it was a happy accident that the handler ever worked).
* HiDPI display controls (or displays where physical pixels do not correspond to a CSS/browser pixel) are fixed.
* JS has been linted.

I have verified the fixes work on Windows. Please let me know whether a freshly compiled Solvespace exports the fixed viewer before accepting.

### DIFF
--- a/res/threejs/SolveSpaceControls.js
+++ b/res/threejs/SolveSpaceControls.js
@@ -154,10 +154,11 @@ SolvespaceControls = function(object, domElement) {
         event.preventDefault();
         event.stopPropagation();
 
+        console.log(event.screenX, event.screenY);
         switch (event.button) {
             case 0:
-                _rotateCur.set(event.screenX / window.devicePixelRatio,
-                               event.screenY / window.devicePixelRatio);
+                _rotateCur.set(event.screenX,
+                               event.screenY);
                 _rotatePrev.copy(_rotateCur);
                 document.addEventListener('mousemove', mousemove_rotate, false);
                 document.addEventListener('mouseup', mouseup, false);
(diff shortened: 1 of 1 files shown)


res/threejs/SolveSpaceControls.js
```diff
@@ -217,29 +217,39 @@ SolvespaceControls = function(object, domElement) {
...
+        var incDiff = new THREE.Vector2().subVectors(_rotateCur, _rotatePrev)
             .multiplyScalar(1 / object.zoomScale);
         object.rotate(-0.3 * Math.PI / 180 * incDiff.x * object.zoomScale,
              -0.3 * Math.PI / 180 * incDiff.y * object.zoomScale);
         _changed = true;
         _rotatePrev.copy(_rotateCur);
     }
 
+    /* eslint-disable no-unused-vars */
     function panstart(event) {
```
whitequark: Can this be replaced with e.g. `function panstart(_event) {` ?

whitequark:
> There's just one more issue I can see: after releasing two fingers after a pinch gesture, the model starts to rotate and this can result in a strange jerk. Do you think you can fix this easily?

Friendly ping--have you had any time to look at this issue?

cr1901:
>Friendly ping--have you had any time to look at this issue?
Not much. The issue doesn't appear in the touchscreen emulation of hammer.js, so I need to hook up my phone to Chrome's console and enable remote debugging. I haven't gotten around to doing that.

whitequark:
Merged manually, thanks.

cr1901:
>There's just one more issue I can see: after releasing two fingers after a pinch gesture, the model starts to rotate and this can result in a strange jerk. Do you think you can fix this easily?

Is this issue still there? I can't fix right now for reasons I discussed, but I'm happy to try once I'm a bit better.

whitequark:
Yeah, still there in master.

PR #289, opened 2017-09-13
Updating dependencies in README

dgrundel:
`libgtkmm` is wrong version is `x11` is missing from the packages to install. All git submodules were needed on my system (or at least more than just `libdxfrw`.)

Might also want to mention in the README that the repo should be cloned, not downloaded from the downloads page, but I'll leave that to someone else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,18 +25,18 @@ Building on Linux
 
 ### Building for Linux
 
-You will need CMake, libpng, zlib, json-c, fontconfig, freetype, gtkmm 2.4,
+You will need CMake, libpng, zlib, json-c, fontconfig, freetype, gtkmm 3.0, x11,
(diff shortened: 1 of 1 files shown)


whitequark:
Sorry, this is not correct. The SolveSpace 2.x branch indeed depends on GTK 2.x, as supplied in README. Are you sure you're running the right version of SolveSpace and/or are submitting the PR  against the right branch?

dgrundel:
You are absolutely correct (of course.) I cloned the master branch but was following the 2.x README during the install. That's never a recipe for success.

FWIW, I did still need to `apt install libx11-dev` and pull in the rest of the submodules before I could build the master branch, but I might have something else going on. (I am running Ubuntu Gnome 16.04)

Thanks for looking.

PR #290, opened 2017-09-13
Yet another Show DOF in text window

bcmpinc:
Shows the group's DOF in the group list. The ok-column is changed into a dof-column. It shows 'ERR' when there is an error, 'ok' when DOF==0, '###' if DOF>999, and otherwise the DOF itself.

Compared to other patches: 
- #200: uses 'D2' instead of 'ok' to show the DOF. If DOF>9 shows 'D+'.
- #204: leaves the ok-column as is and adds a separate dof-column.


whitequark:

I think I like your patch the most of all, so I'm inclined to merge it over the other PRs as soon as it's possible.

whitequark:
Merged with minor modifications as 8192c96, thanks!

PR #307, opened 2018-01-02
Add a linker option to work around the stack overflow on startup

rpavlik:
This was needed to successfully run a build from VS2017. This is basically just putting the patch from #92 into a convenient place in the build system and making a pull request.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -255,6 +255,9 @@ if(MSVC)
     # We rely on these /we flags. They correspond to the GNU-style flags below as
     # follows: /w4062=-Wswitch
     set(WARNING_FLAGS   "${WARNING_FLAGS} /we4062")
+
+    # Needed to avoid crash on start with VS2017, etc.
+    set(CMAKE_EXE_LINKER_FLAGS "/STACK:33554432 ${CMAKE_EXE_LINKER_FLAGS}")
 endif()
 
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")


whitequark:
The reason it's not in master is this needs to be fixed properly (as in by not having this massive data structure in the first place).

rpavlik:
Ah, yes, that would probably also be good :) I just didn't know the codebase well enough to know how to do that, nor why my VS2017 x64 build was stymied by that.

Though, after going through and re-implementing `IdList` to wrap `std::vector` and cut file loading time by 1/3, I probably nearly know it well enough now... 😉

whitequark:
> Though, after going through and re-implementing IdList to wrap std::vector and cut file loading time by 1/3, I probably nearly know it well enough now...

Yeah, the homegrown collection situation is... unfortunate.

rpavlik:
what are your opinions on the code's self-professed style guide of not having private members? Also unfortunate? or current design goal?

I managed to pull off even faking an `n` member by having it be a `std::reference_wrapper<const int>` to a private int that gets updated any time a mutating method was called, so porting from/improving the collection situation isn't actually as tough as it seems (or as tough as I first made it before the proxy object/reference wrapper idea occurred to me)

whitequark:
> what are your opinions on the code's self-professed style guide of not having private members? Also unfortunate? or current design goal?

Privacy at class boundaries is a mistake. Privacy should be at module boundaries (even modern C++ admits that, with the widespread use of the pimpl pattern). The collections being in the same module as all the geometry and graphics logic is unfortunate, but probably unavoidable for now.

> I managed to pull off even faking an n member by having it be a std::reference_wrapper<const int> to a private int that gets updated any time a mutating method was called, so porting from/improving the collection situation isn't actually as tough as it seems (or as tough as I first made it before the proxy object/reference wrapper idea occurred to me)

That's horrifying. And clever. And horrifying. Ideally, I would like SolveSpace to migrate most of its loops to C++11 foreach loops and most of the rest to std algorithms, which should eliminate this problem, but meanwhile this *may* work. I'm not yet completely sure though as to how much I like it.

PR #309, opened 2018-01-02
Replace calls to fabs in util.cpp with calls to std::abs.

rpavlik:
Minor cleanup. As I put in the commit msg:

> Some of these showed up in profiling hotspots. std::abs is
> properly overloaded to not cause undesired conversions, which
> (for numeric values) are often more expensive than you'd expect.
> fabs is not overloaded: it's a C function that only operates on
> doubles.

Confession: when I first started this I thought `fabs` only operated on floats, and thus there were a ton of unnecessary and expensive double->float->double conversions happening. I checked that assumption before making the commit, and thus discovered my misunderstanding. However, this change does not decrease performance, and if nothing else should improve type safety because std::abs is generic (so no matter how the code changes, it won't introduce conversions).

Update: I just noticed this definition in the header: https://github.com/solvespace/solvespace/blob/c6fc0125a21aa27f89ab6e0bb16037f04bc39fd4/src/solvespace.h#L115 - which I imagine is probably actually what the std::abs definition is (vs is acting up on me so I can't double-check it right now).

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -497,7 +497,7 @@ Vector Vector::Normal(int which) const {
 
     // Arbitrarily choose one vector that's normal to us, pivoting
     // appropriately.
-    double xa = fabs(x), ya = fabs(y), za = fabs(z);
+    double xa = std::abs(x), ya = std::abs(y), za = std::abs(z);
     if(this->Equals(Vector::From(0, 0, 1))) {
         // Make DXFs exported in the XY plane work nicely...
         n = Vector::From(1, 0, 0);
@@ -645,7 +645,7 @@ Vector Vector::WithMagnitude(double v) const {
     double m = Magnitude();
     if(EXACT(m == 0)) {
         // We can do a zero vector with zero magnitude, but not any other cases.
-        if(fabs(v) > 1e-100) {
+        if(std::abs(v) > 1e-100) {
             dbp("Vector::WithMagnitude(%g) of zero vector!", v);
         }
         return From(0, 0, 0);
@@ -693,7 +693,7 @@ Vector4 Vector::Project4d() const {
 }
 
 double Vector::DivPivoting(Vector delta) const {
-    double mx = fabs(delta.x), my = fabs(delta.y), mz = fabs(delta.z);
+    double mx = std::abs(delta.x), my = std::abs(delta.y), mz = std::abs(delta.z);
 
     if(mx > my && mx > mz) {
         return x/delta.x;
@@ -705,7 +705,7 @@ double Vector::DivPivoting(Vector delta) const {
 }
 
 Vector Vector::ClosestOrtho() const {
-    double mx = fabs(x), my = fabs(y), mz = fabs(z);
+    double mx = std::abs(x), my = std::abs(y), mz = std::abs(z);
 
     if(mx > my && mx > mz) {
         return From((x > 0) ? 1 : -1, 0, 0);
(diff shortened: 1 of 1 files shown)


whitequark:
Thanks for the pull requests! I'm still sorting out the licensing situation with SolveSpace, so I won't be able to immediately merge them. I am sorry, and I'll get back to you as soon as possible.

rpavlik:
sorry to hear there is a "licensing situation" - I had some other improvements (low-hanging fruit and slightly-less-low-hanging-fruit perf-wise) in the pipeline, will put on hold.

whitequark:
> sorry to hear there is a "licensing situation" - I had some other improvements (low-hanging fruit and slightly-less-low-hanging-fruit perf-wise) in the pipeline, will put on hold.

Please go ahead. The only things that need to be done are:
  * rebasing a few hundreds of commits on the master branch on top of the 2.0 release, and
  * figuring out some sort of website that lets contributors sign CLA.

Assuming you are OK signing a CLA, I will definitely merge your work!

rpavlik:
As long it's not overly onerous, CLAs are fine - the one that Apache uses is pretty commonly used/adapted (even when not using the apache license).

whitequark:
In essence, we want to be able to license SolveSpace to commercial customers, with the fees funding continued development of SolveSpace.

whitequark:

whitequark:
@rpavlik Friendly ping--could you take another look at the CLA?

rpavlik:
Sure, why not. Initially was a bit "hmm" at the terms, but somebody may as well make use of the changes I made.

whitequark:
Great!

> I had some other improvements (low-hanging fruit and slightly-less-low-hanging-fruit perf-wise) in the pipeline

I would be happy to see these improvements.

whitequark:
Thanks!

whitequark:
Hm... I've reverted the PR as it causes build failures on Linux: https://travis-ci.org/solvespace/solvespace/jobs/498495065. Could you take a look? I'm not sure how to disambiguate these in a nice way...

ruevs:
This is weird. Why does GCC (g++) even consider the integer "abs" declarations in `stdlib.h` and `cstdlib` when calling `std::abs(x)` ?

I have no way to test this now but perhaps adding `using std::abs;` here https://github.com/solvespace/solvespace/blob/9d1601eea9039011c16c435b26a680fb4bc35c34/src/solvespace.h#L80
would help (but it does not make any sense to me).


```
/home/travis/build/solvespace/solvespace/src/util.cpp: In member function ‘SolveSpace::Vector SolveSpace::Vector::Normal(int) const’:
/home/travis/build/solvespace/solvespace/src/util.cpp:512:27: error: call of overloaded ‘abs(const double&)’ is ambiguous
     double xa = std::abs(x), ya = std::abs(y), za = std::abs(z);
                           ^
In file included from /home/travis/build/solvespace/solvespace/src/solvespace.h:18:0,
                 from /home/travis/build/solvespace/solvespace/src/util.cpp:7:
/usr/include/stdlib.h:775:12: note: candidate: int abs(int)
 extern int abs (int __x) __THROW __attribute__ ((__const__)) __wur;
            ^
In file included from /usr/include/c++/5/bits/stl_algo.h:59:0,
                 from /usr/include/c++/5/algorithm:62,
                 from /home/travis/build/solvespace/solvespace/src/solvespace.h:20,
                 from /home/travis/build/solvespace/solvespace/src/util.cpp:7:
/usr/include/c++/5/cstdlib:166:3: note: candidate: long int std::abs(long int)
   abs(long __i) { return __builtin_labs(__i); }
   ^
/usr/include/c++/5/cstdlib:174:3: note: candidate: long long int std::abs(long long int)
   abs(long long __x) { return __builtin_llabs (__x); }
   ^
/usr/include/c++/5/cstdlib:179:3: note: candidate: __int128 std::abs(__int128)
   abs(__GLIBCXX_TYPE_INT_N_0 __x) { return __x >= 0 ? __x : -__x; }
   ^
```

Issue #324, opened 2018-05-29
Smarter font selection for Command::TTF_TEXT


henriklaxhuber:
This is not a bug report but a feature request/discussion.

Currently, when creating a new TTF_TEXT object, the font defaults to `arial.ttf` (see [here](https://github.com/solvespace/solvespace/blob/5744d1d599dfaeaa768e1667509eee1fda3b8aea/src/mouse.cpp#L1139)). Especially on linux systems (such as mine), `arial.ttf` may not be present. This means that every time I create a new TTF_TEXT object, I need to manually scroll down the list and select a suitable sans-serif font, which is rather tedious.

To resolve this issue, the default font should be selected in such a way that it actually exists. I can see primarily three options here:

1.) Require `arial.ttf` as a dependency. This is a simple solution and enables cross-platform compatability. However, it is very much against the GPL philosophy (see the [arial licensing info](https://github.com/solvespace/solvespace/blob/5744d1d599dfaeaa768e1667509eee1fda3b8aea/src/mouse.cpp#L1139)).

1.1.) Require/bundle something like FreeSans as a dependency and use it by default.

2.) Use pango to perform automatic font selection. This mean the user gets to see their default sans-serif font being used. However, the font may not be present on other systems opening the save file, or, if only the 'sans-serif' directive is stored in the save file, may look different on other systems.

3.) Persist the last selected font and use it as default when creating a new TTF_TEXT object. The obvious downside is that if the user intents to change the font only for a specific object, he will need to go through the process of scrolling down the list again.

I'm very much willing to create a pull request solving this issue, but I was hoping for some guidance as to which direction to go. Personally, I would prefer the solution 1.1, as it reduces the problem to a one-off issue when first installing solvespace.

whitequark:
Sure, let's add FreeSans (or maybe Bitstream Vera Sans) as a default TTF font.

henriklaxhuber:
I'm supposed to be on holiday, but since this seems like a quick fix I'll make a PR later this week or so.

whitequark:
Thank you! I suggest adding the font under `res/fonts` and then adding a separate item in the TTF font selector, say `Bitstream Vera Sans (built-in)`. This would clearly establish that this font will never go missing.

henriklaxhuber:
Ended up getting around to implementing it tonight.

PR #334, opened 2018-07-12
Built in Bitstream Vera Sans Roman font

henriklaxhuber:
Added Bitstream Vera Sans Roman as a builtin font and made it the default font for the TTF_TEXT entity.

This change adds a method TtfFont::LoadFromResource to handle loading TTF fonts from resources.
To implement this, I moved the freetype2 processing code into its own function, such that LoadFromFile and LoadFromResource only do the loading and then hand off to the new method PostLoadProcess.
 
The lazy loading mechanism for fonts loaded from a file (nameOnly=true) interfered with the resource loading, as LoadFont would call LoadFromFile on a font actually provided by a resource.
To circumvent this, I removed the nameOnly option for resource fonts such that they are fully loaded on initialisation of SolveSpace. This solution is far from ideal (see my comment in the code), but works as long as there are few resources.

This PR solves issue #324


src/ttf.cpp
```diff
@@ -149,6 +155,39 @@ bool TtfFont::LoadFromFile(FT_Library fontLibrary, bool nameOnly) {
...
+    // lazily loaded as a file. This could also be implemented via a `bool
+    // is_resource` member in the TtfFont class. However, as long as there are only
+    // few bundled fonts, this works fine.
+    return PostLoadProcess(false);
+}
+
+//-----------------------------------------------------------------------------
+// Extract font information. We care about the curves that define the letter
+// shapes, and about the mappings that determine which glyph goes with which
+// character.
```
whitequark: This comment is outdated--we actually only care about font name and unit size.

src/ttf.h
```diff
@@ -11,17 +11,21 @@
...
 
     std::string FontFileBaseName() const;
     bool LoadFromFile(FT_LibraryRec_ *fontLibrary, bool nameOnly = true);
+    bool LoadFromResource(FT_LibraryRec_ *fontLibrary);
 
     void PlotString(const std::string &str,
                     SBezierList *sbl, Vector origin, Vector u, Vector v);
     double AspectRatio(const std::string &str);
+
+private:
```
whitequark: SolveSpace does not use C++ visibility mechanisms.

src/ttf.h
```diff
@@ -11,17 +11,21 @@
...
+  bool PostLoadProcess(bool nameOnly);
```
whitequark: Wrong indentation.

res/CMakeLists.txt
```diff
@@ -210,6 +210,7 @@ add_resources(
     fonts/private/6-stipple-dash.png
     fonts/private/7-stipple-zigzag.png
     fonts/unicode.lff.gz
+    fonts/BitstreamVeraSans-Roman-builtin.ttf
```
whitequark: What's the license for Bitstream Vera? Can you please rename `COPYING.txt` to `COPYING-GPL3.txt` and add a new `COPYING.txt` describing which parts are covered by which license? Ideally that should cover `extlib/` but you can just cover the newly added font and I'll fill in the rest.

src/ttf.cpp
```diff
@@ -149,6 +155,39 @@ bool TtfFont::LoadFromFile(FT_Library fontLibrary, bool nameOnly) {
...
+//-----------------------------------------------------------------------------
+bool TtfFont::PostLoadProcess(bool nameOnly) {
```
whitequark: I feel like a better name for this function would be `ExtractInformation` or something.

src/ttf.cpp
```diff
@@ -149,6 +155,39 @@ bool TtfFont::LoadFromFile(FT_Library fontLibrary, bool nameOnly) {
...
+    if(int fterr = FT_New_Memory_Face(fontLibrary, buffer, size, 0, &fontFace)) {
+            dbp("freetype: loading font '%s' from memory failed: %s",
+                fontFile.raw.c_str(), ft_error_string(fterr));
+            return false;
+    }
+
+    // Call with nameOnly = false as a quick hack to prevent the font from being
+    // lazily loaded as a file. This could also be implemented via a `bool
+    // is_resource` member in the TtfFont class. However, as long as there are only
+    // few bundled fonts, this works fine.
```
whitequark: I agree that this is suboptimal. Do you think you could add a prefix for builtin fonts, e.g. `res:`, so that the font you added becomes `res:/fonts/BitstreamVeraSans-Roman-builtin.ttf`? This would also unambiguously distinguish builtin fonts in savefiles (so that system fonts with the same filename don't interfere with them), and also allow us to add new builtin fonts and keep compatibility between versions.

henriklaxhuber: Sure, that makes sense, although I would change it to 'res://' similar to protocol prefixes such as 'ssh://' etc. I feel like that makes the intent very clear.

res/CMakeLists.txt
henriklaxhuber: I'm not sure I understand what you mean. The Bistream Copyright notice should be included in the distribution of SolveSpace; adding it as a resource (i.e. inlining it into the executable's data section) doesn't really seem like the best idea here.
Or do you mean in the project root? Though then I would name the file 'THIRD_PARTIES.txt' and list all third party licenses/copyrights etc there.

src/mouse.cpp
```diff
@@ -1136,7 +1136,7 @@ void GraphicsWindow::MouseLeftDown(double mx, double my) {
                     AddToPending(hr);
                     Request *r = SK.GetRequest(hr);
                     r->str = "Abc";
-                    r->font = "arial.ttf";
+                    r->font = "BitstreamVeraSans-Roman-builtin.ttf";
```
whitequark: This isn't updated to use the `res://`, and the `-builtin` suffix is no longer necessary to disambiguate.

src/ttf.cpp
```diff
@@ -84,11 +92,14 @@ TtfFont *TtfFontList::LoadFont(const std::string &font)
     LoadAll();
 
     TtfFont *tf = std::find_if(l.begin(), l.end(),
-        [&](const TtfFont &tf) { return tf.FontFileBaseName() == font; });
+        [&font](const TtfFont &tf) { return tf.FontFileBaseName() == font; });
 
     if(tf != l.end()) {
         if(tf->fontFace == NULL) {
-            tf->LoadFromFile(fontLibrary, /*nameOnly=*/false);
+            if(tf->fontFile.raw.compare(0, 6, "res://"))
```
whitequark: Here, it seems to me like all logic handling the `res:` prefix ought to be concentrated in the same place. (And if not, at least `LoadFromResource` should use an assertion before chopping off the prefix.)

src/mouse.cpp
henriklaxhuber: The res:// is added to the `fontFile` when loading the fonts. Setting a font is only done by the base name, which does not contain the res://. This is also why I would keep the builtin suffix.

src/ttf.cpp
henriklaxhuber: That seems reasonable.

henriklaxhuber:
What's preventing this from getting merged?

whitequark:
Nothing; I have not noticed your changes. I apologize for that; I'll take a look soon.

whitequark:
I apologize again for the delay. I haven't been able to dedicate as much of my time to maintenance as I wished to.

I've merged your pull request, and I'm looking forward to more great contributions in the future.

PR #339, opened 2018-07-25
Fixed macOS build

Lichtso:
Because of the missing "synthesize acceptsFirstResponder;"
the App crashes immediately on macOS: "unrecognized selector setAcceptsFirstResponder send to SSView".


whitequark:
This has quite a few changes other than the fix you mentioned. And, they break the build for me because I develop on 10.9.

whitequark:
Thank you very much for the PR, merged!

PR #353, opened 2018-09-13
Fix GIT_COMMIT_HASH to autoset

ktrace:
When build from git, GIT_COMMIT_HASH set to correct value.
When build from tarball, GIT_COMMIT_HASH set to 000000 as recommended


whitequark:
That is not what is recommended.

> and instead uncomment the following, **adding the complete git hash of the checkout you are using**

ktrace:
Sorry, but for source-based distro like gentoo we can't recommend _every_ user make changes by hand.

ktrace:
> I guess I could make it possible to specify GIT_COMMIT_HASH via a command-line option to cmake as well.

Yes, you're right

whitequark:
@ktrace Thanks for the report. I've fixed this properly, using `ident` in `.gitattributes`. Now, any git checkout (including archives) can be built, regardless of their source.

Issue #78, opened 2016-10-14
Make Sketch self-contained


whitequark:
Currently, sketch generation requires a lot of global values, which means that two sketches cannot coexist in memory. This is a blocker for many other features, and should be refactored.

Thovthe:
Is this blocked on #75 or could it conceivably be done by someone with the time and inclination?  If so ~ how much time would this take?  And is it straightforward boring work or exciting but immensely difficult?

hinell:
@Thovthe There are tonnes of problems. I think it's best to start from sketch a new SolveSpace based on CAD formats like [STEP](https://en.wikipedia.org/wiki/ISO_10303-21). The current one is quite limited. It's also has weird modelling structure where the next entity depends on the previous one and relationships can't be defined manually.

phkahler:
> Is this blocked on #75 or could it conceivably be done by someone with the time and inclination? If so ~ how much time would this take? **And is it straightforward boring work or exciting but immensely difficult?**

No it's not blocked. In fact I'd say it's almost a requriement to change the internal data structures first. Some serious consideration needs to happen though so we're not passing pointers or references around. OTOH we have to reference the globals in a lot of places now. Refactoring code to reduce or eliminate globals should not affect functionality or file format though. But then a text window (as opposed to *the* text window) would need to know what sketch it's showing, same for a main drawing window. Just my thoughts.

DjogaRo:
I haven't seen any SolveSpace code, yet, so this might be a silly suggestion.
Declaring all those globals thread_local and having one editor thread per sketch?

phkahler:
> I haven't seen any SolveSpace code, yet, so this might be a silly suggestion. Declaring all those globals thread_local and having one editor thread per sketch?

We will still need to access multiple sketches from a single thread to do a hierarchy.

DjogaRo:
That's the part I cannot eyeball, not knowing the code. Such editor threads needed some interface. Depending on the code that's there this might be an easier fix (or not) - to add a keyword to those globals' declarations and implementing such interface - than tracking down all accesses to those globals to adjust them. Given that there are multiple branches being worked on, if I got that right.
If this could be easier, it would get this here TODO out of the way, until with reaching 4.0 everythings consolidated back into master and a more thorough approach would be more feasible.
I know, I have no idea what I'm talking about not knowing this codebase. Just throwing thoughts around.

timbod7:
[...]

Issue #301, opened 2017-11-08
Adding a constraint identical to an existing one should be no-op instead of unsat


azonenberg:
As of now, when you add a horizontal, vertical, etc constraint to a line, and one already exists, the solver declares the design unsatisfiable. The user's intention was probably to make that line go in the direction it's already in, so adding a second identical constraint should be a legal no-op (keep the existing constraint and don't add a new one).

Evil-Spirit:
This is very resonable and we can do the same for other types of solver errors. We can intorduce optional checkbox: Solver silent mode enabled by default (easy for new users).
1. Redundant constraint without value label -> remove immediate with error msgbox
2. Redundant constraint with value label -> add it like reference (actually implemented)
3. Didn't converge while adding constraint -> remove immediate with error msgbox
4. Didn't converge while dragging -> just undo to pervious valid position

whitequark:
Fixed in 406c55e8.

Instead of just being a no-op, I select the existing constraint, so that you can see feedback.

Issue #121, opened 2016-11-26
Implement tool → select → confirm workflow


whitequark:
Originally filed by @pimdegroot:

>The current way to use a tool in solvespace is to select the items you want to modify and then apply a tool.
>
>In other cad programs there is a second workflow where you select a tool, select the items you want to apply it to and then confirm.
>
>One example would be the dimension tool, which in the new workflow could work like this: Nothing selected > press D > select one or multiple lines or points > confirm with enter > have text floating under pointer > confirm location with click.
>
>Extrude could work like this: nothing selected > extrude tool > select areas to extrude > confirm > confirm location with click.

phkahler:
I'm curious how many people actually want this. It's a second way to do the same things, but actually takes *more* clicks than the existing way. We have "noun, verb" today but this asks for "verb, noun, confirm" to also be available. I honestly don't see the advantage and some improvements have been delayed/rejected in part to wait for this to land first.

But rather than just closing, I'm asking how many people would actually want this or not?

ruevs:
@phkahler No one wants is currently it seems. (neither do I)

But since there is even a (not very good) pull request on this I's leave it open for now...

Thovthe:
If you're used to verb>noun it will trip you up a bit when starting to use solvespace but this is a minor quirk.  This is low priority.


To track demand for this feature:
:thumbsup: if you want this added; :eyes: **Eyeballs** if you don't care; :thumbsdown: if it makes you feel icky gross that there be another option

PR #359, opened 2018-09-29
Splitting solvespace.h into separate headers as the first step towards modulization

marthinwurer:
As discussed [here on the forums.](http://solvespace.com/forum.pl?action=viewthread&parent=2444&tt=1538026805) I'm trying to break the system into its component parts as the first part of turning each section into a module so that #78 would be easier. 

This is a first draft. I have only tried compiling on my Ubuntu 16.04 linux machine, so I don't know if the cross-platform stuff still works. 

I need to figure out what to do with FatalError, because it needs to be extracted for AssertFailure to work correctly. Right now I just inlined the linux version to prevent the wonders of circular dependencies. 

Any and all constructive criticism is more than welcome; I don't know what's wrong with my changes if you don't tell me.


whitequark:
There's a ton of important work to be done towards modularization, for example eliminating the global sketch object. But I don't really see how moving class definitions around helps anything...

Evil-Spirit:
@whitequark 
let it be a first step

marthinwurer:
It's the first step towards that. Once headers are all separated, you can see where the circular dependencies are. For example, if we wanted to extract out our NURBS functionality into our own library, we'd have to take out the references to Group. Splitting the headers out lets you see those dependencies, and will make it easier to split the functionality up in the future. I didn't want to do any changes to the actual .cpp files yet; I just wanted to get the ball rolling and make those future changes easier.

whitequark:
@jwesthues What do you think?

jwesthues:
I see how this change would make the code seem easier to understand. I'm less convinced it would actually make the code easier to understand.

@marthinwurer I might suggest that you make your first change to SolveSpace (or other free software projects) an isolated feature or bugfix within the style and structure of the project as it currently exists, rather than a big change to style and structure. @whitequark may have some suggestions, or there's the issue tracker.

A small change is easier for a maintainer to digest, and it's also a good step towards more deeply understanding the codebase. This will help make subsequent bigger changes. Alternatively, you could start by drafting documentation of the current class structure, not just what's obvious from the headers but the way the objects are actually used and interact. This would be helpful both to yourself and to future contributors, and whitequark and others can help or correct where you're unsure. Sorry for not responding earlier on the forums.

whitequark:
Thanks, you have expressed my feelings also well.

Evil-Spirit:
So there is no any way to change things like this? Please describe how this can be changed, because this is really problematic thing for new contributors because modularization can simplify source code understanding and navigation.

whitequark:
@Evil-Spirit In the past, I have *first* modularized code, and *then* plucked it out into different fields. For example, that's how I've done the render/ and `platform/`. That's the way it should be done IMO.

Also, I'm not sure if "one file per class" rule is useful. This isn't Java; being able to get an overview of an entire component's interface, grouped in meaningful blocks, is valuable. (And there's nothing wrong with free functions either.)

marthinwurer:
@jwesthues That makes sense. Would #121 be a reasonable feature to start with?

Evil-Spirit:
@whitequark but why you don't merge my changes which makes sketch non-global? I've made it a long time ago.

whitequark:
> but why you don't merge my changes which makes sketch non-global? I've made it a long time ago.

There is no reason other than my health and general lack of time; I absolutely want to see those changes merged. I do give SolveSpace attention, but I also have to pay my bills...

whitequark:
> That makes sense. Would #121 be a reasonable feature to start with?

That would be a fairly sizable change. If you'd like to take it up, please discuss it with me first; we'd need to work out a plan to migrate tools one by one from the old system to the new one that would allow handling tool workflows in a more structured way.

marthinwurer:
Ok, would #301 be better? If not, do you have any suggestions for a ticket to start on?

whitequark:
@marthinwurer #301 seems good, and useful.

marthinwurer:
Ok, I'll start on that this weekend when I get time.

marthinwurer:
Ok, just got back into development. Some classes 100% need to be broken out into their own files for ease of navigation. Having SolveSpaceUI and Sketch in the same header makes using a C++ IDE (I use CLion) to try to navigate the codebase painful, so I'll be breaking those out into separate headers no matter what. As a new developer to this project, I have no clue where any code components are, and because there's little documentation, it's hard to tell what I need to change or what things do without reading all the code, which is hard when everything is in one massive header. I'll also be setting up the basics for Doxygen, which will let us easily build docs for the codebase javadoc-style. Also, is there a reason why operator overloading (mainly []) is not used for the list classes besides historical reasons?

whitequark:
> As a new developer to this project, I have no clue where any code components are, and because there's little documentation, it's hard to tell what I need to change or what things do without reading all the code, which is hard when everything is in one massive header.

Please consider that after you move all of the headers around, it will be even harder to understand, because right now you can use `git blame` to see what commit introduced the code and why, but after that all the lines will just be attributed to you. This is the reason I haven't refactored anything "just to make the code slightly nicer" so far, and stuck to refactoring where it gives a specific, nontrivial benefit--without being able to consult change history I would have gotten nowhere with this project.

> Also, is there a reason why operator overloading (mainly []) is not used for the list classes besides historical reasons?

Operator overloading, like STL and many other C++ features such as class member visibility, were not used anywhere for historical reasons. I am introducing some of them into the codebase (STL is now used where it makes good sense), but not some others (C++ implements visibility completely wrong and it should preferably not be used anywhere). Adding specifically operator overloading would be a massive, multiple month change that brings absolutely no specific benefits but adds bugs.

whitequark:
> Yeah, you're right, I hadn't considered that aspect.

Anyway, I'd be really happy to *eventually* get SolveSpace in a state where the sources are somewhat more structured, but I hope you can see why other things, like making code *actually* modular (and not just easier to navigate) were higher priority.

> Why doesn't the main program just use the library instead of doing all the fun conditional compilation stuff?

Because the geometric core contains a lot of UI-related code. E.g. compare Entity with EntityBase. All this UI-related code has to be removed in the library build (so that it doesn't drag any desktop dependencies), but everything was written with a GUI frontend in mind, so it's not really structured in a way that lets one easily inject UI dependencies. Thus, conditional compilation. (I am not happy about this at all, but again, other things took priority, e.g. the file format being backwards compatible *and* intricately tied to internal data representation, making either of those nearly impossible to change between versions.)

marthinwurer:
> I hope you can see why other things, like making code actually modular (and not just easier to navigate) were higher priority.

Yep, I'm starting to see that now. I'd like to apologize for simply jumping to conclusions instead of trying to understand how and why the codebase is the way that it is today. Nothing ever happens without a reason, and I should figure those reasons out before I attack them. Also, thank you for being so responsive and answering my questions in detail. 

In that vein, how do you feel about javadoc style comments? If you don't like that style, how should I be putting comments in and where (headers, cpp files)?

whitequark:
> In that vein, how do you feel about javadoc style comments? If you don't like that style, how should I be putting comments in and where (headers, cpp files)?

Doxygen style comments are 100% fine. I would be very happy to see more documentation, though it should of course be actual documentation and not IDE-generated (or even handwritten) comments of the style `/** Performs A. */ void a()`, which I occasionally see in projects that try to achieve 100% documentation coverage no matter what quality...

whitequark:
As per discussion above, I'm closing this PR. @marthinwurer, I will be glad to see your other contributions.

Issue #9, opened 2016-06-19
CI / Travis integration / auto-stats


twitchyliquid64:
Any reason there isnt CI setup ATM?
If not, I might try setting on up.

whitequark:
There is... https://travis-ci.org/solvespace/solvespace https://ci.appveyor.com/project/whitequark/solvespace

twitchyliquid64:
Kek, might wanna get these ones going then

https://docs.travis-ci.com/user/status-images/ etc

whitequark:
I don't see any point in those. They seem to be cargo-culted around open-source projects despite not providing any useful information. As a user of the project there is absolutely no reason for you to care about CI and as a contributor, how often do you look at those badges in README? Do you ever?

twitchyliquid64:
Disagree. The badge itself provides credability that the project follows TDD and has some kind of indicative QA around the state of the build.

whitequark:
See, you are making an excellent example. SolveSpace doesn't even *have tests*. All that the CI being green is currently demonstrating is that the packages can be built.

twitchyliquid64:
Oh. Welp.

whitequark:
(SolveSpace has tests now.)

PR #363, opened 2018-10-18
Adding TravisCI badge to Readme

TheKnarf:


whitequark:
See https://github.com/solvespace/solvespace/issues/9. I'm not sure what the point is, personally... I don't have these on any of my FOSS projects.

TheKnarf:
They function as a link to the CI, without them the only way in is to click the little green mark on a specific commit. While it's not something people need, sometimes I find myself clicking in on the CI to check which platforms and compilers the CI is setup to build on.

The whole idea of these badges is to show some important info about the project up front for people browsing. Like what license you use and if the branch is in a working state.

It's not necessarily that fun to clone a project and try to build it locally just to find out that someone with push access has broken the master branch, since the only way of knowing that is either building it locally or checking the CI status.

But as you said @whitequark it doesn't really do much, if you don't want it in your project feel free to close the PR.

whitequark:
> and if the branch is in a working state.

The thing is that Travis runs a horribly outdated version of Ubuntu that we can't even build the complete application on (we only build the CLI and tests currently), so the green badge is not an indicator of usability.

> They function as a link to the CI, without them the only way in is to click the little green mark on a specific commit. While it's not something people need that often, sometimes I find myself clicking in on the CI to check which platforms and compilers the CI is setup to build on.

Right, but there's another issue here, which is that on branches other than master and forks these badges still point to master of the parent repo. This can get pretty confusing.

We can add explicit links to Travis and Appveyor with a note saying that these are only valid for the master branch of solvespace/solvespace, but I'm not sure if it's easier to follow the commit links instead...

PR #380, opened 2019-02-19
Import styles from linked sketches

bcmpinc:
When a sketch that contains no custom styles links a sketch that does contain a custom style, a new custom style is created with the default format. This patch changes that and creates a new style with the same format as in the linked file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,6 +94,8 @@ Bugs fixed:
   * Paste Transformed with a negative scale does not invert arcs.
   * The tangent arc now modifies the original entities instead of deleting
     them, such that their constraints are retained.
+  * When linking a sketch file, missing custom styles are now imported from 
+    the linked file.
 
 2.x
 ---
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -77,6 +77,8 @@ Other new features:
   * The "=" key is bound to "Zoom In", like "+" key.
   * The numpad decimal separator key is bound to "." regardless of locale.
   * On Windows, full-screen mode is implemented.
+  * When linking a sketch file, missing custom styles are imported from the
+    linked file.
```
whitequark: I think this is a bugfix.

bcmpinc: Ok. I've changed that.

whitequark:
I think you pushed the same commit?

bcmpinc:
I used `git commit --am` and forgot `-a`.

whitequark:
Thanks!

PR #381, opened 2019-02-19
Uninitialized memory errors

bcmpinc:
Fixes two uninitialized memory errors in the toolbar.

### DIFF
--- a/src/toolbar.cpp
+++ b/src/toolbar.cpp
@@ -154,6 +154,9 @@ bool GraphicsWindow::ToolbarDrawOrHitTest(int mx, int my,
 
     bool withinToolbar =
         (mx >= aleft && mx <= aright && my <= atop && my >= abot);
+    
+    // Initialize/clear menuHit.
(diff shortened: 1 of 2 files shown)


src/toolbar.cpp
```diff
@@ -154,6 +154,9 @@ bool GraphicsWindow::ToolbarDrawOrHitTest(int mx, int my,
 
     bool withinToolbar =
         (mx >= aleft && mx <= aright && my <= atop && my >= abot);
+    
+    // Initialize/clear menuHit.
+    if(menuHit) *menuHit = Command::NONE;
```
whitequark: You've moved this above the early return. Are you sure this is sound?

whitequark: Nevermind, I misunderstood.

whitequark:
Thanks!

src/toolbar.cpp
bcmpinc: Yes, it is the early return that triggers the uninitialized memory error.

PR #392, opened 2019-04-06
Fix: disable mnemonic accelerator keys in menu's path labels

mehw:
**Solvespace version**: 3.0~46528bfc
**OS**: Gentoo GNU/Linux

_Already signed Solvespace's Contributor License Agreement._

### Bug

`File > Open Recent` and `New Group > Link Recent` replace/remove strings in file paths and activate mnemonic accelerator keys.

This means that on GNU/Linux the path `/tmp/_foo_&_bar_.slvs` is showed as `/tmp/foo_bar.slvs`.

**settings.json**
```
{
  "RecentFile_0":"\/tmp\/_foo_&_bar_.slvs"
}
```

before

### Proposed solution

Add `mnemonics` argument to `Platform::Menu::AddItem` to treat as raw text label's mnemonic accelerator keys when it is `false` (`src/platform/gui.h`).

~~New pure virtual function `Platform::Menu::AddItemRaw` which treats as raw text label's mnemonic accelerator keys (`src/platform/gui.h`).~~

~~`AddItemRaw `doesn't replace/remove label's text (i.e. `guigtk.cpp '&' -> '_'`), nor it activates mnemonic accelerator keys.~~

after

### Caveats

Solution tested only on Gentoo GNU/Linux.

The code in `src/platform/guiwin.cpp` and `src/platform/guimac.mm` isn't tested and may require amendments.

Does `NegateMnemonics` escape mnemonic accelerator keys correctly (`src/platform/guiwin.cpp`)?

~~Not sure if `AddItemRaw` in `guiwin.cpp` shall allocate a buffer before assigning it to `mii.dwTypeData`. Also, does `SetMenuItemInfoW` operate as expected updating the menu item's label?~~

Can someone review the code, please?

### DIFF
--- a/src/graphicswin.cpp
+++ b/src/graphicswin.cpp
@@ -281,7 +281,7 @@ void GraphicsWindow::PopulateMainMenu() {
         } else if(Menu[i].cmd == Command::LOCALE) {
             Platform::MenuRef localeMenu = currentSubMenu->AddSubMenu(label);
             for(const Locale &locale : Locales()) {
-                localeMenu->AddItem(locale.displayName, [&]() {
+                localeMenu->AddItem(locale.displayName, true, [&]() {
                     SetLocale(locale.Culture());
                     Platform::GetSettings()->FreezeString("Locale", locale.Culture());
 
@@ -344,7 +344,7 @@ static void PopulateMenuWithPathnames(Platform::MenuRef menu,
         menuItem->SetEnabled(false);
     } else {
         for(Platform::Path pathname : pathnames) {
-            Platform::MenuItemRef menuItem = menu->AddItemRaw(pathname.raw);
+            Platform::MenuItemRef menuItem = menu->AddItem(pathname.raw, false);
             menuItem->onTrigger = [=]() { onTrigger(pathname); };
         }
     }
--- a/src/graphicswin.cpp
+++ b/src/graphicswin.cpp
@@ -344,7 +344,7 @@ static void PopulateMenuWithPathnames(Platform::MenuRef menu,
         menuItem->SetEnabled(false);
     } else {
         for(Platform::Path pathname : pathnames) {
-            Platform::MenuItemRef menuItem = menu->AddItem(pathname.raw);
+            Platform::MenuItemRef menuItem = menu->AddItemRaw(pathname.raw);
             menuItem->onTrigger = [=]() { onTrigger(pathname); };
         }
     }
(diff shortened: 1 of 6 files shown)


whitequark:
I would prefer if you added an argument to `AddItem`, `bool mnemonics = true`, instead.

mehw:
Ok, I'll provide an alternative way as new commit. Then, if you approve it, let's squash the commits.

src/graphicswin.cpp
```diff
@@ -281,7 +281,7 @@ void GraphicsWindow::PopulateMainMenu() {
         } else if(Menu[i].cmd == Command::LOCALE) {
             Platform::MenuRef localeMenu = currentSubMenu->AddSubMenu(label);
             for(const Locale &locale : Locales()) {
-                localeMenu->AddItem(locale.displayName, [&]() {
+                localeMenu->AddItem(locale.displayName, true, [&]() {
```
whitequark: Please add the comment on each call site indicating the purpose of the boolean argument, i.e. `/*mnemonic=*/true`, as it's done elsewhere. Otherwise it's quite confusing.

src/platform/gui.h
```diff
@@ -171,7 +171,8 @@ class Menu {
     virtual ~Menu() {}
 
     virtual std::shared_ptr<MenuItem> AddItem(
-        const std::string &label, std::function<void()> onTrigger = std::function<void()>()) = 0;
+        const std::string &label, bool mnemonics = true,
```
whitequark: I think unless you put the `mnemonics` argument after `onTrigger` the default value for it is quite pointless.

mehw:
Pull amended as per @whitequark request.

[First post](https://github.com/solvespace/solvespace/pull/392#issue-268011821) edited to reflect the current behavior.

src/graphicswin.cpp
mehw: Thanks!

src/platform/gui.h
mehw: Thanks!

whitequark:
You did the opposite of what I asked.

whitequark:
Thanks for the bug report and the patch. Fixed in ce0b130.

mehw:
> You did the opposite of what I asked.

I see... I misinterpreted your suggestion...

> I think unless you put the `mnemonics` argument after `onTrigger` the default value for it is quite pointless.

I'm sorry, what I got was "do not use a default value for `mnemonics`"... So, I saw no reason to put `mnemonics` after `onTrigger`...

> Thanks for the bug report and the patch. Fixed in [ce0b130](https://github.com/solvespace/solvespace/commit/ce0b130d6ceb2145c8150eecbffe7380948899d4).

You're welcome. Thanks for fixing the master branch!

PR #394, opened 2019-04-16
Engraving

bcmpinc:
I mostly use solvespace for designing stuff that I then cut out using a laser cutter. For that purpose I always use the section export function. In the past I used Inkscape to add text or line segments, but I would prefer to be able to model these in solvespace, especially when positioning is important. 

For custom line styles, there is an option "export these objects", which I hoped meant that these lines are included in the section export, but they aren't. This patch changes that, such that they are included. To get it to work properly I had to change some parts of the code for which I'm not sure whether I got them correct.

I don't know what "export these objects" is supposed to do, so I guess this is not entirely the right place for configuring a style for section export and that a new flag might be more appropriate.

Please let me know what you think of it.

P.S. I thought that it would be useful to add a way to disable the closed contour check for these custom line styles, but this is already possible by turning them into construction line segments (although those are visually indistinguishable).

### DIFF
--- a/src/export.cpp
+++ b/src/export.cpp
@@ -78,9 +78,8 @@ void SolveSpaceUI::ExportSectionTo(const Platform::Path &filename) {
     g->runningMesh.MakeEdgesInPlaneInto(&el, n, d);
 
     // If there's a shell, then grab the edges and possibly Beziers.
-    g->runningShell.MakeSectionEdgesInto(n, d,
-       &el,
-       (SS.exportPwlCurves || fabs(SS.exportOffset) > LENGTH_EPS) ? NULL : &bl);
+    bool export_as_pwl = SS.exportPwlCurves || fabs(SS.exportOffset) > LENGTH_EPS;
+    g->runningShell.MakeSectionEdgesInto(n, d, &el, export_as_pwl ? NULL : &bl);
 
     // All of these are solid model edges, so use the appropriate style.
     SEdge *se;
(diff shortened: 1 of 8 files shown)


src/export.cpp
```diff
@@ -92,8 +91,27 @@ void SolveSpaceUI::ExportSectionTo(const Platform::Path &filename) {
         sb->auxA = Style::SOLID_EDGE;
     }
 
-    el.CullExtraneousEdges();
-    bl.CullIdenticalBeziers();
+    el.CullExtraneousEdges(true);
```
whitequark: Would be good to have `/*both=*/` here and elsewhere because it looks quite puzzling otherwise.

src/srf/surface.h
```diff
@@ -156,7 +156,7 @@ class SBezierLoopSet {
     static SBezierLoopSet From(SBezierList *spcl, SPolygon *poly,
                                double chordTol,
                                bool *allClosed, SEdge *errorAt,
-                               SBezierList *openContours);
+                               SBezierLoopSet *openContours);
```
whitequark: I don't quite understand why this is needed. Could you elaborate on it? That would also make a good commit message.

bcmpinc: Due to their different style, the engravings end up in the list of open contours during `FindOuterFacesFrom`, regardless of whether they form a closed contour or not. Considering that openContours is a list of segments, open loops have to be added per segment, thereby losing their connectivity. For example, an engraved circle will be exported as 4 separate arcs (at least for SVG). To fix this, I had to change the type of open contours to a loopset, which does retain connectivity.

src/srf/curve.cpp
```diff
@@ -705,12 +706,10 @@ void SBezierLoopSetSet::FindOuterFacesFrom(SBezierList *sbl, SPolygon *spxyz,
         if(loop->tag == USED_LOOP) continue;
 
         if(openContours) {
-            SBezier *sb;
-            for(sb = loop->l.First(); sb; sb = loop->l.NextAfter(sb)) {
-                openContours->l.Add(sb);
-            }
+            openContours->l.Add(loop);
+        } else {
+            loop->Clear();
```
bcmpinc: Here the open loops are exploded into separate segments.

src/srf/surface.h
whitequark: I'm still confused. Aren't open contours, by definition, not bezier loops?

bcmpinc: I guess. Though `SBezierLoop::FromCurves` returns open loops and there is `SBezierLoop::IsClosed()` to tell if a loop is closed. So I assumed that it's okay for loops to be open. And in every other aspect, that class does exactly what I need from it. If you want, I can rename `SBezierLoop` to `SBezierContour`, to make it more consistent.

whitequark: Oh, you're right. SBezierLoop doesn't have to be closed. That's OK then.

whitequark:
@bcmpinc Thank you for the PR, and sorry for the delay in merging.

PR #396, opened 2019-04-21
Add a lookup for ../share/solvespace

probonopd:
This is useful for relocatable bundles (e.g., AppImages).

See https://github.com/solvespace/solvespace/issues/347#issuecomment-485226547


whitequark:
Thanks!

Issue #81, opened 2016-10-14
"Fixed or Smaller (Bigger)" constraint for length and angle


Symbian9:
Add constraint that give a way set parametre as ≤ or ≥ to inserted value.

jwesthues:
In brief, SolveSpace's solver could be hacked to kind of support inequalities, but it was never made for that. Some earlier threads:

http://solvespace.com/forum.pl?action=viewthread&parent=211
http://solvespace.com/forum.pl?action=viewthread&parent=886

Symbian9:
Also, some time its needed prevent to change vector (for example, "line could be enlarged only from start point in left direction, and not in reverse")

Evil-Spirit:
@jwesthues, This actually was made and works fine. I even don't see any "hacks" in this solution, just the pure mathematics :)

damian0815:
Are they any plans to bring this into the core? I would love this feature.

whitequark:
No immediate plans, other tasks have a higher priority.

Symbian9:
> This actually was made and works fine. I even don't see any "hacks" in this solution, just the pure mathematics :)

@jwesthues & @Evil-Spirit, so if it work in test, why not implemented yet?

Evil-Spirit:
@jwesthues,
@whitequark,
@Symbian9,
Uff... I made it!
https://github.com/Evil-Spirit/solvespace-master/commit/6b7e99b068ac7daeef87d2aab808538fb1e38a4a

Evil-Spirit:
`case Op::NORM: return a->PartialWrt(p)->Times(a->Abs()->Sgn());`

> This actually was made and works fine. I even don't see any "hacks" in this solution, just the pure mathematics :)

> I even don't see any "hacks"

> "hacks"

Symbian9:
> Uff... I made it!

@Evil-Spirit, you are the man! Do not forget call @probonopd to this party too!

Harvie:
Can you please merge this? Looks very nice!

https://github.com/KmolYuan/python-solvespace/issues/3

whitequark:
Yeah, it's in my review queue. I did not forget about this patch though.

Harvie:
BTW if we could just enter the inequality or range like this, it would be smooth:

![image](https://user-images.githubusercontent.com/276504/58156770-2a0da300-7c77-11e9-88a3-09e06b6b0af0.png)

(it already supports some kind of basic math like subtraction, so this can be nice addition)

PR #399, opened 2019-05-08
Geometry

phkahler:
This creates a new group type "Revolve" similar to Lathe, but it allows sweeping a sketch through less than 360 degrees. The surface creation code also has some support for helical extrusions but that is not a supported feature yet.

### DIFF
--- a/src/graphicswin.cpp
+++ b/src/graphicswin.cpp
@@ -106,6 +106,7 @@ const MenuEntry Menu[] = {
 { 1, NULL,                              Command::NONE,             0,       KN, NULL   },
 { 1, N_("E&xtrude"),                    Command::GROUP_EXTRUDE,    S|'x',   KN, mGrp   },
 { 1, N_("&Lathe"),                      Command::GROUP_LATHE,      S|'l',   KN, mGrp   },
+{ 1, N_("Re&volve"),                     Command::GROUP_REVOLVE,    S|'l',   KN, mGrp   },
 { 1, NULL,                              Command::NONE,             0,       KN, NULL   },
 { 1, N_("Link / Assemble..."),          Command::GROUP_LINK,       S|'i',   KN, mGrp   },
 { 1, N_("Link Recent"),                 Command::GROUP_RECENT,     0,       KN, mGrp   },
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -532,20 +532,28 @@ void Group::Generate(IdList<Entity,hEntity> *entity,
 
             for(i = 0; i < entity->n; i++) {
                 Entity *e = &(entity->elem[i]);
+                if(e->group.v != opA.v) continue;
+
+                e->CalculateNumerical(/*forExport=*/false);
                 hEntity he = e->h;
+
+                CopyEntity(entity, SK.GetEntity(predef.origin), 0, ai,
+                    NO_PARAM, NO_PARAM, NO_PARAM,
+                    NO_PARAM, NO_PARAM, NO_PARAM, NO_PARAM,
+                    CopyAs::NUMERIC);
+
                 for(a = 0; a < 2; a++) {
                     if(e->group.v != opA.v) continue;
 
                     e->CalculateNumerical(false);
                     CopyEntity(entity, e,
                         a*2 - (subtype == Subtype::ONE_SIDED ? 0 : (n-1)),
-                        (a == (n - 1)) ? REMAP_LAST : a,
+                        (a == (n - 1)) ? REMAP_LATHE_END : REMAP_LATHE_START,
                         h.param(0), h.param(1), h.param(2),
                         h.param(3), h.param(4), h.param(5), h.param(6),
                         CopyAs::N_ROT_AA);
                 }
-// TODO: make arcs for Rotate groups
-//                MakeLatheCircles(entity, param, he, axis_pos, axis_dir, ai);
+                MakeLatheCircles(entity, param, he, axis_pos, axis_dir, ai);
                 MakeLatheSurfacesSelectable(entity, he, axis_dir);
                 ai++;
             }
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -203,6 +203,8 @@ void Group::MenuGroup(Command id, Platform::Path linkFile) {
             }
             g.type = Type::REVOLVE;
             g.opA = SS.GW.activeGroup;
+            g.valA = 2;
+            g.subtype = Subtype::ONE_SIDED;
             g.name = C_("group-name", "revolve");
             break;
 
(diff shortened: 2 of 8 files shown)


whitequark:
Ideally the permanent history of the project would not have any "fix" commits, because to debug problems, it is useful to use `git bisect`, and if each commit is self-contained, then one can bisect much easier. Do you think you could squash the commits to be more self-contained?

phkahler:
I suspect the conflict is caused by that fix I offered in a comment to another issue. I also fixed it in my branch, but you also fixed it in master.

whitequark:
> If you revert the change from April 22 this should be able to merge properly.

This really isn't the right way to solve this problem. Do you know how to use `git rebase`?

phkahler:
The problem (I think) is that I made that change in the same commit as some others, so my patch won't apply cleanly now. No, not familiar with rebase but I'll read about it.

whitequark:
Rebase allows you to apply your commits on top of some other point in the history, like current master. If you rebase your commits on top of master in this case, the changes that are already in master will be removed (if possible to do automatically) or you'll get a merge conflict that you can resolve to remove them.

Similarly, git's "interactive rebase" feature allows combining commits together, like I've asked you do in [here](https://github.com/solvespace/solvespace/pull/399#issuecomment-490304227). With some effort it lets you split commits as well, although that's a bit tricky.

phkahler:
@whitequark Please try this again soon. If there are any issues I don't want to leave thing in their present state for long. Thanks.

whitequark:
@phkahler Thanks, I will take a look at the earliest opportunity.

whitequark:
@phkahler Looks great. I've tried to torture it a bit and it seems to hold up nicely, with only a few NURBS failures around the usual pain points.

Have you thought about adding an inequality constraint (see #81 for implementation) that limits the range of the movement for the revolution? It seems that going over 360° tends to really degrade performance, so I'm worried about the sketch contorting itself into a state where you can't untangle it back easily.

Is it hard to add the remaining derived entities, i.e. arcs and faces?

phkahler:
> Have you thought about adding an inequality constraint

Once we add axial displacement for helical revolutions you may want to go way beyond 360 degrees. In its present form that just creates overlapping surfaces which are buggy and slow.

> Is it hard to add the remaining derived entities, i.e. arcs and faces?

Arcs I don't know. The code for Circles is still in there but commented out so we can see where that kind of thing would go. There are no helical entities either.

Faces - I'm still trying to figure out why they are not selectable. When I first made a helix from a rectangle I could select the top and bottom faces even though they were not flat, so I took that out. I thought it was back in, but it's not and I'm not sure why.

I wanted to do more with this, but it may be a while.

Nageki:
> > Have you thought about adding an inequality constraint
> 
> Once we add axial displacement for helical revolutions you may want to go way beyond 360 degrees. In its present form that just creates overlapping surfaces which are buggy and slow.

Pardon, but would n-dimensional generalizations be too hard, or am I completely missing the point? (building things gets exponentially easier as things go into higher dimensions, idk about computational complexity tho,)

whitequark:
> Once we add axial displacement for helical revolutions you may want to go way beyond 360 degrees. In its present form that just creates overlapping surfaces which are buggy and slow.

I was thinking about having a separate Revolve and Helix commands as you've suggested before. Former would have an inequality constraint, latter would not.

phkahler:
> I was thinking about having a separate Revolve and Helix commands as you've suggested before. Former would have an inequality constraint, latter would not.

These are details to be worked out. Maybe we can get some user feedback? Can you accept/close this pull request?

whitequark:
> Can you accept/close this pull request?

I can merge it if you rebase it so that each commit is independent and self-contained. If you are not able to do that then I will have to do it myself and I don't know when I will have time for it.

phkahler:
I did rebase it. The series of commits shows the progressive implementation of the feature. Each step can be run. If you want to flatten it go ahead, but I prefer to be able to read the commit messages and have finer grained steps so it's easier to see what changes added to or fixed each thing.

phkahler:
Something is wrong. I did the rebase on another machine, so now I re-cloned my github back home and it won't build and neither will the master branch - I also pulled from solvespace master.  I'm getting:

fatal error: gtkmm/filechoosernative.h: No such file or directory
    include <gtkmm/filechoosernative.h>

I rebased on 5/11 but now I should locally be up to date while my github repo remains as it was after rebase. BTW I'm on Fedora if that makes any difference.

phkahler:
@whitequark 
I just checked and both Master and Geometry seem to build properly if I comment out:
`
define HAVE_GTK_FILECHOOSERNATIVE
`
In file guigtk.cpp which which was recently changed (in 838126f81f9feb5cf16d133693055493df30f0fe). I can't tell what changed with the file chooser, I can still open and save when it builds. This breakage seems totally independent of this PR. Please let me know how I/you/we should handle this.

whitequark:
@phkahler No action is needed from you wrt file chooser in this pull request as it is completely independent. That said, I would be curious as to why it happens--have you tried removing the entire build directory and trying again? I did change the build system a bit in the meantime, so it might be that an old value is cached in config.h somehow. Also, what exactly is your GTK/gtkmm versions?

Regarding this PR I will be able to give it attention in a day or two.

phkahler:
@whitequark  My installed and devel versions of GTK may have gotten out of sync (one was 22 and one 24?). I Upgraded the whole system to Fedora 30 and both branches (master and geometry) build fine now. I must have accepted some udpates around the time of the rebase. Sorry for the panic.

whitequark:
> My installed and devel versions of GTK may have gotten out of sync (one was 22 and one 24?).

Yes, that would do it. CMake caches certain version information and it has no good way to be notified of updates, so I would expect this.

whitequark:
@phkahler Sorry, had not much time this week and I have some other obligations I need to do soon. Next week hopefully.

phkahler:
FYI  I tried a change where going over 360 degrees would create the same shell as lathe. It prevents the self intersecting and performance issues, but does not keep you from winding up the ends. A ranged constraint seems like a better idea.

whitequark:
Thanks for the recent fixes! I have some time to address this now, so let me know when you're done and I'll merge it then.

phkahler:
I've disabled the arcs on Revolve. The fix is not going to be easy and may require creating a new entity type with a different definition - it's a chirality issue.

whitequark:
@phkahler Thank you for this wonderful contribution, and sorry again for the delay. I've merged your commits as 5df53fc5, with one minor semantic change: I fixed the hotkey for Group > Revolve. I've also ran clang-format on the diff to match the project's existing code style.

PR #403, opened 2019-05-13
GTK: fix build on Ubuntu 18.04 by removing debug define

zhuowei:
Ubuntu 18.04 uses GTKMM 3.22.2-2, which doesn't support native file chooser.

Commit bc3e09edbfeb4859218bf51476e34635f45cf249 checks if native file chooser
is available, but the result is overridden with a hardcoded define,
probably for debugging.

Removing the debugging code fixes build on Ubuntu 18.04.

### DIFF
--- a/src/platform/guigtk.cpp
+++ b/src/platform/guigtk.cpp
@@ -15,7 +15,6 @@
 #include <gtkmm/cssprovider.h>
 #include <gtkmm/entry.h>
 #include <gtkmm/filechooserdialog.h>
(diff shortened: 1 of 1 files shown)


whitequark:
Whoops. I was thinking when committing that "chances are I will mess it up somewhere but someone will report that quickly so it's OK". Well, that's exactly what happened :)

zhuowei:
@whitequark Thanks!

PR #404, opened 2019-05-13
macOS Mojave: fix hang on launch

troughton:
On Mojave, [NSApp menuBar] doesn't exist until after applicationDidFinishLaunching, causing `nsMenuBar` to be nil (and `Clear()` to never exit since `numberOfItems` is 0). Rather than restructuring the lifecycle of the app, just create our own menu bar and set it as the main menu.

### DIFF
--- a/src/platform/guimac.mm
+++ b/src/platform/guimac.mm
@@ -306,7 +306,8 @@ public:
     std::vector<std::shared_ptr<MenuImplCocoa>> subMenus;
 
     MenuBarImplCocoa() {
-        nsMenuBar = [NSApp mainMenu];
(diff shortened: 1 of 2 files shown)


whitequark:
The reason it was using the existing main menu is that it's the one created from [this nib](https://github.com/solvespace/solvespace/blob/master/res/cocoa/MainMenu.xib) which IIRC was required for the default items in the system menu to work properly. E.g. that's where the `Preferences…` item lives. Does that still work?

troughton:
Okay, it does still work using the .nib if the Xcode project is set up properly and using the `make`-based build system – it's only the generated Xcode project that has issues. I'm going to close this, but the `while([nsMenuBar numberOfItems] > 1)` change might be worthwhile on its own in case others run into the same issue.

PR #410, opened 2019-05-15
Fix tests by correcting detection of build path sep.

rpavlik:
Before, would guess incorrectly if the CMake source tree was specified
via a relative path (since then the path would not start with /).
Now, checks for any backslashes, using them as an indicator that
the backslash is the separator.

Without this, I can't get the test suite to run on Linux, and I'm pretty sure I couldn't get it to run on Windows before.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -7,6 +7,10 @@ foreach(pkg_config_lib CAIRO)
     link_directories(${${pkg_config_lib}_LIBRARY_DIRS})
 endforeach()
 
+if(${CMAKE_HOST_SYSTEM_NAME} STREQUAL "Windows")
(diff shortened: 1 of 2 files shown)


whitequark:
This condition should be inverted. Backslash is legal as a separator in Linux (though misguided) but forward slash is not legal as a separator on Windows.

rpavlik:
Hmm. It's only forbidden on Windows in some APIs - I regularly use forward slash.

Perhaps the better way would be setting a define from CMake if the build system is win32, and completely dodge trying to do it at runtime.

whitequark:
Yeah, that would be nicer. No idea why I went with `__FILE__` in first place.

rpavlik:
ok, updated accordingly.

whitequark:
Thanks!

PR #413, opened 2019-05-20
Hopefully-Uncontroversial cleanups

rpavlik:
Went through my old branches and pulled out the commits that look reasonably safe and innocuous. (My switch to use std::vector inexplicably is causing test failures, and my performance improvement to bulk-add to non-vector collections is not very pretty.)

All tests pass on Debian Buster. Feel free to cherry-pick if some of them don't appeal to you - they're mostly independent or in brief sequences (e.g. the CountIf and GetNumConstraints changes are related, the WRAP changes are related, the rest are pretty independent though I don't know what would happen if you re-ordered the "convert to range-for" commits.)

### DIFF
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -424,6 +424,7 @@ void Group::Generate(IdList<Entity,hEntity> *entity,
             // Get some arbitrary point in the sketch, that will be used
             // as a reference when defining top and bottom faces.
             hEntity pt = { 0 };
+            // Not using range-for here because we're changing the size of entity in the loop.
             for(i = 0; i < entity->n; i++) {
                 Entity *e = &(entity->elem[i]);
                 if(e->group.v != opA.v) continue;
@@ -457,6 +458,7 @@ void Group::Generate(IdList<Entity,hEntity> *entity,
             // Remapped entity index.
             int ai = 1;
 
+            // Not using range-for here because we're changing the size of entity in the loop.
             for(i = 0; i < entity->n; i++) {
                 Entity *e = &(entity->elem[i]);
                 if(e->group.v != opA.v) continue;
@@ -502,6 +504,7 @@ void Group::Generate(IdList<Entity,hEntity> *entity,
             }
 
             for(a = a0; a < n; a++) {
+                // Not using range-for here because we're changing the size of entity in the loop.
                 for(i = 0; i < entity->n; i++) {
                     Entity *e = &(entity->elem[i]);
                     if(e->group.v != opA.v) continue;
@@ -537,6 +540,7 @@ void Group::Generate(IdList<Entity,hEntity> *entity,
             }
 
             for(a = a0; a < n; a++) {
+                // Not using range-for here because we're changing the size of entity in the loop.
                 for(i = 0; i < entity->n; i++) {
                     Entity *e = &(entity->elem[i]);
                     if(e->group.v != opA.v) continue;
@@ -563,6 +567,7 @@ void Group::Generate(IdList<Entity,hEntity> *entity,
             AddParam(param, h.param(5), 0);
             AddParam(param, h.param(6), 0);
 
+            // Not using range-for here because we're changing the size of entity in the loop.
             for(i = 0; i < impEntity.n; i++) {
                 Entity *ie = &(impEntity.elem[i]);
                 CopyEntity(entity, ie, 0, 0,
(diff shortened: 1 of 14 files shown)


rpavlik:
updated to fix a style issue.

src/group.cpp
```diff
@@ -425,6 +425,7 @@ void Group::Generate(IdList<Entity,hEntity> *entity,
             // Get some arbitrary point in the sketch, that will be used
             // as a reference when defining top and bottom faces.
             hEntity pt = { 0 };
+            // Not using range-for here because we're changing the size of entity in the loop.
```
whitequark: Oh *of course* that's why my own range-for patch (in a local branch that got buried years ago) didn't work. Iterator invalidation, the second worst thing in C++ that "modern" C++ does absolutely nothing to help with...

src/undoredo.cpp
```diff
@@ -48,6 +48,7 @@ void SolveSpaceUI::PushFromCurrentOnto(UndoStack *uk) {
     UndoState *ut = &(uk->d[uk->write]);
     *ut = {};
     ut->group.ReserveMore(SK.group.n);
+    // Not using range-for in most of the below because it's unclear if sizes are changed.
```
whitequark: The size of anything in `SK` is not changed.

src/group.cpp
whitequark: Can we add iterators that don't get invalidated when inserting into the collection, maybe? Store an index as well as pointer to the element.

whitequark:
I've merged about half of the commits as f885daf7, 43c9cba7 and 39c3480.

I'd like to merge the range-for stuff but I think we need to do something about iterator invalidation while adding to the collection.

src/mesh.cpp
```diff
@@ -580,16 +580,20 @@ void SKdNode::SnapToVertex(Vector v, SMesh *extras) {
             bool mightHit = true;
 
             for(k = 0; k < 3; k++) {
-                if((tr->a).Element(k) < v.Element(k) - KDTREE_EPS &&
-                   (tr->b).Element(k) < v.Element(k) - KDTREE_EPS &&
-                   (tr->c).Element(k) < v.Element(k) - KDTREE_EPS)
+                const auto trA = (tr->a).Element(k);
```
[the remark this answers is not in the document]
rpavlik: No, probably not. Not sure why I did that 1.5 yrs ago :) Think I was in "auto all the things, and const everything you can" mode

src/group.cpp
rpavlik: maybe as "future work". I did have some patches that make an IdListAdditions type, which basically queues up things to add so it can be bulk-added at the end via a merge. Would also avoid invalidating iterators if iterating an idlist, but not sure the clarity tradeoff is worth it. (I reverted the patch on my dev branch after doing the port to std::vector as a backend, since that gave the equivalent performance improvement. Think it was allocation behavior that was gettting me.)

rpavlik:
Cool, thanks.  Will rebase with the changes suggested.

src/undoredo.cpp
rpavlik: Updated in the new push.

rpavlik:
OK, rebased the remaining changes, tried to make my commit messages more like yours and combine some commits.

src/group.cpp
rpavlik: Here's that branch rebased on this, and trimmed. https://github.com/rpavlik/solvespace/commits/idlist-additions

See in particular this: https://github.com/rpavlik/solvespace/commit/93f5e410e586281e641f52056d409de2dd81263d  I only did this in the few places where I saw it coming up a lot on profiling data, but I think this could be used to avoid iterator invalidation in loops as well - instead of adding directly, add to an "additions" object, then "MoveIntoList" after iteration is complete.

src/solvespace.h
```diff
@@ -92,10 +92,11 @@ __attribute__((__format__ (__printf__, 1, 2)))
 #endif
 std::string ssprintf(const char *fmt, ...);
 
-inline int WRAP(int v, int n) {
+template<typename T>
+inline T WRAP(T v, T n) {
```
whitequark: (I think you missed my earlier question.) I don't understand this change--what's the point of making WRAP generic if it's only used with ints anyway *and* you need to cast it each time?

whitequark:
Merged f7f0a93, adding a similar assert for SMesh as well.

src/solvespace.h
rpavlik: Ah, I did miss your question. To be honest, I have no idea, that's the risk of me leaving commits un-submitted for so long.

rpavlik:
Closing since most of this got in

whitequark:
I actually very much want for loops, that will enable some IdList optimizations, see #430.

whitequark:
Ah, you've opened #428. Sorry for the noise.

Issue #414, opened 2019-05-20
clang-format config?


rpavlik:
In working on the code, I tend to accidentally not follow the style guide because it's a little unique (mostly the "no spaces in `if(condition)`, etc" item)  If this style is still the desired one and important, would be great if someone could make a clang-format (or other formatter) config file that enforced it.

whitequark:
I generally agree, but I'm not sure how clang-format deals with partial reformats. As been discussed before, bulk formatting changes are a no-go, and if clang-format can only do one file at a time its usefulness seems limited here.

rpavlik:
I tend to use git clang-format to just reformat the changes currently staged, for projects where I can't format the whole thing.

whitequark:
Whoa, I had no idea that existed. I'm completely in favor of having a `.clang-format` file then and suggesting its use in guidelines. I'm not sure when I will get around it, so if you feel like making one, go ahead please.

PR #416, opened 2019-05-20
Clang format

rpavlik:
Follow-up from #414 - this is my best attempt at getting one using my typical approach of populating with the explicitly-stated preferences, then iterating through "apply to file" and "correct config causing most apparent spurious changes". There are substantial inconsistencies that mean that whatever decision I make for some of these, it will introduce changes if it were applied all-over, but since that's not on the table anyway, this should be at least something. There were also a few where clang-format couldn't quite express the subtle distinctions I saw in the existing formatting, so I picked the overall less-intrusive one.

The file is commented to help you modify it as desired.

I also added a clang-tidy config file because the contributing doc says "braces around all the things" and that's not something clang-format can do (for some reason), you have to use clang-tidy. Doing so exposed me to the reality that it's a guideline very rarely followed - but if you'd like to enforce it, that clang-tidy file is a good way to do so. Something like this, assuming you have a build tree in `build/`...

```sh
find src test \(-name "*.h" -o -name "*.cpp" \) -print0 | xargs --null clang-tidy-7 -p build -fix 
git add -u
git clang-format
```

(Can't use run-clang-tidy here because it will result in things getting multiple sets of brackets, etc. Sadly must do it single-threaded like this, or have it save out the fixes and manually choose the ones to apply.)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,43 @@
+---
+# Things explicitly enumerated in CONTRIBUTING.md
+IndentWidth: 4
+UseTab: Never
+SpaceBeforeParens: Never
+BreakBeforeBraces: Attach
+ColumnLimit: 100
+# Docs say 100, but some places look more like 80.
+#ColumnLimit: 80
+
+# Based on minimizing diff when applying clang-format
+AccessModifierOffset: -4
+DerivePointerAlignment: true
+AlignConsecutiveAssignments: true
+FixNamespaceComments: true
+NamespaceIndentation: Inner
+AllowShortLoopsOnASingleLine: true
+AlwaysBreakTemplateDeclarations: Yes # MultiLine
+SpaceAfterTemplateKeyword: false
+MaxEmptyLinesToKeep: 2
+IndentPPDirectives: AfterHash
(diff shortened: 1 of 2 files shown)


whitequark:
> I also added a clang-tidy config file because the contributing doc says "braces around all the things" and that's not something clang-format can do (for some reason), you have to use clang-tidy. Doing so exposed me to the reality that it's a guideline very rarely followed - but if you'd like to enforce it, that clang-tidy file is a good way to do so.

Thanks for investigating this. I don't think it's a very important guideline, and I appear to have significantly overestimated how often that's the case when writing CONTRIBUTING.md. So, let's just drop it instead, as if I understand you correctly, that means we don't really need clang-tidy anymore without it (is that right?)

rpavlik:
Well, certainly not to enforce code style. It's a decent idea to run from time to time as a linter.

whitequark:
I pushed the clang-format change as ffef006 and updated CONTRIBUTING.md, thank you very much. I might look into clang-tidy some time later but for now let's omit it.

rpavlik:
Thanks!

PR #420, opened 2019-05-21
Automatically resize the Three.js view to fit window if the size was not forced

nabijaczleweli:
Resizing the browser window had no effect, leading to requiring a reload. With this patch the view size follows the browser's, if not overridden in `params`.

I arrived at this code by blind-hacking at the export, so am not sure if all is up to standard (the `render()` call in particular).


whitequark:
This looks good but incomplete, because while resizing recenters the model, it doesn't adjust its scale. Do you think you could try to handle that as well?

nabijaczleweli:
Not sure what the state-of-the-art correct way of handling this is, but code from new commit behaves pretty much like I'd expect scaling on resize to behave?

whitequark:
Thank you, merged as e243396.

PR #428, opened 2019-05-23
More cleanups and simplification

rpavlik:
This mostly cleans up the remaining pieces I had in flight, plus some extra cleanups. tests pass, the containers aren't very different (still allocated the same), but this lets the "port" to backing by std::vector be fairly simple in addition to making things cleaner and more uniform.

PR #429, opened 2019-05-23
Port idlist

rpavlik:
Builds on #428 - replaces the manual memory management with something less manual. Had to go with a shared_ptr<vector<T>> instead of just a vector<T> because there's a lot of code that relies on assignment/copy being shallow. The tests pass here. (Unlike its related branch where I also ported List<T>, that fails tests by being off by a few pixels. https://github.com/rpavlik/solvespace/commits/port-both )

### DIFF
--- a/src/dsc.h
+++ b/src/dsc.h
@@ -337,21 +337,64 @@ struct CompareId {
 // id.
 template <class T, class H>
 class IdList {
-    T *elem            = nullptr;
-    int elemsAllocated = 0;
+    std::shared_ptr<std::vector<T>> vec;
+    int internalN = 0;
+    void UpdateSize() {
+        if(!vec) {
+            internalN = 0;
+        } else {
+            internalN = static_cast<int>(vec->size());
+        }
+    }
+    void AllocForOneMore() {
+        if(!vec) {
+            vec = std::make_shared<std::vector<T>>();
+        }
+        if(vec->capacity() == vec->size()) {
+            // Let the built-in growth policy deal with this.
+            vec->reserve(vec->size() + 1);
+        }
+    }
+
 public:
-    int n = 0;
+    std::reference_wrapper<const int> n = std::cref(internalN);
 
     using Compare = CompareId<T, H>;
 
-    bool IsEmpty() const {
-        return n == 0;
+    bool IsEmpty() const { return !vec || vec->empty(); }
+
+    IdList() : n(std::cref(internalN)) {}
+
+    // Makes a shallow copy that shares the underlying list!
+    IdList(IdList const &other) : vec(other.vec), internalN(other.internalN), n(std::cref(internalN)) {}
+
+    // Assigns a shallow copy that shares the underlying list!
+    IdList &operator=(IdList const &other) {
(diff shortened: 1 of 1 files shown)


rpavlik:
ah, the windows build env. in use doesn't support defaulting move constructors.

whitequark:
Is this obsoleted by #433?

rpavlik:
no, #433 (ideally) builds on this. However, at least in my rebased version of this branch, neither one succeeds at the tests - this branch crashes, while the other fails tests and crashes.

rpavlik:
OK, when I remove the merge-idlists patch from this branch, I do pass all the tests and can open something with a lathe object, though I get a double-free on shutdown.

This is probably the best shot so far at such a port, but I'm not sure the value of it at this point since things break when I poke at stuff with varied ownership semantics.

whitequark:
yeah it's unfortunately not easy at all

Evil-Spirit:
@rpavlik Have you tested performance? There will probably be significant decrease.

rpavlik:
Why would it be expected to be slower? (Is MemAlloc a pool? I can't remember. If so, that makes sense.)

I have no great allegiance here: if we somehow manage to avoid leaking/double-freeing/use-after-freeing memory with the current code, it's fine. Raw pointers owning memory just makes me anxious so I did something about it once. (and again, and again, when rebasing and testing ;) )

Evil-Spirit:
@rpavlik I am not sure, but it always worth testing before such significant changes. I said so, because almost every change of original containers by @jwesthues since 2.0 lead to performance decrease. std::things is not a very good at performance because should work everywhere and support all the data types. You can just look into id-list-indexing branch, the work I made on improvement of performace which is really significant, if you will test on really big examples like houses [modelisation.slvs](https://github.com/whitequark/solvespace-testcases/blob/master/modelisation.slvs)

ruevs:
@phkahler @rpavlik @pjanx Hi guys - this is the old attempt by @rpavlik mentioned here https://github.com/solvespace/solvespace/issues/939#issuecomment-811141395 . I think after either https://github.com/solvespace/solvespace/pull/1002 or https://github.com/solvespace/solvespace/pull/992 is merged we can safely close this. What do you think?

Discussions are in https://github.com/solvespace/solvespace/issues/939 and https://github.com/solvespace/solvespace/issues/932 .

pjanx:
The idea could be revisited after the high-impact optimisations are in. But as it is, this PR isn't getting merged anytime soon, especially if it's unstable. There seems to be a battle of approaches in Solvespace, which originally was much more C-like.

ruevs:
> 
> There seems to be a battle of approaches in Solvespace, which originally was much more C-like.

The project is from 2008 when Jonathan Westhues started it based on his [SkethFlat](https://cq.cx/sketchflat.pl) and up to 2.0 in 2013 it was not even open source. And it was pretty much only him until 2015 when Whitequark and EvilSpirit joined.
C++11 did not exist in the beginning ;-)

The code is updated/modernized from time to time - usually hen there is another reason to touch it - as with the IdList in our case.

Issue #386, opened 2019-03-03
System::CalculateRank() is slow


bmorel:
### System information

SolveSpace version: master, 9d1601eea9039011c16c435b26a680fb4bc35c34
Operating system: Debian stable

In a file I'm working on (I'll attach it tomorrow when I'll have more proofs), adding a 'same length' constraint between the top of the 1st square of 1st line and the top of the 1st square of 2nd line is extremely slow (on my low end computer).

Using LD_PRELOAD, I've nailed the problem to the method in the title, and implemented some optimisations.
I don't have enough tests to have reliable numbers, but the number of samples gperftools took moved from 47.3% of 3009 to 36.8% of 827 (those numbers make it obvious to me that I need more numbers).

I am opening this ticket in advance to ask for hints about how to automate the runs, which would provide me more reliable numbers, and to provide a 1st patch of the optimized code I have so far (too lazy to do a usual fork/clone/create branch/commit/merge/push/pull request, for just 10 lines of code):

```C++
diff --git a/src/system.cpp b/src/system.cpp
index e90611a..24d9ebe 100644
--- a/src/system.cpp
+++ b/src/system.cpp
@@ -146,22 +146,31 @@ int System::CalculateRank() {
     for(i = 0; i < mat.m; i++) {
         // Subtract off this row's component in the direction of any
         // previous rows
+        double* curr_mat_num;
+        double* prev_mat_num;
+        const auto num_mat = mat.n;
         for(iprev = 0; iprev < i; iprev++) {
             if(rowMag[iprev] <= tol) continue; // ignore zero rows
 
             double dot = 0;
-            for(j = 0; j < mat.n; j++) {
-                dot += (mat.A.num[iprev][j]) * (mat.A.num[i][j]);
+            prev_mat_num = mat.A.num[iprev];
+            curr_mat_num = mat.A.num[i];
+            for(j = 0; j < num_mat; ++j, ++curr_mat_num, ++prev_mat_num) {
+                dot += (*prev_mat_num) * (*curr_mat_num);
             }
-            for(j = 0; j < mat.n; j++) {
-                mat.A.num[i][j] -= (dot/rowMag[iprev])*mat.A.num[iprev][j];
+            double tmp = dot/rowMag[iprev];
+            prev_mat_num = mat.A.num[iprev];
+            curr_mat_num = mat.A.num[i];
+            for(j = 0; j < num_mat; ++j, ++curr_mat_num, ++prev_mat_num) {
[...]

Issue #225, opened 2017-04-05
Solve equation systems with Eigen to improve convergence


whitequark:
See https://drive.google.com/file/d/0B15RHGNY8nBpYnNwcG82WDdseDQ/view. The newly added arc segment has a pt-coincident constraint that causes the failure.

Evil-Spirit:
@whitequark 
I see allowed redundant constraints here, so I've found what 12.7 pt-pt-distances appied twice like as two pt-on-circle. If we are delete all the redundant, we can converge here fine.

Evil-Spirit:
@whitequark, Somehow too many redundant constraints can't be converged. I've figured out what 3xPT_ON_CIRCLE can't converge (just try to make it on empty sketch). After disallowing redundant, too many redundants threated as non-converged for no reason. This is probably bug, but I am not sure whether rank test can work fine or not for such amount of redundants.

whitequark:
@Evil-Spirit Yes, I did notice that many redundant constraints tend to degrade convergence. Please figure out why. This makes such a mode much less useful.

Evil-Spirit:
@whitequark, I have researched a lot and made some debug functions (visualizing solve way and equations). Finally I've linked Eigen and this solves the problem. I think the problem is what the Gauss linear system solving is not so precise as it should be.

Evil-Spirit:
here is some beta
https://github.com/Evil-Spirit/solvespace-master/commit/703bccde0970ee39803d5af8bed4d04539d4820d

whitequark:
1. What is "solve way"?
2. Are there any downsides to using Eigen?

Evil-Spirit:
@whitequark, this is path of points during newton steps like here https://youtu.be/72Dnqvn_uTk

whitequark:
OK. Please prepare a commit that just adds Eigen, and separately these debugging capabilities.

Evil-Spirit:
2. I can try to work with Gauss Method to find any ways to imrove... But I think it is better to move in direction of Eigen integration because of better stability and convergence and a lots of methods to choose.

whitequark:
@Evil-Spirit I see no problem with Eigen. Just want to know if there is anything it's bad at.

Evil-Spirit:
@whitequark 
here is eigen https://github.com/Evil-Spirit/solvespace-master/commits/eigen
here is debug https://github.com/Evil-Spirit/solvespace-master/commits/solver-debug

[...]

PR #432, opened 2019-05-23
Some additional Eigen porting

rpavlik:
Just putting this out there instead of letting it fester, totally understand if you aren't interested in it. I originally made this to attack some of the biggest time-sinks in profiling some pathological cases. In my testing now, it is about equal to 10% faster than master when running the test suite in relwithdebinfo mode. The downside of the Eigen magic is that a pure debug build is slower (4 seconds instead of 3).

There's more use of Eigen stuff that could be done, this was just what I had. Let me know if there's any interest so I can stop forward-porting it if it's not useful :D  (I haven't had to use those pathological files in a while anyway, so the itch is not as strong.)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -115,6 +115,7 @@ add_library(solvespace-core STATIC
     sketch.h
     solvespace.h
     ui.h
+    utileigen.h
     platform/platform.h
     render/render.h
     render/gl3shader.h
--- a/src/polygon.cpp
+++ b/src/polygon.cpp
@@ -4,6 +4,7 @@
 // Copyright 2008-2013 Jonathan Westhues.
 //-----------------------------------------------------------------------------
 #include "solvespace.h"
+#include "utileigen.h"
 
 Vector STriangle::Normal() const {
     Vector ab = b.Minus(a), bc = c.Minus(b);
@@ -11,9 +12,9 @@ Vector STriangle::Normal() const {
 }
 
 double STriangle::MinAltitude() const {
-    double altA = a.DistanceToLine(b, c.Minus(b)),
-           altB = b.DistanceToLine(c, a.Minus(c)),
-           altC = c.DistanceToLine(a, b.Minus(a));
+    double altA = DistanceToLineFromEndpoints(a, b, c),
+           altB = DistanceToLineFromEndpoints(b, c, a),
+           altC = DistanceToLineFromEndpoints(c, a, b);
 
     return min(altA, min(altB, altC));
 }
--- a/src/solvespace.h
+++ b/src/solvespace.h
@@ -246,7 +246,7 @@ public:
         int m, n;
         struct {
             // This only observes the Expr - does not own them!
-            Eigen::SparseMatrix<Expr *> sym;
+            std::vector<Eigen::Triplet<Expr *>> sym;
             Eigen::SparseMatrix<double> num;
         } A;
 
--- a/src/srf/ratpoly.cpp
+++ b/src/srf/ratpoly.cpp
@@ -6,6 +6,7 @@
 // Copyright 2008-2013 Jonathan Westhues.
 //-----------------------------------------------------------------------------
 #include "../solvespace.h"
+#include "../utileigen.h"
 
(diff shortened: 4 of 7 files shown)


whitequark:
Have you seen @Evil-Spirit's Eigen port in #225? I was planning to merge that today, actually...

whitequark:
Regardless of anything else, you really shouldn't drop the entire source of Eigen right into this repository. Use https://github.com/eigenteam/eigen-git-mirror as a submodule.

rpavlik:
oops, no, I hadn't...  Yeah, my bad on the dropping the source in. Not sure why I did that last year. When I went through and fixed it, it was just easier to drop a new one instead of adding the submodule.

rpavlik:
Rebased this (along with fixing the extlib thing, now using a submodule)

phkahler:
@rpavlik  could you have a look at PR #721 ?  I know he did a more extensive update and had lengthy discussion with JW about it, but adding new dependencies without breaking CI is currently way outside my territory.

rpavlik:
So I shall look into this eventually, but I definitely wouldn't block 3.0 on it.

phkahler:
@rpavlik  No problem, 3.0 is imminent. Soon after that I'll probably try to merge the evil spirit solver improvements that don't involve Eigen and then there will be plenty of time for the rest. BTW I wanted to email you but can't find contact info. Feel free to ping me - same ID at g mail.

rpavlik:
OK, updated this.

vespakoen:
I honestly don't know enough about the solver to give any helpful feedback.
All I know is that I also run into the limitations of the solver quite quickly (even on a M1 Pro) and for that reason think it is a great idea to bring in Eigen and it's benefits.

I am hoping this can be combined with Evil-Spirit's changes, but am afraid this will make the merge of those changes harder? (I didn't try or look at it extensively though)

I'll probably give a stab at combining the 2 Eigen PR's soon.

ruevs:
> I'll probably give a stab at combining the 2 Eigen PR's soon.

@vespakoen don't.

@rpavlik did already! https://github.com/solvespace/solvespace/issues/386#issuecomment-999720840

Better review his.

rpavlik:
OK, I have updated this to contain the un-merged commits. These are all pretty simple replacements, I think most of these profiled as a "problem" back in the day when I did them.

rpavlik:
OK, so using the benchmark tool (nice! I don't remember that being there previously) and a "normal" model for me, this reduces the time to load from 0.185s to 0.165s. We did def. pick up a regression, probably in that openmp jacobian, since previous master here gave me 0.161, but I'm bisecting now.

rpavlik:
ok, now I can't repro the 0.185. Trying a release openmp build now...

rpavlik:
Yeah I should just not do performance testing on a laptop on battery I guess.

Rebased on top of #1183 to avoid that regression.

Anyway, I think these are at least performance neutral compared to #1183 as far as I can tell, possibly a performance improvement. They were a performance improvement back in pre-3.0, but even if they're just a readability improvement: Eigen allows pretty clear, concise expressions which is lovely.

phkahler:
@rpavlik  a while back, @ruevs made changes to our vector class adding C++ 11 style return { ... } to the methods. That was a solid win in performance.  I'm not in favor of adding Eigen functions in place of ours unless there is a clear advantage. OTOH the faster this gets, the more significant each improvement gets!  So I guess I'm saying to retest for performance wins when adding Eigen functions.

rpavlik:
I don't have data for these anymore, but they made a visible interactive improvement while editing as well as a profiling improvement back when I wrote these in 2016. Since then I've changed jobs, had another kid, done many other things, and rebased this countless times, so I no longer have that data. But I clearly remember both the min altitude and tangents at being big bottlenecks at that time. I imagine I was using tangent arcs (fillets) which were not used in my test model yesterday because the whole thing was round and because I basically learned they don't work well in stock Solvespace. If you don't want to merge these that's fine, but please close this pr then so I stop spending time rebasing it and trying to submit it for review and integration, I'll just carry it on a local tree.

phkahler:
I'll close it then. I believe there is more improvement in ratpoly to be had without introducing a new API.

PR #436, opened 2019-05-31
Linux: add support for building snap packages

ppd1990:
This is a "copy & paste" integration of the solvespace snap repo @ https://github.com/ppd1990/solvespace-snap as discussed in https://github.com/ppd1990/solvespace-snap/issues/2

## A few things left to discuss

1. ~~The integration resides in ```$ROOT/snap```. This is what build.snapcraft.io requires. If using the Ubuntu provided build server (and thus easy auto-build on commit) is not wanted, we could move it to mirror the flatpak integration~~

2. ~~The svg icon should probably find its place in ```res/freedesktop``` or we could just use a png icon. At the moment snap packages need an absolute path to an icon.~~
The svg icon was proposed for inclusion in #441 

3. ~~The snap is declared ```grade: stable```. Maybe that should only apply to tagged versions. Some extra logic would be required~~

## The next step

Publishing of the snap should be done under the flag of the upstream project. Therefore I propose a name transfer to a newly created "SolveSpace" publisher. This has to be requested on forum.snapcraft.io.

## Testing the build process
```
# optional for a faster build (adjust values for virtual build machine)
export SNAPCRAFT_BUILD_ENVIRONMENT_MEMORY=4G
export SNAPCRAFT_BUILD_ENVIRONMENT_CPU=6

snapcraft
...
snap install --dangerous solvespace_3.0~9ac55f39_amd64.snap 
```

### DIFF
--- a/pkg/snap/snap/snapcraft.yaml
+++ b/pkg/snap/snap/snapcraft.yaml
@@ -12,7 +12,6 @@ description: |
   * mechanism design — use the constraint solver to simulate planar or spatial linkages, with pin, ball, or slide joints
   * plane and solid geometry — replace hand-solved trigonometry and spreadsheets with a live dimensioned drawing
 
-grade: stable
 confinement: strict
 
 layout:
@@ -40,6 +39,8 @@ parts:
       snapcraftctl pull
       version="3.0~$(git rev-parse --short=8 HEAD)"
       snapcraftctl set-version "$version"
+      git describe --exact-match HEAD && grade="stable" || grade="devel"
+      snapcraftctl set-grade "$grade"
       git submodule update --init extlib/libdxfrw extlib/flatbuffers extlib/q3d
     configflags:
       - -DCMAKE_INSTALL_PREFIX=/usr
--- a/pkg/snap/snap/snapcraft.yaml
+++ b/pkg/snap/snap/snapcraft.yaml
@@ -38,7 +38,9 @@ parts:
     source-type: local
     override-pull: |
       snapcraftctl pull
-      version="3.0~$(git rev-parse --short=8 HEAD)"
+      version_major=$(grep "solvespace_VERSION_MAJOR" CMakeLists.txt | tr -d "()" | cut -d" " -f2)
+      version_minor=$(grep "solvespace_VERSION_MINOR" CMakeLists.txt | tr -d "()" | cut -d" " -f2)
+      version="$version_major.$version_minor~$(git rev-parse --short=8 HEAD)"
       snapcraftctl set-version "$version"
       git describe --exact-match HEAD && grade="stable" || grade="devel"
       snapcraftctl set-grade "$grade"
--- a/pkg/snap/snap/snapcraft.yaml
+++ b/pkg/snap/snap/snapcraft.yaml
@@ -1,0 +1,72 @@
+name: solvespace
+base: core18
+summary: Parametric 2d/3d CAD
+adopt-info: solvespace
+description: |
+  SOLVESPACE is a free (GPLv3) parametric 3d CAD tool.
+  Applications include
+  * modeling 3d parts — draw with extrudes, revolves, and Boolean (union / difference) operations
+  * modeling 2d parts — draw the part as a single section, and export DXF, PDF, SVG; use 3d assembly to verify fit
+  * 3d-printed parts — export the STL or other triangle mesh expected by most 3d printers
+  * preparing CAM data — export 2d vector art for a waterjet machine or laser cutter; or generate STEP or STL, for import into third-party CAM software for machining
+  * mechanism design — use the constraint solver to simulate planar or spatial linkages, with pin, ball, or slide joints
+  * plane and solid geometry — replace hand-solved trigonometry and spreadsheets with a live dimensioned drawing
+
+grade: stable
+confinement: strict
+
+layout:
+  /usr/share/solvespace:
+    bind: $SNAP/usr/share/solvespace
+
+apps:
+  solvespace:
+    command: usr/bin/solvespace
+    desktop: solvespace.desktop
+    extensions: [gnome-3-28]
+    plugs: [opengl, unity7, home, removable-media, gsettings, network]
+    environment:
+      __EGL_VENDOR_LIBRARY_DIRS: $SNAP/usr/share/glvnd/egl_vendor.d
+  solvespace-cli:
+    command: usr/bin/solvespace-cli
+    plugs: [home, removable-media, network]
+
+parts:
+  solvespace:
+    plugin: cmake
+    source: ./solvespace-snap-src
+    source-type: local
+    override-pull: |
+      snapcraftctl pull
+      version="3.0~$(git rev-parse --short=8 HEAD)"
+      snapcraftctl set-version "$version"
+      git submodule update --init extlib/libdxfrw extlib/flatbuffers extlib/q3d
+    configflags:
+      - -DCMAKE_INSTALL_PREFIX=/usr
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DENABLE_TESTS=OFF
+      - -DSNAP=ON
+    build-packages:
+      - zlib1g-dev
+      - libpng-dev
+      - libcairo2-dev
+      - libfreetype6-dev
+      - libjson-c-dev
+      - libfontconfig1-dev
+      - libgtkmm-3.0-dev
+      - libpangomm-1.4-dev
+      - libgl-dev
+      - libglu-dev
+      - libspnav-dev
+      - git
+    stage-packages:
+      - libspnav0
+      - libatkmm-1.6-1v5
+      - libcairomm-1.0-1v5
+      - libgtkmm-3.0-1v5
+      - libglibmm-2.4-1v5
+      - libpangomm-1.4-1v5
+      - libsigc++-2.0-0v5
+      - libglew2.0
+      - libegl-mesa0
+      - libdrm2
--- a/pkg/snap/snap/snapcraft.yaml
+++ b/pkg/snap/snap/snapcraft.yaml
@@ -13,6 +13,7 @@ description: |
   * plane and solid geometry — replace hand-solved trigonometry and spreadsheets with a live dimensioned drawing
 
 confinement: strict
+license: GPL-3.0
 
 layout:
   /usr/share/solvespace:
--- a/pkg/snap/snap/snapcraft.yaml
+++ b/pkg/snap/snap/snapcraft.yaml
@@ -27,7 +27,7 @@ apps:
     plugs: [opengl, unity7, home, removable-media, gsettings, network]
     environment:
       __EGL_VENDOR_LIBRARY_DIRS: $SNAP/usr/share/glvnd/egl_vendor.d
-  solvespace-cli:
+  cli:
     command: usr/bin/solvespace-cli
     plugs: [home, removable-media, network]
 
--- a/pkg/snap/.gitignore
+++ b/pkg/snap/.gitignore
@@ -1,0 +1,3 @@
+*.snap
+solvespace-snap-src
+squashfs-root
--- a/pkg/snap/build.sh
+++ b/pkg/snap/build.sh
@@ -7,6 +7,6 @@ trap "rm -rf $solvespace_snap_src" EXIT
 cd "$dir"
 
 git_root="$(git rev-parse --show-toplevel)"
-rsync -r "$git_root"/ "$solvespace_snap_src"
+rsync --filter=":- .gitignore" -r "$git_root"/ "$solvespace_snap_src"
 
 snapcraft "$@"
--- a/pkg/snap/build.sh
+++ b/pkg/snap/build.sh
@@ -1,0 +1,12 @@
+#!/bin/sh -xe
+
+dir="$(dirname "$(readlink -f "$0")")"
+solvespace_snap_src="$dir/solvespace-snap-src"
+trap "rm -rf $solvespace_snap_src" EXIT
+
+cd "$dir"
+
+git_root="$(git rev-parse --show-toplevel)"
+rsync -r "$git_root"/ "$solvespace_snap_src"
+
+snapcraft "$@"
--- a/res/CMakeLists.txt
+++ b/res/CMakeLists.txt
@@ -153,6 +153,22 @@ else()
                 DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/${SIZE}/mimetypes
                 RENAME      com.solvespace.SolveSpace.png)
         endforeach()
+    elseif(SNAP)
+        install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/freedesktop/solvespace-snap.desktop
+            DESTINATION /
+            RENAME      solvespace.desktop)
+        
+        # snapd does not support registering new mime types
+
+        install(FILES freedesktop/solvespace-scalable.svg
+            DESTINATION /meta/icons/hicolor/scalable/apps
+            RENAME      snap.solvespace.svg)
+
+        foreach(SIZE 16x16 24x24 32x32 48x48)
+            install(FILES freedesktop/solvespace-${SIZE}.png
+                DESTINATION /meta/icons/hicolor/${SIZE}/apps
+                RENAME      snap.solvespace.png)
+        endforeach()
     else()
         configure_file(
             ${CMAKE_CURRENT_SOURCE_DIR}/freedesktop/solvespace.desktop.in
--- a/res/freedesktop/solvespace-snap.desktop
+++ b/res/freedesktop/solvespace-snap.desktop
@@ -1,0 +1,10 @@
+[Desktop Entry]
+Version=1.0
+Name=SolveSpace
+Comment=A parametric 2d/3d CAD
+Exec=solvespace
+MimeType=application/x-solvespace
+Icon=${SNAP}/meta/icons/hicolor/scalable/apps/snap.solvespace.svg
+Type=Application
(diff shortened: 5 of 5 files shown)


whitequark:
> * The integration resides in `$ROOT/snap`. This is what build.snapcraft.io requires. If using the Ubuntu provided build server (and thus easy auto-build on commit) is not wanted, we could move it to mirror the flatpak integration

Hmm that's a bit annoying. Maybe both flatpak and snap packaging should be moved out into separate repositories? But then auto-build on commit still wouldn't work when a new commit to solvespace/solvespace is made. I'm not sure how to address this.

> * The svg icon should probably find its place in `res/freedesktop` or we could just use a png icon. At the moment snap packages need an absolute path to an icon.

An `svg` icon for freedesktop platforms would be very nice. Would you be able to do this as a separate PR? There would be no problems merging that.

> 3\. The snap is declared `grade: stable`. Maybe that should only apply to tagged versions. Some extra logic would be required

Yes ideally that would be done. The Travis CI configuration already has the necessary logic, but that presumes we'll build with Travis, which is not currently actually possible, since Travis builds turn GUI off due to an extremely outdated GTK there.

whitequark:
> Documentation exists to build snap packages from a Docker container under Travis (https://tutorials.ubuntu.com/tutorial/continuous-snap-delivery-from-travis-ci). So skipping the build server is probably an option.

I'd say: let's put it under `pkg/snap` and use Docker to build on Travis CI.

ppd1990:
```pkg/snap``` seems a bit problematic due to this reason:

snapcraft mounts the working directory into the build container like so:
```
$ mount
[...]
/home/mfederle/Desktop/solvespace/pkg on /root/project type fuse.sshfs [...]
```
Thus, we cannot access the solvespace src with a relative path in snapcraft.yaml. 

Options?
1. Don't move into ```pkg```
2. Use a git URL in snapcraft.yaml. Not very elegant
3. Separate repository

**Comment by whitequark** [comment]
That's uh, pretty obnoxious on the part of snapcraft. I guess I'll have to go through every option it has with a fine-toothed comb...

**Comment by ppd1990** [comment]
It's very clearly intended to be in a separate repo or to reside in the first-level subdirectory ```snap```. I guess that's also KISS in a way...

Here is an example of a snap that uses a build-trigger file to fire the build:
https://github.com/snapcrafters/android-studio

**Comment by whitequark** [comment]
Well, neither of Canonical's supposed solutions works for me, and this isn't an issue with pretty much any other packaging system, so...

**Comment by whitequark** [comment]
If there is absolutely no way to get `snapcraft` to refer to `..`, then I suggest a wrapper script that uses `git worktree add pkg/snap/worktree` to make an additional checkout, and removes it once everything's done. Would that work?

**Comment by ppd1990** [comment]
A worktree does not work with the ```snapcraft.yaml``` as it is, because the worktree references a path outside of the virtual machine:

```
snapcraft-solvespace ../parts/solvespace/src# cat .git
gitdir: /home/mfederle/Desktop/solvespace/.git/worktrees/solvespace-worktree
```
This then fails the git commands.

What works, is to just clone from the parent path like so:

```
git clone .. solvespace-snap-src
```

That's certainly an option for staying close to the paradigm of not generating ```snapcraft.yaml``` with cmake, but containing the snapcraft specific logic as much as possible within ```snapcraft.yaml```.

I'll investigate whether it is maybe even possible to somehow contain the setup and cleanup commands in ```snapcraft.yaml``` itself.

**Comment by whitequark** [comment]
Thank you!

**Comment by ppd1990** [comment]
I think the build script is a sensible choice.

As I understand it, the build environment for snapcraft is basically either ```host``` or ```multipass```, where ```host``` would apply to the Docker case, and thus to Travis. So maybe the relative path would even work in such an environment, but the resulting ```snapcraft.yaml``` wouldn't be environment-agnostic and hardly well-formed.

**Comment by ppd1990** [comment]
The fun continues: https://hub.docker.com/r/snapcore/snapcraft is the official snapcraft docker image and it is still based on Ubuntu 16.04 which makes it a bit useless for compiling ```base: core18``` snaps.

I'll investigate that further.

**Comment by ppd1990** [comment]
Another option: Run a Travis job that syncs this repo's ```snapcraft.yaml``` to another repository whose sole function is to serve as a source for build.snapcraft.io.

In the ```snapcraft.yaml``` itself we'd add support for building from the source tree itself (i.e. local) as well as from the git repository.

**Comment by whitequark** [comment]
> What do you think?

Sorry, I missed your edit. I'll take a look soon!

**Comment by ppd1990** [comment]
Note to ourselves:

Snapcraft supports remote builds on Launchpad: https://forum.snapcraft.io/t/preview-snapcraft-remote-build/11541

A possible pipeline: Github -> Travis --- (Snapcraft) ---> Launchpad => snaps for multiple architectures

This would offload all building again and of course the issue of compiling for other architectures (if so desired) is not our concern.

**Comment by whitequark** [comment]
I think that's a reasonable option. What should we do for that? Separate repo, an encrypted SSH key on Travis, and trigger builds from there?

**Comment by ppd1990** [comment]
Oh. I'm sorry I haven't replied yet. I'll investigate and create a prototype as soon as I'm not swamped with work any more...

**Comment by ppd1990** [comment]
Coming back to this, a few observations:

1. Custom docker images won't be necessary for snapcraft work as parametrized docker images are being proposed atm: https://github.com/snapcore/snapcraft/pull/2673

2. Remote build is still in beta: https://forum.snapcraft.io/t/preview-snapcraft-remote-build/11541/10?u=ppd

> I think that's a reasonable option. What should we do for that? Separate repo, an encrypted SSH key on Travis, and trigger builds from there?

The SSH key and the snapcraft credentials must be stored on Travis. 

From the snapcraft example:
```
openssl aes-256-cbc -K <travis-key> -iv <travis-iv>
      -in .snapcraft/travis_snapcraft.cfg
      -out .snapcraft/snapcraft.cfg -d
```

Build & release then something like this:

```
docker run -v $(pwd):/solvespace mfederle/solvespace-snapcraft:latest sh -c 
"apt update && sh /solvespace/pkg/snap/build.sh" &&
snapcraft push *.snap --release edge
```

To summarize, we need:

1. A snapcraft version with remote-build enabled
2. Launchpad credentials in Travis
3. Snapcraft cfg in Travis

I'll report back when I have a usable Travis cfg.

**Comment by ppd1990** [comment]
@whitequark I have a sample travis config at https://github.com/ppd1990/solvespace-travis that makes use of snapcraft remote-build and this branch

Builds are successful, only the deploy step in Travis threw me an error. But that's probably a spurious issue on the server side...

**Comment by ppd1990** [comment]
Note regarding maintainability: One of the very next snapcraft releases will ship the gnome/desktop extension and will thus render lots of the fluff (the extra part, those gtk-common-themes plugs...) in snapcraft.yaml obsolete, making everything look a lot cleaner.

**Comment by whitequark** [comment]
Is that happening soon? If so we might just wait a bit.

**Comment by ppd1990** [comment]
I guess so. Seems like they do about one release per month and the last non-minor release was on 2019-07-23.

We're also still waiting for the remote-build functionality that is still not in a stable release of snapcraft. So we might soon get the new release but the remote-build branch of snapcraft is not guaranteed to instantly rebase on this release.

**Comment by whitequark** [comment]
Gotcha. Thanks for following through on all this, I appreciate it.

**Comment by ppd1990** [comment]
We have time. It's already in the store and working. So let's make this foolproof first.

**Comment by ppd1990** [comment]
Added various cleanups and small fixes. 

The main update is the use of the gnome-3-28 extension instead of setting up everything manually.
This cleans up snapcraft.yaml quite a bit, but is not yet compatible with remote builds.

File: pkg/snap/snap/snapcraft.yaml [hunk 1]
```diff
@@ -0,0 +1,76 @@
...
+    override-pull: |
+      snapcraftctl pull
+      version="3.0~$(git rev-parse --short=8 HEAD)"
+      snapcraftctl set-version "$version"
+      git submodule update --init extlib/libdxfrw extlib/flatbuffers extlib/q3d
+    override-stage: |
+      snapcraftctl stage
+      # patch desktop file with correct icon path
+      SED_CMD="sed -i -E s|^Icon=(.*)|Icon=\${SNAP}/usr/share/icons/hicolor/scalable/apps/solvespace.svg|g"
+      $SED_CMD usr/share/applications/solvespace.desktop
```
whitequark: We already have a `-DFLATPAK=ON` flag that installs a different desktop file. I think you should add the same for snap (`-DSNAP=ON`) and get rid of this override.

pkg/snap/snap/snapcraft.yaml
```diff
@@ -0,0 +1,76 @@
...
+
+layout:
+  /usr/share/solvespace:
+    bind: $SNAP/usr/share/solvespace
+
+apps:
+  solvespace:
+    command: usr/bin/solvespace
+    extensions: [gnome-3-28]
+    plugs: [opengl, unity7, home, removable-media, gsettings, network]
```
whitequark: `unity7`?

pkg/snap/snap/snapcraft.yaml
```diff
@@ -0,0 +1,76 @@
...
+    desktop: usr/share/applications/solvespace.desktop
+    environment:
+      __EGL_VENDOR_LIBRARY_DIRS: $SNAP/usr/share/glvnd/egl_vendor.d
+  solvespace-cli:
+    command: usr/bin/solvespace-cli
+    plugs: [home, removable-media, network]
```
whitequark: Does the CLI actually work with strict confinement? As in, have you tried specifying paths?

pkg/snap/snap/snapcraft.yaml
ppd1990: From: https://snapcraft.io/docs/unity7-interface I quote:

> Condition to Connect: If your snap has a graphical user interface and is expected to be launched on Unity desktop environment(the default DE on Ubuntu 16.04), you should connect this interface.

And https://snapcraft.io/docs/desktop-interfaces tells us
> the unity7 interface remains for compatibility for typical, non-Wayland capable DEs

and explains its possible use for a
> Desktop app that can run anywhere

So probably a good idea to keep around for maximum compatibility with all sorts of desktop environments

whitequark: Strange name, but OK.

pkg/snap/snap/snapcraft.yaml
ppd1990: That works:

```bash
solvespace.solvespace-cli export-mesh -o test.stl -t 0.01 Desktop/testsketch.slvs 
Missing (absent) translation for group-name'#references'
Missing (absent) translation for group-name'sketch-in-plane'
Written 'test.stl'.
```

So path-wise that is OK. However I've got me some segfaults when trying to use the export-view command. I have to read a little about how to properly use gdb in snaps...

whitequark: What about absolute paths?

ppd1990: Something a bit more exciting:

```bash
mfederle@maxi:~$ solvespace.solvespace-cli export-mesh -o /run/user/1095800001/gvfs/sftp:host=*****,user=pi/home/pi/test.stl -t 0.01 /home/mfederle/Desktop/testsketch.slvs 
Missing (absent) translation for group-name'#references'
Missing (absent) translation for group-name'sketch-in-plane'
Written '/run/user/1095800001/gvfs/sftp:host=hermes.internal.linergy.de,user=pi/home/pi/test.stl'

mfederle@maxi:~$ stat /run/user/1095800001/gvfs/sftp:host=*****,user=pi/home/pi/test.stl
  File: /run/user/1095800001/gvfs/sftp:host=hermes.internal.linergy.de,user=pi/home/pi/test.stl
  Size: 684       	Blocks: 2          IO Block: 4096   regular file
Device: 46h/70d	Inode: 16851       Links: 1
Access: (0644/-rw-r--r--)  Uid: (1095800001/mfederle)   Gid: (1095800001/mfederle)
Access: 2019-09-10 15:01:08.000000000 +0200
Modify: 2019-09-10 15:01:08.000000000 +0200
Change: 2019-09-10 15:01:08.000000000 +0200
 Birth: -
```

Something noteworthy, for the forbidden path ```/tmp/test.stl``` we get this:

```bash
mfederle@maxi:~$ solvespace.solvespace-cli export-mesh -o /tmp/test.stl -t 0.01 /home/mfederle/Desktop/testsketch.slvs 
Missing (absent) translation for group-name'#references'
Missing (absent) translation for group-name'sketch-in-plane'
Written '/tmp/test.stl'.

mfederle@maxi:~$ stat /tmp/test.stl
stat: cannot stat '/tmp/test.stl': No such file or directory
```

whitequark: OK, I have no idea *how* it works (it seems impossible to me! clearly, I don't know something important about Linux containers) but I'm fine just knowing it *does* work.

ppd1990: The snap program is run by the snapd runner/daemon/whatever. A snap does not live in its container, but more or less has its permissions restricted by apparmor (sort of "access nothing by default"). Its plugs in the app declaration allow access to certain paths (r, rw, rwx. Whatever is appropriate) or sockets thereafter.

So /home/peter/myfile.txt is the same for every snap, but not every snap has permissions to access it. Some paths however behave like you would expect in a container. In our case the layout definition in snapcraft.yaml effects an overlay mount in the snap's mount-namespace .

pkg/snap/snap/snapcraft.yaml
ppd1990: Snapd is in the process of gaining support for icon themes:
https://forum.snapcraft.io/t/proposal-support-the-icon-theme-spec-for-desktop-icons/10676
https://github.com/snapcore/snapd/pull/6767

So I'll add complete snap specific section in CMakeLists.txt in preparation for this functionality and remove this hack.

ppd1990: Done

ppd1990:
Implemented changes:

- added SNAP CMake option to install snap-specific desktop file
- Install icons in correct location in preparation for when snapd gains icon theme support
- Replace git clone with rsync in build.sh to allow building uncommited code

whitequark:
Looks great. The only blocker I see right now is `grade`; we can merge this PR and add remote builds when they're available.

whitequark:
Can you adjust `rsync` so that it only syncs the files in git, or at least skips the `build` folder? Mine is 5 GB in size.

ppd1990:
> Can you adjust `rsync` so that it only syncs the files in git, or at least skips the `build` folder? Mine is 5 GB in size.

Oh yeah. That is a rather naive implementation so far. I'll look into it.

> Looks great. The only blocker I see right now is grade

What's your preferred strategy? ```grade = devel``` mainly prevents you from releasing to a stable channel. That's about it.

whitequark:
> > 
> 
> Oh yeah. That is a rather naive implementation so far. I'll look into it.

Looks like `--filter=':- .gitignore'` works

whitequark:
> What's your preferred strategy? `grade = devel` mainly prevents you from releasing to a stable channel. That's about it.

Ideally check if the commit corresponds to a tag, and select grade based on that.

whitequark:
Is it possible to make the CLI install as `solvespace.cli` instead of `solvespace.solvespace-cli`?

ppd1990:
That's easy. I'll update the PR soon

ppd1990:
I addressed the requests and also added license info to the snapcraft declaration.

One thing that still bugs me a bit: The hardcoded version scheme ```3.0~$hash```. We could extract it from CMakeLists.txt instead, but that is a bit hacky. I'd love to use git describe or similar, but given that we have no tag like ```3.0-dev``` or similar, that is not really an option.

pkg/snap/snap/snapcraft.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+parts:
+  solvespace:
+    plugin: cmake
+    source: ./solvespace-snap-src
+    source-type: local
+    override-pull: |
+      snapcraftctl pull
+      version="3.0~$(git rev-parse --short=8 HEAD)"
```
ppd1990:
```suggestion
      version_major=$(grep "solvespace_VERSION_MAJOR" CMakeLists.txt | tr -d "()" | cut -d" " -f2)
      version_minor=$(grep "solvespace_VERSION_MINOR" CMakeLists.txt | tr -d "()" | cut -d" " -f2)
      version="$version_major.$version_minor~$(git rev-parse --short=8 HEAD)"
```
How about that? Not great, not terrible...

whitequark: Seems fine to me.

whitequark:
Thanks.

whitequark:
I'm curious: why is the `network` plug enabled? SolveSpace shouldn't really be talking over network.

ppd1990:
The reason is a bit convoluted: By default, snaps will not have access to exotic PAM providers like SSSD and thus will not recognize "non-traditional" users like those configured via LDAP etc.

A workaround, as proposed in https://forum.snapcraft.io/t/accessing-unix-users-and-groups-from-a-strict-snap/9035/7, is to install a user info proxy on the machine like `nscd`. 
To access that proxy's socket within the confinement, we need the `network` plug.

whitequark:
OK but ... SolveSpace doesn't care one bit about UNIX groups and users.

whitequark:
Oh right, we do care about `$HOME` and that's from `passwd` and we don't use the Portal API because we have to access a file at a relative path for assemblies. Sigh.

ppd1990:
Solvespace doesn't, but its desktop integration does. The `gnome-3-28` extension bit sets up some softlinks that enable the user to see their "real" home folder and such from the file management dialogs despite it having its own $HOME within confinement.

PR #441, opened 2019-06-14
freedesktop: Add svg icon

ppd1990:
As discussed in #436: Add a scalable icon for freedesktop platforms


whitequark:
Thanks. I merged this manually after rescaling the image to 256x256 and using the "Save optimized SVG" option in Inkscape, with all unnecessary stuff removed, and digits after decimal set to 0. [As you can see](https://github.com/solvespace/solvespace/blob/master/res/freedesktop/solvespace-scalable.svg?short_path=7be996e) the result is a bit more compact :)

ppd1990:
Oh great. That was indeed an oversight.

PR #446, opened 2019-07-07
Brighter blue and yellow in text window for higher contrast/readability.

phkahler:
Higher contrast colours in the text window for readability. Much lighter blue and somewhat lighter yellow.


whitequark:
For reference to others, here's how it looks with this PR:
Screenshot_20190709_004632
... and before it:
Screenshot_20190709_004708

whitequark:
I like it! These are beautiful colors. And they combine well with other parts of the UI like the ok/err mark.

phkahler:
Thanks! There was actually a lot of thought in this despite the PR being just 2 numbers in the end.

PR #447, opened 2019-07-09
Handle operators

rpavlik:
Handles seem like the kind of thing you should be able to compare, and the .v scattered throughout the codebase was an impediment to me learning it. Finally did a little modification I meant to do for a while: provide `==`, `!=`, `<` and `explicit operator bool` (aka "use in `if` statements") operators for all the handle types. Easiest, least messy way was with a curiously-recurring-template-pattern base class so there is only one implementation of the operators, yet the handles are still type-safe (can't compare across handle types without using `.v` to get the uint32_t).

If it seems useless to you, no hard feelings if you just reject this (or drop parts of it, like the operator bool perhaps, since that seems less obviously-useful than the comparison operators).  FWIW, I also have a version of this based on #428 (actually I devved it there first then rebased it onto master), at https://github.com/rpavlik/solvespace/tree/handle-operators-after-simplify

Could revise to not use CRTP, but would then need a "tag type" for each handle type to keep them distinct, so this seemed simplest.

### DIFF
--- a/src/clipboard.cpp
+++ b/src/clipboard.cpp
@@ -12,16 +12,16 @@ void SolveSpaceUI::Clipboard::Clear() {
 }
 
 bool SolveSpaceUI::Clipboard::ContainsEntity(hEntity he) {
-    if(he.v == Entity::NO_ENTITY.v)
+    if(he == Entity::NO_ENTITY)
         return true;
 
     ClipboardRequest *cr;
     for(cr = r.First(); cr; cr = r.NextAfter(cr)) {
-        if(cr->oldEnt.v == he.v)
+        if(cr->oldEnt == he)
             return true;
 
         for(int i = 0; i < MAX_POINTS_IN_ENTITY; i++) {
-            if(cr->oldPointEnt[i].v == he.v)
+            if(cr->oldPointEnt[i] == he)
                 return true;
         }
     }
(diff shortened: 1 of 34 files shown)


whitequark:
I think this doesn't work, technically speaking. The reason is that according to C++11, only classes without base classes count as POD, and handles really ought to stay POD. In particular, the last time I changed some existing classes to be non-POD, I immediately reaped a bunch of problems with default-initialization and zero-initialization not working as expected--IIRC it was related to default constructor not zero-initializing POD members or something like that but my memory is hazy.

whitequark:
Oh yeah and it breaks horribly on MSVC for the same reasons. But that's probably the least of the worries here, VS2013 is six years old and we're getting pretty close to the point where upgrading it makes good sense.

whitequark:
I just tried to bump the MSVC requirement to 2015 and of course the testsuite [crashes](https://ci.appveyor.com/project/whitequark/solvespace/builds/25853567) with `Command exited with code -1073741571`. Not sure why and I don't have that MSVC version locally... help with this is much appreciated.

rpavlik:
ah, I figured I must have changed something fundamental for pod initialization to stop working here with GCC. (Pretty sure newer compilers also have xp support in a separate toolchain?)

whitequark:
Oh, how'd traits work here? That sounds nicer than CRTP.

rpavlik:
err, like this... https://github.com/rpavlik/solvespace/tree/handle-operators-2

Don't know how to inject the "explicit operator bool" this way, but it at least gives you equality/inequality.

whitequark:
I think this approach is not really viable, but #448 might be good.

PR #448, opened 2019-07-09
Handle operators 2

rpavlik:
Alternate to #447 - doesn't have the operator bool stuff, but you at least get ==, !=, and <. Let's see what CI thinks.

### DIFF
--- a/src/clipboard.cpp
+++ b/src/clipboard.cpp
@@ -12,16 +12,16 @@ void SolveSpaceUI::Clipboard::Clear() {
 }
 
 bool SolveSpaceUI::Clipboard::ContainsEntity(hEntity he) {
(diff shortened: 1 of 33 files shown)


whitequark:
@jwesthues Any strong opinions on this change? It adds a bit of type safety in that it's an error to compare unrelated handles, and it makes maps with handles as keys a bit nicer. And although I'm very much not fond of the way C++ manages to provide basic functionality, this implementation is not invasive and shouldn't cause us any problems like doing this with CRTP.

jwesthues:
Seems reasonable to me.

whitequark:
Thanks @rpavlik!

phkahler:
Yes, thanks @rpavlik. I like the readability improvements this brings.

PR #456, opened 2019-08-09
Export transparency correctly in VRML

nabijaczleweli:
Opted for an explicit span container for clarity but that can be easily replaced with explicit iterator-`for`s if you're not a fan

screenshot

### DIFF
--- a/src/export.cpp
+++ b/src/export.cpp
@@ -1175,10 +1175,10 @@ void SolveSpaceUI::ExportMeshAsThreeJsTo(FILE *f, const Platform::Path &filename
 //-----------------------------------------------------------------------------
 void SolveSpaceUI::ExportMeshAsVrmlTo(FILE *f, const Platform::Path &filename, SMesh *sm) {
     struct STriangleSpan {
-        const STriangle *first, *past_last;
+        STriangle *first, *past_last;
 
-        const STriangle *begin() const { return first; }
-        const STriangle *end() const { return past_last; }
+        STriangle *begin() const { return first; }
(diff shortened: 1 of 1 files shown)


src/export.cpp
```diff
@@ -1174,6 +1174,14 @@ void SolveSpaceUI::ExportMeshAsThreeJsTo(FILE *f, const Platform::Path &filename
 // Export the mesh as a VRML text file / WRL.
 //-----------------------------------------------------------------------------
 void SolveSpaceUI::ExportMeshAsVrmlTo(FILE *f, const Platform::Path &filename, SMesh *sm) {
+    struct STriangleSpan {
+        const STriangle *first, *past_last;
+
+        const STriangle *begin() const { return first; }
```
whitequark: Nit: this should be properly called `cbegin` since it returns a const iterator, but it's probably better to ditch `const` here since it doesn't really do anything useful. (`const` in SolveSpace is only used for replacing by-value passing with by-const-reference.)

nabijaczleweli: I just copied the `List<>` API here tbh
And, FWIW, `begin()` has [both `const` and non-`const` overloads](https://en.cppreference.com/w/cpp/container/vector/begin) in the STL, and ranged-for uses only [`begin()`](http://eel.is/c++draft/stmt.ranged#1.3.2)

Anyway, ditched

whitequark: Oh, you're right, my bad.

whitequark:
Thanks!

PR #459, opened 2019-08-20
Add proxy/iterator to iterate groups in order.

rpavlik:
Simplifies an idiom seen all over the codebase, of iterating
thru groupOrder then using it to look up a group.

### DIFF
--- a/src/solvespace.h
+++ b/src/solvespace.h
@@ -485,6 +485,48 @@ public:
     bool CanOutputMesh() const override { return false; }
 };
 
+class Sketch;
+
+class OrderedGroupsIterator {
+public:
+    OrderedGroupsIterator() = default;
+    OrderedGroupsIterator(Sketch &sketch, hGroup *currentGroup)
+        : sk(&sketch), current(currentGroup) {}
+
+    bool IsEnd() const { return current == nullptr; }
+
+    Group *operator*() const;
+
+    // Pre-increment operator
+    // Definition below that for Sketch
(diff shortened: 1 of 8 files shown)


src/solvespace.h
```diff
@@ -485,6 +485,48 @@ class GCodeFileWriter : public VectorFileWriter {
...
+
+    bool operator==(OrderedGroupsIterator const &other) const {
+        if(IsEnd() && other.IsEnd()) {
+            return true;
+        }
+        return sk == other.sk && current == other.current;
+    }
+    bool operator!=(OrderedGroupsIterator const &other) const { return !((*this) == other); }
+
+private:
```
whitequark: We don't really use C++ visibility.

src/solvespace.h
```diff
@@ -485,6 +485,48 @@ class GCodeFileWriter : public VectorFileWriter {
...
+    Sketch *sk      = nullptr;
+    hGroup *current = nullptr;
+};
+
+
+class OrderedGroupsProxy {
```
whitequark: Is there any way to simplify this? Returning a pair or something. As it is I feel like this patch adds more complexity than it removes.

whitequark: I mean, +70 lines -27 lines.

rpavlik: I wish there was - but ranges haven't made it into the standard yet. (Maybe c++20?)

No hard feelings, I can close this. Just had it sitting around here.

PR #460, opened 2019-08-20
Simplify and clang tidy

rpavlik:
Take or leave any subset of these patches. They're all split by topic for easy cherry-picking.


whitequark:
Thanks, I've cherry-picked most of this commit. Could you please rebase it so we can discuss the rest? There are a few things I'd like to see changed.

The main of them is the clang-tidy configuration that causes it to put small functions and loops/conditionals entirely on one line (which IMO makes the code more annoying to read); ideally it'd not reformat the functions that way at all, and maybe match loops/conditionals to the existing style better, but the second part isn't as important.

The rest are something I'd like to discuss in review.

whitequark:
@rpavlik Also, looks like tests crash on Windows now. Could you please take a look? https://ci.appveyor.com/project/whitequark/solvespace/builds/27299022

whitequark:
Looks like the crash is caused by "Range-for conversions and improvements. NFC."

rpavlik:
Yeah, I have hung my machine twice now attempting to run the tests for that commit with sanitizers from GCC enabled. (It was all fine with Clang...)  I've dropped that patch from this series, will figure it out separately.

whitequark:
@rpavlik This PR still fails on Windows.

src/polygon.h
```diff
@@ -400,11 +400,10 @@ class PolylineBuilder {
...
     Edge *AddEdge(const Vector &p0, const Vector &p1, uint32_t kind, uintptr_t data = 0);
-    void Generate(
-        std::function<void(Vertex *start, Vertex *next, Edge *edge)> startFunc,
-        std::function<void(Vertex *next, Edge *edge)> nextFunc,
-        std::function<void(Edge *)> aloneFunc,
-        std::function<void()> endFunc = [](){});
+    void Generate(std::function<void(Vertex *start, Vertex *next, Edge *edge)> const &startFunc,
+                  std::function<void(Vertex *next, Edge *edge)> const &nextFunc,
+                  std::function<void(Edge *)> const &aloneFunc,
+                  std::function<void()> const &endFunc = []() {});
```
whitequark: What does this actually guarantee?

src/polygon.cpp
```diff
@@ -296,16 +296,10 @@ bool SEdgeList::AssemblePolygon(SPolygon *dest, SEdge *errorAt, bool keepDir) co
...
-    for(const SEdge *se = l.First(); se; se = l.NextAfter(se)) {
-        if(se->EdgeCrosses(a, b, ppi, spl)) {
-            cnt++;
-        }
-    }
-    return cnt;
+int SEdgeList::AnyEdgeCrossings(Vector a, Vector b, Vector *ppi, SPointList *spl) const {
+    auto cnt = std::count_if(l.begin(), l.end(),
+                             [&](SEdge const &se) { return se.EdgeCrosses(a, b, ppi, spl); });
+    return static_cast<int>(cnt);
```
whitequark: Can we not use `static_cast` for trivial integer casts?

whitequark:
Merged a few more commits from here. In "Simplify returns and booleans. NFC.", most of the changes make the code less readable, and a few comments no longer explain what's happening, so you can drop that one.

src/polygon.cpp
rpavlik: as in, you'd prefer `(int)cnt` ?

whitequark: Yep. I think e.g. `const_cast` makes sense when you're making it explicit that you're removing a `const` modifier, but doing simple integer casts with `static_cast` is just noise.

whitequark: Although in this case maybe the function just needs to return `size_t`.

rpavlik: yeah, there are a bunch of places where it would make sense to return size_t but we return int.

whitequark: I'm fine fixing them to return `size_t`.

whitequark:
Merged everything except FindIndex changes. Do we really have no better approach here than returning -1 if it's not found? Morally it seems like FindIndex ought to return size_t.

nabijaczleweli:
FWIW, the [`std::string::find()`](https://en.cppreference.com/w/cpp/string/basic_string/find) API just returns `(size_t)-1` when the needle wasn't found, which has the added benefit of both returning the correct type and having an explicitly-named sentinel (`npos`, in that case), rather than an unclear `< 0` check.

whitequark:
Yeah that'd probably be fine.

rpavlik:
Rebased to leave the one commit not cherry-picked yet.

PR #461, opened 2019-08-20
Add the ability to merge IdLists, and use to banish more non-range-for.

rpavlik:
Eliminates a class of reasons for not using range-for: Adding
during iteration. Instead, just add to a new IdList that you
merge in after iteration is complete.

In general I didn't convert the new types of solids to use this, just because I'm still new at using them so I couldn't tell for sure if it worked right. But, all tests pass.

### DIFF
--- a/src/dsc.h
+++ b/src/dsc.h
@@ -420,6 +420,27 @@ public:
         std::inplace_merge(begin(), end() - 1, end(), Compare());
     }
 
+    void MergeInto(IdList *dest) {
+        dest->ReserveMore(n);
+
+        const auto oldEnd = dest->elem + dest->n;
+        auto outIter      = oldEnd;
+
+        for(auto &elt : *this) {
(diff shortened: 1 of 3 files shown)


src/dsc.h
```diff
@@ -420,6 +420,27 @@ class IdList {
...
+        const auto oldEnd = dest->elem + dest->n;
+        auto outIter      = oldEnd;
+
+        for(auto &elt : *this) {
+            // Copy-construct at the end of the list.
+            new(outIter) T(elt);
+            ++outIter;
+        }
+
+        /// @todo Look to see if we already have something with the same handle value.
```
whitequark: I'm not really comfortable merging this without an assert here.

rpavlik: yeah I just realized this messes up a bunch of things if you actually try using a lathe. I'm dropping it, sorry.

Issue #465, opened 2019-08-22
Internal: replace struct "Revolved"


phkahler:
In surface .cpp there is a struct defined as:

`

    typedef struct {
        hSSurface d[100];
    } Revolved;
`
This was originally used in Lathe groups to temporarily collect 4 90 degree sections of a surface. I expanded it to 100 to support more sections in Helix groups, but having a fixed number in there is not so great.

Also, in one of my branches (draft) I wanted to pass a list of these to a function and so had to move the definition into surface.h which to me is also a bit undesirable.

A better container for these would be nice. Is the right answer just a list? I'm not fan of doing things like list of lists in C++ but even that is better than making this a fixed size. @rpavlik ?

rpavlik:
Well, if it needs the slightly-weird semantics of List<> (that is, calling clear on everything before removing it), then use List<> otherwise use std::vector directly.

whitequark:
We don't really have a policy on using `List` vs `std::vector` in existing code. I've used `std::vector` in parts that are self-contained and completely new (like new export functions), but I'd say if it's a part of the geometry code then `List` would be less annoying to work with.

But this is fairly minor and just my opinion.

phkahler:
std::vector worked out nicely because we know what length is needed when they're created.

phkahler:
This was done in PR #466 commits ced879c and cad7002

Issue #458, opened 2019-08-14
Revolve shouldn't be possible for 3D groups


rpavlik:
### System information

SolveSpace version: 3.0~97c8cb7d

Operating system: Debian buster

### Expected behavior

I tried to use the new "Revolve" feature, initially trying to sweep a solid. It appears that is not supported (only sweeping a 2D sketch), so it shouldn't have let me do it.

### Actual behavior

It let me do it, but the result didn't make any sense.

phkahler:
Lathe isn't very useful for solids either.

rpavlik:
Probably pop up the same kind of window that I get when trying to make a new workspace group without selecting an origin, etc. (or just grey out that option?)

phkahler:
The check for that would go in Group::MenuGroup() under the cases for those group types. Question on Localization, is it OK if we use the exact same string for the error message in multiple places in the code? If so, should we?

The test would be similar to the one used in Extrude groups and should be in Revolve and Helix groups too. It occurs to me that in weird cases, doing this with Lathe might create circles that are relevant to someone somewhere.

whitequark:
> Question on Localization, is it OK if we use the exact same string for the error message in multiple places in the code? If so, should we?

Completely identical messages get grouped into a single group by gettext so that's fine.

> It occurs to me that in weird cases, doing this with Lathe might create circles that are relevant to someone somewhere.

Fortunately, existing files with such strange constructs will continue to work correctly. And if someone complains and it turns out we missed an important use case, we can always revert.

phkahler:
Implemented Fix in commit 2dd50d0 of PR #466

PR #466, opened 2019-08-24
Changes for  #458 and #465

phkahler:


whitequark:
(I can't merge this PR as-is because the last commit does two completely unrelated things, so it's going to wait until I have time to manually untangle it.)

phkahler:
@whitequark  Regarding the 3rd commit:  The Revolved struct was fixed length, so by using std::vector there is no need to limit number of sections, and that was the last use of the struct so its definition can be deleted entirely.

I would have done the last 2 commits as one but I wanted to get the lathe one in first in case I messed up the bigger one (which turned out to be small too).

If anything, the first commit is completely unrelated to the other two. But when I committed those to my repo, github wanted to automatically lump all three together. Wasn't sure how to get the second two as a separate PR so I just edited it. To me this might have been 2 PRs, but IMHO there is no problem with the individual commits.

whitequark:
> To me this might have been 2 PRs, but IMHO there is no problem with the individual commits.

Ah, sorry, I misunderstood. Indeed there is no problem.

Issue #472, opened 2019-09-09
Skewed NURBS issue


phkahler:
NURBS surfaces are more likely to have problems when the derivatives along U and V are not orthogonal. This is only likely if we lathe or revolve around an axis that does not lie in the sketch plane. Also, all helical extrusions have this non-orthogonality in the U,V space and tend not to do well in intersection as shown here:

![Helix_issue](https://user-images.githubusercontent.com/14852918/64497347-62afa780-d27b-11e9-925b-d04b8a91d9b2.png)

The errors go away if the axial displacement of the helix is reduced - as you get closer to a plain revolve.

The problem is that the function SSurface::ClosestPointNewton() assumes the derivatives are orthogonal. I have a fix in the works for this (works but I'm not sure it's quite correct). This may also be in play for #471

Issue #471, opened 2019-09-09
NURBS issue with torus


phkahler:
By using lathe around a point that is not in the sketch plane, we can create a torus with an oval cross section. By dragging that point we can change the aspect ration of that section, but as shown here this can result in NURBS errors. Even worse, this will also happen if we force to triangle mesh.

![torus_issue](https://user-images.githubusercontent.com/14852918/64497250-71e22580-d27a-11e9-9932-a4f06b087dc2.png)

phkahler:
Fixed with #473

Issue #315, opened 2018-02-22
Naked edges in bezier lathe difference


wavexx:
### System information

SolveSpace version: 3.0~33b6e517

Operating system: Debian Unstable

### Expected behavior

Here's a clip I've been working on. It has a pass-through funnel made with a lathed bezier curve.
The difference between the upper face and the lathed surface is causing a naked edge:

![2018-02-22t171642](https://user-images.githubusercontent.com/1017726/36550189-0094f506-17f5-11e8-8a58-fed9da1583ea.png)

As visible on the screenshot, the naked edge is only between the top face and the first visible point, the rest of the intersection is correctly computed.

Union is also correctly computed, and I can also get a correct difference by forcing a mesh geometry:

![2018-02-22t171706](https://user-images.githubusercontent.com/1017726/36550453-a22e1370-17f5-11e8-946f-6e72c7be0633.png)

So maybe it's something easy to fix.
Test file: [clip.zip](https://github.com/solvespace/solvespace/files/1748684/clip.zip)

I tried to reduce the test case to a simple cube and a lathed bezier, but I wasn't able to reproduce it.

whitequark:
It's not anything easy to fix. It's a known issue with the NURBS backend. The workaround is to use mesh.

wavexx:
One can dream. Is there some sort of master issue no. to track NURBS issues?
I could find several, but all showing "corner" cases.

whitequark:
Not really. Fixing the NURBS backend is a many man-years problem requiring the sort of highly specialized experience that cannot be easily contracted out. At least, I wasn't able to find anyone with a set of skills even remotely approaching what's required to make the NURBS boolean code robust. Even if I did, the SolveSpace project does not have the sort of budget required to fund this feature.

wavexx:
Some time ago I was also looking into this, and stumbled upon your question here: https://softwarerecs.stackexchange.com/questions/27546/a-c-nurbs-library-with-boolean-operations-that-is-not-opencascade

Did you discover anything else since then? Did you, by pure chance, had a look at the NURBS support in brlcad? They mention some mixed nurbs/srep support, but I never had the time to investigate if there is some code we could make use of, or if it's just CSG behind the scenes.

whitequark:
> Did you discover anything else since then?

No. I'm pretty sure nothing OSS besides OpenCASCADE exists, at this point.

> Did you, by pure chance, had a look at the NURBS support in brlcad?

[...]

PR #473, opened 2019-09-09
Nurbs

phkahler:
Fixes for NURBS issues in #471 and #472.
471 also had issues with triangle mesh.

### DIFF
--- a/src/srf/ratpoly.cpp
+++ b/src/srf/ratpoly.cpp
@@ -439,9 +439,14 @@ void SSurface::ClosestPointTo(Vector p, double *u, double *v, bool mustConverge)
                bu   = (ctrl[1][0]).Minus(orig),
                bv   = (ctrl[0][1]).Minus(orig);
         if((ctrl[1][1]).Equals(orig.Plus(bu).Plus(bv))) {
+
+            Vector n = bu.Cross(bv);
+            Vector ty = n.Cross(bu).ScaledBy(1.0/bu.MagSquared());
+            Vector tx = bv.Cross(n).ScaledBy(1.0/bv.MagSquared());
+
             Vector dp = p.Minus(orig);
-            *u = dp.Dot(bu) / bu.MagSquared();
-            *v = dp.Dot(bv) / bv.MagSquared();
+            *u = dp.Dot(bu) / tx.MagSquared();
+            *v = dp.Dot(bv) / ty.MagSquared();
             return;
         }
     }
@@ -598,10 +603,14 @@ Vector SSurface::ClosestPointOnThisAndSurface(SSurface *srf2, Vector p) {
 
         // Adjust our guess and iterate
         for(j = 0; j < 2; j++) {
+            Vector n = tu[j].Cross(tv[j]);
+            Vector ty = n.Cross(tu[j]).ScaledBy(1.0/tu[j].MagSquared());
+            Vector tx = tv[j].Cross(n).ScaledBy(1.0/tv[j].MagSquared());
+
             Vector dc = pc.Minus(cp[j]);
-            double du = dc.Dot(tu[j]), dv = dc.Dot(tv[j]);
-            puv[j].x += du / ((tu[j]).MagSquared());
-            puv[j].y += dv / ((tv[j]).MagSquared());
+            double du = dc.Dot(tx), dv = dc.Dot(ty);
+            puv[j].x += du / tx.MagSquared();
+            puv[j].y += dv / ty.MagSquared();
         }
     }
     if(i >= 10) {
--- a/src/srf/ratpoly.cpp
+++ b/src/srf/ratpoly.cpp
@@ -511,6 +511,11 @@ bool SSurface::ClosestPointNewton(Vector p, double *u, double *v, bool mustConve
         double du = dp.Dot(tu), dv = dp.Dot(tv);
         *u += du / (tu.MagSquared());
         *v += dv / (tv.MagSquared());
+
+        if (*u < 0.0) *u = 0.0;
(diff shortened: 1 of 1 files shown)


phkahler:
Some explanation with a diagram:

skew_nurbs

In the existing code, the point P is projected on to the tu and tv vectors at points 1 and 4. Normally that's just a dot product and division by the magnitude of the vector, but that's in world space not u,v so we need to divide by the magnitude again which is where the magnitude squared came from (yay no square roots!). And is fine if u and v are orthogonal. In this picture they are not.

What we need is p projected on V at point 3 and on U at point 5. Lets just worry about V for now. By constructing W in the plane and orthogonal to U we get this picture. Point 2 is just P projected on W, but it's the same fraction of W that point 3 is of V but that projection requires the division by |W| and we still need to divide by |v|. The later patch calculates the sin of the angle so we can just divide by |V| squared and the sin.

Finding the fraction of U looks like more work, but since there isn't really a distinction between U and V, we can compute that in exactly the same way using another vector perpendicular to V. So I renamed W -> Y in the code and replicated it all with X for the U calculations. The 3rd patch is just optimization.

This is a really short explanation that may not be very clear. Maybe it will help.

phkahler:
I believe I still made a mistake in all that. If the numbered point are P1-P5, what we need is the ratio |P3| / |V|  which is the same as |P2| / |W|  I'll think about that for a bit as see if anyone else can weigh in.

phkahler:
We will definitely want to get this right. It looks like similar issue might exist in 3 more functions:

PointIntersectingLine()
ClosestPointOnThisAndSurface()
PointOnSurfaces()

I wish this were a major find for NURBS issues, but the way SolveSpace usually creates NURBS surfaces (straight extrusion or lathe) the U and V derivatives should always be orthogonal with the exception of doing lathe or revolve around an out-of-plane axis. Helix will always be affected, as will some of the possible future things.

phkahler:
Pushed another update with a math fix. Also used the new math in 2 more functions. Helical extrusions work dramatically better now. I'd still like to update the other couple places.

phkahler:
There should be 5 commits here. I'm not sure why sometimes it shows 3. AFAIK this is done. I don't want to add the U,V range limiting in other places until I have more test cases.

whitequark:
The math here is over my head, so I'd like @jwesthues to review this.

@phkahler Do I understand correctly that f961a71b323124a876cf321dc4e83bfbcb08dbd9, 0de8618fafed941883f3ade8507eea1bb6b3dfc5, c76d373bca0effeb6c0c95ac60202828f2b35d9e all fix the same bug in different functions?

phkahler:
If @jwesthues is going weigh in, I'd like your thoughts on changing the implementation of the function Vector::DivPivoting(). It's used in about 16 places, but I noticed it in ratpoly where trying to find a particular point on a curve. Very similar to what we're doing here finding a point on a surface.

The implementation seem to be avoiding the square root of a projection by selecting an axis to compute the ratio of a vector and "this" vector projected on it. The whole function can be replaced with:

`    return this->Dot(delta) / delta.Dot(delta);
`

This is mathematically nicer than what's in there and seems to work fine. I suspect it's faster too because straight-line code vs. branching. I would probably want to expand the dot products to:

`

    return (x*delta.x + y*delta.y + z*delta.z)

         / (delta.x*delta.x + delta.y*delta.y +delta.z * delta.z);
`

Which seems more likely to be inlined by the compiler. But then there is the issue of the name. It's not a bad name, but it wouldn't fit the intent with this change. Maybe DivProjDistByLengthOf?

As a math guy and a bit of a micro-optimizer this function kinda triggers me ;-)

jwesthues:
Sorry for the delay, traveling. I agree that your `DivProjecting()` is nicer than `DivPivoting()`. I have no objection to the change, though I'd be surprised/concerned if that changed any result. I haven't deeply reviewed the rest but it seems generally plausible to me, thanks.

maz3max:
I think I found another example that only works when `forcing NURBS surfaces to triangle meshes`.
I tested it with master and this version and both don't work. Is maybe my approach wrong or do you think its due to bugs?

fillet

[fillet.zip](https://github.com/solvespace/solvespace/files/3849334/fillet.zip)

phkahler:
@maz3max that seems to be the same problem as issue #315. This PR does not address that.

phkahler:
@whitequark Ping.  It looks like there may be question about the first change (limit U,V range)??. If that causes any problems it would likely be with planar surfaces, and we can put conditions around it to only apply to higher order surfaces (or remove it, it's independent of the others). I haven't seen any issues with it though, and neither has @ruevs as far as I know. This is really critical to helix booleans.

whitequark:
@phkahler The reason I haven't merged this yet is that, in a project with long history and many contributors, I expect the commits to be (a) self-contained and complete and (b) have messages that provide a detailed explanation of the change: why was it needed, what was changed, what problem does it solve, what is the full context. This allows anyone in the future who encounters any sort of issue with the changeset to easily see what happened and why. You can go through the git history to see what sort of goal I aim towards.

Now, not everyone tends to make their changesets the same way, and that's fine--normally I carefully review the commit, ask the author if necessary, and add all the missing details as well as merging or splitting commits as necessary. Unfortunately it takes a lot of time and lately I've had less time than I wish for maintenance, which results in delays. Perhaps I simply expect too much.

phkahler:
@whitequark that's fair. I didn't realize you were waiting on me. The ideal is probably to split out the first commit into its own PR to fix #471. The other 4 could be squashed into one with a single message "Change the math for projecting a point onto a plane to work better with non-orthogonal U,V derivatives. Fixes #472.". I'll try to make those changes soonish.

PR #474, opened 2019-09-09
Fix 468

### LINKED ISSUE #468: Crash on revolved sketch change
### System information

Recent master
Operating system: Fedora 30

### Expected behavior

File->New
Create a rectangle to the right of the origin.
Select the vertical axis and the origin.
Shift->V to revolve the rectangle
Select the sketch group in the text window
Select a corner of the rectangle and Shift->A to round the corner.

What should have happened?
The corner should be rounded
Going back to the revolved group the edge should be rounded.

### Actual behavior

What actually happened?
It crashed when trying to round the corner on the sketch.
Sometimes it's the first corner, sometimes it's the opposite corner.

### Additional information
/solvespace/src/dsc.h, line 425, function FindById:
Assertion failed: t != NULL.

rpavlik:
Fixes #468 by removing use of pointers after an invalidation-causing action (CopyEntity).

I briefly reviewed the other code in this massive switch, and the rest (besides helix and revolve) appears to be safe: there are a few other places where the address of a list element is taken, but it's either done in the tightest loop and not used after the invalidation, and/or something like `hEntity he = e->h; e = NULL;` is done, where then `SK.GetEntity(he)` is used after in place of `e`.  This seems sufficient, if a bit overkill: it's putting in a log(n) lookup by handle where there's already a good-enough index available (good enough that it's being used to control iteration, anyway.)

The last commit here is a matter of style preference: can't use an invalidated pointer if you don't take one in the first place. :) Take or leave, no worries.

The middle commit highlights some things I got confused about when reading the code, as well as adds the comments that got missed due to the feature being in-flight when I added them.

### DIFF
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -566,27 +566,26 @@ void Group::Generate(IdList<Entity,hEntity> *entity,
             AddParam(param, h.param(6), axis_dir.z);
 
             int ai = 1;
+            EntityList& ent = *entity;
 
             // Not using range-for here because we're changing the size of entity in the loop.
             for(i = 0; i < entity->n; i++) {
-                Entity *e = &(entity->Get(i));
-                if(e->group != opA)
+                if(ent[i].group != opA)
                     continue;
 
-                e->CalculateNumerical(/*forExport=*/false);
-                hEntity he = e->h;
+                ent[i].CalculateNumerical(/*forExport=*/false);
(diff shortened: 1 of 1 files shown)


rpavlik:
Looks like some of the CI issues at least are issues getting FreeType cloned.

whitequark:
Merged, thanks.

PR #484, opened 2019-09-12
Simplify UNION and DIFFERENCE boolean operations.

ruevs:
Union and difference are optimized by replacing the expression
  `(!inShell && !inFace)`
which is equivqlent to
  `(!inShell && !inSame && !inOpp)`
with
  `outSide`
which is qeuivalent, since `SShell::Class::OUTSIDE` is the only remaining possibility.


whitequark:
Thank you!

ruevs:
Thank you! For maintaining this beautiful project. I still need to go back to the NURBS code and really figure it out in order to make intersections work without hacks. And then perhaps start fixing boolean problems... but the code is complicated and... life happens :-)

whitequark:
Totally understandable. Life happens over here too...

Issue #292, opened 2017-09-14
Some question about the solver and *.slvs format


KmolYuan:
Hello. I made a Python 3 wrapper for the solver,

+ https://github.com/KmolYuan/python-solvespace

This project is using newest source code and compile without CMake. It can made `slvs.h` be import then used simply in a Python script (shown as [PyDemo](https://github.com/KmolYuan/python-solvespace/blob/master/PyDemo.py)).

If allowed, I would like to add it to the "solver" branch.

On the other hand, I want external programs can read and write *.slvs format, with fully Link / Assemble support, then solvespace files can be generated from C or Python.

Best regard.

Symbian9:
For the record, soon it could be used for implement SolveSpace solver in Blender 2.7x or later
* https://github.com/KmolYuan/python-solvespace/issues/2

whitequark:
@KmolYuan The "solver" branch exists because of copyright reasons, not for technical ones. If you are aiming for integrating your Python bindings, that should be done in the "master" branch, where active development happens. Are you fine with that?

KmolYuan:
Sure!

If it can be integrated into the CMake build process it would be better.

-----

By the way, the file [file.cpp] is using information from SK ([Sketch class]) of [SolveSpaceUI class], It seems to be unrelated to solver parts.

Is possible to separate SK and UI parts, or just using SolveSpaceUI directly to write a *.slvs format?

[file.cpp]: https://github.com/solvespace/solvespace/blob/master/src/file.cpp
[Sketch class]: https://github.com/solvespace/solvespace/blob/master/src/solvespace.h#L562
[SolveSpaceUI class]: https://github.com/solvespace/solvespace/blob/master/src/solvespace.h#L591

whitequark:
OK. I will ask you to sign a CLA very soon and then I will be able to integrate your code. I will handle CMake and basic testing for you.

Right now the solver that is exposed via the external interface can only solve for one group, and files have many. So, the exposed solver cannot interact with files. I'm not sure what's the best way to fix this, I've not really considered the binding evolution much.

KmolYuan:
For the file format section, maybe there should be a new external interface with the internal data format.

Because I want to put the results of the external design process into Solvespace. There would be used as auto-assemble function.

whitequark:
> For the file format section, maybe there should be a new external interface with the internal data format.

[...]

PR #419, opened 2019-05-21
WIP: Add an Emscripten port

whitequark:
## Overview

The purpose of this pull request is to give some visibility to the Emscripten port that has so far quietly existed as a branch. Its status is **an early preview**; enough things work that it demonstrates the concept is sound, but it cannot be used for anything useful yet. [Try it out!](https://files.whitequark.org/SolveSpace/solvespace.html)

Other than the issues listed below, everything works to my knowledge, and it's reasonably fast. If you encounter a bug not listed, or pathological slowness, please report it in a comment.

## Known issues

The major missing features are:
  - [x] Tooltips are not implemented.
  - [x] Preferences are not saved.
  - [ ] Any way to open or save files. This is particularly bad because of the way SolveSpace handles assemblies. It is likely that #75 will be required before this port is usable in any meaningful way.
  - [ ] Some message dialogs (the ones that require a choice, specifically) are written in a blocking style and ~need to be~ may be rewritten with callbacks. It is also possible to use "new Asyncify" as of 2020, and might be preferable to do so.
  - [ ] There's no way to move the "split" between main window and property browser.
  - [ ] Scrollbars are not implemented, so there's no way to scroll the property browser.

The following are known bugs:
  - [x] Menu mnemonics are broken, the Alt+... hotkeys don't trigger them. No idea why.
  - [x] Pressing Esc does not dismiss message dialogs.
  - [x] Pressing Esc while editing a property also invokes the Esc keybinding, changing the open page in the property browser.
  - [ ] Fullscreen is broken. The browser hard-stretches the entire canvas over the entire screen, ignoring element sizing. This is apparently required by the spec?

Additionally, there are some Firefox-specific issues:
  - [ ] Fractional devicePixelRatio factors don't work (result in broken rendering).
  - [ ] Something about the styling and/or \<canvas\> doesn't work the same, and the canvas is stretched vertically under the menu.
  - [ ] Shift-RMB cannot be used to rotate the sketch as that is hardcoded to bring up the browser menu, as a security feature. (It still rotates but the menu is in the way and input handling code can get confused.)

## Working on the port

The missing features can all be quickly located in the codebase by searching for the regex `FIXME\((async|emscripten)\)`. The instructions for building it are located in the [updated `README.md`](https://github.com/solvespace/solvespace/blob/emscripten/README.md) in this branch.

If you implement a missing feature (please do!) you can send it as a PR against this branch.

## Note on JavaScript libraries

[...]

PR #493, opened 2019-10-12
Python wrapper of the solver

KmolYuan:
#292 The first pull request of Python wrapper branch.

New folder:
+ `cython`

Modified files:
+ ` .gitignore`: Ignore the project settings of JetBrains IDE.
+ `.travis.yml` and `appveyor.yml`: Deployment of PyPI.
+ `src/platform/platform.cpp`: Add a type cast for MinGW and newer version compiler.


ruevs:
@vespakoen are you aware of this?
https://github.com/solvespace/solvespace/pull/419
https://files.whitequark.org/SolveSpace/solvespace.html

vespakoen:
Yes I have seen the webassembly version, but I would like to package up the **core only** as a Node.js / browser package.

blooop:
I'd just like to say that I am very interested in python bindings for an up to date version of solvespace, so this pr looks great.  

Thanks

vespakoen:
I think providing SolveSpace's solver as a library would be great for many other projects, and also hopefully would get more people working / improving on it.

It would be great if the SolveSpace CI can produce these libraries automatically for every platform.

The most "universal" way I think is by using pybind (For Python), embind (For the web) and n-api (For node.js), they are all very similar, and work the best by wrapping a class.

Also, by moving as much logic as we can into a C++ class, the bindings can be smaller, and there is less "wrapper code" that we would have to maintain.

I would happily take on the job of hooking it into our CI to have it produce the libraries for all the platforms, I just don't know what this class should look like exactly.

Anyways, this is definitely something I want to take a look at in the near future.

blooop:
That sounds great.  I agree about pybind11 for python bindings,iI found its very easy to generate bindings with minimal extra code/maintenance. 

I have been looking into cadquery for code based CAD, but am still struggling to work out the right code to do things.  
I read this paper:
https://onlinelibrary.wiley.com/doi/full/10.1111/cgf.14046

which basically says, gui cad is quick and easy to use, but not as robust as code, but code is difficult to understand.  They come up with a dual system which allows gui creation but also saves a code representation so gets the best of both worlds.  

I was thinking of trying to do something similar with solvespace but I'm guessing it will be a lot of work.

vespakoen:
CadQuery is a very cool project indeed, there is also OpenSCAD (and OpenJSCAD) which might be of interest to you.

And I guess we should not forget "SWIG" for generating external bindings to SolveSpace, maybe it's even the least invasive but I have no experience using it yet.

blooop:
I started out with openSCAD, but it doesn't have operations like fillet or shell and the language itself is not as nice a python.  

Cadquery fixes those problems but doesn't have constraints so is still rather procedural.  Freecad in theory has everything, but when I started implementing basic programmatic 2D sketching constraints everything was very slow and crashed all the time.

  
And now we come to solvespace, which appears to have everything.  I'm still very early in the research stage so I need to investigate the solvespace code more, but I've been using the editor as my CAD of choice for a while now and have been impressed so far. 

In terms of SWIG, I have not used it, but did not like the look of it compared to pybind11 which I've been using on a large project quite happily for over a year now.

phkahler:
@vespakoen  The reason this is still not merged is I'm not sure what to make of it, and not planning to use it soon. Not even sure what the use cases are yet - to bring the solver to python programs, or to bring python to solvespace? It obviously useful to someone! If you want to take it on we can merge it - probably after 3.0.

It would also be good to squash most of the commits. Down to 5ish?

blooop:
In terms of the use case of bringing the solver to python, it would be useful because then I could add constraints to Cadquery relatively easily which removes one of its biggest weaknesses.

ruevs:
> we can merge it - probably after 3.0.
> 
> It would also be good to squash most of the commits. Down to 5ish?

Does a python whapper of the solver belong in the same repository as the application?

robnee:
> Does a python whapper of the solver belong in the same repository as the application?

The python module seems to depend on mostly the same src/ files as libslvs.  Could the python module depend on libslvs instead?

vespakoen:
Hello Yuan Chang, thanks a lot for this work!

I'd really like to merge this into SolveSpace, so we can have a "officially supported" SolveSpace PyPi package.

It would have been nice if this used the project's CMake, but since it is currently very hard to build the "solver only" with CMake, I can totally see why that has been avoided.
This might be something we can improve in the future, but for now, your setup.py seems to work very well.

One suggestion I have is to use "cibuildwheel" to compile the packages.
This uses the manylinux docker images that have an old glibc version, to ensure compatibility with older linux versions.
I have added it in a fork, and it seems to work pretty well, see: https://github.com/vespakoen/solvespace/tree/python

@jwesthues Can you create an account on PyPi (https://pypi.org/account/register/) and on Test PyPi (https://test.pypi.org/account/register/) and add the API keys to the CI secrets? (as `PYPI_API_TOKEN` and `TEST_PYPI_API_TOKEN`)

@ruevs @phkahler are you ok to merge this?

jwesthues:
Very cool, thanks. I'll await @phkahler's thoughts on merging, and create those accounts assuming it all makes sense to him.

phkahler:
@vespakoen  @jwesthues   I've let it sit because I don't want the maintenance burden. If someone else is willing to deal with this I'm fine with merging it - assuming it makes sense. I've looked into the details of this exactly zero minutes ;-)

vespakoen:
To me this looks like the most up-to-date, minimal implementation for a python wrapper of the solver.

It is also a part of SolveSpace that is very loved, and I think it would be good to have it in the official repository.

I am also fine spending some time on it to make sure it works and keeps working in the future.

phkahler:
@vespakoen  rebase and merge it and you have the job ;-)

vespakoen:
Thanks a lot for this contribution!
A variant of this got merged to master, so I believe this can now be closed

PR #503, opened 2019-11-23
Fix CMake Xcode generator.

vespakoen:
By setting `CMAKE_RUNTIME_OUTPUT_DIRECTORY` to `$<1:${CMAKE_BINARY_DIR}/bin>`

Without this, the resources did not end up into the SolveSpace.app file, causing a crash on startup (where it was unable to load the locales.txt file)

The reason is that the XCode project puts things into the "bin/Debug/SolveSpace.app" or "bin/Release/SolveSpace.app" folders (depending on the target), and some files ended up in "bin/SolveSpace.app" instead, this seems to fix the locales ending up in the wrong place, but I still see some files (libraries) ending up in "bin/Debug" or "bin/Release" folders.

Here is a list of them:

```
$ ls bin/Debug
libcairo.a               libpixman-1.a            libslvs.dylib
libdxfrw.a               libpng16d.a              libsolvespace-core.a
libflatbuffers.a         libslvs.1.dylib          libsolvespace-headless.a
libfreetyped.a           libslvs.3.0.dylib        libz.a
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,9 @@ set(OPENGL 3 CACHE STRING "OpenGL version to use (one of: 1 3)")
 
 set(EXECUTABLE_OUTPUT_PATH ${CMAKE_BINARY_DIR}/bin)
 set(LIBRARY_OUTPUT_PATH    ${CMAKE_BINARY_DIR}/bin)
+if("${CMAKE_GENERATOR}" STREQUAL "Xcode")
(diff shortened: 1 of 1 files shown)


whitequark:
Thank you for the PR. The SolveSpace project has a CLA. It looks like our CLA tracking tool requires your commit email to be added to the GitHub account—can you take a look at that?

Regarding libraries ending up in `bin/Debug`, none of those are runtime dependencies of `SolveSpace.app`, so it doesn't matter where they go.

PR #502, opened 2019-11-23
Skip temporary copying to framebuffer.

vespakoen:
This is work done by @troughton (see: https://github.com/troughton/solvespace/commit/42f187aac1585de4e274118823b368edafd02b27).

See his comment in this issue: https://github.com/solvespace/solvespace/issues/354
I just wanted to turn it into a PR on this repository so more people can see it, and possibly try it out / continue working on it.

I also fixed some deprecation warnings I got in XCode on Catalina.

I can compile and run this, it make SolveSpace very fast again!
However, there is some flashing in the toolbar occurring now and then, and there are probably other issues that I haven't stumbled into yet.

Notice that this PR does not bring in the changes that @troughton made to `src/render/gl3shader.cpp` since I am not sure what the purpose of these changes are.

Can someone tell me how to crash it / what needs to be done to get this working reliably?

Issue #354, opened 2018-09-27
Significant performance regression in macOS 10.14


prof-foster:
Performance on ...

- iMac Pro
- macOS 10.14

... with both a fresh install of SolveSpace version 2.3 and a fresh build from GitHub has regressed to the point of not being usable. Creating a line and moving the second point results in a redraw time of roughly 1 sec.

Apple has announced that OpenGL is being deprecated starting with 10.14, but I haven't seen any references online to performance issues. In the short term I'm wondering about the possibility of a quick performance fix. In the longer term is SolveSpace still going to be feasible on macOS?

whitequark:
Unfortunately it is not possible for me to do any fixes to SolveSpace under macOS 10.14 because I do not own any Apple hardware and so I use a VM ([OSXKVM](https://www.contrib.andrew.cmu.edu/~somlo/OSXKVM/)) to do all development work and there's currently no way to run anything newer than 10.12 in a VM. Worse still, anything newer than 10.9 is absurdly slow when using software rendering, and I do not have GPU passthrough working yet.

whitequark:
That said, if you could use Xcode to profile SolveSpace, I might be able to fix it without even having access to a macOS 10.14 machine.

prof-foster:
I'm new to profiling but happy to help. I'm currently trying to follow the build instructions using the "generate an XCode project" path in order to generate a binary that I can profile at the function call level. I had to start again from "scratch" as CMake error-ed out.

Looks like the error is continuing:

-- Performing Test USE_LOONGSON_MMI - Failed
CMake Error at extlib/pixman/CMakeLists.txt:95 (if):
  if given arguments:

    "EQUAL" "4"

  Unknown arguments specified

This is after a `brew reinstall cmake`

whitequark:
What is your cmake version?

prof-foster:
cmake version 3.12.2 installed as follows:
`
$ brew reinstall cmake
==> Reinstalling cmake 
==> Downloading https://homebrew.bintray.com/bottles/cmake-3.12.2.mojave.bottle.tar.gz
Already downloaded: /Users/jafoster/Library/Caches/Homebrew/downloads/39c2f688611df67dad1815faf296167cafc669472e60e74b900d819111d8b456--cmake-3.12.2.mojave.bottle.tar.gz
==> Pouring cmake-3.12.2.mojave.bottle.tar.gz
`

[...]

PR #504, opened 2019-11-23
Remove GL framebuffer on macOS

vespakoen:
This is work done by @troughton (see: troughton@42f187a).

See his comment in this issue: #354
I just wanted to turn it into a PR on this repository so more people can see it, and possibly try it out / continue working on it.

I can compile and run this, it make SolveSpace very fast again!
However, there is some flashing in the toolbar occurring now and then, and there are probably other issues that I haven't stumbled into yet.

Notice that this PR does not bring in the changes that @troughton made to src/render/gl3shader.cpp.
Those seemed to be another optimisation that I don't fully understand, and we might be able to do without.

Can someone tell me how to crash it / what needs to be done to get this working reliably?

This PR is supersedes #502

### DIFF
--- a/src/platform/guimac.mm
+++ b/src/platform/guimac.mm
@@ -344,7 +344,7 @@ MenuBarRef GetOrCreateMainMenu(bool *unique) {
 // Cocoa NSView and NSWindow extensions
 //-----------------------------------------------------------------------------
 
-@interface SSView : NSView
+@interface SSView : NSOpenGLView
 @property Platform::Window *receiver;
 
 @property BOOL acceptsFirstResponder;
@@ -362,8 +362,6 @@ MenuBarRef GetOrCreateMainMenu(bool *unique) {
 
 @implementation SSView
 {
-    GlOffscreen         offscreen;
-    NSOpenGLContext    *glContext;
     NSTrackingArea     *trackingArea;
     NSTextField        *editor;
 }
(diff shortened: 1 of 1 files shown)


vespakoen:
From the other PR:
> If you add a constraint, and double click on the text, can you edit it? Or does it disappear / become black eventually? What about scrollbars and tooltips? That's the reason it uses offscreen rendering right now.

@whitequark I don't see any of this above issues, the only thing I see that is strange is that the toolbar sometimes disappears for a frame or 2.

whitequark:
> I don't see any of this above issues

Great! I assume they finally fixed whatever issue there was with GL layers. I'm going to recheck if this still works on Mavericks, I think.

Feel free to remove all traces of `GlOffscreen`, since it was only used for Gtk 2 and macOS and we have long migrated off Gtk 2.

whitequark:
> the only thing I see that is strange is that the toolbar sometimes disappears for a frame or 2.

Is it just the toolbar? Or does selection marquee and fps counter also blink?

whitequark:
I think your patch at least breaks non-HiDPI screens.

whitequark:
That was caused by a missing
```
self.wantsBestResolutionOpenGLSurface = YES;
```

vespakoen:
Cool! Shall I add it to this PR or will you do it along with the other changes you made?

whitequark:
I'll do it.

vespakoen:
Great, thanks a lot! it's great to have a butter smooth SolveSpace again =)

whitequark:
Fixed in 9b61f988. Let me know if there are any issues.

vespakoen:
Working great on my end! Thanks a lot!

vespakoen:
Here is a macOS build from the latest master 54015b6777d2195d54f1738d80c5aa20f194a6b8 
I signed and notarised it using my Apple Developer account for anyone interested to check it out.

[solvespace.zip](https://github.com/solvespace/solvespace/files/3882638/solvespace.zip)

PR #505, opened 2019-11-23
Fix Xcode auto-fixable issues

vespakoen:
These are the auto-fixable fixes that Xcode could do on macOS Catalina (Xcode v11.1).
There are still a bunch of warnings but they are all related to OpenGL being deprecated.

### DIFF
--- a/src/platform/guimac.mm
+++ b/src/platform/guimac.mm
@@ -225,9 +225,9 @@ public:
 
         NSUInteger modifierMask = 0;
         if(accel.shiftDown)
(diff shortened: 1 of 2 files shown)


whitequark:
Thank you. Do you happen to know how far back this breaks compatibility with? I don't have a physical machine with macOS and unfortunately VM support [became largely impractical by Sierra](https://www.contrib.andrew.cmu.edu/~somlo/OSXKVM/). The newest VM I have that still actually boots is Mavericks.

Issue #483, opened 2019-09-11
Snap Travis integration


ppd1990:
As outlined in https://github.com/solvespace/solvespace/pull/436 we'd like  to have build on commit and also release to the edge channel.

## Using ```snapcraft remote-build```

The current plan involves using ```snapcraft remote-build``` to build for all possible architectures. This feature is still in beta @ https://github.com/snapcore/snapcraft/tree/remote-build and will receive further simplifications, so I propose waiting for that to be properly finished.

A sample implementation is here: https://github.com/ppd1990/solvespace-travis

## Using docker

If that does not work out so well for any reason, we can still use docker to build for Travis' native architectures https://github.com/solvespace/solvespace/pull/436#issuecomment-502370983

Cross-builds for other architectures should be possible with qemu, but I haven't looked into that enough.

Multiarch support for docker: https://github.com/multiarch/qemu-user-static
Travis with Docker and QEMU: https://developer.ibm.com/linuxonpower/2017/07/28/travis-multi-architecture-ci-workflow/

## Architectures to support

It wouldn't hurt to define what architectures we'd like to build for. ```amd64``` and ```i386```for sure, ```armhf``` and ```arm64``` make sense too for a desktop application.

ppd1990:
And here we go: https://forum.snapcraft.io/t/call-for-testing-snapcraft-3-9/13944

Snapcraft 3.9 is pre-released and includes remote-build support. Let's do it.

whitequark:
Sounds good. Let me know once it becomes stable.

ppd1990:
Certainly. We're going to need a solvespace launchpad account for remote builds though. I'll test with mine in the meantime

ppd1990:
Is there a particular reason for ```dist: trusty```? The Travis snap addon requires xenial or above and I'd rather use that if possible.

whitequark:
I don't recall if there was. It should be fine to upgrade to Xenial.

ppd1990:
This was implemented by: 
1. a60d4df
2. d5351c4
3. 2dcd81d

Snaps are being built and deployed correctly: https://travis-ci.org/solvespace/solvespace/jobs/640723069

I consider this completed.

whitequark:
Is this process supposed to upload 4 separate releases, each supporting 1 architecture? Seems suboptimal.

ppd1990:
The revision is a strictly increasing integer per project and applies to all uploaded snaps. So yes, it is indeed meant to be that way.

[...]

Issue #482, opened 2019-09-11
Transfer snap ownership to solvespace organization


ppd1990:
With https://github.com/solvespace/solvespace/pull/436 being almost ready for merging, I propose the following steps regarding the publishing of the snap:

- [x] Create a ```solvespace ``` account on snapcraft.io
This gives the solvespace snap store listing an official touch and at some point in the future, it can be verified by the store and appear with a little badge or checkmark.
- [x] I request to transfer ownership of the snap to the newly created account on the snapcraft forum: https://forum.snapcraft.io/c/store-requests
- [x] If necessary, the snap's administration can be shared with other accounts. A shared secret is not necessary. The link for that would be https://dashboard.snapcraft.io/snaps/solvespace/collaboration/

whitequark:
@jwesthues Could you create an account on snapcraft.io for SolveSpace and add me as a collaborator?

jwesthues:
Cool, `solvespace` account created. Let me know when the transfer is complete, and I'll add you (and my own personal account, and whoever else we should have) as collaborators.

ppd1990:
The transfer request was filed here:
https://forum.snapcraft.io/t/please-transfer-solvespace-to-solvespace-account/14342

ppd1990:
@jwesthues Have you been contacted by someone from the Snap Store? I presume they'll require your OK first. You could probably also claim the `solvespace` name from your account. That'll get the ball rolling a little faster, I think.

jwesthues:
No contact yet. I'm traveling now, but will try claiming it if no action by my return Sunday.

jwesthues:
No contact yet, sent the claim today.

ppd1990:
Nothing yet. Will poke the store admins some more...

ppd1990:
Poking was successful. I've been contacted already and have confirmed the transfer request. @jwesthues You should you have mail from Canonical

ppd1990:
The transfer has been completed. Please note that I am still (by default) an active collaborator of the snap. So don't forget to remove me if necessary.

jwesthues:
I got the transfer, added @whitequark as a collaborator, left @ppd1990 just in case. Thanks, and let me know if any other action required.

whitequark:
@jwesthues It looks like I can't access the snap--can you recheck?

ppd1990:
@whitequark Your invite is still pending. You need to respond to the email

whitequark:
Thanks, that worked.

[...]

PR #506, opened 2019-11-24
snap: Build snaps on Travis via remote-build

ppd1990:
This PR introduces building & publishing of snaps into the CI pipeline.
Implements #483 

~A bit of refactoring while at it:~
- ~Fix invalid osx_image xcode8.2~
- ~Drop depreceated sudo keyword~
- ~Run debian build on bionic image & re-enable building with ui~
- ~Split build stage into separate jobs for better separation of concerns~

~I guess the unrelated changes should be split out into separate PRs if the general approach is considered sane.~

The above was merged in #510 

## What's left to do
- [x] Wait for snapcraft 3.9.x to make it into stable with remote-build included
- [x] Create a solvespace launchpad account and replace my personal credentials with it
- [x] Transfer ownership of the solvespace snap  #482 
- [x] Use someone else's snapcraft token for deploying:
```
snapcraft export-login --snaps=solvespace --channels=edge snapcraft-login
travis env set SNAP_TOKEN "$(cat snapcraft-login)"
```
- [x] Decide on what architectures we want to build snaps for

##


ppd1990:
An exemplary build: https://travis-ci.org/ppd1990/solvespace/builds/616292318

.travis.yml
```diff
@@ -1,24 +1,50 @@
...
+      deploy:
+        provider: releases
+        api_key:
+          secure: dDlkIawHcODlW9B/20/cQCtzeoocvs0hKuNngRKXKqzXLWTRq33oq/B7+39tAixWbmv6exTpijiKrRNFiSCW5Z4iwHLwaRD4XJznxw63e/Hus/dxg2Tvqx7XFpkCz8mT1Z+gZQE5YxAngeZPpI/sZbZtF1UO3yH5eLeeokZ15p26ZskQUPoYuzrTgTzYL3XfpG3F+20rNBawH1ycsCTVD/08/n31d2m3CrKAsbW7er92ek6w4fzKr7NW8WeXjrPJETVpw5fQg1Od3pRGW8dPQaJcvKQEogMp8Mm0ETYd0qigg89/giBz7QwOgmAWQ4dH+DfZH4Ojl//127QztBolMvyDMQBykWrtJoGcij05sT6K2IJr2FHeUBO12MAEdjiVvhQj3DtTzjPiZAHHDBSLWxLKWWhlhHE4pq7g1MQhqXkaAHI2BLNzwLmaowbMT0bECf9yfz6xx18h6XPQFX44oOktraobVALFlyHqeKa8zdcUt22LF6uAL1m5dxL0tny3eXCIPE4UH/RZgua/cHV9G3cUvKQa/QnFSLRhvWVSbGB+7YsHouBJcsUOOW1gmd5442XuC7mpppccRldh+GSxUk6TBJRAx7TeQ0ybDUaoco9MUqp2twv3KreR2+8Q12PDaAhfQVNEGdF3wTm1sShImjCN4VN3eSLlBEbve1QRQXM=
+        skip_cleanup: true
+        file: build/SolveSpace.dmg
+        on:
+          repo: solvespace/solvespace
+          tags: true
+          condition: "$TRAVIS_OS_NAME == osx"
```
whitequark: This condition seems redundant.

ppd1990: Good old copy & paste. I concur, it is redundant.

whitequark:
Great work refactoring the Travis config! I would prefer if it went in as a separate PR so we can get the benefits of it before snapcraft releases their changes.

ppd1990:
Do you think we should refine the Travis config further into ```test``` & ```deploy``` stages or do you feel that's overkill for now? At the moment, a OSX or Debian test fail wouldn't stop a snap package from being built & released (to edge that is).

whitequark:
> Do you think we should refine the Travis config further into `test` & `deploy` stages or do you feel that's overkill for now?

That would be great! We definitely shouldn't publish packages that don't pass tests.

ppd1990:
We have no easy build artefact sharing between Travis stages though. Suppose we want something like this:

## Stage: Test
- Debian test build
- OSX test build

## Stage: Deploy (run only if all jobs in Test succeed)
- OSX production build -> deploy 
- Snap production build -> deploy

Then we need to do the OSX build twice.

Also: Do we want to block snap builds if OSX fails? Do we want to block all deploy jobs when a single one of the test jobs fails?

whitequark:
> Do we want to block snap builds if OSX fails?

No.

> Do we want to block all deploy jobs when a single one of the test jobs fails?

The tests are supposed to be completely platform-independent. (There's a few tricky parts, mostly having to do with FP precision issues and such.) So I think we should run tests on every platform, but as a redundant mechanism to alert us if they somehow are *not* platform-independent; they ideally always should be.

ppd1990:
With the currently merged Travis config, we run sequentially like so:

```
1. Test on Debian 
2. Test & Deploy on OSX
```

Adding snap:

```
1. Test on Debian 

In parallel:
2.1 Test & Deploy on OSX
2.2 Build & Deploy Snap
```

So OSX & Snap are blocked by Debian/Ubuntu test fails and of course by their own fails. I feel that Snap deployments are rather uncritical as we'd only ever manually promote revisions to ```candidate```, ```stable``` and ```beta```.

.travis/install-macos.sh
```diff
@@ -2,3 +2,4 @@
 
 brew update
 brew install freetype cairo
```
vespakoen: I noticed the macOS build is failing, and it might be fixed (and become faster) by removing the brew update & brew install statements.

I don't think we need to update brew and also don't need freetype & cairo since they are already bundled as git submodules.

It seems we only need git & cmake which should already be installed by default on Travis.

whitequark: > I don't think we need to update brew and also don't need freetype & cairo since they are already bundled as git submodules.

My position is that libraries that are available on the system should not be linked as submodules, and also should be linked dynamically, if possible at all.

vespakoen: Agreed, however, cairo and freetype are not installed by default as far as I know, so I think we should make sure they are included (on macOS and windows deliverables) also, the CI should probably test the dependencies that are linked as submodules to make sure they work.

I have seen many OSS failing on macOS due to some missing library and otool workarounds (even on this repo if I remember correctly) and I guess that is not what we want.

As an added bonus, I guess the libraries that are in use are not that big at all.

whitequark: > I have seen many OSS failing on macOS due to some missing library and otool workarounds (even on this repo if I remember correctly) and I guess that is not what we want.

They're now linked statically instead, which is what I should have done in the first place.

> the CI should probably test the dependencies that are linked as submodules to make sure they work.

Our CI is already quite slow and testing the dependency version that is updated once a few months at best is a waste of time.

ppd1990:
Note to myself: `--launchpad-user` will not be necessary in the future

https://github.com/snapcore/snapcraft/pull/2825

ppd1990:
Snapcraft 3.9.2 is in the stable channel

.travis/build-snap.sh
```diff
@@ -0,0 +1,15 @@
...
+mkdir -p "$lp_data_dir"
+openssl aes-256-cbc -K $encrypted_02880a344e9c_key -iv $encrypted_02880a344e9c_iv \
+  -in .travis/launchpad-credentials.enc \
+  -out "$lp_credentials" -d
+chmod 600 "$lp_credentials" 
+
+./pkg/snap/build.sh remote-build \
+  --launchpad-user ppd \
+  --launchpad-accept-public-upload \
+  --build-on=amd64,arm64,armhf,i386
```
ppd1990: I feel those arches are quite sufficient to cover most desktop-like Linux machines in the wild. @whitequark Any strong opinion from your side?

.travis.yml
```diff
@@ -25,3 +25,19 @@ jobs:
...
+      addons:
+        snaps:
+          - name: snapcraft
+            confinement: classic
+      script: ./.travis/build-snap.sh
+      deploy:
+        provider: snap
+        skip_cleanup: true
+        snap: pkg/snap/*.snap
+        channel: edge
```
ppd1990: We could also try promoting tagged builds to `candidate` or `beta` directly

.travis/build-snap.sh
whitequark: LGTM

.travis.yml
whitequark: `beta` seems fine

ppd1990:
A little closer to the finish line: https://travis-ci.org/ppd1990/solvespace/builds/639505699

Nothing much left to do, I guess. Snapcraft 3.9.7 is stable, Launchpad is happy, Travis is happy. Are we happy?

whitequark:
Yup all looks good. Let me know how I should merge this and we're done.

ppd1990:
1. We need a launchpad account for building. Just register one solely for solvespace.
Then you need to use it once via remote-build:

```bash
./pkg/snap/build.sh remote-build \
  --launchpad-user solvespace \
  --launchpad-accept-public-upload \
```
You will now have those credentials stored in `$HOME/.local/share/snapcraft/provider/launchpad/credentials`. 

2. Encrypt them in Travis: `travis encrypt-file credentials` and replace my `launchpad-credentials.enc` with the encrypted file in this merge request. Adjust the openssl command in `build-snap.sh` accordingly.

3. Login in snapcraft into the solvespace account: `snapcraft login` and then do
```bash
snapcraft export-login --snaps=solvespace --channels=beta,edge snapcraft-login
travis env set SNAP_TOKEN "$(cat snapcraft-login)"
```
Those are the credentials for pushing & releasing to the store.

ppd1990:
Re the failed to deploy error: I found this in `dpl` (deploy tools for Travis):

```ruby
      def validate
        error :no_snaps if snaps.empty?
        error :multiple_snaps if snaps.size > 1
      end
```

I guess the deployment is meant to only handle one snap at a time.

ppd1990:
See #543

PR #507, opened 2019-11-24
Adding .csv trace file import functionality

vespakoen:
Don't know how useful this is, but if it is, let me know and I will fix the translations.

Also, we might be able to add a menu option that turns the trace directly into a line in stead of having to go through the save trace -> import flow.

Also, I am not a professional C++ programmer so let me know if I am doing anything wrong.

Cheers

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -183,6 +183,7 @@ set(solvespace_core_SOURCES
     graphicswin.cpp
     group.cpp
     groupmesh.cpp
+    importcsv.cpp
     importdxf.cpp
     mesh.cpp
     modify.cpp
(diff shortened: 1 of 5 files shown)


whitequark:
What's the use case?

vespakoen:
Here is a good example: https://www.youtube.com/watch?v=i6tDWJsNsok

whitequark:
Hmm, interesting. How do you deal with dragging producing a large number of points, causing a slowdown? Do you use "Step Dimension"?

Otherwise, I think it is reasonable to support this but it would probably be better with a menu option.

vespakoen:
Yeah I am using step dimension over here but the code ignores duplicate
coordinates which might help already, we could perhaps add a minimum
distance of points for the import.

PR #508, opened 2019-11-24
Fix crash on macOS Catalina where GL info is missing

vespakoen:
Opening the preferences no macOS Catalina crashes SolveSpace.
This is due to the way the gl_vendor / gl_renderer and gl_version is retrieved.
We use `glGetString` in `rendergl3.cpp` (in the `getIdent` function).

Somehow this doesn't seem to work anymore in Catalina.

Let me know if anything needs to change.

### DIFF
--- a/src/confscreen.cpp
+++ b/src/confscreen.cpp
@@ -350,10 +350,14 @@ void TextWindow::ShowConfiguration() {
     if(canvas) {
         const char *gl_vendor, *gl_renderer, *gl_version;
         canvas->GetIdent(&gl_vendor, &gl_renderer, &gl_version);
-        Printf(false, "");
-        Printf(false, " %Ftgl vendor   %E%s", gl_vendor);
-        Printf(false, " %Ft   renderer %E%s", gl_renderer);
-        Printf(false, " %Ft   version  %E%s", gl_version);
+        if (!gl_vendor) {
+            Printf(false, "  gl info not found");
+        } else {
+            Printf(false, "");
+            Printf(false, " %Ftgl vendor   %E%s", gl_vendor);
+            Printf(false, " %Ft   renderer %E%s", gl_renderer);
+            Printf(false, " %Ft   version  %E%s", gl_version);
+        }
     }
 }
 


whitequark:
`glGetString` can return `NULL` if for some reason there is no current OpenGL context. But that should not be possible by the time when you click on the preferences link.

Could you catch the crash with a debugger (or just put a breakpoint at `GetIdent`), and print the value of `[NSOpenGLContext currentContext]` at that point? It should not be `nil`.

vespakoen:
So it looks like our change that removed the framebuffer caused this, the code below fixes that:

I'll make a new PR for it if you are happy with it.

Screenshot 2019-11-25 at 10 48 42


```
diff --git a/src/platform/guimac.mm b/src/platform/guimac.mm
index 33901ff..b8aa938 100644
--- a/src/platform/guimac.mm
+++ b/src/platform/guimac.mm
@@ -385,6 +385,7 @@ - (id)initWithFrame:(NSRect)frameRect {
         editor.bezeled = NO;
         editor.target = self;
         editor.action = @selector(didEdit:);
+        [[self openGLContext] makeCurrentContext];
     }
     return self;
 }
(END)
```

whitequark:
Your fix is a bit dirty. The platform code does not really guarantee that there will be an active GL context outside of onRender, and it should not. What about this?
```patch
diff --git a/src/render/rendergl3.cpp b/src/render/rendergl3.cpp
index b3ae1fa6..e4faf4d8 100644
--- a/src/render/rendergl3.cpp
+++ b/src/render/rendergl3.cpp
@@ -87,6 +87,7 @@ public:
         Fill       *fill;
         std::weak_ptr<const Pixmap> texture;
     } current;
+    const char *vendor, *renderer, *version;
 
     // List-initialize current to work around MSVC bug 746973.
     OpenGl3Renderer() :
@@ -440,6 +441,10 @@ void OpenGl3Renderer::Init() {
     meshRenderer.Init();
     imeshRenderer.Init();
 
+    vendor   = (const char *)glGetString(GL_VENDOR);
+    renderer = (const char *)glGetString(GL_RENDERER);
+    version  = (const char *)glGetString(GL_VERSION);
+
 #if !defined(HAVE_GLES) && !defined(__APPLE__)
     GLuint array;
     glGenVertexArrays(1, &array);
@@ -696,9 +701,9 @@ std::shared_ptr<Pixmap> OpenGl3Renderer::ReadFrame() {
 }
 
 void OpenGl3Renderer::GetIdent(const char **vendor, const char **renderer, const char **version) {
-    *vendor   = (const char *)glGetString(GL_VENDOR);
-    *renderer = (const char *)glGetString(GL_RENDERER);
-    *version  = (const char *)glGetString(GL_VERSION);
+    *vendor   = this->vendor;
+    *renderer = this->renderer;
+    *version  = this->version;
 }
 
 void OpenGl3Renderer::SetCamera(const Camera &c) {
```

vespakoen:
Cool! Yes that looks a lot better. I will test it after work (in about 4 hours)

vespakoen:
Yes that worked as well! Shall I make a PR for it or will you apply it yourself?

whitequark:
Fixed in 47c00f9a.

PR #512, opened 2019-11-26
Immediately edit dimension after adding it

vespakoen:
Having to double click the dimensions after adding them can be quite annoying when adding a lot of dimensions, I think the default should be to directly show the input field when adding a dimension.

This PR changes that.

Hope it's helpful.

### DIFF
--- a/src/constraint.cpp
+++ b/src/constraint.cpp
@@ -196,6 +196,7 @@ void Constraint::MenuConstrain(Command id) {
             c.valA = 0;
             c.ModifyToSatisfy();
             AddConstraint(&c);
+            SS.GW.EditConstraint(c.h);
             break;
         }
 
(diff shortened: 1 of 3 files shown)


neuks:
This works good, and more user friendly, I have added it to my local source pool, thanks.

whitequark:
I don't think changing the defaults to benefit one person's choice of workflow at the expense of everyone else adjusting their workflow in response is appropriate.

vespakoen:
Would love to hear the alternate workflow that I am unaware of then, I have seen some references of a grid in the code, so I guess that has to do with it?

My girlfriend gave up on my little solvespace tutorial once she was adding in dimensions.

Directly entering the value also is the default of most other CAD programs as far as I know.

I think this should be reconsidered or added as an option to the config.

whitequark:
> Would love to hear the alternate workflow that I am unaware of then

Double-clicking the dimension after adding it.

A configuration option for this seems fine.

vespakoen:
Ok I will add a configuration option for it later.

whitequark:
I've reopened the PR and I'll merge it with an option added.

vespakoen:
See #515 514

PR #513, opened 2019-11-26
Remove missing rendergl.cpp file from CMakeLists.txt

vespakoen:
macOS build broke with 07992cecaaeeeb0584f3c655eaaf32ba09bd3e6d this fixes that.


whitequark:
Thanks!

PR #514, opened 2019-11-26
Adding config option to immediately edit dimensions

vespakoen:
Here it is again, this time with config option.

Let me know if the default should be on or off, and if the variables / place in the preferences pane are fine with you.

The option does now also work on the angle constraint.

### DIFF
--- a/src/confscreen.cpp
+++ b/src/confscreen.cpp
@@ -102,6 +102,11 @@ void TextWindow::ScreenChangeTurntableNav(int link, uint32_t v) {
     }
 }
 
+void TextWindow::ScreenChangeImmediatelyEditDimension(int link, uint32_t v) {
+    SS.immediatelyEditDimension = !SS.immediatelyEditDimension;
+    SS.GW.Invalidate(/*clearPersistent=*/true);
+}
+
 void TextWindow::ScreenChangeShowContourAreas(int link, uint32_t v) {
     SS.showContourAreas = !SS.showContourAreas;
     SS.GW.Invalidate();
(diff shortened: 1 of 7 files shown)


src/confscreen.cpp
```diff
@@ -342,6 +347,9 @@ void TextWindow::ShowConfiguration() {
         SS.automaticLineConstraints ? CHECK_TRUE : CHECK_FALSE);
     Printf(false, "  %Fd%f%Ll%s  use turntable mouse navigation%E", &ScreenChangeTurntableNav,
         SS.turntableNav ? CHECK_TRUE : CHECK_FALSE);
+    Printf(false, "  %Fd%f%Ll%s  immediately edit dimensions%E",
```
whitequark: What about `edit newly added dimensions`? The variable naming is fine, but the preference is a bit confusing.

whitequark:
> Let me know if the default should be on or off

The existing behavior should not change for people who are already using SolveSpace.

vespakoen:
Changed the default to false and used your suggested label.
Just have to figure out how to squash these commits together now =)

Will take a bit...

whitequark:
I can just squash while merging in a simple case like this.

vespakoen:
Cool! then I leave it =)
Was just running
```
git reset --soft HEAD~2
```

Which seemed to work allright =)

whitequark:
Usually people use `git rebase -i` which is a bit more flexible, but that works too.

vespakoen:
Indeed, I am not very familiar with git I notices since I mostly only commit and push things with it, should read the manual again...

vespakoen:
Force pushed it anyway ;)

whitequark:
This needs a changelog addition as well.

PR #515, opened 2019-11-26
Fix preferences not opening when closed on macOS

vespakoen:
On macOS, when closing the preferences window, then clicking `Preferences` in the Menu, does not open the preferences window.

This PR fixes that.

### DIFF
--- a/src/platform/guimac.mm
+++ b/src/platform/guimac.mm
@@ -1409,6 +1409,9 @@ void OpenInBrowser(const std::string &url) {
 
 @implementation SSApplicationDelegate
 - (IBAction)preferences:(id)sender {
+    if (!SS.GW.showTextWindow) {
+        SolveSpace::SS.GW.MenuView(SolveSpace::Command::SHOW_TEXT_WND);
+    }
     SolveSpace::SS.TW.GoToScreen(SolveSpace::TextWindow::Screen::CONFIGURATION);
     SolveSpace::SS.ScheduleShowTW();
 }


src/platform/guimac.mm
```diff
@@ -1409,6 +1409,7 @@ - (NSApplicationTerminateReply)applicationShouldTerminate:(NSApplication *)sende
 
 @implementation SSApplicationDelegate
 - (IBAction)preferences:(id)sender {
+    SolveSpace::SS.GW.MenuView(SolveSpace::Command::SHOW_TEXT_WND);
```
whitequark: This toggles the text window visibility.

vespakoen: Oops, good to know haha, I will add a check around it to see if it's actually off

whitequark: Just do `SS.GW.showTextWindow = true;`

vespakoen: But that would not update the checkmark if it's open or not, correct? e.g.:

```
    if(SS.TW.window) SS.TW.window->SetVisible(SS.GW.showTextWindow);
    showTextWndMenuItem->SetActive(SS.GW.showTextWindow);
```

whitequark: Oh yes, the horrible `EnsureValidActives` function. Go ahead with your plan.

vespakoen:
Went ahead with the plan, hope it's fine now =)

PR #516, opened 2019-11-26
Fix tooltips not closing on macOS Catalina

vespakoen:
This is the last bug I have found on Catalina.
Tooltips are not closed when hovering off something.
The private API seems to have changed, and implementing tooltips seems to be a pain in the butt, so I checked for the availability of the method and fall back to the old method for now.


whitequark:
Looks good, and thank you for the investigation and fix! I know figuring out private APIs can be quite frustrating.

vespakoen:
I got quite lucky figuring out the new magic method to close the tooltip,
and with a little help of "class-dump" ofcourse :)

PR #517, opened 2019-11-28
[WIP] macOS travis builds + sigining and notarizing

vespakoen:
Added macOS build for travis, tried it out on my own fork so some settings will have to be restored, also some secrets will have to be added to the travis build.

Travis Environment variables:

Screenshot 2019-11-28 at 14 01 22

MACOS_APPSTORE_USERNAME - The Apple ID email adress
MACOS_APPSTORE_APP_PASSWORD - And "App Specific Password". See **How to manage app-specific passwords** on https://support.apple.com/en-us/HT204397
MACOS_CERTIFICATE_P12 - A base64 encoded .p12 certificate (export from "Developer ID Application: Your Name" using Keychain Assistant, e.g.: `cat cert.p12 | base64 | pbcopy`)
MACOS_CERTIFICATE_PASSWORD - The password used during exporting of the .p12 file
MACOS_DEVELOPER_ID - See image, make sure to wrap it in quotes!

Will merge all commits into a single one once I have verified it it working well.


whitequark:
@jwesthues Do you have a macOS signing certificate? If not, it effectively costs $100/year. Since macOS Catalina, unsigned applications will not run by default, although it isn't entirely clear to me right now how burdensome it is to run them anyway--I don't have anything that runs Catalina. I assume Apple plans to make it increasingly burdensome in the future, anyway.

vespakoen:
Yes I do have a certificate and added it to the travis ci build on my fork.

I am happy to share it with this project if you like, or produce the
executables for this project.

Whatever is easiest for you.

This PR also does the notarizing / all the voodoo apple needs to be happy.

I had to drop support for pre-mavericks though but could add a job for that
as well if needed.

vespakoen:
The first build just landed actually: https://github.com/vespakoen/solvespace/releases

Will have to test it on a computer that is not "infected" with my certificates to test it out properly though. Will do that later

vespakoen:
https://travis-ci.org/vespakoen/solvespace/builds/618185752

.travis/build-macos.sh
```diff
@@ -4,6 +4,68 @@ if echo $TRAVIS_TAG | grep ^v; then BUILD_TYPE=RelWithDebInfo; else BUILD_TYPE=D
 
 mkdir build
 cd build
-cmake -DCMAKE_OSX_DEPLOYMENT_TARGET=10.7 -DCMAKE_BUILD_TYPE=$BUILD_TYPE ..
-make VERBOSE=1
+cmake -DCMAKE_OSX_DEPLOYMENT_TARGET=10.9 -DCMAKE_BUILD_TYPE=$BUILD_TYPE ..
+make -j 2 VERBOSE=1
 make test_solvespace
+
+ls bin
```
whitequark: Can you split this into a different file, e.g. `.travis/sign-macos.sh`?

.travis/build-macos.sh
```diff
@@ -4,6 +4,68 @@ if echo $TRAVIS_TAG | grep ^v; then BUILD_TYPE=RelWithDebInfo; else BUILD_TYPE=D
...
+codesign -s "${MACOS_DEVELOPER_ID}" --timestamp --options runtime -f --deep "${app}"
+
+# create the .dmg from the signed .app
+hdiutil create -srcfolder "${app}" "${dmg}"
+
+# sign the .dmg
+codesign -s "${MACOS_DEVELOPER_ID}" --timestamp --options runtime -f "${dmg}"
+
+# notarize and store request uuid in variable
+notarize_uuid=$(xcrun altool --notarize-app --primary-bundle-id "${bundle_id}" --username "${MACOS_APPSTORE_USERNAME}" --password "${MACOS_APPSTORE_APP_PASSWORD}" --file "${dmg}" 2>&1 | grep RequestUUID | awk '{print $3'})
```
whitequark: I'm a bit confused by this. Does XCode offer no way to block until notarization is finished?

vespakoen: You are correct =) There is no way to block indeed, a very anti-CI approach. The voodoo is necessary unfortunately...

.travis/build-macos.sh
vespakoen: Allright, will do!

vespakoen:
Strangely the Travis builds on this PR seem to get built differently, I see colors where my build does not have that, and the build for this repository seems to stop half way during the copying of assets.

Cannot imagine this has to do with the missing secrets, so that it really strange... Perhaps there are some limitations on what travis builds when it's in a PR?

Anyways, I incorporated your suggestions...

vespakoen:
Nevermind that last comment, now is seemed to get further, and actually seems to crash in the sign-macos.sh file, which is expected without the secrets.

Shall I hand my secrets over to you for the signing? Or shall I do the releases separately and deliver the .dmg's to you?

vespakoen:
The latest master doesn't build correctly on macOS anymore.
I will git bisect it too see where it broke.

vespakoen:
I am making a new PR in favor of this one. I got the build working again by simply moving these lines:

```
if(APPLE OR CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
endif()
```

Higher up the CMakeLists.txt file, before:

```
include(GetGitCommitHash)
```

PR #528, opened 2019-12-09
Cut down some usages of cached pointers

rpavlik:
Found this laying around. I know the previous use of cached pointers had led to a crash, which is presumably why I did this change. Haven't done anything with it since making it except a rebase.

### DIFF
--- a/src/generate.cpp
+++ b/src/generate.cpp
@@ -216,31 +216,31 @@ void SolveSpaceUI::GenerateAll(Generate type, bool andFindFree, bool genForBBox)
 
     // Not using range-for because we're using the index inside the loop.
     for(i = 0; i < SK.groupOrder.n; i++) {
-        Group *g = SK.GetGroup(SK.groupOrder[i]);
+        hGroup hg = SK.groupOrder[i];
 
         // The group may depend on entities or other groups, to define its
         // workplane geometry or for its operands. Those must already exist
         // in a previous group, so check them before generating.
-        if(PruneGroups(g->h))
+        if(PruneGroups(hg))
             goto pruned;
 
         for(auto &req : SK.request) {
             Request *r = &req;
-            if(r->group != g->h) continue;
+            if(r->group != hg) continue;
 
             r->Generate(&(SK.entity), &(SK.param));
         }
         for(auto &con : SK.constraint) {
             Constraint *c = &con;
-            if(c->group != g->h) continue;
+            if(c->group != hg) continue;
 
             c->Generate(&(SK.param));
         }
-        g->Generate(&(SK.entity), &(SK.param));
+        SK.GetGroup(hg)->Generate(&(SK.entity), &(SK.param));
 
         // The requests and constraints depend on stuff in this or the
         // previous group, so check them after generating.
-        if(PruneRequests(g->h) || PruneConstraints(g->h))
+        if(PruneRequests(hg) || PruneConstraints(hg))
             goto pruned;
 
         // Use the previous values for params that we've seen before, as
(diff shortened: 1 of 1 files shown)


phkahler:
@rpavlik 
What's the best way for me to test this? I cloned the repo but failed to build with:

CMake Error at CMakeLists.txt:100 (add_subdirectory):
  The source directory

    /home/paul/projects/testing/solvespace/extlib/libdxfrw

  does not contain a CMakeLists.txt file.


-- Configuring incomplete, errors occurred!

Or if you've been running with this change long enough to have confidence we should just merge it.

nabijaczleweli:
FWIW: the following works for me (note the `--recursive`):
```
git clone https://github.com/rpavlik/solvespace -b reduce-cached-pointer --recursive
mkdir -p solvespace/build && cd solvespace/build
cmake -GNinja .. && ninja
```

whitequark:
I think this PR shouldn't be necessary--unless I'm missing something, the use of cached pointers here dates back to the initial commit. @phkahler I expect that testing will show no functional changes other than a minor slowdown.

phkahler:
@whitequark   It's not needed unless there is an effort to change function calls from taking pointers to taking handles (which is not done here). If that's the goal this moves in that direction. If we just prefer to use hg rather than g->h that's just a style change (one I like BTW). I like to move toward the future when possible, but if that's not what this is then lets drop it. If it's moving toward something we'd like let's merge it - It looks fine.

@rpavlik what was your goal here?

phkahler:
I just dislike the syntax of dereferencing the pointer g->h. IMHO pointers are actually preferable but they can not be written to a file. They may also have issues with regeneration - I'm still not clear on the whole remap thing. @whitequark if it ever happens, what would be preferred for hierarchical sketches? That's a future worth moving toward.

whitequark:
> if it ever happens, what would be preferred for hierarchical sketches? That's a future worth moving toward.

Globally, we can't move away from handles (they are deeply baked into the architecture of SolveSpace), and even if we could, we shouldn't--they serve multiple important roles. That is, they turn dangling pointers (safety error) into dangling handles (logic error); they make remapping possible; they simplify savefiles (although they are not strictly required there--we could assign transient identifiers just for savefiles, and turn them back into pointers on load).

Regarding hierarchical sketches, IMO it is far too early to talk about them. There are many steps we have to take first, and they involve a significant amount of work. For example, new file format and eliminating the global sketch would be absolutely necessary. I don't think there is any rush to replace the handle system with something because, even if we do that, we still have to deal with the legacy files--so let's solve problems in the order they arrive.

rpavlik:
If it makes it easier for the other devs of the project, this can be dropped. I saw it as a safety thing, mostly, for the reasons @whitequark mentioned. I had recently fixed some issues where a pointer was cached, then some operation invalidated the pointers but they continued to be used. (Well, that, and it looked awfully upside-down to me to look up a pointer by handle, then immediately dereference that pointer to get the corresponding handle.) I don't remember offhand the backing structure for the lookup of handles to pointers, if it's ordered or unordered, to know the overhead. But, now that I fixed the various "stale pointer" issues, this is not really critical. Shouldn't happen again until we get another new entity :-P

phkahler:
I think the first function changed is good and should be taken for all the reasons above that prefer handles to pointers. The second one will experience a performance hit, but probably not perceptible so I'd just merge it all.

whitequark:
> I don't remember offhand the backing structure for the lookup of handles to pointers, if it's ordered or unordered, to know the overhead.

I believe it's binary search.

> But, now that I fixed the various "stale pointer" issues, this is not really critical. Shouldn't happen again until we get another new entity :-P

Does asan find these issues? We should already be running tests under asan on Travis, though the test coverage is low.

> I think the first function changed is good and should be taken for all the reasons above that prefer handles to pointers. The second one will experience a performance hit, but probably not perceptible so I'd just merge it all.

I'm confused--isn't there just one function?

whitequark:
Merged with minor changes as e51fdf6f. I hope that looks good to everyone.

PR #530, opened 2019-12-14
>=2 objects in eq and parallel constraints

77maxikov:


whitequark:
@phkahler I'm tempted to avoid changing the way we handle adding constraints until we do something to solve #121 properly, which requires even more infrastructural work. (Unfortunately #492 has problems similar to the problems with this PR.)

phkahler:
After looking at this problem a lot, I believe this is the way. @77maxikov  if you rebase this to master I'll merge it and then build on it a little.  I'll probably do it myself this weekend if nobody else does - I'm getting the itch to work on solvespace again!

77maxikov:
https://github.com/solvespace/solvespace/pull/1314

phkahler:
Closing since 1314 was used.

PR #531, opened 2019-12-17
Fix eight trivial "implicit conversion 'double' to 'int'" warnings. NFC.

ruevs:
Found when compiling with clang 9.


whitequark:
Thanks! We keep fixing those. I think before clang 9 (evidently) only MSVC was reporting these consistently, but that requires going through Appveyor logs.

ruevs:
I'll submit another PR when I get to the "other" machine that I have VS 2019 installed on. There are a bunch about "float to double".

PR #533, opened 2019-12-20
Win32: Drastically reduce stack usage at startup. NFC?

ruevs:
See:
https://github.com/solvespace/solvespace/issues/92#issuecomment-567831112

The problem was first introduced here:
https://github.com/solvespace/solvespace/commit/dabd57847edf142f145491175fb203455023d9d7

and later "fixed" here:
https://github.com/solvespace/solvespace/commit/f324477dd06c15022670a4486bf83be3abbf48b5
by setting the stack size to /STACK:33554432

Solvespace now starts up even with /STACK:554432

According to this:
https://en.cppreference.com/w/cpp/language/value_initialization
```
"2) if T is a class type with a default constructor that is neither user-provided nor deleted (that is, it may be a class with an implicitly-defined or defaulted default constructor), the object is zero-initialized and then it is default-initialized if it has a non-trivial default constructor; "
```
So removing the `{}` should leave both the `System` and `TextWindow` class instances properly initialized.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -340,7 +340,7 @@ if(ENABLE_GUI)
 
     if(MSVC)
         set_target_properties(solvespace PROPERTIES
-            LINK_FLAGS "/MANIFEST:NO /SAFESEH:NO /INCREMENTAL:NO /OPT:REF /STACK:33554432")
+            LINK_FLAGS "/MANIFEST:NO /SAFESEH:NO /INCREMENTAL:NO /OPT:REF")
     elseif(APPLE)
         set_target_properties(solvespace PROPERTIES
             OUTPUT_NAME SolveSpace)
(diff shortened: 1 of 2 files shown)


whitequark:
> The only question now is whether omitting the {} in TextWindow and System constructor calls in the initializer list of the SolveSpaceUI constructor has any adverse effect.

Default-initialization and list-initialization are different. On a cursory check, System is POD, so default-initializing it would leave garbage inside. TextWindow is not, so it should be benign there. However, VS12 that we use for Appveyor builds could have bugs beyond this, so you should check that the CI-built binary works.

ruevs:
The appveyor artifact runs on my Win10 64 machine, so does my locally compiled copy... but that does not mean much.

ruevs:
Merge? ;-)

whitequark:
Thank you!

PR #539, opened 2019-12-23
Win32: Open a console in debug mode to display stderr debug messages.

ruevs:
It may be useful when running a debug build directly (instead of using a debugger that captures and displays `stderr`).


whitequark:
Sorry, I really don't think this is useful. All of our Appveyor builds are debug, for example.

ruevs:
OK, makes sense.

PR #543, opened 2020-01-22
snap: Release snaps to store via script instead of snap provider

ppd1990:
dpl-snap only supports pushing one snap at a time.
Instead of many repetitive deploy statements, we use our own script to loop over the snaps to be released.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,17 +35,15 @@ jobs:
             confinement: classic
       script: ./.travis/build-snap.sh
       deploy:
-        - provider: snap
+        - provider: script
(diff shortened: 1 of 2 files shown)


whitequark:
@ppd1990 Argh:
```
There seems to be a network error: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response',))
```

ppd1990:
It looks quite alright: https://travis-ci.org/solvespace/solvespace/jobs/640723069
Where did you get that error?

whitequark:
I restarted the build; looks like it worked the second time.

PR #544, opened 2020-01-23
README: Add snap as an installation option

ppd1990:


README.md
```diff
@@ -39,6 +39,20 @@ and cannot guarantee their functionality.
 
 [notesalexp]: https://notesalexp.org/packages/en/source/solvespace/
 
+### Via Snap Store
```
whitequark: Let's put this section above "unofficial binary packages", since these are, in fact, official, and preferred as we control them and can guarantee they're good.

ppd1990: Right

README.md
```diff
@@ -39,6 +39,20 @@ and cannot guarantee their functionality.
...
+
+Builds from master are automatically released to the `edge` channel in the Snap Store. Those packages receive no manual scrutiny and do not constitute official releases.
```
whitequark: In fact, the `master` branch of SolveSpace is generally quite stable and well suited for daily use in spite of occasional bugs. We don't have to warn so strongly against it. I would say something like...

> Those packages contain the latest improvements, but receive less testing than release builds.

whitequark: There's not been a release for very long for various regrettable reasons, and 2.3 is *very* outdated, so de facto, `master` is currently what I recommend using to many people. But that's not a very satisfying situation either, so let's not just say that outright in README.

ppd1990: Less harsh words + the future prospects of an official release seems like a nice compromise in this section

Issue #538, opened 2019-12-23
RFE: Default "export" filename stem to same as slvs file


rpavlik:
Just a UI nitpick. If I get the chance I'll look into doing this myself.

### System information

SolveSpace version: 3.0~64c0f62b

Operating system: Debian Buster

### Expected behavior

I have a solvespace file. I go to export it to trimesh (to get an STL to 3d print). The dialog probably should open to the location of the slvs file and defaulting to the same stem but with the appropriate extension (`.stl`)

### Actual behavior

It defaults to "untitled" (`untitled.stl`) in some other directory - think it's the last directory I used that export option in.

whitequark:
> It defaults to "untitled" (`untitled.stl`) in some other directory - think it's the last directory I used that export option in.

That's on purpose: suppose you are exporting many files in some other directory (like "Outputs" in the root of your projects).

rpavlik:
ah, so it's a workflow problem on my part that I tend to have my stl files live next to the slvs files and edit one model at a time :) 

Would defaulting the name to match the model make sense, though?

whitequark:
Yep. That's how it worked before 6b5db5897155ca9bdc5daa3e791c07723a6b7d9c; it looks like I forgot to add functionality that sets the file stem. You should be able to use `Platform::Path` and `FileDialog::SetCurrentName` to add this back easily.

whitequark:
Regarding your workflow, I would expect you to use solvespace-cli, seeing as you're (I assume) comfortable with command line.

PR #545, opened 2020-02-02
Make default "export" filename stem same as slvs file.

RavenFire:
Implements #538. 

This works on Linux, but I don't currently have the ability to test on Windows or Mac.


nabijaczleweli:
Unfortunately I can't really debug beyond this at the moment, but the pre-populated filename is `untitled` on my Windows, like before the patch.

whitequark:
This doesn't seem like a very elegant solution; main application code isn't supposed to reach into whatever `dialog->FreezeChoices` does. That's probably why it doesn't work on Windows as well.

phkahler:
We're probably going with PR #562 which still doesn't have the GTK version implemented.

PR #546, opened 2020-02-04
Fix immediate editing of reference constraints

vespakoen:
As promised, the fix for immediately editing reference dimensions.

The issue was that the editor was opened before redundant constraint check was done.


whitequark:
@ppd1990 Looks like Snap builds break Travis on every PR, could you please fix that?

whitequark:
@vespakoen Thanks!

ppd1990:
@whitequark I've been quite busy in the last weeks, so my apologies for the delay. I'll get right on proposing a fix.

ppd1990:
@whitequark See #547

PR #551, opened 2020-02-12
packaging: Rename pkg to build-aux

ppd1990:
Snapcraft 3.3 introduced support for snap directories
at build-aux/snap instead of only at root level.

Take advantage of this by renaming pkg/ to build-aux/
which enables us to drop the hacky indirection of snap/build.sh,
resulting in less complexity and a cleaner directory layout.

When I created the original snap packaging, I overlooked this feature which had been recently introduced. It's a neat opportunity to simplify stuff but it also moves the flatpak packaging. Not strictly necessary but still a very good idea, I feel.

Note: `build-aux` for packaging/building is very common within the GNOME ecosystem


whitequark:
I quite dislike this change. I don't think that a single packaging mechanism should dictate the directory layout in a repository of a cross-platform application.

whitequark:
It is also not even a good name; I use `build-*` directories to organize builds for a variety of platforms and toolchains, like `build-mingw32` and so on, as you can see in the gitignore file. This would make `rm -rf build*` unusable.

ppd1990:
I like `pkg` better too, by a lot. It's just a thing I learned about snapcraft an hour ago, and so I felt it was sensible to at least discuss it.

whitequark:
I think your original approach works perfectly fine.

Issue #490, opened 2019-09-21
Crash instead of "can't do that operation" dialog in AppVeyor build


rpavlik:
### System information

SolveSpace version: 5d78f993cea3749b6644e0a2f9ddf89c9754e377 - Windows artifact from AppVeyor

Operating system: Windows 10

### Expected behavior

When I accidentally clicked "other supplementary angle" instead of "constraint angle" (with 2 line segments selected) I assume it was supposed to tell me I goofed with a message box.

### Actual behavior

It crashed.

### Additional information

EDIT: OK, this is actually reproducible. Open "bothparts" and on the topmost sketch, click these two lines and then hit the "other angle" button. Crashes on Windows at least. 
![image](https://user-images.githubusercontent.com/61129/65379088-5cacc400-dc88-11e9-8d5a-4b13edfaef33.png)
[circuit box.zip](https://github.com/solvespace/solvespace/files/3639079/circuit.box.zip)

whitequark:
Can't reproduce here. Could you grab a backtrace?

rpavlik:
ah ha. I can't reproduce in my local build (with vs2019) but this, as well as other things that should pop up a dialog (drawing  a rectangle without a workplane, etc) all crash the AppVeyor build. Looks like there's something wrong with that build which is causing crashes instead of popups.

whitequark:
There's debug information uploaded from AppVeyor, do you think you can use that in MSVC to grab a backtrace?

rpavlik:
ah, yeah probably, will give it a shot

rpavlik:
"Unhandled exception at 0x014C0713 in solvespace.exe: An invalid parameter was passed to a function that considers invalid parameters fatal. occurred"

```
>	solvespace.exe!_invoke_watson(const wchar_t * pszExpression, const wchar_t * pszFunction, const wchar_t * pszFile, unsigned int nLine, unsigned int pReserved) Line 132	C++
 	solvespace.exe!_invalid_parameter(const wchar_t * pszExpression, const wchar_t * pszFunction, const wchar_t * pszFile, unsigned int nLine, unsigned int pReserved) Line 86	C++
 	solvespace.exe!std::_String_const_iterator<std::_String_val<std::_Simple_types<char>>>::operator*() Line 80	C++
 	solvespace.exe!std::_String_iterator<std::_String_val<std::_Simple_types<char>>>::operator*() Line 336	C++
[...]

PR #560, opened 2020-03-12
Solve crash when trying to show error with bad separator.

rorlork:
This should solve #490 

The problem was that the message box tries to split the `description` string after the first dot or colon, but if there is only a single dot at the very end of the string, when calling `isspace(*it)`, the value of `*it` is `\0` and crashes.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -138,9 +138,11 @@ static void MessageBox(const char *fmt, va_list va, bool error,
         }
     }
 
-    std::string::iterator it = description.begin();
-    while(isspace(*it)) it++;
-    description = description.substr(it - description.begin());
+    if(description.length() > 0) {
+        std::string::iterator it = description.begin();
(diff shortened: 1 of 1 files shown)


whitequark:
Thanks!

rorlork:
Happy to help!

I've been using this for some days now and it's the simplest/most functional thing out there that does both 2D and 3D, and I really like that it doesn't try to do *everything* while having broken stuff everywhere. For its simplicity I think I can even contribute to it without spending weeks trying to understand the source, so expect some more PRs in the future, possibly :)

whitequark:
> For its simplicity I think I can even contribute to it without spending weeks trying to understand the source, so expect some more PRs in the future, possibly :)

A word of caution: it has a highly interdepenent implementation with a lot of hidden assumptions. Hack away, but note that anything you do that touches the core will probably have unintended consequences. I did my best to improve the situation but it's not much (yet).

Issue #559, opened 2020-03-09
Build failing on MSYS2 due to zlibstatic and png_static libraries not being found


nabijaczleweli:
### System information

SolveSpace version: current HEAD – 6017ecdaccda6bc9ebda1470714e17f8255ea858

Operating system: Windows 10

### Expected behavior

SolveSpace builds.

### Actual behavior

Given the following setup:
  * CMake from the website, and
  * the rest of the dependencies from MSYS2.

The following error occurs:

```
C:/PP/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lzlibstatic
C:/PP/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lpng_static
collect2.exe: error: ld returned 1 exit status
```

The commandline is

```
cd /C/PP/solvespace/build/src && /C/PP/msys64/mingw64/bin/g++.exe  -ffile-prefix-map=C:/PP/solvespace=. -static-libgcc -static-libstdc++ -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Werror=switch   -Wl,--whole-archive CMakeFiles/solvespace-cli.dir/objects.a -Wl,--no-whole-archive  -o ../bin/solvespace-cli.exe -Wl,--out-implib,../bin/libsolvespace-cli.dll.a -Wl,--major-image-version,0,--minor-image-version,0 @CMakeFiles/solvespace-cli.dir/linklibs.rsp
```

With the following in `CMakeFiles/solvespace-cli.dir/linklibs.rsp`:
```
../bin/libsolvespace-core.a ../bin/libsolvespace-headless.a ../bin/libsolvespace-core.a ../bin/libdxfrw.a -lzlibstatic -lpng_static -lfreetype ../bin/libflatbuffers.a ../bin/libcairo.a ../bin/libpixman-1.a -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 
```

When configured for Ninja, the linker arguments are inlined, but to the same effect.

I found these specified on L153, under an `if(WIN32)`:
```cmake
    find_vendored_package(ZLIB zlib
        ZLIB_LIBRARY            zlibstatic
[...]

PR #561, opened 2020-03-14
Force-vendor libraries on Win32 like on Apple and update ANGLE, fixing some Windows build envs

nabijaczleweli:
As discussed in #559


whitequark:
Thanks!

PR #562, opened 2020-03-21
Pre-populate file dialogs with sensible defaults

nabijaczleweli:
Example: screenshot

Tested on Windows only, with the assumption that other platforms' `Platform::FileDialog` behaves the same.

Ref: #538

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -904,11 +904,13 @@ try_again:
         } else if(linkMap.count(g.linkFile) == 0) {
             dbp("Missing file for group: %s", g.name.c_str());
             // The file was moved; prompt the user for its new location.
-            switch(LocateImportedFile(g.linkFile.RelativeTo(saveFile), canCancel)) {
+            const auto linkFileRelative = g.linkFile.RelativeTo(saveFile);
+            switch(LocateImportedFile(linkFileRelative, canCancel)) {
                 case Platform::MessageDialog::Response::YES: {
                     Platform::FileDialogRef dialog = Platform::CreateOpenFileDialog(SS.GW.window);
                     dialog->AddFilters(Platform::SolveSpaceModelFileFilters);
                     dialog->ThawChoices(settings, "LinkSketch");
+                    dialog->SuggestFilename(linkFileRelative);
                     if(dialog->RunModal()) {
                         dialog->FreezeChoices(settings, "LinkSketch");
                         linkMap[g.linkFile] = dialog->GetFilename();
@@ -985,6 +987,7 @@ bool SolveSpaceUI::ReloadLinkedImage(const Platform::Path &saveFile,
         Platform::FileDialogRef dialog = Platform::CreateOpenFileDialog(SS.GW.window);
         dialog->AddFilters(Platform::RasterFileFilters);
         dialog->ThawChoices(settings, "LinkImage");
+        dialog->SuggestFilename(filename->RelativeTo(saveFile));
         if(dialog->RunModal()) {
             dialog->FreezeChoices(settings, "LinkImage");
             *filename = dialog->GetFilename();
--- a/src/platform/gui.h
+++ b/src/platform/gui.h
@@ -356,6 +356,7 @@ public:
 
     virtual Platform::Path GetFilename() = 0;
     virtual void SetFilename(Platform::Path path) = 0;
+    virtual void SuggestFilename(Platform::Path path) = 0;
 
     virtual void AddFilter(std::string name, std::vector<std::string> extensions) = 0;
     void AddFilter(const FileFilter &filter);
--- a/src/platform/guigtk.cpp
+++ b/src/platform/guigtk.cpp
@@ -1246,6 +1246,10 @@ public:
         gtkChooser->set_filename(path.raw);
     }
 
+    void SuggestFilename(Platform::Path path) override {
+        SetFilename(path.WithExtension(""));  // TODO
+    }
+
     void AddFilter(std::string name, std::vector<std::string> extensions) override {
         Glib::RefPtr<Gtk::FileFilter> gtkFilter = Gtk::FileFilter::create();
         Glib::ustring desc;
--- a/src/platform/guimac.mm
+++ b/src/platform/guimac.mm
@@ -1274,6 +1274,10 @@ public:
         nsPanel.nameFieldStringValue = Wrap(path.FileStem());
     }
 
+    void SuggestFilename(Platform::Path path) override {
+        SetFilename(path.WithExtension(""));
+    }
+
     void FreezeChoices(SettingsRef settings, const std::string &key) override {
         settings->FreezeString("Dialog_" + key + "_Folder",
                                [nsPanel.directoryURL.absoluteString UTF8String]);
--- a/src/platform/guiwin.cpp
+++ b/src/platform/guiwin.cpp
@@ -1582,6 +1582,10 @@ public:
         wcsncpy(filenameWC, Widen(path.raw).c_str(), sizeof(filenameWC) / sizeof(wchar_t) - 1);
     }
 
+    void SuggestFilename(Platform::Path path) override {
+        SetFilename(Platform::Path::From(path.FileStem()));
+    }
+
     void AddFilter(std::string name, std::vector<std::string> extensions) override {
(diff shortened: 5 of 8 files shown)


whitequark:
Thanks for the PR! I'm going to test it on Linux a bit later and merge.

whitequark:
Could you please test it on Windows with Explorer configured to show file extensions?

nabijaczleweli:
This does indeed look wrong and prompts to override `uwu.slvs` on save, so not just visual difference:
screenshot

Passing `pp.WithExtension("png/csv/&c.")` to `SetFilename()` should solve this I think? Will amend in a bit.

(Closed the PR by accident, sorry!)

whitequark:
> Passing `pp.WithExtension("png/csv/&c.")` to `SetFilename()` should solve this I think?

That won't work quite right either because the file type selector is remembered between uses of a dialog, so the extension should actually be taken from the selector. I think this might need a new method, something like `SuggestFilename()`.

nabijaczleweli:
Two changes:
  * I made `Path::SetExtension("")` strip the extension entirely (as I feel would be the expected behaviour), and
  * replaced the `SetFilename()` calls with a new `SuggestFilename()` function, which calls `SetFilename()` with a stripped extension, which behaves like I'd expect on Windows:
screenshot

Will rebase the commits around to be more presentable after some testing to reduce noise.

I wanted to make sure my updated tests were good but I got hit with this both before and after my changes (but maybe I'm just running the tests wrong? I'll do a clean rebuild in the mean-time but that'll take some time):
```
D:\Programy\solvespace\build\bin>solvespace-testsuite
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::erase: __pos (which is 18446744073709551615) > this->size() (which is 44)

This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.
```

```
D:\Programy\solvespace\build\bin>ninja test_solvespace
[109/109] Testing SolveSpace
FAILED: test/CMakeFiles/test_solvespace
cmd.exe /C "cd /D D:\Programy\solvespace\build\test && D:\Programy\solvespace\build\bin\solvespace-testsuite.exe"

This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::erase: __pos (which is 18446744073709551615) > this->size() (which is 44)
ninja: build stopped: subcommand failed.
```

Starting `solvespace-testsuite` via gdb didn't yield a backtrace.

nabijaczleweli:
I've managed to build SolveSpace on my laptop, and the GTK backend doesn't seem to honour the `SetFilename()`/`gtkChooser->set_filename()` call at all?

This working diff:
```diff
diff --git a/src/platform/guigtk.cpp b/src/platform/guigtk.cpp
index 73fb01e..81e0c0d 100644
--- a/src/platform/guigtk.cpp
+++ b/src/platform/guigtk.cpp
@@ -1242,7 +1242,11 @@ public:
     }
 
     void SetFilename(Platform::Path path) override {
+        printf("SetFilename1(%s -> %s)\n", gtkChooser->get_filename().c_str(), path.raw.c_str());
         gtkChooser->set_filename(path.raw);
+        // printf("SetFilename1(%s -> %s)\n", gtkChooser->get_filename().c_str(), path.FileStem().c_str());
+        // gtkChooser->set_filename(path.FileStem());
+        printf("SetFilename2(%s)\n", gtkChooser->get_filename().c_str());
     }
 
     void AddFilter(std::string name, std::vector<std::string> extensions) override {
@@ -1296,6 +1300,7 @@ public:
             return;
 
         Platform::Path path = GetFilename();
+        printf("FilterChanged(%s -> %s)\n", path.raw.c_str(), extension.c_str());
         SetCurrentName(path.WithExtension(extension).FileName());
     }
 
@@ -1311,10 +1316,12 @@ public:
     }
 
     void CheckForUntitledFile() {
+        printf("CheckForUntitledFile()1: %s\n", gtkChooser->get_current_name().c_str());
         if(gtkChooser->get_action() == Gtk::FILE_CHOOSER_ACTION_SAVE &&
                 Path::From(gtkChooser->get_current_name()).FileStem().empty()) {
             gtkChooser->set_current_name(std::string(_("untitled")) + "." + GetExtension());
         }
+        printf("CheckForUntitledFile()2: %s\n", gtkChooser->get_current_name().c_str());
     }
 };
```

Which yields
```
nabijaczleweli@nabtop:~/src/solvespace/build$ bin/solvespace /tmp/owo.slvs

(solvespace:2628): dbind-WARNING **: 21:48:35.517: AT-SPI: Error retrieving accessibility
ded by any .service files
SolveSpace!

Missing (absent) translation for '&Helix'
Missing (absent) translation for 'Re&volve'
Missing (absent) translation for 'Show &Underconstrained Points'
Missing (absent) translation for file-type'PNG image'
FilterChanged( -> png)
SetFilename1( -> /tmp/owo)
SetFilename2()
CheckForUntitledFile()1: .png
CheckForUntitledFile()2: untitled.png
```
and likewise with the commented block being in.

nabijaczleweli:
Quoting https://developer.gnome.org/gtkmm/stable/classGtk_1_1FileChooser.html#a35bbea274ddb0a7b948e6f23b70b9bed:
> Sets filename as the current filename for the file chooser, by changing to the file’s parent folder and actually selecting the file in list; all other files will be unselected.
>
> If the chooser is in `Gtk::FILE_CHOOSER_ACTION_SAVE` mode, the file’s base name will also appear in the dialog’s file name entry.
> 
> Note that the file must exist, or nothing will be done except for the directory change.

The documentation then goes on to reference an example, which seems to be missing from the header as well as the doc.

I don't really know what to do about this…

whitequark:
> the GTK backend doesn't seem to honour the `SetFilename()`/`gtkChooser->set_filename()` call at all?

It does. The problem is that the semantics of `SetFilename` are not what you think: that function *selects an existing file*. That's the reason I asked you to make a separate `SuggestFilename` function, which would not require that.

nabijaczleweli:
Right, I made `SuggestFilename()` fully virtual and copied the implementation over to the platform impls. The GTK one is marked TODO for now as it clearly doesn't work. I don't think I can test the Cocoa one, but a glance at the documentation suggests it should.

phkahler:
@nabijaczleweli  where do we stand with this one?

nabijaczleweli:
Rebased and fixed on Win32. See TODO in `guigtk.cpp` and discussion above for Linux, I don't have an understanding of GTK or currently a way to port to and test there, unfortunately. No idea about Darwin, but, again, docs say it should be fine(?).

phkahler:
I can test on Linux/GTK, but it will probably have to wait until tomorrow. Fixing it if it isn't right is another story... So you have OSX and Windows covered?

nabijaczleweli:
Windows should be fine. The OSX implementation is a stub, which *should* work going by the Cocoa docs, but I don't even know if it builds.

phkahler:
I tried this on Linux / GTK and it doesn't solve the problem there. The suggested export filenames I get remain the same as before:  "untitled.stl" or "untitled.step".

nabijaczleweli:
Indeed, the GTK implementation does need to be written.

phkahler:
@vespakoen and @ruevs  can you guys have a look at this and possibly #545 and #650. Somewhere in these 3 is a good thing - I think mostly in this one.

ruevs:
@phkahler As a start I rebased it on master.

Builds and runs on Windows. All the Export options offer a file name the same as the opened .slvs and with the correct extension.

Now I need to look at the actual code and what it does...

phkahler:
On Linux I tried:
```
    void SuggestFilename(Platform::Path path) override {
//        SetFilename(path.WithExtension(""));  // TODO
        gtkChooser->set_current_name(path.FileStem());
    }
```
But that suggests the file name with no extension at all. The original commented out give unknown.xxx where xxx is the correct extension. Some of these function names make little sense ;-)

phkahler:
OK, this looks like it works with GTK:
```
        gtkChooser->set_current_name(path.FileStem()+"."+GetExtension());
```

phkahler:
@vespakoen  can you verify/fix the macOS version of this?

Issue #565, opened 2020-03-23
Tests crash on i386


nabijaczleweli:
### System information

SolveSpace version: current HEAD (e355a095c4c3456487be8bc48675757ab79d44d3)

Operating system: Debian sid i386 + libc6 2.30-2

### Expected behavior

Tests run

### Actual behavior

```
nabijaczleweli@nabtop:~/src/solvespace/build$ bin/solvespace-testsuite
[...]
test/constraint/pt_face_distance/test.cpp:21: FAILED: render doesn't match reference; 2 (0.00%) pixels differ
  NG   test constraint/pt_face_distance/reference_roundtrip
[...]
File ../test/harness.cpp, line 112, function PrepareSavefile:
Assertion failed: value.size() == newValue.size().
Message: Expected no change in value length.
Backtrace support not compiled in.
Aborted
```

### Additional information

```
nabijaczleweli@nabtop:~/src/solvespace/build$ bin/solvespace-testsuite
Missing (absent) translation for group-name'#references'
Missing (absent) translation for group-name'sketch-in-plane'
  OK   test constraint/length_difference/normal_migrate_from_v22
Vector::WithMagnitude(1) of zero vector!
Vector::WithMagnitude(1) of zero vector!
Vector::WithMagnitude(1) of zero vector!
Vector::WithMagnitude(1) of zero vector!
Vector::WithMagnitude(1) of zero vector!
Vector::WithMagnitude(1) of zero vector!
Vector::WithMagnitude(1) of zero vector!
Vector::WithMagnitude(1) of zero vector!
Vector::WithMagnitude(1) of zero vector!
  OK   test constraint/pt_line_distance/free_in_3d_roundtrip
  OK   test core/path/from_raw
  OK   test constraint/proj_pt_distance/reference_migrate_from_v22
  OK   test constraint/where_dragged/free_in_3d_roundtrip
  OK   test constraint/horizontal/line_migrate_from_v22
  OK   test core/path/extension
  OK   test constraint/length_difference/normal_roundtrip
  OK   test constraint/same_orientation/normal_migrate_from_v22
  OK   test constraint/equal_angle/other_roundtrip
  OK   test constraint/symmetric/normal_migrate_from_v22
  OK   test constraint/pt_pt_distance/reference_roundtrip
  OK   test constraint/equal_line_arc_len/tau
  OK   test core/expr/literal
  OK   test constraint/pt_on_line/normal_migrate_from_v22
[...]

PR #566, opened 2020-03-24
Force SSE2 floats for i686 targets

nabijaczleweli:
Just GCC and Clang, I don't have a way to test MSVC behaviour.

Ref: #565

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,11 +80,20 @@ endif()
 if(MINGW)
     set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} -static-libgcc")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -static-libgcc -static-libstdc++")
+endif()
 
-    if(TRIPLE STREQUAL "i686-w64-mingw32")
-        set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} -msse2")
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse2")
+# Ensure that all platforms use 64-bit IEEE floating point operations for consistency;
+# this is most important for the testsuite, which compares savefiles directly
+# and depends on consistent rounding of intermediate results.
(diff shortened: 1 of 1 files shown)


ruevs:
The use of x87 fp does _not_ cause loss of precision. On the countrary!
https://gcc.gnu.org/wiki/x87note
Therefore we should treat the crash as a bug and figure out why it is happening with x87 floating point instead of hiding it with SSE.

nabijaczleweli:
Right

CMakeLists.txt
```diff
@@ -80,11 +80,17 @@ endif()
 if(MINGW)
     set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} -static-libgcc")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -static-libgcc -static-libstdc++")
+endif()
```
whitequark: A comment here would help. Something like:
```suggestion

# Ensure that all platforms use 64-bit IEEE floating point operations for consistency; this is
# most important for the testsuite, which compares savefiles directly and depends on consistent
# rounding of intermediate results.
```

CMakeLists.txt
```diff
@@ -80,11 +80,17 @@ endif()
...
 
-    if(TRIPLE STREQUAL "i686-w64-mingw32")
-        set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} -msse2")
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse2")
+if((CMAKE_SYSTEM_PROCESSOR STREQUAL "i686" OR CMAKE_SYSTEM_PROCESSOR STREQUAL "X86"))
```
whitequark:
```suggestion
if(CMAKE_SYSTEM_PROCESSOR STREQUAL "i686" OR CMAKE_SYSTEM_PROCESSOR STREQUAL "X86")
```

ruevs:
OK makes sense. The "compares savefiles directly" explains it.

nabijaczleweli:
Review addressed

whitequark:
Thanks!

PR #574, opened 2020-03-28
Fix more segfaults in mouse.cpp lambdas

Timmmm:
This changes all the lambdas to have explicit captures, since the use of implicit captures has led to some memory errors, especially segfaults in the right-click menu.

I'm not 100% sure that the code is correct anyway - it really needs auditing to ensure all referenced values are still valid when the menu item is clicked (e.g. can you change stuff with keyboard shortcuts while the context menu is visible?), but it should at least be *more* correct!


whitequark:
> it really needs auditing to ensure all referenced values are still valid when the menu item is clicked

I wrote that code and the values should be still valid. You should *definitely* not be able to change anything with keyboard shortcuts, at least.

PR #575, opened 2020-03-29
Add menu options for line styles / view / configuration

Timmmm:
This is to make them a bit easier to find, since the natural place to look for config options is somewhere like File->Preferences or Edit->Settings. On Mac the Configuration should be under the "App" menu but I'm not sure how to do that.

### DIFF
--- a/src/graphicswin.cpp
+++ b/src/graphicswin.cpp
@@ -77,7 +77,9 @@ const MenuEntry Menu[] = {
 { 1,  NULL,                             Command::NONE,             0,       KN, NULL   },
 { 1, N_("&Line Styles..."),             Command::EDIT_LINE_STYLES, 0,       KN, mEdit  },
 { 1, N_("&View Projection..."),         Command::VIEW_PROJECTION,  0,       KN, mEdit  },
+#ifndef __APPLE__
 { 1, N_("Con&figuration..."),           Command::CONFIGURATION,    0,       KN, mEdit  },
+#endif
 
 { 0, N_("&View"),                       Command::NONE,             0,       KN, mView  },
 { 1, N_("Zoom &In"),                    Command::ZOOM_IN,          '+',     KN, mView  },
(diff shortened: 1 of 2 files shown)


whitequark:
> On Mac the Configuration should be under the "App" menu but I'm not sure how to do that.

It should already be there, as a special case I added in the platform code. Have you checked it?

src/graphicswin.cpp
```diff
@@ -74,6 +74,10 @@ const MenuEntry Menu[] = {
 { 1, N_("Select &Edge Chain"),          Command::SELECT_CHAIN,     C|'e',   KN, mEdit  },
 { 1, N_("Select &All"),                 Command::SELECT_ALL,       C|'a',   KN, mEdit  },
 { 1, N_("&Unselect All"),               Command::UNSELECT_ALL,     '\x1b',  KN, mEdit  },
+{ 1,  NULL,                             Command::NONE,             0,       KN, NULL   },
+{ 1, N_("&Line Styles"),                Command::EDIT_LINE_STYLES, 0,       KN, mEdit  },
+{ 1, N_("&View Projection"),            Command::VIEW_PROJECTION,  0,       KN, mEdit  },
+{ 1, N_("&Configuration"),              Command::CONFIGURATION,    0,       KN, mEdit  },
```
whitequark: This mnemonic conflicts with `&Copy`.

src/graphicswin.cpp
```diff
@@ -74,6 +74,10 @@ const MenuEntry Menu[] = {
 { 1, N_("Select &Edge Chain"),          Command::SELECT_CHAIN,     C|'e',   KN, mEdit  },
 { 1, N_("Select &All"),                 Command::SELECT_ALL,       C|'a',   KN, mEdit  },
 { 1, N_("&Unselect All"),               Command::UNSELECT_ALL,     '\x1b',  KN, mEdit  },
+{ 1,  NULL,                             Command::NONE,             0,       KN, NULL   },
+{ 1, N_("&Line Styles"),                Command::EDIT_LINE_STYLES, 0,       KN, mEdit  },
```
whitequark: All three of these should have a `...` at the end, since they open a new dialog (which in this case gets displayed in the property browser).

Timmmm:
> It should already be there, as a special case I added in the platform code. Have you checked it?

Ah no I haven't. Don't have it running on my Mac at the moment.

> At least you should #ifdef away the Configuration item, but maybe all three would be more appropriate.

Opted for just the configuration item so there is still a way to get to the other two from menus on Mac. Hope that's ok!

whitequark:
> Opted for just the configuration item so there is still a way to get to the other two from menus on Mac.

Seems OK to me, but I'm not sure what the macOS interface guidelines say about it. Regardless this is not a blocker for merging.

PR #579, opened 2020-04-16
snap: Use symlink instead of bind-mount for layout

ppd:
As per https://snapcraft.io/docs/snap-layouts,
bind-mounts significantly increase the startup time of the snap.
Use symlink instead for better performance.


whitequark:
Thanks!

ppd:
Holy hell you're quick

PR #580, opened 2020-04-17
snap: Upgrade from gnome-3-28 extension to gnome-3-34

ppd:
Using gnome-3-34 gives us a newer stack in general
and also provides us with a sdk build-snap, so we don't need
to specify as many build-packages and stage-packages.

This makes the snap slimmer and delegates the responsibility
for security updates to the gnome platform snap for even more
packages.


whitequark:
Looks great!

Issue #549, opened 2020-02-09
helix -> Assertion failed: FindByIdNoOops(t->h) == nullptr


andres-erbsen:
### System information

dcc80de on Arch Linux

### Expected behavior

symmetry axis -> helix -> either a working helix or an error message explaining the issue

### Actual behavior

symmetry axis -> helix -> assertion failed

```
Missing (absent) translation for group-name'helix'
File ./src/dsc.h, line 418, function Add:
Assertion failed: FindByIdNoOops(t->h) == nullptr.
Message: Handle isn't unique.
[1]    501906 abort (core dumped)  solvespace sillygear.slvs
```

### Additional information

[sillygear.zip](https://github.com/solvespace/solvespace/files/4176401/sillygear.zip)

Putting construction and non-construction lines in different workspaces and applying helix only to the non-construction lines works.

phkahler:
I made a simple shape with some construction and non-construction lines, did a rotate and helix. It works fine until I increase the number of copies in the rotate group. Just 30 little octagons with a couple construction lines to the middle works, 36 fails. Either one can be extruded straight, only the helix operation fails.
[helix_test.zip](https://github.com/solvespace/solvespace/files/4489723/helix_test.zip)

phkahler:
I'm starting to think this is an issue in Remap itself. As I understand thing:

EntityMap is used to track which entities are copied from which others. Example would be extruding a sketch where 2 copies of the entities are created with a translation from the originals. Remap is used to get handle for a new entity given the handle of the original and a "copy number". For extrusions, the copy numbers are REMAP_TOP and REMAP_BOTTOM. For a Lathe they are LATHE_START and LATHE_END. But wait, Lathe actually creates 3 copies and the first one actually uses a running counter for the copy number rather than a fixed value. Why I don't know.

If a copy is requested but already exists, no new entity is created and remap returns the handle to the existing entity. I believe this is so when an underlying sketch is modified, all remaining original entities and their copies are preserved when the modified extrusion is created. That's important because other groups may have dependencies (constraints) built on those entities so they should not be destroyed and recreated.

This a fairly simple setup, but it's using tempates and parts of the standard library that I'm not familiar with. There is an EntityKeyHash function that I believe XORs the handle and the copy number, but that may produce collisions and I'm not sure at all how that is handled. There is also and EntityKeyEqual function that seem to compare both the handle and the copy number (as opposed to comparing hashes of them) so ordering and equality are not based on the same thing?

[...]

PR #581, opened 2020-04-20
No extra copies of entities in Revolve and Helix groups.

phkahler:
fixes issue #549.
Might break some older files with those groups.


phkahler:
@whitequark any idea what's failing here?

whitequark:
> any idea what's failing here?

Fluke? The cut off build log is indicative of that. Travis sometimes has issues like that on macOS, unfortunately.

whitequark:
Thanks for the fix!

PR #582, opened 2020-04-21
Eliminate some entity copies in lathe groups.

phkahler:
Eliminates a crash due to copy numbers going over 1000.
May break some older files with constraints on lathed entities.

### DIFF
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -516,9 +516,6 @@ void Group::Generate(IdList<Entity,hEntity> *entity,
             Vector axis_pos = SK.GetEntity(predef.origin)->PointGetNum();
             Vector axis_dir = SK.GetEntity(predef.entityB)->VectorGetNum();
 
-            // Remapped entity index.
-            int ai = 1;
-
             // Not using range-for here because we're changing the size of entity in the loop.
             for(i = 0; i < entity->n; i++) {
                 Entity *e = &(entity->Get(i));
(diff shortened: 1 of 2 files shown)


phkahler:
Oops. This also has changes from #581

phkahler:
Rebased.

whitequark:
Thanks for the fix!

Can I ask you to format the git commits a bit differently? The usual convention is to put the summary on the first line, then a blank line, then a detailed description below. This makes commit messages much easier to read in tools like `gitk`, e.g. here's how this commit this commit looks there:

Screenshot_20200421_051602

and here's how your previous one looks (I took the liberty of reformatting it while merging):

Screenshot_20200421_051645

(They'll also be a bit easier to read on GitHub as well, though the difference is much smaller there.)

phkahler:
No problem. When I put a single line comment in the commit. When I created a PR it looked bad on github so I edited there. Probably best to leave it alone.

whitequark:
Oh yeah, I agree it's pretty annoying that commit messages look best with manual wrapping if you're viewing them in gitk, but the very same commit messages look a lot worse on GitHub when it automatically copies them into the PR body. I expect a lot more people to read git history than pull request comments so I spend effort to make that readable. Sometimes I edit things in the PR body to make them look nicer but that's only if a lot of people will be potentially looking at it.

PR #584, opened 2020-04-22
Allow Revolve and Helix end surface selection and constraints.

phkahler:


phkahler:
Among other things, this makes a 180 degree arc easy by selecting a point on one end and constraining to the surface of the other end. That's the correct DoF of constraint on Revolve. I'm still not entirely sure what the remapping is doing here, it was copied from some other places but seems to work after I revisited and tweaked today.

Issue #587, opened 2020-04-25
Frequent build timeouts for arm64 snap


ppd:
Travis "deploy" stage for arm64 snap fails often with message: "No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself."

It's always in the CMake install stage, which is invoked by snapcraft. No build-error, just no output for ten minutes. There's nothing extraordinary going on and the hardware seems "OK fast", so it shouldn't choke on a copy of a font or icon.

WIP, obviously...

phkahler:
@ppd Thank you for spending so much time making SolveSpace available via Snap. I build from source, but the more ways there are for other people to get it the better!

ppd:
There's something wrong with stdout/stderr capturing on arm64. It's not really stalling, as even output from simple echo loops fails to show after a while. 

It would be simple to redirect snapcrafts output to a logfile, one would think. Then there's less output and maybe that will help. But no; apparmor confines its execution and makes delegation of file descriptors impossible, which means no piping etc.

It's not a huge deal. Travis will eventually fix their stuff without a doubt. Until then, it's maybe wise to pause the edge builds for arm64. It will fail all the time for no good reason and that's not so helpful, I think.

whitequark:
@ppd By any chance, would [this](https://github.com/solvespace/solvespace/blob/master/.travis.yml#L18) also help arm64?

ppd:
@whitequark Same problem as with the logfile, I think. Apparmor will shoot it down with sth like `Log: apparmor="DENIED" operation="file_inherit" profile="/usr/lib/snapd/snap-confine" [...]`

whitequark:
> Until then, it's maybe wise to pause the edge builds for arm64.

Let's do that. Let's also remove the builds for arm64 from the store so that people don't unintentionally download outdated buggy versions. (There's probably very few people who can run it in the first place at the moment, too.)

ppd:
Support request regarding snapcraft output redirection: https://forum.snapcraft.io/t/redirecting-snapcraft-output/16948

ppd:
Related Travis issue: https://travis-ci.community/t/output-is-truncated-heavily-in-arm64-when-a-command-hangs/7630

ppd:
@whitequark What about adding `allow_failures` to Travis cfg to prevent this hickup from polluting the CI results?

[...]

PR #589, opened 2020-04-27
Travis: allow Snap arm64 to fail

ppd:
arm64 builds on Travis are not yet fully mature and this
causes a high failure rate that is not caused by our code.

Continue building and deploying arm64 to the edge channel,
but don't consider the result for the success of the whole
build job.

Alleviates #587


ppd:
A touch of irony: Snap arm64 did not fail with this commit

PR #591, opened 2020-05-01
Bernstein polynomials with no branching.

phkahler:
This is cleaner and smaller, but I was not able to measure a performance impact.

### DIFF
--- a/src/srf/ratpoly.cpp
+++ b/src/srf/ratpoly.cpp
@@ -13,84 +13,31 @@
 // and convergence should be fast by now.
 #define RATPOLY_EPS (LENGTH_EPS/(1e2))
 
-double SolveSpace::Bernstein(int k, int deg, double t)
+static double Bernstein(int k, int deg, double t)
 {
-    if(k > deg || k < 0) return 0;
+// indexed by [degree][k][exponent]
+    static const double bernstein_coeff[4][4][4] = {
+    { { 1.0,0.0,0.0,0.0 },  { 1.0,0.0,0.0,0.0 }, { 1.0,0.0,0.0,0.0 }, { 1.0,0.0,0.0,0.0 } },
+    { { 1.0,-1.0,0.0,0.0 }, { 0.0,1.0,0.0,0.0 }, { 0.0,0.0,0.0,0.0 }, { 0.0,0.0,0.0,0.0 } },
+    { { 1.0,-2.0,1.0,0.0 }, { 0.0,2.0,-2.0,0.0 },{ 0.0,0.0,1.0,0.0 }, { 0.0,0.0,0.0,0.0 } },
+    { { 1.0,-3.0,3.0,-1.0 },{ 0.0,3.0,-6.0,3.0 },{ 0.0,0.0,3.0,-3.0}, { 0.0,0.0,0.0,1.0 } } };
 
-    switch(deg) {
-        case 0:
-            return 1;
-
-        case 1:
-            if(k == 0) {
-                return (1 - t);
-            } else if(k == 1) {
-                return t;
-            }
-            break;
-
-        case 2:
-            if(k == 0) {
-                return (1 - t)*(1 - t);
-            } else if(k == 1) {
-                return 2*(1 - t)*t;
-            } else if(k == 2) {
-                return t*t;
-            }
-            break;
-
-        case 3:
-            if(k == 0) {
-                return (1 - t)*(1 - t)*(1 - t);
-            } else if(k == 1) {
-                return 3*(1 - t)*(1 - t)*t;
-            } else if(k == 2) {
-                return 3*(1 - t)*t*t;
-            } else if(k == 3) {
-                return t*t*t;
-            }
-            break;
-    }
-    ssassert(false, "Unexpected degree of spline");
+    const double *c;
+    c = bernstein_coeff[deg][k];
+    return (((c[3]*t+c[2])*t)+c[1])*t+c[0];
 }
 
-double SolveSpace::BernsteinDerivative(int k, int deg, double t)
+static double BernsteinDerivative(int k, int deg, double t)
 {
-    switch(deg) {
-        case 0:
-            return 0;
+    static const double bernstein_derivative_coeff[4][4][3] = {
+    { { 0.0,0.0,0.0 },  { 0.0,0.0,0.0 }, { 0.0,0.0,0.0 }, { 0.0,0.0,0.0 } },
+    { { -1.0,0.0,0.0 }, { 1.0,0.0,0.0 }, { 0.0,0.0,0.0 }, { 0.0,0.0,0.0 } },
+    { { -2.0,2.0,0.0 }, { 2.0,-4.0,0.0 },{ 0.0,2.0,0.0 }, { 0.0,0.0,0.0 } },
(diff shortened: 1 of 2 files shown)


src/srf/ratpoly.cpp
```diff
@@ -13,84 +13,31 @@
 // and convergence should be fast by now.
 #define RATPOLY_EPS (LENGTH_EPS/(1e2))
 
-double SolveSpace::Bernstein(int k, int deg, double t)
+// indexed by [degree][k][exponent]
+const double bernstein_coeff[4][4][4] = {
```
whitequark, line 17: These should probably be `static` too, for the same reason the functions are.

phkahler: Updated.

whitequark:
(Thanks for keeping the git history clean!)

I'm not sure why the Windows build breaks--does our MSVC version miss `constexpr`? I thought it should have C++11...

phkahler:
Looks like MS didn't add constexpr until VS2015. See the bottom of page:
https://docs.microsoft.com/en-us/cpp/cpp/constexpr-cpp?view=vs-2019
I'll make them static const.

src/srf/ratpoly.cpp
```diff
@@ -13,84 +13,31 @@
...
-            } else if(k == 1) {
-                return 3*(1 - t)*(1 - t)*t;
-            } else if(k == 2) {
-                return 3*(1 - t)*t*t;
-            } else if(k == 3) {
-                return t*t*t;
-            }
-            break;
-    }
-    ssassert(false, "Unexpected degree of spline");
```
ruevs, line 54: Hm I was going to complain about the lack of range checking but it seems that when the calls are evaluated at compile time nothing "bad" happens:
https://godbolt.org/z/oLA-gN
Maybe leave an assert for debug move (-O0)?

src/srf/ratpoly.cpp
```diff
@@ -13,84 +13,31 @@
...
+    { { 1.0,-2.0,1.0,0.0 }, { 0.0,2.0,-2.0,0.0 },{ 0.0,0.0,1.0,0.0 }, { 0.0,0.0,0.0,0.0 } },
+    { { 1.0,-3.0,3.0,-1.0 },{ 0.0,3.0,-6.0,3.0 },{ 0.0,0.0,3.0,-3.0}, { 0.0,0.0,0.0,1.0 } } };
+
+static const double bernstein_derivative_coeff[4][4][3] = {
+    { { 0.0,0.0,0.0 },  { 0.0,0.0,0.0 }, { 0.0,0.0,0.0 }, { 0.0,0.0,0.0 } },
+    { { -1.0,0.0,0.0 }, { 1.0,0.0,0.0 }, { 0.0,0.0,0.0 }, { 0.0,0.0,0.0 } },
+    { { -2.0,2.0,0.0 }, { 2.0,-4.0,0.0 },{ 0.0,2.0,0.0 }, { 0.0,0.0,0.0 } },
+    { { -3.0,6.0,-3.0 },{ 3.0,-12.0,9.0 },{ 0.0,6.0,-9.0}, { 0.0,0.0,3.0 } } };
+
+static double Bernstein(int k, int deg, double t)
```
ruevs, line 29: Maybe
`static double Bernstein(cons int k, const int deg, const double t)`
But the optimizer figures it out anyway...

src/srf/ratpoly.cpp
```diff
@@ -13,84 +13,31 @@
...
-            break;
-    }
-    ssassert(false, "Unexpected degree of spline");
+    const double *c;
+    c = bernstein_coeff[deg][k];
+    return (((c[3]*t+c[2])*t)+c[1])*t+c[0];
 }
 
-double SolveSpace::BernsteinDerivative(int k, int deg, double t)
+static double BernsteinDerivative(int k, int deg, double t)
```
ruevs, line 36: Maybe
`static double BernsteinDerivative(cons int k, const int deg, const double t)`
But the optimizer figures it out anyway...

src/srf/ratpoly.cpp
```diff
@@ -13,84 +13,31 @@
 // and convergence should be fast by now.
 #define RATPOLY_EPS (LENGTH_EPS/(1e2))
 
-double SolveSpace::Bernstein(int k, int deg, double t)
+// indexed by [degree][k][exponent]
+static const double bernstein_coeff[4][4][4] = {
```
ruevs, line 17: I would even move these inside the respective functions. (as `static const` of course).

ruevs:
By the way, you were probably not able to measure a performance difference, because modern compilers are pretty good. Take a look at this:
https://godbolt.org/z/XxefSH

src/srf/ratpoly.cpp
phkahler: @ruevs I was a little concerned about the range checking too, but these are only called by other functions in ratpoly.cpp and those are lacking range checking.

phkahler:
@ruevs with the new structure we could do even better. These two functions are always called from within a loop over the k-th polynomial. We could transpose the inner 2 dimensions of the array so we'd have vectors of coefficients for each power. Then write fixed-size loops 0-3 to do each term of the polynomials. SSE or AVX optimizations would compute all 4 polynomials in about the time this does one. It could return (or fill in) an array of 4 results and take the call out of the loops in higher level functions (which might then be more ready for the compiler to optimize).

But that all seems like premature optimisation. I think this PR reduces complexity ;-)

src/srf/ratpoly.cpp
phkahler: I moved them inside thinking it would look worse, but now I like it.

phkahler:
@ruevs I did the parallel implementation on my bernstein branch if you want to see. I used plain for-loops to go over all 4 polynomials one term at a time. From what I've read that's what modern vectorizing compilers like. It slowed things down dramatically (test went from 127ms per frame to 137) when I first changed it but only returned 1 of the 4 values. I didn't expect that, thinking AVX could do all 4 in the time it would normally do only 1. Then I realized I'm not sure how to confirm my compiler setting when building SolveSpace - it's probably using Fedora defaults? IDK.  Then I modified it to the latest form and changed PointAt() to make only 1 call and that got the performance back to where it was. I then modified the other 3 functions to use the vector form of the functions and ended up back at 127ms, but it's not doing 4x the work any more. I still think it could run much faster and probably isn't being properly optimized (I'm on a Zen+ core).

Anyway, I have a new version that theoretically could be faster, but I'm considering it shelved until I figure out how to do better measurements and get a handle on what GCC is doing or not.

Lesson: SolveSpace does make a LOT of calls to PointAt() and will suffer measurably if that function is implemented poorly.

whitequark:
> Then I realized I'm not sure how to confirm my compiler setting when building SolveSpace - it's probably using Fedora defaults? IDK.

I suspect you'll need to pass `-march=native` explicitly to see any use of AVX. Also, AVX isn't necessarily going to result in an overall throughput increase due to AVX offset reducing turbo frequency on Intel CPUs.

phkahler:
@whitequark where do I put that? This build system is new to me and I can't find where the compiler flags get set.

rpavlik:
CMAKE_CXX_FLAGS

PR #592, opened 2020-05-01
WIP Initial mirror groups implementation.

phkahler:
WIP so this isn't ready. I am looking for some help.

This implements mirror groups (SHIFT-M) which can be a sketch or a solid. It creates a non-visible mirror plane between the original and copy of the sketch/model. In 3D you get 3DoF. In 2D it constrains the mirror to be orthogonal to the sketch plane so only 2DoF. That's what works. What I'd like help with is the following:

1) Faces of the solid are not selectable for either the original or copy.

2) After mirroring a 2D sketch, you can't extrude the resulting 2D entities.

3) In 2D I'd like to optionally select 2 points or a point-and-vector to constrain the mirror plane. Not sure where or how to store the selection information for use later when constraints on the mirror are created.

4) Same issue as #3 but for 3D. I'd like to be able to select a face and constrain the mirror plane to that. Optionally if one is selected when the group is created.

I don't think it wise to merge this until these issues are resolved. Any advice or code would be appreciated. Oh, or comments on how it should be done differently.

### DIFF
--- a/src/drawentity.cpp
+++ b/src/drawentity.cpp
@@ -86,6 +86,7 @@ void Entity::GetReferencePoints(std::vector<Vector> *refs) {
         case Type::POINT_N_ROT_TRANS:
         case Type::POINT_N_ROT_AA:
         case Type::POINT_N_ROT_AXIS_TRANS:
+        case Type::POINT_MIRROR:
         case Type::POINT_IN_3D:
         case Type::POINT_IN_2D:
             refs->push_back(PointGetNum());
@@ -96,6 +97,7 @@ void Entity::GetReferencePoints(std::vector<Vector> *refs) {
         case Type::NORMAL_N_ROT_AA:
         case Type::NORMAL_IN_3D:
         case Type::NORMAL_IN_2D:
+        case Type::NORMAL_MIRROR:
         case Type::WORKPLANE:
         case Type::CIRCLE:
         case Type::ARC_OF_CIRCLE:
@@ -121,6 +123,7 @@ void Entity::GetReferencePoints(std::vector<Vector> *refs) {
         case Type::FACE_N_TRANS:
         case Type::FACE_N_ROT_AA:
         case Type::FACE_ROT_NORMAL_PT:
+        case Type::FACE_MIRROR:
             break;
     }
 }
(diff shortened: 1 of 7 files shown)


phkahler:
Going to close this PR but I'll leave my mirror branch around for a while.  I've decided the feature is probably inferior. Wouldn't a better idea be to make mirrored copy of a group that has a full 6 DoF instead of 3? And since that would just be a copy with scale of -1, wouldn't it be better to allow the user to specify the scale of the copy? I think that could be done without introducing a new copy type in the solver. This might also be like importing multiple instances of a part into a model without the need for linking another file.

Thoughts on that?

whitequark:
I think I lack some of the context to understand your proposal well. How would mirror groups would have six DoF?

phkahler:
In the original idea and the prototype they only have 3DoF. That's because the group is copied across a mirror plane and that plane only has 3DoF.  It's very interesting to drag the copy around because the orientation changes as you move it. But why should a copy not have a full 6DoF? What if you want to make a mirror image copy but reposition it independent of orientation? If that's the case we can just make a generic group copy and use the scale = -1  to flip it if you want a mirror image.

Even if we want the fancy mirror plane, I think the implementation should be a full copy with 7 parameters (quaternion with length=1 and offset). We won't need a new copy type (I think POINT_N_ROT_TRANS will do), just use constraints to enforce the virtual mirror condition and reduce it to 3DoF.  Much like Extrude is implemented with a full 3 parameters for translation but uses 2 constrains to keep it perpendicular to the sketch plane.

Having the generic copy capability would be nice, and having it scalable would too.

It also has implications for my extrude along a path plan, but that's another story. I keep wanting to make a video to show how the 2 are related.

whitequark:
I agree that it makes a lot of sense to make the copy capability generic internally. I think that it also makes a lot of sense to have a dedicated mirror group because removing the unnecessary degrees of freedom is quite annoying and probably the worst usability issue in the import feature as it currently exists.

phkahler:
Or more simply. You're making a robot and create the right hand. You can make the left hand with a mirror copy of the right, but that doesn't mean you want their positions correlated in any way.

phkahler:
For mirror I was going to have a few ways to construct it. If you mirror a sketch it adds a constraint to the copy doesn't come out of the plane. In 3d I wanted to allow selection of a face and fully constrain across that. Those types of things will still be possible, but the internals should probably be a generic copy.

Issue #537, opened 2019-12-23
Intersecting differences of triangular prisms flips faces


rpavlik:
### System information

SolveSpace version: 3.0~64c0f62b

Operating system: Debian Buster

### Expected behavior

I have essentially a rectangular prism that I wanted to cut two triangular grooves in, one "vertically" and one "horizontally" (same height, and intersecting - making vise jaws).  So, one 2d sketch + extrude for positive solid, one 2d sketch and extrude as difference for one groove, then perpindicular 2d sketch and extrude as difference for other groove. It should look normal.

### Actual behavior

Several of the faces are red, indicating they are back-faces. (The solver didn't fail - it thinks it succeeded.) It actually looks like I'm seeing thru missing faces, when I went and redrew it...

![difference-fail](https://user-images.githubusercontent.com/61129/71367578-1eff5600-256b-11ea-851a-f47a20415834.png)

![image](https://user-images.githubusercontent.com/61129/71367605-2fafcc00-256b-11ea-836c-3e6c5fb52a69.png)

![image](https://user-images.githubusercontent.com/61129/71367630-422a0580-256b-11ea-85db-68145c867543.png)

Forcing to trimesh does fix it (despite the lack of a solver fail message), but I would have guessed that this was a fairly easy geometry to handle - nothing curved :) 

![image](https://user-images.githubusercontent.com/61129/71367687-684fa580-256b-11ea-9b5f-14b477eaa698.png)

### Additional information

open `jaw-covers.slvs` - the other is my reference linked in. (I'm not sure how to, or if I can, "re-seat" the origin of sketches to be able to simplify this one without redrawing) 

[jaw-.zip](https://github.com/solvespace/solvespace/files/3995496/jaw-.zip)

OK, decided to not be lazy and re-draw a simpler test case.

![Peek 2019-12-23 10-06](https://user-images.githubusercontent.com/61129/71367882-f1ff7300-256b-11ea-9e7e-5a43b29150be.gif)

[example2.zip](https://github.com/solvespace/solvespace/files/3995520/example2.zip)

Symbian9:
As a workaround, just lower one of triangular to `0.001` mm

- https://gist.github.com/Symbian9/c72d515ccba9fa939f57aaf73aa48700

![pic.1](https://i.imgur.com/0kY1Zfq.png)

ruevs:
This seems to not be NURBS failure, but rather a failure in triangulating the "surfaces" (planes). It fails twice here:
[...]

Issue #481, opened 2019-09-11
Question/Suggestion Parallel tesselation.


phkahler:
It looks like converting surfaces to triangles for display can take a lot of time. SolveSpace doesn't make use of multi-core CPUs as far as I can tell. One place we might try this is in the amazingly short function:

```
void SShell::TriangulateInto(SMesh *sm) {

    SSurface *s;
    for(s = surface.First(); s; s = surface.NextAfter(s)) {
        s->TriangulateInto(this, sm);
    }
}
```
All shells will have multiple surfaces. I'm not sure that function call can be done in parallel. Maybe we could have a look. If so, we'd just create a vector of triangle lists and use OpenMP parallel for (or equivalent?) to go over handing each surface its own list to populate and then concatenate the triangle lists after.

This hinges on weather that is a pure function. If so, what's the preferred way to do things in parallel? I'm not aware of any place SS does that.

whitequark:
We'd have to investigate what sort of parallel constructs are available (MSVC looks like it could be the problem here), but it's an option.

baryluk:
C++17 `std::execution` is probably the best choice. It looks to be available in MSVC too.

whitequark:
It'd require bumping our MSVC version which historically has been a problem. Definitely solvable, just needs time and effort.

phkahler:
@baryluk if you wanted to prove the concept and report back any performance improvement, that would be awesome. We can decide on which parallel construct is best, or what MSVC version may be needed once there is a demonstrated improvement.

phkahler:
This is not parallel, but it does confirm that it could be run that way. Not sure what the best construct is to use, but tm could be a list/vector/etc of SSmesh objects whose count is known from surface.l.length or something like that. Not the best way to combine the lists either.

`
void SShell::TriangulateInto(SMesh *sm) {

    SSurface *s;

    SMesh tm = {};

    STriangle *st;

    for(s = surface.First(); s; s = surface.NextAfter(s)) {

        s->TriangulateInto(this, &tm);

        for (st = tm.l.First(); st; st = tm.l.NextAfter(st)) {

            sm->AddTriangle(st);

        }

        tm.Clear();

    }

}
`

[...]

PR #594, opened 2020-05-07
Don't create redundant edges in UvGridTriangulate into.

phkahler:
Redundant edges were culled by a function with O(n^2) time. The number of edges was 4xy so not needing to cull them increases frame rate on surfaces of degree 2x2 or higher while dragging.

### DIFF
--- a/src/srf/triangulate.cpp
+++ b/src/srf/triangulate.cpp
@@ -437,68 +437,88 @@ void SPolygon::UvGridTriangulateInto(SMesh *mesh, SSurface *srf) {
     srf->MakeTriangulationGridInto(&li, 0, 1, /*swapped=*/true);
     lj.Add(&v);
     srf->MakeTriangulationGridInto(&lj, 0, 1, /*swapped=*/false);
+    if ((li.n > 3) && (lj.n > 3)) {
+        // Now iterate over each quad in the grid. If it's outside the polygon,
+        // or if it intersects the polygon, then we discard it. Otherwise we
+        // generate two triangles in the mesh, and cut it out of our polygon.
+        // Quads around the perimeter would be rejected by AnyEdgeCrossings.
+        std::vector<bool> bottom(lj.n, false); // did we use this quad?
+        int i, j;
+        for(i = 1; i < (li.n-1); i++) {
+            bool prev_flag = false;
+            for(j = 1; j < (lj.n-1); j++) {
+                bool this_flag = true;
+                double us = li[i], uf = li[i+1],
+                       vs = lj[j], vf = lj[j+1];
 
-    // Now iterate over each quad in the grid. If it's outside the polygon,
-    // or if it intersects the polygon, then we discard it. Otherwise we
-    // generate two triangles in the mesh, and cut it out of our polygon.
-    int i, j;
-    for(i = 0; i < (li.n - 1); i++) {
-        for(j = 0; j < (lj.n - 1); j++) {
-            double us = li[i], uf = li[i+1],
-                   vs = lj[j], vf = lj[j+1];
+                Vector a = Vector::From(us, vs, 0),
+                       b = Vector::From(us, vf, 0),
+                       c = Vector::From(uf, vf, 0),
+                       d = Vector::From(uf, vs, 0);
 
-            Vector a = Vector::From(us, vs, 0),
-                   b = Vector::From(us, vf, 0),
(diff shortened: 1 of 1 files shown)


phkahler:
Dragging my test model went from 125-130ms per frame or 7FPS, to around 90ms or 11FPS.

phkahler:
This seems to add a bug with this file and chord tolerance = 0.1.  I'll have a look.
[intersect_ct1.zip](https://github.com/solvespace/solvespace/files/4590263/intersect_ct1.zip)

phkahler:
@ruevs Could I convince you to have a look at this? The gist is this:

In the UvGridTriangulateInto (in triangulate.cpp) there is a grid created in U,V space over a surface. First the spacing is determined in U and V, then 2 nested loops generate quads between 4 points on the grid. If any of the polygon edges (piecewise linear trim curves) cross the perimeter of the quad it is rejected. Otherwise 2 triangles are created and 4 new edges are created in "holes" which get added to the polygon later (after all the holes are identified). Once all the holes are have been cut and covered in triangles the holes are merged into the polygon and then it calls UvTriangulateInto() which fills in the rest with triangles (That's all thats used for planar surfaces / no grid).

The inefficiency is that the quads (holes) create 4 edges each which are added to the holes polygon. Since quads are adjacent, any 2 adjacent quads will create a pair of edges running in opposite directions. After all the holes are cut a call is made to eliminate the redundant edges and that function has O(n^2) time which is really bad when there are 4*U*V edges. The fix here is to never create the redundant edges in the first place and eliminate that call.

I also made a shortcut by not making quads for the perimeter. It turns out those are rejected anyway because they have edges from the original perimeter.

Everything worked fine. Then I realized that quads that completely contain edges will not be rejected. even in the original code. The test model proved that, but also showed that the new implementation has problems. It's also much faster. The test model I uploaded is good to show the issue.

I was up way too late last night and couldn't figure out what's wrong.

ruevs:
I will look - when (as right now) I am not up way too late (finished the intersection) :-)
I looked at his beautiful and "simple" function (with math that I have totally "not" forgotten 20 years ago :-D ) back in December (I came to it from `CullExtraneousEdges` by way of `boolean.cpp`) and I could not fully figure it out. However it was clear enough that it is doing a lot of unnecessary work.
You on the other hand have done a great job describing what it does. So I will try again :-)

Yesterday I downloaded your model and played with it today while doing the intersection.

ruevs:
By the way - since you are looking at  `SPolygon::UvTriangulateInto` I can offer you two very simple models that cause "couldn't find an ear! fail" (https://github.com/solvespace/solvespace/blob/master/src/srf/triangulate.cpp#L353)
I know `SContour::UvTriangulateInto` it is not exactly on topic, but this has been bugging me for at least 9 months and is the cause for some of the "NURBS" failures :-) For example https://github.com/solvespace/solvespace/issues/537
[CouldntFindAnEar_NURBSTests.zip](https://github.com/solvespace/solvespace/files/4596617/CouldntFindAnEar_NURBSTests.zip)

phkahler:
@ruevs   I found it. Turns out the line "hp.l.Clear();" is inside a loop when it should be outside. I had noticed that it works when the hole is connected to the outer ring of non-quads but it fails when the hole in the grid of quads should be creating a separate inner perimeter. Oops.

Thank you anyway!   I will have a look at what you suggest tomorrow. This is an area I have not looked at much. I just learned what those ears are earlier today.

ruevs:
Great! And it works (fetched your branch, merged with intersection).
Roughly 2.8 times faster when switching from intersection to difference on my machine with your intersect_ct1.slvs and chord tolerance 0.01 (75s vs. 27s).

I will still read and understand your changes... tomorrow.

phkahler:
This came out of my looking at parallel triangulation. For #481 I wanted to be sure those functions were pure - not modifying shared data structures which won't work in parallel. I did some experiments disabling code that wouldn't be needed for simple curved surfaces and stumbled into a performance issue. So far this fix improves performance more than running it in parallel on 4 cores.

If lowering chord tolerance increases edge segments by N, and surface triangles by N^2, the removal of duplicate edges was taking N^4. It seemed best not to create redundant edges in the first place.

What I find odd is that now this isn't even close to being the performance bottleneck, at least on my simple test models.

If this is up to SS standards for the code I'll merge it after a little more testing.

Issue #35, opened 2016-08-23
Add "intersection" boolean operation for combining solids


Symbian9:
Add this types for logic how solid model combine.
- "intersect" - make visible only those region where solid models intersect 
- "XOR" (exclusive or) - make visible only those regions of solid models that not intersect  (inverted to "intersect")

Evil-Spirit:
I don't see real cases where we need XOR. @Symbian9, any samples?

Symbian9:
XOR usefull for design parts of complex shape, for example in engine and watch industry.

mmiscool:
If this is the feature I am thinking where the solid result is where the 2 solds intersected. This is particularly useful in many circumstances. 

The image link below shows one possible use for such an operation. 
http://help.autodesk.com/cloudhelp/2016/ENU/AutoCAD-Core/images/GUID-1C31FCED-9999-4012-B6F9-518D00711466.gif

whitequark:
There is no question that an intersect ("AND") operation is useful. I don't see any point in "XOR" though.

Evil-Spirit:
@whitequark,
I think this is a lot of unpredictable work. I actually have tried do the same thing for NotSolveSpace, but some problems force me to decide stop doing this.
If @jwesthues can help, we can try again.

whitequark:
@Evil-Spirit OK.

ghoss:
**FYI:** I've implemented the "intersection" operation in [my own fork](https://github.com/ghoss/solvespace) now. I needed this function for my 3D printing, where I frequently wish to extract part of a whole object in order to make a test print of that part only.

I circumvented most of the "unpredictable work" mentioned by @Evil-Spirit by rewriting the intersection operation **A * B** as the result of two difference operations **A - (A - B)**. This incurs some overhead of course, but with the benefit of less unpredictability :)

Implementation was relatively straightforward this way and mainly consisted of modifications to `srf/boolean.cpp` for the non-mesh solids:

```
void SShell::MakeFromIntersectionOf(SShell *a, SShell *b) {
   SShell c = {};
   c.MakeFromBoolean(a, b, SSurface::CombineAs::DIFFERENCE);
   MakeFromBoolean(a, &c, SSurface::CombineAs::DIFFERENCE);
   c.Clear();
}
```

and to `mesh.cpp` for the meshes:

```
void SMesh::MakeFromIntersectionOf(SMesh *a, SMesh *b) {
   SMesh c = {};
[...]

PR #595, opened 2020-05-08
NURBS: Add intersection boolean operation.

ruevs:
The NURBS operation is properly implemented.

ToDo: The mesh operation in SMesh::MakeFromIntersectionOf
is still done as C=A-(A-B).

Implements: https://github.com/solvespace/solvespace/issues/35

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changelog
 ---
 
 New sketch features:
+  * New intersection boolean operation for solid models.
   * New groups, revolution and helical extrusion.
   * Extrude, lathe, translate and rotate groups can use the "assembly"
     boolean operation, to increase performance.
--- a/src/bsp.cpp
+++ b/src/bsp.cpp
@@ -62,14 +62,17 @@ void SBsp3::InsertInPlane(bool pos2, STriangle *tr, SMesh *m) {
         ll = ll->more;
     }
 
-    if(m->flipNormal && ((!pos2 && !onFace) ||
-                                   (onFace && !sameNormal && m->keepCoplanar)))
-    {
-        m->AddTriangle(tr->meta, tr->c, tr->b, tr->a);
-    } else if(!(m->flipNormal) && ((pos2 && !onFace) ||
-                                   (onFace && sameNormal && m->keepCoplanar)))
-    {
-        m->AddTriangle(tr->meta, tr->a, tr->b, tr->c);
+    if((!onFace && ((m->keepInsideOtherShell && !pos2) ||
+                    (!m->keepInsideOtherShell && pos2))) ||
+       (onFace && ((m->keepCoplanar && m->flipNormal && !sameNormal) ||
+                   (m->keepCoplanar && !m->flipNormal && sameNormal)))) {
+        // We have decided that we need to keep a triangle either inside,
+        // outside or on the other shell. So add it and flip it if requested.
+        if(!(m->flipNormal)) {
+            m->AddTriangle(tr->meta, tr->a, tr->b, tr->c);
+        } else {
+            m->AddTriangle(tr->meta, tr->c, tr->b, tr->a);
+        }
     } else {
         m->atLeastOneDiscarded = true;
     }
@@ -101,10 +104,15 @@ void SBsp3::InsertHow(BspClass how, STriangle *tr, SMesh *instead) {
     return;
 
 alt:
-    if(how == BspClass::POS && !(instead->flipNormal)) {
-        instead->AddTriangle(tr->meta, tr->a, tr->b, tr->c);
-    } else if(how == BspClass::NEG && instead->flipNormal) {
-        instead->AddTriangle(tr->meta, tr->c, tr->b, tr->a);
+    if(((BspClass::POS == how) && !instead->keepInsideOtherShell) ||
+       ((BspClass::NEG == how) && instead->keepInsideOtherShell)) {
+        // We have decided that we need to keep a triangle (either inside or
+        // outside the other shell. So add it and flip it if requested.
+        if(!(instead->flipNormal)) {
+            instead->AddTriangle(tr->meta, tr->a, tr->b, tr->c);
+        } else {
+            instead->AddTriangle(tr->meta, tr->c, tr->b, tr->a);
+        }
     } else if(how == BspClass::COPLANAR) {
         if(edges) {
             edges->InsertTriangle(tr, instead, this);
@@ -431,6 +439,9 @@ SBsp3 *SBsp3::InsertConvex(STriMeta meta, Vector *vertex, size_t cnt, SMesh *ins
 SBsp3 *SBsp3::InsertOrCreate(SBsp3 *where, STriangle *tr, SMesh *instead) {
     if(where == NULL) {
         if(instead) {
+            // ruevs: I do not think this code is reachable, but in
+            // principle should we use instead->keepInsideOtherShell
+            // in place of instead->flipNormal ?
             if(instead->flipNormal) {
                 instead->atLeastOneDiscarded = true;
             } else {
--- a/src/groupmesh.cpp
+++ b/src/groupmesh.cpp
@@ -143,7 +143,7 @@ void Group::GenerateForStepAndRepeat(T *steps, T *outs, Group::CombineAs forWhat
         // And tack this transformed copy on to the return.
         if(soFar->IsEmpty()) {
             scratch->MakeFromCopyOf(&transd);
-        } else if (forWhat == CombineAs::ASSEMBLE) {
+        } else if(forWhat == CombineAs::ASSEMBLE) {
             scratch->MakeFromAssemblyOf(soFar, &transd);
         } else {
             scratch->MakeFromUnionOf(soFar, &transd);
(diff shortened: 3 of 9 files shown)


phkahler:
This is incredibly clean given what it took to figure it out!

whitequark:
Excellent work! If you don't mind, I would prefer to hold off merging this until the mesh intersection is implemented properly, since there is nothing so permanent as a temporary workaround.

(Also, does the CHANGELOG include some unrelated items by accident?)

ruevs:
I don't mind at all. Hopefully this time I won't take me another 9 months :-D
The CHANGELOG.md was not by mistake - just some additions I thought worthy while digging in the history. But you are right, I'll make a separate pull request for it. Amended.

ruevs:
> 
> 
> This is incredibly clean given what it took to figure it out!

Since Solvespace is incredibly clean and beautifull I am very glad it came out like this. As @jwesthues  said [back in May 2018 literally "eight lines of code"](https://github.com/solvespace/solvespace/issues/35#issuecomment-390127695). Of course [the real history is different](https://github.com/ruevs/solvespace/tree/wip-intersection) but the master branch does not deserve this :-) I mostly agree with @whitequark  that the history in the master branch should be clean and beautiful as if geniuses with OCD wrote it :-D

whitequark:
> clean and beautiful as if geniuses with OCD wrote it :-D

It's not so much about clean and beautiful as about the raw practicalities of combing through `git blame` to understand some feature and using `git bisect` to find where a bug comes from. A project with *excellent* continuous integration and test suite like rustc can afford to have a messy history, we can't.

phkahler:
@ruevs  I took a look at Mesh booleans in mesh.cpp.   I just modified SMesh::MakeFromDifference a bit a found the following:

A new empty shell is created and then surfaces are added from mesh B, followed by some from mesh A. For Union that's it, so I figured the BSP is rejecting surfaces inside the shell. For difference, the normals of B are flipped, so it keeps surfaces from A that are outside B (inside based on flipped normals).

Modifying that to set flipnormal = true for both operations give surfaces from B that are inside A, and surfaces from A that are inside B. That's intersection. There are only 2 problems. The resulting mesh is RED - inside out. And I'm not sure what to do for coincident surfaces. We need to keep them but that's not happening. Most of the logic is already there in some form.

Seems like this is going to be another "easy" thing once the "right" way is found.

phkahler:
Sometimes I ramble. Couldn't remember doing much special for Shells in my mirror branch, so I checked and it does make mirror images when forced to mesh. Funny thing is mirroring involves doing a scale of -1 so I figured it would also flip the normals back, but there is no such operation done. The scaling is called from inside a templated function I had to write for mirroring, so that led me to SMesh::MakeFromTransformationOf  which does the scaling. Right there in the middle it checks for a negative scale and swaps two verticies.

ruevs:
> 
> 
> Modifying that to set flipnormal = true for both operations give surfaces from B that are inside A, and surfaces from A that are inside B. That's intersection. There are only 2 problems. The resulting mesh is RED - inside out. And I'm not sure what to do for coincident surfaces. We need to keep them but that's not happening. Most of the logic is already there in some form.

You are right. I also figured out how to make red (inside out) mesh intersections back in September last year. But I never committed - because of the problems you listed.

```
void SMesh::MakeFromIntersectionOf(SMesh *a, SMesh *b) {
    SBsp3 *bspa = SBsp3::FromMesh(a);
    SBsp3 *bspb = SBsp3::FromMesh(b);

    flipNormal = true;
//    keepCoplanar = true;  // Maybe? Does not help :-(
    AddAgainstBsp(a, bspb);

    flipNormal = true;
    keepCoplanar = true;
    AddAgainstBsp(b, bspa);
}
```

ruevs:
By the way the interesting code seems to be [here](https://github.com/solvespace/solvespace/blob/master/src/bsp.cpp#L65) and [here](https://github.com/solvespace/solvespace/blob/master/src/bsp.cpp#L104).

phkahler:
Is there a way to build a BSP with flipped normals?  A*B  = A - !B

jwesthues:
Currently, we have:

operand | `flipNormal` | `keepCoplanar`
------------ | ------------- | ------
union A | false | true
union B | false | false
difference A | false | false
difference B | true | true

Note that even though the union operation is commutative with respect to volumes, the two operands are treated differently. This is because we need to keep one copy of any coincident bounding surfaces, not both. It's arbitrary which copy we keep.

`flipNormal` turns the mesh inside out by flipping the normals (by reversing the order of the triangle vertices), as the name suggests; but it also changes the rules on what pieces to keep and discard. For intersection, those rules are what you want but the flipping is not. So I think you need to split those two behaviors into two flags, by adding something like `keepInsideOtherShell`. Then you'd have:

operand | `flipNormal` | `keepCoplanar` | `keepInsideOtherShell`
------------ | ------------- | ------ | ----
union A | false | true | false
union B | false | false | false
difference A | false | false | false
difference B | true | true | true
intersection A | false | false | true
intersection B | false | true | true

Or you could replace the three boolean flags with just one enumeration of `UNION_OP_A`, etc.

ruevs:
Thanks! I hope tomorrow I'll have some time and look.

phkahler:
@ruevs I applied a patch to my master based on the commit here. Then I got compiler error in boolean.cpp complaining about  inFace  not being declared in scope.

ruevs:
@jwesthues  Jonathan your hint was really helpful. I implemented the triangle mesh boolean intersection. While doing it I also changed the union to keep the same coplanar faces as the NURBS one does.

@whitequark in my opinion this is done.

@phkahler 
>
>
> I got compiler error in boolean.cpp complaining about inFace not being declared in scope.

You did not merge properly. I removed `inFace` in this commit https://github.com/solvespace/solvespace/commit/ec3056773e2b84803d2acf3b47d653a0fc8cfc46 back in November.

Despite all I have a (really nasty) test model with only flat faces that breaks all boolean operations in both NURBS and mesh mode. I will open a new issue "Nasty models that break booleans" as a convenient place for us to continue work on the boolean engine(s).

whitequark:
@ruevs Thanks. I really like @jwesthues' suggestion of replacing the flags with a enum. Could you do that as well in this PR? We've been replacing booleans with enums for a while and it tends to make logic a lot easier to understand.

ruevs:
In my opinion using an enum with six values will make the code more complex and less readable in this case. Take a look at how the flags are used.

What is more the three flags combined with the `a` and `b` operands keep the information about the type of operation in `mesh.cpp`. If we do enums we "pollute" `bsp.c` with "knowledge" about boolean operations. I think this is bad.

whitequark:
That's reasonable.

jwesthues:
Strictly `bsp.c` is already a bit polluted, with logic like when to keep coincident surfaces with normals in the same vs. opposite directions. I guess you could go the other direction and add more flags, `keepInsideOtherShell`, `keepOutside...`, `keepCoincidentSameNormals`, etc.? Or make them a bit field or something? That would lend itself more to the "stupid" model of building Booleans as experimental lookup tables instead of trying to write intuitive logic (that may later mislead). No strong feelings personally though.

ruevs:
You are right. For example (I think?) with the available flags it is impossible to implement "XOR" since there is no way to express "keep only stuff outside of _both_ shells".

However abstracting it out will be a bigger change with a bit more risk of breaking things. Maybe we should leave it for when/if we need it?

whitequark:
I'm OK with leaving that code as-is for now.

phkahler:
Booleans will continue to get investigated. Lots of test cases that fail, and now more people looking at it.

On the performance side I was having a look. The BSP creation for each surface in a shell can run in parallel (tried it), but I saw no obvious benefit - this will want to be included in the frame time. But I think the majority of the time is going to SShell::CopySurfacesTrimAgainst() which looks mostly thread safe at the per-surface level but not quite. I want to see about untangling "into" and see if a per-surface list or similar can be used and combined at the end. That function and its call tree nest at least 3 loops over edge lists, so it's probably a good place to look. OTOH it's not good to disrupt things before they work correctly.  correct > fast.

ruevs:
>
>
> I'm OK with leaving that code as-is for now.

> 
> 
> [..] it's not good to disrupt things before they work correctly. correct > fast.

I agree. Normally I am not this reticent about refactoring, But in this case [no one has touched `boolean.cpp` since Augus 2009 when Jonathan last worked on it](https://github.com/solvespace/solvespace/commit/f7f9000c68a837a15b6d0d0fa07f21a14d8eb90d) (apart from cleanups and "NFC" changes) so I am being extra careful.

ruevs:
By the way can someone explain why the history of `boolean.cpp` is not visible in `git log` before this 0a24cf40f07ef64ae28a19a82051fda925e20a68 ?
```
0a24cf40f07ef64ae28a19a82051fda925e20a68
Author: Daniel Richard G <skunk@iSKUNK.ORG>
Date:   Wed Nov 13 02:20:31 2013 -0500

    Moved most of the source into a src/ subdirectory
```
I can get to older commits only with `blame` - for example [the above commit](https://github.com/solvespace/solvespace/commit/f7f9000c68a837a15b6d0d0fa07f21a14d8eb90d) from 2009-08-21. It is the same here on GitHub.   Does it have something to d with the moving?

At the same time the repository as a whole goes happily back to the first commit ever from 2008-03-25 67139236fc0b63e06eaa1efa8fe0022df5612bd2

nabijaczleweli:
You will most likely need the `--follow` argument to `git log`:
```
--follow
   Continue listing the history of a file beyond renames (works only for a single file).
```

On my checkout, `git log --follow src/srf/boolean.cpp` ends at bb4b767e99e95940086d1db21e760570966c1801:
```
commit bb4b767e99e95940086d1db21e760570966c1801
Author: Jonathan Westhues <jwesthues@cq.cx>
Date:   Thu Jan 22 19:30:30 2009 -0800

    Tear everything apart, moving away from meshes and toward shells.
    Add stubs for functions to perform Booleans, and get rid of mesh
    stuff, including the kd tree accelerated snap to vertex (which
    should not be required if the shell triangulation performs as it
    should).

    Also check that a sketch is not self-intersecting before extruding
    it or whatever. This is dead slow, needs n*log(n) implementation.

    [git-p4: depot-paths = "//depot/solvespace/": change = 1902]
```

Whereas `git blame` does this by default:
> The origin of lines is automatically followed across whole-file renames (currently there is no option to turn the rename-following off). To follow lines moved from one file to another, or to follow lines that were copied and pasted from another file, etc., see the `-C` and `-M` options.

ruevs:
Thanks! I learned something new.

phkahler:
This works really well. My only suggestion would be putting the radio buttons on 2 rows:

union ...........difference
intersection...assemble

Just because that is now the widest thing in the text window.

ruevs:
Good idea - done. But I did
```
union assemble
difference intersection
```
looks more logical to me. But if you don't like it I can change it. @whitequark ?

whitequark:
Looks great, thanks everyone!

@ruevs I like your placement more too.

PR #607, opened 2020-05-11
Split quads based on angle.

phkahler:
Produces a higher quality mesh in areas covered in quads by spitting the other way if needed to prevent angles over 90 degrees.

### DIFF
--- a/src/srf/triangulate.cpp
+++ b/src/srf/triangulate.cpp
@@ -443,6 +443,7 @@ void SPolygon::UvGridTriangulateInto(SMesh *mesh, SSurface *srf) {
         // generate two triangles in the mesh, and cut it out of our polygon.
         // Quads around the perimeter would be rejected by AnyEdgeCrossings.
         std::vector<bool> bottom(lj.n, false); // did we use this quad?
+        Vector tu = {0,0,0}, tv = {0,0,0};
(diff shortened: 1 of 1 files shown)


whitequark:
Nice! Any before/after?

phkahler:
And YES, it does do a better job on the silhouette edges in those areas too. I had considered doing the check just once in the middle of the surface, but there are cases where part of the grid is split one way and another area goes the other. Performance impact is negligible.

Before:
before
After
after

phkahler:
Before:
bushes1

After:    (No bushes growing in the corners)
bushes2

whitequark:
By any chance, does this improve donuts?

phkahler:
> By any chance, does this improve donuts?

Nope. In fact they split somewhat randomly because the u,v derivatives are exactly orthogonal. That should be OK since the quads are actually flat, it won't show any difference in 3D printing. If you want to get rid of the bushes on a donut you need to lower chord tolerance a little.

whitequark:
Right, I knew lowering chord tolerances fixed it, but I wasn't sure if that's the only way.

ruevs:
It's nice, but it can be improved ;-)
https://github.com/solvespace/solvespace/pull/613

PR #609, opened 2020-05-11
Use C++ instead of C versions of standard library headers

ppd:
Should be more correct + snaps don't fail to build

### DIFF
--- a/src/solvespace.h
+++ b/src/solvespace.h
@@ -7,16 +7,16 @@
 #ifndef SOLVESPACE_H
 #define SOLVESPACE_H
 
-#include <ctype.h>
-#include <limits.h>
-#include <math.h>
-#include <setjmp.h>
-#include <stdarg.h>
-#include <stddef.h>
-#include <stdint.h>
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
+#include <cctype>
+#include <climits>
+#include <cmath>
+#include <csetjmp>
+#include <cstdarg>
+#include <cstddef>
+#include <cstdint>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
 #include <algorithm>
 #include <chrono>
 #include <functional>


whitequark:
What broke the snaps?

ppd:
With math.h, fabs is not in namespace std

whitequark:
Ah, hm. Could you fix all of the other headers while we're at it?

ppd:
Ah yes. Let's give it a little more ++

phkahler:
Well I just learned something. When are we gonna RiiR? hehehe.

whitequark:
> When are we gonna RiiR? hehehe.

This was actually discussed a long time ago and it was clearly unviable to me. Aleksey wouldn't have contributed then, for example. It is still my judgement that RIIR would be a phenomenally bad idea for this project.

ppd:
Modern C++ is good stuff while still giving you plenty of rope to hang yourself or at least find creative new and old ways to shoot yourself in the foot.

whitequark:
I'm not a fan of "modern C++" either. It's touted as a solution to all of the problems of C++ but in reality it mostly just brings layers upon layers of additional complexity with marginal benefit. We can just continue picking and choosing the C++ features we like, like we did for all this time, rather than following fads.

phkahler:
I was mostly kidding about RiiR. As with most software that would be too big a project to ever get done and it would just die. Is there any language related reason or just the obvious practical and people?

whitequark:
> As with most software that would be too big a project to ever get done and it would just die.

SolveSpace isn't *that* big; bigger projects have been successfully RIIR'd or written from scratch to address existing issues.

> Is there any language related reason

There is: Rust isn't very good for UI programming. In fact, for all I can tell (I've been following it closely and I've used it quite a bit) there still aren't any solid, well-established patterns for implementing GUI applications; if there's one thing OO is actually good for, it's GUIs*, and Rust lacking any has made it really hard to write GUIs the way everyone was writing them.

* Yes I know about React.

> or just the obvious practical and people?

Primarily it's the fact that SolveSpace is a very complicated for its size and solely implementation-defined application that at the moment didn't even have any tests, so rewriting it in any significant way, even just refactoring existing code, was and largely still is technically infeasible.

Secondarily it's the fact that a technically successful rewrite would achieve quite little in practical terms and cost a lot of time as well as opportunities to reuse C++ libraries, yeah.

Evil-Spirit:
> so rewriting it in any significant way, even just refactoring existing code, was and largely still is technically infeasible.

That's why I've started my own CAD project. Rewriting costs even more than writing from stratch because we should support legacy file formats, UI etc. But, unfortunately, I have a limited time, so now NoteCAD less than useable. Only 2d is fully supported.

ruevs:
> 
> 
> Well I just learned something. When are we gonna RiiR? hehehe.

:-D I also learned something - had to search what RiiR meant :-D
https://github.com/ansuz/RIIR
http://transitiontech.ca/random/RIIR
I've been "following" the language for maybe 5 years but still am at "Hello world." :-)

Evil-Spirit:
Let's do it now!

whitequark:
> But, unfortunately, I have a limited time, so now NoteCAD less than useable.

See, this is why I am against a rewrite. A rewrite will be useful *someday* if you have unlimited time. I don't have unlimited time, and neither do you. But with the very limited time that I have, SolveSpace is still improving, and is useful to many people *today*.

PR #610, opened 2020-05-11
Renamed isnan() to IsReasonable

### LINKED ISSUE #603: Unusual definition of isnan() and its implications
Commit ea6db67c7 added an unusual `isnan` macro:

```c++
#define isnan(x) (((x) != (x)) || (x > 1e11) || (x < -1e11))
```

Commit 8bc322eb4 adds a preprocessor guard that *looks* like it would cause the `isnan` function from `math.h` to be preferred, but doesn't actually do that on many platforms, e.g. glibc:

```c++
#ifndef isnan
#   define isnan(x) (((x) != (x)) || (x > 1e11) || (x < -1e11))
#endif
```

@jwesthues I'm not really sure what's the intent here--is it to check for NaN and infinity, or is it really a comparison with `VERY_POSITIVE` and `VERY_NEGATIVE`? Regardless, we need a better name for the function that doesn't clash with `math.h`, and to get rid of the misleading preprocessor junk.

phkahler:
Fixes #603.

### DIFF
--- a/src/solvespace.h
+++ b/src/solvespace.h
@@ -97,19 +97,19 @@ std::string ssprintf(const char *fmt, ...);
 static inline bool IsReasonable(double x) {
     return (((x) != (x)) || (x > 1e11) || (x < -1e11));
 }
-inline int WRAP(int v, int n) {
+static inline int WRAP(int v, int n) {
     // Clamp it to the range [0, n)
     while(v >= n) v -= n;
     while(v < 0) v += n;
     return v;
 }
-inline double WRAP_NOT_0(double v, double n) {
+static inline double WRAP_NOT_0(double v, double n) {
     // Clamp it to the range (0, n]
     while(v > n) v -= n;
     while(v <= 0) v += n;
     return v;
 }
-inline double WRAP_SYMMETRIC(double v, double n) {
+static inline double WRAP_SYMMETRIC(double v, double n) {
     // Clamp it to the range (-n/2, n/2]
     while(v >   n/2) v -= n;
     while(v <= -n/2) v += n;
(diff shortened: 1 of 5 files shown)


src/solvespace.h
```diff
@@ -76,8 +76,8 @@ typedef struct _cairo_surface cairo_surface_t;
     dbp("tri: (%.3f %.3f %.3f) (%.3f %.3f %.3f) (%.3f %.3f %.3f)", \
         CO((tri).a), CO((tri).b), CO((tri).c))
 
-#ifndef isnan
-#   define isnan(x) (((x) != (x)) || (x > 1e11) || (x < -1e11))
+#ifndef IsReasonable
+#   define IsReasonable(x) (((x) != (x)) || (x > 1e11) || (x < -1e11))
```
whitequark, line 80: The whole point of renaming would be to get rid of a name collision and all this preprocessor junk. This function should be `static inline` inside `namespace SolveSpace`, like the `WRAP` ones.

phkahler: Oh, ok.

phkahler: You want the WRAP one to be static too?

whitequark: Ah, I misremembered that it was already static. It makes zero difference either way.

phkahler: Done.  unless you want the 3 wrap functions inline too.

whitequark:
@phkahler I applied the rename in c1cd315d. I did it myself because in this PR you did not include either the issue number or the explanation of the change, so for anyone reading the history it would be a mystery why it was done.

Issue #489, opened 2019-09-20
Helix difference extrusion yeilds staircase effect


kalle:
### System information

SolveSpace version: 3.0-7f9117b2bf0

Operating system: Fedora

### Expected behavior

When selecting "force NURBS surfaces to triangle mesh" for a helix extrusion I though "chord tolerance" in the configuration menu would somehow influence the size of the triangles since it makes cylinders etc smoother. 

### Actual behavior

The slope of the helix in the attached model is very coarse, almost a staircase. In this model I have lowered the tolerance to 0.05% to no avail.

### Additional information

If there is a way to control this, it should perhaps be more prominently positioned.

whitequark:
There is no attached model.

kalle:
[model.zip](https://github.com/solvespace/solvespace/files/3637334/model.zip)
That was fast! :-) Here it is! Sorry

phkahler:
My suspicion on this is the piece-wise edge down the middle of the helix is a good fit for the actual edge - they're both straight. That edge needs to be broken into more sections due to the twist of the surface, but twist is not taken into account by the chord tolerance setting.

If you use the icon at the top of the text window to display the mesh this issue is present even with regular NURBS rendering. It's probably less visible because the triangles are not flat-shaded in that mode.

ruevs:
With current master the NURBS booleans actually work better than the triangle mesh ones for this particular model. This is a first :-)

Symbian9:
I still see "stairs edges" on Helix surface:

- chord tolerance: `0.01 %`
  - ![pic.1](https://i.imgur.com/pfKWoXN.png)
- chord tolerance: `0.10 %`
  - ![pic.2](https://i.imgur.com/VtCsjD5.png)
- chord tolerance: `0.001 %`
  - ![pic.3](https://i.imgur.com/JaMfEfO.png)

phkahler:
Those pictures are just rendering artifacts. Try showing the green triangle mesh via the icon on the text window toolbar. Everything is converted to triangles for display. The original problem looked like this:
![auger6](https://user-images.githubusercontent.com/14852918/81476380-b2b28580-91df-11ea-8b72-d8b973419c7c.png)

Ideally the triangle mesh would have more detail down the central axis and not produce those large triangles that form stair-steps. If you export this one as STL for 3D printing, or OBJ for use in something else, those stair steps will actually be there, and I do consider that an issue.

[...]

Issue #175, opened 2017-01-28
forcing NURBS to mesh causes "naked edges"


wpwrak:
SolveSpace at commit d4b052d34d94a62d9a9921f9c7916c4821bb325a

The following design produces complaints about 864 naked edges, which makes SolveSpace very unhappy:
https://gitlab.com/anelok/mexp/raw/016367ea70960daac4d8bb94a5740d84a3f5e900/alt13/aaa-minus-weirdly-broken.slvs

I found that turning off NURBS to mesh in g00b-spring makes the naked edges go away.

Before I noticed that I had NURBS to mesh selected, I tried removing the rectangular blocks on the side, the hole, the "dimple", ... none of which helped to get rid of the naked edges.

whitequark:
Triage: still an issue after baf9dc0aae573f017b8477e5d438665e90e2d5ae, although I get many fewer problematic edges (different chord tolerance?).

whitequark:
Triage: per discussion in https://github.com/Evil-Spirit/solvespace-master/commit/51616bc339d4a980dc2ae73c8a140be0fdefdcc0#commitcomment-21277921, switching from ear-triangulation to Delaunay will help here.

Evil-Spirit:
@whitequark 
What about using https://www.geometrictools.com/Samples/Geometrics.html?
Occasionally, my friend @bazhenovc made cmake building for this https://github.com/bazhenovc/WildMagic/commits/master

Evil-Spirit:
@bazhenovc
What do you know about this library? Is this the best choice?

bazhenovc:
Oh hello there.

I can say that this library does the job and the code is of high quality - so yes, I'd recommend it. I'd suggest to extract the needed classes rather then using the whole WM5/GTE - it's quite huge and you'll probably never need the whole package.

whitequark:
@bazhenovc Thanks for chiming in!

@Evil-Spirit This uses the Boost license, so it's compatible. It will remove a massive chunk of code from SolveSpace, if I understand correctly (exactly which?). Seems easy to integrate, though I think using a submodule is preferable to copying files (history is lost, upgrades are painful, etc).

bazhenovc:
History will be lost no matter what you'll do - their source code is distributed in zip archives.

whitequark:
In these cases I normally import releases into git version by version (which is quite easy with a script), so we can still easily bisect, and also easily track/rebase our changes on top of their classes.

[...]

PR #613, opened 2020-05-12
 Improve triangle mesh (splitting of quads based on angle).

ruevs:
When checking the dot product of the tangents `tu` and `tv` to decide in which direction to split a quad compare it to to `LENGTH_EPS` instead of zero to avoid alternating triangle "orientations" when the tangents are orthogonal (revolve, lathe etc.). This improves the quality of the resulting triangle mesh.
Before:
QadSplitOriginal
After:
QadSplitImprove

Model:
[LatheSpline.zip](https://github.com/solvespace/solvespace/files/4618604/LatheSpline.zip)

### DIFF
--- a/src/srf/triangulate.cpp
+++ b/src/srf/triangulate.cpp
@@ -482,7 +482,11 @@ void SPolygon::UvGridTriangulateInto(SMesh *mesh, SSurface *srf) {
                 if (this_flag) {
                     // Add the quad to our mesh
                     srf->TangentsAt(us,vs, &tu,&tv);
-                    if (tu.Dot(tv) < 0.0) {  //split the other way if angle>90
+                    if(tu.Dot(tv) < LENGTH_EPS) {
+                        /* Split "the other way" if angle>90
+                           compare to LENGTH_EPS instead of zero to avoid alternating triangle
+                           "orientations" when the tangents are orthogonal (revolve, lathe etc.)
+                           this results in a higher quality mesh. */
                         STriangle tr = {};
                         tr.a = a;
                         tr.b = b;


ruevs:
It seems to me that `ANGLE_COS_EPS`  the more logical variable to use but it is used in only one place (exactly for checking 90 degrees) and `LENGTH_EPS` is used in 152 places, and the values are the same. What do you think?

phkahler:
The "random" split directions are technically OK because they happen on flat quads. This is much nicer looking though.

phkahler:
On the top surface there is a missing triangle in the lower left. You can tell the edge has a vertex there because the vertical surface triangles connect to it, but the top do not. That seems suspicious. Not related to the PR of course.

Evil-Spirit:
Ofc the problem is exist, buyer should be solved in a different way. One of the solutions is check distance from centre of two edge variants to actual surface and take variant with minimal distance

Evil-Spirit:
Ofc the problem is exist, buyer should be solved in a different way. One of the solutions is check distance from centre of two edge variants to actual surface and take variant with minimal distance

Evil-Spirit:
This approach will probably solve #489

Evil-Spirit:
And probably we should apply this to whole surface triangulation instead of quads.

phkahler:
@Evil-Spirit Thanks for your input. Maybe I merged this too soon. The goal here was to eliminate the randomness in triangles by using a tiny value other than zero. @ruevs and @whitequark to that end it didn't matter which small constant was used. Better to use the common one rather than allow people to thing the one that is hardly used is somehow more important than it is.

I like the idea of checking the distance of the 2 possible midpoints to the surface. It seems like more calculation and may not make a better mesh - your test results and PR are certainly welcome! It will not fix the stair-step effect because that is due to helix-axis line not getting subdivided enough. It might do something useful for the other issue of extra silhouette line, but those are also needed near one end but not across the entire "quad" which is twisted - they are too long, which AFAICT can only be fixed by more subdividing the surface along the radial axis - another problem where the edge being a straight line prevents more subdivision.

This is all very interesting to me BTW and I welcome the discussion!

Evil-Spirit:
@phkahler 
Look at my new comment at #489. I mean, we will solve the problem with thorus and so on. 
The helicoid surface which produced by line, not needs to get more subdivisions. The only problem is what we should turn edges not only inside quads but also quad border edges to get better (more precise) surface approximation. This is not an easy task, so the other way to solve - is choose the better parametrization for u and v, ofc if it is bad. I mean there is an issue about circles. By some reason, circle is split into 4 peices, but actually it can be exactly presented by only one rational bezier segment. The reason why it is split is because good parametrization for t is not performed (the problem is described e.g. here https://fjorgedigital.com/insights/blog/can-bezier-curves-be-quickly-parameterized-by-arc-length/)

Evil-Spirit:
I sayd about circe which can be exactly presented by NURBS. But helical surface, unfortunately, can't be only approximated by NURBS.

ruevs:
> 
> 
> On the top surface there is a missing triangle in the lower left. You can tell the edge has a vertex there because the vertical surface triangles connect to it, but the top do not. That seems suspicious. Not related to the PR of course.

I noticed it as well - it is suspicious - but I though I should look at it a bit closer and create a simpler test case before opening an issue.

ruevs:
It is not missing. Just small and... strange.
QadSplitImproveZoom

phkahler:
Oh. I'm curious why that happens, but wouldn't worry about it too much. The best solution is to use Delaunay triangulation. Not that it's an easy thing to do, but eventually SS really should have it ;-)

ruevs:
I have no idea why it happens, debugging it would be...interesting... :-)

Previous discussions of Delaunay triangulation
Here:
https://github.com/solvespace/solvespace/issues/175#issuecomment-285961826
Here:
https://github.com/Evil-Spirit/solvespace-master/commit/51616bc339d4a980dc2ae73c8a140be0fdefdcc0

phkahler:
Was looking at the triangulation code. That was probably just a fluke because it's an ear. I'm thinking the bridge to the hole is near there too. Really getting to understand how it's done, the code is actually very readable. Except maybe when this made me laugh:

`if(l[ear].ear == EarType::EAR) {`

PR #623, opened 2020-06-02
Fix SSurface::EdgeNormalsWithinSurface() to handle non-orthogonal u,v derivatives correctly.

phkahler:
This is similar to the corrections in #473 but affects calculation of edge normals in boolean.cpp. I don't know a specific issue that this fixes, so I'd like to hold off merging until we have an example issue that it corrects. Such problems are likely to occur with booleans involving Helix extrusions.


phkahler:
I've gone back and forth with a number of models and can't find any regressions or improvements with this. I'd like to merge it if that's OK because the math is more correct than without it.

Issue #295, opened 2017-09-28
Circle rendering is very conservative


Timmmm:
SolveSpace version: 2.3
Operating system: MacOS 10.13

When sketching a circle it seems like the rendering accuracy (i.e. how many line segments it is split into) depends on the physical size of the circle, rather than its on-screen size. The result is if you draw a small circle and zoom in it looks like this:

![screen shot 2017-09-28 at 15 50 09](https://user-images.githubusercontent.com/376842/30973339-bdbe8fc0-a464-11e7-80cf-3b3c6b0733ff.png)

I.e. not very circular. Nobody is running this on a 386 so I would suggest as a simple hacky fix just increase the minimum number of segments from 8 to something a bit more reasonable like 128.

Timmmm:
Hmm so I had a look in the source code and discovered the chord tolerance setting, which obviously helps. Reducing it from 0.5% to 0.01% basically fixes this for circles down to 1mm diameter, which is probably about the limit that I imagine most people would care about.

To be honest I would have expected the code to work by representing all the geometry as true curves, and then turning them into polygons for display. In which case it should be easy (well "easy") to set the chord accuracy dynamically based on how big the object is on screen. Or in other words, the chord accuracy should be measured in screen-space pixels, not world-space mm.

You know more than me though so maybe it isn't that simple. :-)

I just checked how Blender handles this. It seems like it uses a fixed number of divisions of each bezier irrespective of their size. Also not ideal.

Timmmm:
I guess [this](https://github.com/solvespace/solvespace/blob/master/src/srf/ratpoly.cpp#L260) is the relevant code?

```
void SBezier::MakePwlInto(List<Vector> *l, double chordTol) const {
    if(EXACT(chordTol == 0)) {
        // Use the default chord tolerance.
        chordTol = SS.ChordTolMm();
    }
    l->Add(&(ctrl[0]));
    if(deg == 1) {
        l->Add(&(ctrl[1]));
    } else {
        // Never do fewer than one intermediate point; people seem to get
        // unhappy when their circles turn into squares, but maybe less
        // unhappy with octagons.
        MakePwlInitialWorker(l, 0.0, 0.5, chordTol);
        MakePwlInitialWorker(l, 0.5, 1.0, chordTol);
    }
}
```

Maybe as a not-ideal fix it could just be changed so that the minimum number of segments is more than 2.

jwesthues:
Ideas:
[...]

PR #626, opened 2020-06-04
Fix issue #295 keep circles to 16 edges minimum.

phkahler:


whitequark:
Code looks reasonable, do you think you could post some before/after? I'll check it out locally and play with it a bit later.

phkahler:
Before: even at this level holes look bad:
Before
After: much smaller hole diameter zoomed in:
After
With this change, people shouldn't be changing chord tolerance just to prevent octagons. When the holes look bad, any other curves will too, and changing tolerance should be the right thing to do. Also good for those smallish rounded corners.

Issue #629, opened 2020-06-06
SolveSpace won't import 3D circles/arcs from DXF


johreh:
### System information

SolveSpace version: 3.0~225f82a2 

Operating system: Arch 

### Actual behavior
Trying to import a DXF of this exported from Fusion 360
![test1 v2](https://user-images.githubusercontent.com/1417219/83955330-0b625600-a852-11ea-9208-be27c598ccf9.png)
results in
![test1 v2_solvespace](https://user-images.githubusercontent.com/1417219/83955353-5d0ae080-a852-11ea-8aa0-6f0b3a6f41d3.png)
The circles and arcs are imported and show up in the request list of the group, but are not drawn and cause the solver to fail.

### Expected behavior
Something like
![test1 v2_solvespace_fix](https://user-images.githubusercontent.com/1417219/83955668-a446a080-a855-11ea-8257-0341ddaa0581.png)

### Additional information

DXF file manifesting problem.
[test1 v2.zip](https://github.com/solvespace/solvespace/files/4741051/test1.v2.zip)

johreh:
As could be guessed from the screenshot above, I have started on a patch for this 5c472fd568de548762ddc2c15a81c7ca2815fc75

whitequark:
Your modified import function creates entities in a way that is not reproducible via the UI. I'm not sure if this is the right approach, and I would be hesitant to merge this patch.

johreh:
Can you clarify what is not reproducible via the UI?

phkahler:
@johreh I think what whitequark is saying is:  The output .slvs file your code is going to produce can not be created with unmodified SolveSpace using the normal construction tools. In other words, you're creating new internal structures or functions just for this capability instead of reading the file and creating SS equivalent entities.

BTW, I appreciate the approach of opening the issue when you decided to work on it. This provides a place for the dialog that is about to happen to get it right. Not right - optimally useful.

johreh:
I still don't quite follow, as I don't see what could not be created using normal SolveSpace tools.

The patch introduces the following changes:
- Applies object coordinates to world coordinates transformation before creating SS objects using the extrusion vector as in the DXF specification
- Creates new workplanes for arcs not in the baseplane
- Applies a normal vector to circles (also to arcs, but I realize that is not actually necessary as they're locked to workplanes)

All of the operations are already achievable from the UI as far as I can tell.

Granted, there's also changes to createDiametric, but that was just to make the code compile for testing and needs to be figured out properly.

[...]

PR #640, opened 2020-06-23
Allow DXF import of 3D arcs and circles

johreh:
Extrusion direction (normal) of arcs and circles were not taken into
account when importing.

- Add method for calculating a quaternion from extrusion direction
  according to DXF arbitrary axis algorithm
- Add required workplanes for arcs not on XY origin plane
- Adjust addDimRadial and addDimDiametric to include normal when
  creating associated circle request

### DIFF
--- a/src/importdxf.cpp
+++ b/src/importdxf.cpp
@@ -9,6 +9,21 @@ static std::string ToUpper(std::string str) {
     return str;
 }
 
+static Quaternion NormalFromExtPoint(Vector extPoint) {
+    // DXF arbitrary axis algorithm for transforming a Z-vector into a rotated
+    // coordinate system
+    Vector ax, ay;
+    Vector az = extPoint.WithMagnitude(1.0);
+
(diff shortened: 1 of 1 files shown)


johreh:
Fix for #629 
Not that it might be prudent to merge pull request https://github.com/solvespace/libdxfrw/pull/13 before this one since it otherwise won't import 3D circle-constraints correctly. It also makes it possible to remove the temporary copies in those methods.

whitequark:
Code looks essentially reasonable, I'll take a closer look again later. Can you attach a few files you're testing this on?

phkahler:
@johreh  As I recall this worked, but I am unable to compile locally after applying the patch now. Seems to be related to the libdxfrw  change, but that looks like it got merged. It passed the CI test though so I'm not sure what to make of the situation.

johreh:
@phkahler I just tried rebuilding this rebased on top of latest master. I did not get any build errors.

@whitequark Oops, forgot about this. I'll try to find some of the files I used.

Issue #550, opened 2020-02-09
Helix tool almost always results in invalid mesh


andres-erbsen:
### System information

dcc80de on Arch Linux

### Expected behavior

A manifold mesh representing the helical extrusion, for example

![c](https://user-images.githubusercontent.com/918403/74096987-4d61d200-4ad4-11ea-9d40-6f1e745e67ae.png)

### Actual behavior

If I increase the angle just a bit I get a self-intersecting mesh:

![d](https://user-images.githubusercontent.com/918403/74097001-944fc780-4ad4-11ea-9ca5-4739a5b0e2ca.png)

and increasing the angle just slightly more gives naked edges instead:

![b](https://user-images.githubusercontent.com/918403/74096990-5ce11b00-4ad4-11ea-809f-cbd822363ab7.png)

### Additional information

Note that the issues are present even with "force NURBS surfaces to triangle mesh".

This example (unit square at unit distance from axis): [helix.zip](https://github.com/solvespace/solvespace/files/4176517/helix.zip)
Non-trivial example (8x rotated 2 beziers approximating involute that is almost a triangle): [gear90.zip](https://github.com/solvespace/solvespace/files/4176518/gear90.zip)
This also occurs when applying helix to a circle.

#472 seems relevant.

phkahler:
The problems with helix.zip seem to be solved by #473 which is very helpful for extrusions with a lot of twist. No naked edges.

Basic rendering with gear.zip is also better with those changes. There are fewer errors when using Analyze -> show naked edges but there are still a lot of them with that model. I'm not sure what causes them in that case - lowering the chord tolerance causes more of them.

mkyral:
Hi,
I've played with Helix (really useful tool) in and I've generated this problematic shape: [rounded_edges_issue.zip](https://github.com/solvespace/solvespace/files/4664827/rounded_edges_issue.zip)

![obrazek](https://user-images.githubusercontent.com/155542/82607196-b0293600-9bb8-11ea-8d4f-e171cf2d2aa8.png)
![obrazek](https://user-images.githubusercontent.com/155542/82607447-1910ae00-9bb9-11ea-92e4-2130868bd8ac.png)
![obrazek](https://user-images.githubusercontent.com/155542/82607737-a48a3f00-9bb9-11ea-9893-42e569bdac3d.png)

phkahler:
[...]

PR #643, opened 2020-07-07
Better helical triangulation #489 and #550 fixed.

phkahler:
…4 segments and forcing grid triangulation to use a minimum of 4x4 quads. This helps with issue #489. Also add a max_dt parameter for PWL creation and use that for helical edges.

### DIFF
--- a/src/srf/ratpoly.cpp
+++ b/src/srf/ratpoly.cpp
@@ -173,18 +173,18 @@ void SBezier::SplitAt(double t, SBezier *bef, SBezier *aft) const {
     }
 }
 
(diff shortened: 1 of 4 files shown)


phkahler:
I want to hold off on this a bit more. Sorry for going too fast.

whitequark:
Take all the time you need! There's no harm in having half-done pull requests (it may even help a bit by showing who works on what, and perhaps inspiring collaboration).

phkahler:
@whitequark this seems to fix #550 as well. If you're OK with it I am.

PR #644, opened 2020-07-08
WIP:  Triangle Fans

phkahler:
This is still experimental. I noticed that ear triangulation calls RemoveTagged() for every ear and that function scans (copies?) the entire list of verticies. It also has to check IsEar() for the adjacent verticies afterward. I figured we can make a pass looking for chains of ears that form a sort of meta-ear and triangulate that independently, and only make one call to RemoveTagged() to dispose all those ears. This runs prior to the existing triangulation code (and only for flat surfaces) so it might be faster overall for very complex surfaces. It does tend to make better-formed triangles.

Still some bugs and odd cases to figure out.
The logic is not clear yet.
Haven't proven any practical benefit like speed improvement.

One odd thing I did notice. If you look at the mesh for a circle it can look less round when showing the triangles. It's not, but your eyes can trick you due to the structured lines.

### DIFF
--- a/src/polygon.h
+++ b/src/polygon.h
@@ -131,6 +131,7 @@ public:
     void FindPointWithMinX();
     Vector AnyEdgeMidpoint() const;
 
+    bool IsEmptyTriangle(int ap, int bp, int cp, double scaledEPS) const;
     bool IsEar(int bp, double scaledEps) const;
     bool BridgeToContour(SContour *sc, SEdgeList *el, List<Vector> *vl);
     void ClipEarInto(SMesh *m, int bp, double scaledEps);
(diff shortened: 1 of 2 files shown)


phkahler:
Tri-fans

phkahler:
It's working a lot better. On one model this actually reduces the number of naked edges. Maybe chance, or maybe because the triangles are of better shape?
Also removed a redundant pass over all the verticies. That was an n^2 but not the only one so not an O(n) reduction. I left that as a separate commit but it would conflict with the first one so added to this PR.
triangle_fans

phkahler:
@ruevs  can you try this out? It's very subjective weather this one should go in or not.

ruevs:
I'll try but probably next week.

PR #646, opened 2020-07-09
Use mimalloc for {Alloc,FreeAll}Temporary()

### LINKED ISSUE #642: RFC: Use mimalloc to avoid freelist woes on *nix
Right now we implement `AllocTemporary` and `FreeAllTemporary` on *nix using a freelist. This does work but it is strictly worse than what happens on Windows, which is unfortunate because I'd like to not privilege any platforms. There are also some multithreading related issues that are currently solved in not especially elegant way.

It looks like [mimalloc](https://github.com/microsoft/mimalloc) is pretty much perfect:
  * works on all of our platforms;
  * uses CMake so easy to integrate;
  * supports heaps so we can efficiently deallocate temporary objects.

One thing to note is their performance claims. It's not a bad allocator but the revolutionary claims should be taken with a grain of salt.

nabijaczleweli:
Closes #642

### DIFF
--- a/src/platform/platform.cpp
+++ b/src/platform/platform.cpp
@@ -10,6 +10,7 @@
 #   include <CoreFoundation/CFBundle.h>
 #endif
 #include "solvespace.h"
+#include "mimalloc.h"
 #include "config.h"
 #if defined(WIN32)
 // Conversely, include Microsoft headers after solvespace.h to avoid clashes.
@@ -18,7 +19,6 @@
 #else
 #   include <unistd.h>
 #   include <sys/stat.h>
-#   include <mutex>
 #endif
 
 namespace SolveSpace {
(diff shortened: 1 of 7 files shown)


src/platform/platform.cpp
```diff
@@ -680,64 +680,32 @@ void DebugPrint(const char *fmt, ...) {
...
-void *AllocTemporary(size_t size)
-{
-    if(!TempArena)
-        TempArena = HeapCreate(0, 0, 0);
-    void *ptr = HeapAlloc(TempArena, HEAP_ZERO_MEMORY, size);
-    ssassert(ptr != NULL, "out of memory");
-    return ptr;
-}
+struct MimallocHeap {
+    mi_heap_t *heap = NULL;
```
whitequark, line 687: I think it'd be a little cleaner if the heap was allocated here and not in `AllocTemporary`?

src/platform/platform.cpp
```diff
@@ -680,64 +680,30 @@ void DebugPrint(const char *fmt, ...) {
...
-//-----------------------------------------------------------------------------
-// Temporary arena, on Linux.
-//-----------------------------------------------------------------------------
-
-#if !defined(WIN32)
-
-struct ArenaChunk {
-    ArenaChunk *next;
+    ~MimallocHeap() {
+        if(heap)
```
whitequark, line 690: This is always true now

whitequark:
Can you squash the first two commits too? There's not much point in adding mimalloc if it's not used. Then this PR is good to go.

src/platform/platform.cpp
nabijaczleweli: It might not be, considering that `mi_heap_new()` heads with 
```c
mi_heap_t* mi_heap_new(void) {
  mi_heap_t* bheap = mi_heap_get_backing();
  mi_heap_t* heap = mi_heap_malloc_tp(bheap, mi_heap_t);  // todo: OS allocate in secure mode?
  if (heap==NULL) return NULL;
```

Seeing as `mi_heap_malloc()` (and `mi_heap_malloc_small()` similarly) starts with
```c
extern inline mi_decl_restrict void* mi_heap_malloc(mi_heap_t* heap, size_t size) mi_attr_noexcept {
  if (mi_likely(size <= MI_SMALL_SIZE_MAX)) {
    return mi_heap_malloc_small(heap, size);
  }
  else {
    mi_assert(heap!=NULL);
```
and potentially aborts under us, I opted for a `ssassert()` in the constructor instead.

whitequark: Ah yeah good catch!

PR #650, opened 2020-07-12
fix gtk file chooser set name methods

xzcvczx:
use set_filename for existing files based on documentation
change set_current_name to use a Glib::ustring rather than a std::string


whitequark:
See also #545 and #562--I think this would break some of the desirable behavior. I didn't realize this would lead to crashes though, so I'll take a look at this sooner.

phkahler:
Closing because we're going to base changes on one of the other similar PRs.

Issue #524, opened 2019-12-03
Flat colors render mode


Symbian9:
### System information

- SolveSpace version: `3.0~690f87c`
- Operating system: Debian 9.x stretch (x86_64)

### Expected behavior

Add an option *"Show flat color view of solid model"* to enable "flat colors" render:  (direct RGB values without any gradient shades; independent of light direction)

### Actual behavior

Actual render has gradient shades, according to default "light direction" setting.

![pic.1](https://i.imgur.com/S6aqbD4.png)

![pic.2](https://i.imgur.com/pw2fKQT.png)

And it's non-trivial task to get near "flat colors" render in SolveSpace.

![pic.3](https://i.imgur.com/y6kd2vL.png)

As workaround I use next settings:

```
light direction                 intensity     
  #0    ( 0.00, 0.00, 10000.00)[c]  1.00 [c]
  #1    ( 0.00, 0.00, 0.00)[c]  0.50 [c]
```

(`10000` used for `#0` light source Z, because with `10`, `100` and `1000` value gradient shades still appear visually)

![pic.4](https://i.imgur.com/LYtPQ10.png)

### Additional information

As side note, seems like there no way to reset "light direction" setting to default values:

```
light direction                 intensity     
  #0    (-1.00, 1.00, 0.00)[c]  1.00 [c]
  #1    ( 1.00, 0.00, 0.00)[c]  0.50 [c]
```

So, it would be good to add button for reset "light direction" too.

![pic.5](https://i.imgur.com/qLABq2w.png)

Symbian9:
Think, such mode would be especially useful for export 2D view to PDF and SVG.

Symbian9:
I just wondered when tried to place reference image behind solid model while *"Hide shaded view of solid model"* is toggled — it looks like "flat render" which just use background color for triangle mesh of this solid model.

If so, maybe this is the start point to *"Show flat color view of solid model"*?

![pic.1](https://i.imgur.com/BMdr6ah.png)

phkahler:
One option is to set the intensity of both light sources to 0.0.   There is still an ambient light which will show the model but will not be as bright as you may want (white shows as a uniform grey). Proper flat shading might just be turning off the existing lights and changing the ambient brightness which is no a UI option.

phkahler:
[...]

PR #653, opened 2020-07-17
Implement #524 Allow flat colors. Adds an edit control for Ambient li…

phkahler:
@whitequark  can you have a look at the text window code here? The change[c] link isn't working.

When that's fixed we should be able to turn off the 2 directional lights (set intensity to 0) and turn up ambient to 1 to get flat colors.

### DIFF
--- a/src/confscreen.cpp
+++ b/src/confscreen.cpp
@@ -18,6 +18,12 @@ void TextWindow::ScreenChangeLightIntensity(int link, uint32_t v) {
     SS.TW.edit.i = v;
 }
 
+void TextWindow::ScreenChangeLightAmbient(int link, uint32_t v) {
+    SS.TW.ShowEditControl(31, ssprintf("%.2f", SS.ambientIntensity));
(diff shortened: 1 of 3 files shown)


phkahler:
@whitequark never mind, got it.  It seems to work. Setting both lights brightness to zero and the ambient to 1.0 yields flat coloring.

src/confscreen.cpp
```diff
@@ -218,6 +224,10 @@ void TextWindow::ShowConfiguration() {
             CO(SS.lightDir[i]), i, &ScreenChangeLightDirection,
             SS.lightIntensity[i], i, &ScreenChangeLightIntensity);
     }
+    Printf(false, "%Bp         Ambient Lighting     "
```
whitequark, line 227: Actually, this should be lowercase to follow the style of that dialog.
```suggestion
    Printf(false, "%Bp         ambient lighting     "
```

phkahler: Fixed. Good call it's all lower case.

whitequark:
Thanks!

PR #655, opened 2020-07-19
NFC: Remove a redundant pass over all verticies in ear triangulation.

phkahler:
A simple optimization in triangulation code. Tested on several models.


whitequark:
@phkahler Sorry, I think I lack the domain knowledge to review this.

PR #658, opened 2020-07-22
Fix issue174 if selection for dragging is not in a previous group, us…

phkahler:
This needs some testing and it seems like the logic could maybe be simpler.

I figured the reason ChooseFromHoverToDrag() exists is to allow dragging entites from previous groups which doesn't happen with ChooseFromHoverToSelect().  If that's the case, we can just check if the drag selection found something from a previous group. If so use it, otherwise use the selection logic for dragging.

### DIFF
--- a/src/draw.cpp
+++ b/src/draw.cpp
@@ -353,13 +353,29 @@ GraphicsWindow::Selection GraphicsWindow::ChooseFromHoverToSelect() {
     return sel;
 }
 
+// This is different from ChooseFromHoverToSelect() by allowing us to drag
+// an entity from a previous group. However it doesn't behave quite the
+// same in the current group, so if it fails at selecting a underlying
+// entity we shall use ChooseFromHoverToSelect instead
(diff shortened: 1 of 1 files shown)


phkahler:
I can squash those if you want.  Having drags fail because it tried to select something from group zero has been very annoying for me, hence the second commit.

whitequark:
Thanks for the fix! I'm going to need to look closer into it.

phkahler:
@whitequark can we merge this? It really does fix this problem, and it doesn't touch code for selection or hovering at all. All it does is prevent the new rules for dragging earlier groups from applying the same (first in the list) alternate logic to the current group. I'm still thinking about ways to solve the other dragging related issue.

whitequark:
Sorry, I still haven't found the time to look at this. I'm currently on vacation until the end of the next week, so I'll take a look around then. (Yes, I know I answered to some issues... old habits die slowly)

phkahler:
Closing as superseded by #672.

PR #661, opened 2020-07-26
Fix issue659 - Problems constraining to ends of Helix.

phkahler:
Create a new copy type for faces that includes the translation aspect of helical extrusions. Also swap the end remappings when the shell is inside out - this was also affecting some Revolve extrusions.

### DIFF
--- a/src/drawentity.cpp
+++ b/src/drawentity.cpp
@@ -121,6 +121,7 @@ void Entity::GetReferencePoints(std::vector<Vector> *refs) {
         case Type::FACE_N_TRANS:
         case Type::FACE_N_ROT_AA:
         case Type::FACE_ROT_NORMAL_PT:
+        case Type::FACE_N_ROT_AXIS_TRANS:
             break;
     }
 }
@@ -757,6 +758,7 @@ void Entity::Draw(DrawAs how, Canvas *canvas) {
         case Type::FACE_N_TRANS:
         case Type::FACE_N_ROT_AA:
         case Type::FACE_ROT_NORMAL_PT:
+        case Type::FACE_N_ROT_AXIS_TRANS:
             // Do nothing; these are drawn with the triangle mesh
             return;
     }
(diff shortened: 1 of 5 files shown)


whitequark:
Hm, looks like mimalloc causes all of our tests to fail.

phkahler:
> Hm, looks like mimalloc causes all of our tests to fail.

Did you respond to the wrong PR?

whitequark:
Nope, I was looking into why tests fail.

whitequark:
I filed https://github.com/microsoft/mimalloc/issues/279.

whitequark:
Seems like they pushed a fix upstream, do you think you can check if their commit fixes the issue?

phkahler:
> Seems like they pushed a fix upstream, do you think you can check if their commit fixes the issue?

How do I even test that? No sure how to update the link to a newer version of mimalloc (or other external libs) so that Travis will use it.

nabijaczleweli:
`git pull https://github.com/microsoft/mimalloc dev` in `extlibs/mimalloc` will update the submodule, and `git log` there should show the commit with the fix. Then `git diff` of `extlibs/mimalloc` should show the new SHA.

phkahler:
@nabijaczleweli Thanks, but I'm still missing something. First, I can't reproduce the issue here, the  tests pass. It would be cool to issue a PR using a newer mimalloc and let travis test it but I'm not sure how to update the external link itself. Locally I have folders full of code, not links with hashes.

nabijaczleweli:
And those folders full of code are git repositories of their own, with the HEAD showing up if you diff them from the parent project:
image

I pushed the commit to a branch here for demonstration purposes: https://github.com/nabijaczleweli/solvespace/tree/submodules-how-do-they-work

phkahler:
Thanks!  I tried that here and then made a PR to see if it fixes the problem.

PR #662, opened 2020-07-26
Update mimalloc

phkahler:
Let see if this fixes the CI failures.


phkahler:
Fixes the Linux issue. MacOS is still failing on..... sqlite ?

whitequark:
@phkahler Ignore that, transient errors. We should merge this PR once mimalloc pulls the fix into master.

Issue #165, opened 2017-01-22
make arc-center relation "explorable"


wpwrak:
Now that we can follow the relation between items and constraints, it would be nice if one could do the same for arcs and their center point. (The same would apply for circles, and possibly other items with "special" points.)

Right now, the center point is shown in the text window just as "POINT", without indication that it is the center of an arc. It would be nice if one could 1) tell that it is special, e.g., by calling it "CENTER POINT" or such, or maybe displaying a center-of-arc constraint (if that makes sense), and 2) reach the point from the arc and vice versa, much like we can now reach items through the constraints that connect them.

wpwrak:
I forget an even more common case: the endpoints of lines.

Evil-Spirit:
@whitequark, implemented here

wpwrak:
Thanks ! Here's a fix for a little typo:
https://github.com/wpwrak/solvespace/commit/782a5bbbe61c0a1e5326d55218de33e6a7682f26

I also found this extra semicolon, which already existed before #165. I'm not sure if this is a safe change, though, or if something may depend on that typo:
https://github.com/wpwrak/solvespace/commit/a77cedbfd8dedd147fb59ebf9af936f4259ff31b

Evil-Spirit:
@wpwrak, thank you :) We've found this semicolon a long time ago, but I forgot to fix it.

wpwrak:
By the way, along the lines of explorability, how about something like this ? https://github.com/wpwrak/solvespace/commit/41e55b3d79f7d2df61fb9db0b494ff09a37d6add
(My ultimate goal in this quest of mine is to be able to perform the deletion in #182 without losing any groups, and this should bring me a little closer to being able to rearrange the dependencies.)

whitequark:
FWIW I haven't merged the patch because the UI work in it is horrible. It's completely inconsistent with the rest of text window stuff.

Evil-Spirit:
@whitequark, what do you mean?

ruevs:
I agree with @whitequark that the proposed implementation was too intrusive and inconsistent with the style of the UI code for the test window. Here https://github.com/solvespace/solvespace/pull/1182 is a clean implementation with minimal intrusiveness.

Whether the style of the UI code is nice or not is another question. I'm split on this...

Symbian9:
> I agree with `@whitequark`

@ruevs, just a reminder, `@whitequark` asked to not cast (mention/trigger) she in SolveSpace related threads anymore.

[...]

PR #664, opened 2020-07-30
Improve text screens for selected constraint and entity

whitequark:
Improve text screen for a selected constraint:
  * Show the type of a constraint.
  * Show the workplane a constraint is in, for constraints which can be both projected to workplane or be free in 3d space.
  * Clearly distinguish reference from non-reference constraints.
  * Add a checkbox for toggling the reference option.
  * When showing requests a constraint applies to, highlight on hover the specific entity being constrained, not just the first one.

Improve text screen for a selected entity:
  * Add a link to its request, unless it's the first entity in that request (which would just select the same entity again).
  * Add a link to its group.
  * Add a link to its workplane.
  * Add a link to its style; and hide the style row for entities that aren't stylable.
  * Show constraints and measurements (reference constraints) in separate lists.
  * For curve entities, show constraints that apply to the points related to the curve, not just to the curve itself.

Screenshot_20200730_094122

### DIFF
--- a/src/constrainteq.cpp
+++ b/src/constrainteq.cpp
@@ -28,6 +28,50 @@ bool ConstraintBase::HasLabel() const {
     }
 }
 
+bool ConstraintBase::IsProjectible() const {
+    switch(type) {
+        case Type::POINTS_COINCIDENT:
+        case Type::PT_PT_DISTANCE:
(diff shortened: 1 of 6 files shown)


whitequark:
Related to #165 (but does not implement everything in that patch).

PR #665, opened 2020-08-01
Fix #656 by making SSurface::SplitInHalf thread safe.

### LINKED ISSUE #656: OpenMP issue with complex NURBS
### System information

SolveSpace version: 3.0~188b2e2

Operating system: Fedora 31

CPU AMD 2400G - 4 core 8 thread.

### Expected behavior

Software should work when compiled with OpenMP

### Actual behavior

The attached model seems to work fine until the depth of the hole is dragged. When dragging, the top surface (which is 2 patches) can get distorted. Not just naked edges, but the geometry of the surface changes as seen in the picture:

parallel_bug

[Revolve120.zip](https://github.com/solvespace/solvespace/files/4954955/Revolve120.zip)

### Additional information

This model was created to demonstrate NURBS boolean issues. Failures normally occur when the hole intersects the seam between the two top surface patches - that's an expected failure. The issue here is that the 2 surface patches change shape when the hole *depth* is changed. As soon as the hole *position* is moved even a little the surfaces revert to normal. It is also strange that both surfaces can change shape even if only one of them contains the hole. This distortion only happens when using OpenMP.

Reverting to commit 9802b5d which is prior to the mimalloc changes did not help - The problem exists there as well (possibly worse but that's terribly subjective).

Compiling single-threaded "fixes" the issue.

phkahler:
The surface was being modified in SSurfaceSplitInHalf. Make a local copy instead.

### DIFF
--- a/src/srf/raycast.cpp
+++ b/src/srf/raycast.cpp
@@ -130,39 +130,38 @@ void SSurface::SplitInHalf(bool byU, SSurface *sa, SSurface *sb) {
     sa->degn = sb->degn = degn;
 
     // by de Casteljau's algorithm in a projective space; so we must work
-    // on points (w*x, w*y, w*z, w)
-    WeightControlPoints();
+    // on points (w*x, w*y, w*z, w) so create a temporary copy
+    SSurface st;
+    st = *this;
+    st.WeightControlPoints();
 
     switch(byU ? degm : degn) {
         case 1:
-            sa->CopyRowOrCol (byU, 0, this, 0);
-            sb->CopyRowOrCol (byU, 1, this, 1);
+            sa->CopyRowOrCol (byU, 0, &st, 0);
+            sb->CopyRowOrCol (byU, 1, &st, 1);
 
-            sa->BlendRowOrCol(byU, 1, this, 0, this, 1);
-            sb->BlendRowOrCol(byU, 0, this, 0, this, 1);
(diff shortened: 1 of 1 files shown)


ruevs:
Still on a phone - otherwise I would test it myself.

So which is faster? Putting a critical section around this or making an on the stack copy of each surface multiple times per boolean operation?

Obsuously this may depend on many things, CPU, compiler, RAM but if the difference is drastic on some "typycal" system it may help to choose.

ruevs:
A different question is which is cleaner...

phkahler:
@ruevs  You had some model that showed significant performance change with OpenMP. You might try both methods with that. Just put a "omp crtical" with { } around most of this function from before WeightControlPoints() to after the last UnWeightControlPoints().  compare that to this PR.  I'm in favor of this approach because it fixes the underlying problem, and also modifying a surface and putting it back just seems like a bad idea, and this doesn't pollute the code with an extra #pragma

If you want to do some tests you might also try using OMP in SShell::CopyCurvesSplitAgainst()  I think the last 2 lines need to be inside a critical (or not the last one). I tried this once before and the problem with surfaces distorting was even worse but that will be fixed with this PR.

In the mean time I'm probably going to merge this one.

ruevs:
Yes, that model is attached here https://github.com/solvespace/solvespace/pull/613
With very small chord tolerance and large number of segments the difference was very noticeable. I will experiment when I get some time. Also have not forgotten about your triangulation changes.

PR #673, opened 2020-08-12
NURBS boolean improvement.

phkahler:
I'd like to get input on this from @jwesthues  @ruevs and @Evil-Spirit 

The code in Boolean tries to intersect a curve with a surface. It uses the PWL representation of the curve (the more general case) and the 2 surfaces it trims. To do the intersection it calls SSurface::PointOnSurfaces() which finds the U,V intersection of the 3 surfaces. I noticed that function will not converge (it gives up) if 2 of the surfaces are tangent along the curve, which is the case for Lathe and Helical surfaces joined every 90 or less degrees. In that case we always have an exact representation of the curve, so I created a new function to intersect an exact curve with a surface which can be used in the (common) special case that we have and exact curve. I thought this may result in better accuracy finding the intersection points and might fix some of the boolean issues.

Unfortunately I have not seen any broken boolean start working, but I do see this broken one being more consistent:
Bug1_before
Bug1_after

The first picture is before this change. The red lines tend to flicker between the upper missing 90 degree surface patch and the lower one, but very rarely will all the curves show in red. It flickers back and forth as you drag the cylinder. With the change it very consistently shows all the curves. In neither case does it ever show the surfaces. The thing is I'm not sure this is consistently better than before or consistently worse. From a theoretical point of view this change should be better.

What do you think?

BTW I have not seen any regressions with 10-20 sketches.
BTW2 the above sketch is a lathed object (spool) with a circle/cylinder cut out to demo boolean failure.

### DIFF
--- a/src/srf/boolean.cpp
+++ b/src/srf/boolean.cpp
@@ -88,8 +88,15 @@ SCurve SCurve::MakeCopySplitAgainst(SShell *agnstA, SShell *agnstB,
                     }
                 }
 
-                // We're keeping the intersection, so actually refine it.
-                (pi->srf)->PointOnSurfaces(srfA, srfB, &(puv.x), &(puv.y));
+                // We're keeping the intersection, so actually refine it. Finding the intersection
+                // to within EPS is important to match the ends of different chopped trim curves.
+                // The general 3-surface intersection fails to refine for trims where surfaces
+                // are tangent at the curve, but those trims are usually exact, so…
+                if(isExact) {
+                    (pi->srf)->PointOnCurve(&exact, &(puv.x), &(puv.y));
+                } else {
+                    (pi->srf)->PointOnSurfaces(srfA, srfB, &(puv.x), &(puv.y));
+                }
                 pi->p = (pi->srf)->PointAt(puv);
             }
             il.RemoveTagged();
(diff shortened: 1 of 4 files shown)


jwesthues:
I agree that your change to use the exact curve where available should be a strict improvement. For debugging, you could try printing/plotting the results of both your new curve-surface intersection and the old surface-surface-surface intersection, to see where they differ.

phkahler:
After some more tweaks and using the exact version over in surfinter.cpp this fixes the above test model. It also fixes another one I have and a simplified version of issue #315   My head needs a rest now. I'll clean this up a bit tomorrow or this weekend and see if it fixes any other github issues.

phkahler:
@whitequark why can't I link this to issue 315?

phkahler:
BTW, not sure if it's measureable but this will result in a reduction in runtime vs the more general function. Performance and correctness!

jwesthues:
Indeed. I suspect that commercial kernels get a lot of their robustness from special cases, a lot of them much specialer than your change (specific code for circle-cylinder intersection, circle-cone, cylinder-cylinder, etc.). I see no downside to such special cases beyond the considerable work to develop and test them.

phkahler:
@whitequark There are a couple incidental changes in this too. I can split them out if you want, otherwise I it should be good to go.

whitequark:
> I can split them out if you want, otherwise I it should be good to go.

I'd prefer if you split them, if this change causes any issues, this will make bisecting work better.

phkahler:
Done.

Issue #303, opened 2017-11-18
Tangential circles in a rectangle can break contour filling


Natronium:
### System information

SolveSpace version: 3.0~13695be0
Operating system: Gentoo Linux

### Expected behavior
The contour gets filled in

### Actual behavior
This happens:
![](https://user-images.githubusercontent.com/25114018/32985401-d0a80062-cc87-11e7-8fef-e8c6c91101e3.png)

Additionally, extruding the shape creates naked edges at the points where that same circle touches its neighbours:
![](https://user-images.githubusercontent.com/25114018/32985436-fdd419bc-cc88-11e7-8c4f-b360aa044ed9.png)

All of this is accompanied by the same error that whitequark pointed out in #296:
```
couldn't find an ear! fail
```
however, unlike #296, enabling 'force NURB surfaces to triangle mesh' doesn't fix the issue in either case.
### Additional information

The file in question: [tangent-circles-break-contour-fill.zip](https://github.com/solvespace/solvespace/files/1485089/tangent-circles-break-contour-fill.zip)

whitequark:
Yeah, needs a better triangulation algorithm.

Evil-Spirit:
Tangential should be handled as error like countour self-intersection. In case of tangential the result mesh will be non-manifold (this is bad for some reason)

whitequark:
> Tangential should be handled as error like countour self-intersection.

I agree. Do you think you can write the fix? I'm not good with numeric code...

Evil-Spirit:
This not look hard to me.

phkahler:
Initially could not reproduce using the attached .slvs file of similar construction.
[Tangent_Circle_Test.zip](https://github.com/solvespace/solvespace/files/4828794/Tangent_Circle_Test.zip)

But... It happens if the construction lines attached to the upper right circle-center are H and V constrained. In that configuration 4 Beziers end at a single point where the circles meet. Pinging @ruevs because he's got some interest in NURBS issues and this is a case where something 2D breaks down prior to any surface creation.

phkahler:
PR #675 only addresses the first part of this, so I'm not going to close this yet. The naked edges remain.

whitequark:
@phkahler Did you close this by accident?

phkahler:
@Whitequark Yes by accident, I thought I deleted the reference from the PR before I merged it but this still closed. It's still a good case of naked edge problems though.

phkahler:
[...]

PR #675, opened 2020-08-14
Don't duplicate points when connecting contours with zero length brid…

phkahler:
This seems to fix issue #303  you may want to test it a bit more than I did.

### DIFF
--- a/src/srf/triangulate.cpp
+++ b/src/srf/triangulate.cpp
@@ -109,6 +109,7 @@ bool SContour::BridgeToContour(SContour *sc,
                                SEdgeList *avoidEdges, List<Vector> *avoidPts)
 {
     int i, j;
+    bool withbridge = true;
 
     // Start looking for a bridge on our new hole near its leftmost (min x)
     // point.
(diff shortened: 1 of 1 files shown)


phkahler:
In testing this I may have stumbled on a simplified test case for 296. Verified it is not a regression with this PR.

phkahler:
The story: When combining contours into a single polygon, islands or loops inside the outer perimeter are connected by a bridge which goes from a point on the existing contour to a point on the interior one (hole) and then after going around the hole we go back across the bridge (between the same 2 points but in reverse) and continue with points on the existing contour. So if the outer is points A,B,C,D, ..... Z and the inner is points a,b,c,d,e .... n maybe the combined contour connects a bridge from C to k which would go:  A,B,C, k,l,m,n,a,b,c,d...k,C,D,E, ..... Z  notice that because of the bridge C and k are in the combined list twice. In the case with issue #303 the bridge is between points that are coincident. That means the polygon will have 2 points C,k on top of each other and then another pair k,C later. This would cause problems in ear triangulation, however there is a pass made over the contour to eliminate duplicate points. This patch just tweaks the bridge code to not insert point C at all, which eliminates the need to delete it later which apparently wasn't happening for reasons?!?!

In other words, the basis for this is sound and consistent with the other code so much so that it should not be needed. But for the same reasons it should be fine to use it.

phkahler:
Updated brace style.

PR #690, opened 2020-09-01
NFC: performance improvements

phkahler:
First commit changes the order of booleans in repeat-groups. This reduces the average time for the booleans by combining equal size pieces. It also does the transforms in parallel when build with OMP.

Second commit brings OpenMP to the remaining time-consuming part of boolean operations.

### DIFF
--- a/src/groupmesh.cpp
+++ b/src/groupmesh.cpp
@@ -106,25 +106,27 @@ void SMesh::RemapFaces(Group *g, int remap) {
 
 template<class T>
 void Group::GenerateForStepAndRepeat(T *steps, T *outs, Group::CombineAs forWhat) {
-    T workA, workB;
-    workA = {};
-    workB = {};
-    T *soFar = &workA, *scratch = &workB;
 
     int n = (int)valA, a0 = 0;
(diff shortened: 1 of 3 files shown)


src/groupmesh.cpp
```diff
@@ -149,7 +149,7 @@ void Group::GenerateForStepAndRepeat(T *steps, T *outs, Group::CombineAs forWhat
     std::vector<T> *scratch = &workA;
     // do the boolean operations
     while(n > 1) {
-#pragma omp parallel for
+//#pragma omp parallel for
```
whitequark, line 152: Should this be commented?

phkahler: It should probably be removed. Running that loop in parallel does work, but it ends up giving only 1 thread to each boolean operation which slows them down.

phkahler:
Address Sanitizer fails, probably an issue in SSurface::AddExactIntersectionCurve(). I'll try pulling the first commit into a separate PR, that one is simple enough it should pass.

phkahler:
Closing this, as it's been split into two separate PRs - one works, but the "simple" one still fails ASAN.

PR #691, opened 2020-09-01
NFC: Performance. For step-and-repeat groups, create the copies first…

phkahler:
… (in parallel) and then combine them using unions of equal size shells to reduce the total time spent on booleans.

This was pulled out of the previous PR that failed ASAN.

### DIFF
--- a/src/groupmesh.cpp
+++ b/src/groupmesh.cpp
@@ -118,7 +118,7 @@ void Group::GenerateForStepAndRepeat(T *steps, T *outs, Group::CombineAs forWhat
     std::vector <T> workA(n);
     workA[0] = {};
     // first generate a shell/mesh with each transformed copy
-#pragma omp parallel for
+
     for(a = a0; a < n; a++) {
         transd[a] = {};
         workA[a] = {};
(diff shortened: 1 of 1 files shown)


phkahler:
@whitequark  Looks like I need some help with initialization and cleanup of these vectors of shells. What is the right way?

phkahler:
I think I'm going to scrap this change. A better way to copy a shell N times involves less than 2*log(N) boolean operations vs the N-1 used now and also by this PR. Still need to figure out how to remap faces, but it should be significantly faster than the 10-30 percent improvement from this.

rpavlik:
Sorry I apparently failed to submit my comment earlier: it was "I suspect you may need to clear one of the vectors here (or maybe clear scratch at the top of the loop) - not std::vector::clear() but the equivalent thing that is in the internal list classes, that calls clear on each element."

phkahler:
> Sorry I apparently failed to submit my comment earlier: it was "I suspect you may need to clear one of the vectors here (or maybe clear scratch at the top of the loop) - not std::vector::clear() but the equivalent thing that is in the internal list classes, that calls clear on each element."

@rpavlik  Thanks. I didn't know there was such a thing :-)  I found a solution in PR #709 which seems to not leak. It still seems like the old version leaves one sshell lying around, as does my new one. That must clean up when it goes out of scope.

PR #697, opened 2020-09-12
Fix #696. Account for multiple coincident edges when looking for nake…

### LINKED ISSUE #696: Coincident edges show as "Naked Edges"
### System information

SolveSpace version: 3.0~d8f5a8da

Operating system: Fedora 32

### Expected behavior

Make a sketch with 2 rectangles with one point coincident. Extrude. Analyze->Show Naked Edges

There should be no naked edges.

### Actual behavior

The coincident edges are shown red and there are 4 naked edges found.

### Additional information

You get similar behaviour even if the 2 extrusions are from different sketches as shown here
Coincident_edges

This is similar to the second problem reported in issue #303. It is also loosely related to issue #171.

phkahler:
…d edges.

Also closes out the second issue reported in #303.

### DIFF
--- a/src/mesh.cpp
+++ b/src/mesh.cpp
@@ -971,9 +971,15 @@ void SKdNode::MakeCertainEdgesInto(SEdgeList *sel, EdgeKind how, bool coplanarIs
 
             switch(how) {
                 case EdgeKind::NAKED_OR_SELF_INTER:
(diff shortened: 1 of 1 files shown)


phkahler:
There is something strange going on. See my comment in issue 696.

PR #705, opened 2020-09-18
de_DE: missing translations

rurban:


phkahler:
Is there not a key binding for "Extrudieren"? That one seems important.

rurban:
Yes, I added some more shortcuts

phkahler:
Thank you for this update!

phkahler:
@rurban  I'm new to translations. How do we update the .pot file with strings that are currently missing? The en_US translation is missing several strings so I was going to update that but don't they need to be in the .pot file first? Is there an automated way to update the .pot?

rurban:
The normal autotools way is to do `make -C po update-po`
Haven't checked that yet though. All the line numbers are off eg.
There is an automated way, yes. Bis xgettext and msgmerge, I think. But it's about 30 years since I last did that extensively.

PR #710, opened 2020-09-21
translate_solvespace updates

rurban:
for all languages. See PR #705


rurban:
wrong rebase, will submit a new one

PR #711, opened 2020-09-21
translate_solvespace updates

rurban:
for all languages. See PR #705

my russian is not good enough to fix the remaining ru and uk strings.
german and french are now complete.


phkahler:
It might be slightly premature. I almost opened an issue like this yesterday, but there may be some string changes/addition/deletions due to the remaining issues before 3.0. After that we need to update the .pot file and then ask people to update the localizations. Having said that, it probably wouldn't hurt to update them whenever people are willing to do so.

I would also suggest that hot-keys try to be consistent across languages unless they're really awkward. I know English seems to get preferential treatment, but even there we have some oddities. reVolve got "V" because "R" was taken by rotate.

PR #713, opened 2020-09-21
Add a cmake option ENABLE_LTO for link time optimization on Release b…

phkahler:
…uilds.

LTO builds take significantly longer but produce smaller executables with better performance.


whitequark:
I think the condition isn't useful; for example it prevents `RelWithDebInfo` builds from using LTO.

phkahler:
@whitequark what would be better? When developing I prefer release builds, but LTO takes several minutes longer. I don't think we want it for CI builds.

whitequark:
Just have no condition? When developing you should use two build trees to avoid spurious rebuilds, anyway.

phkahler:
> Just have no condition? When developing you should use two build trees to avoid spurious rebuilds, anyway.

Do you really want it for CI? we already take too long there sometimes. Also, what would two build trees get me?

whitequark:
> Do you really want it for CI? we already take too long there sometimes.

Well, you're adding an option, not unconditionally enabling LTO builds. On CI (and presumably by default) the option would be off.

phkahler:
Got it, you meant just the inner condition. Done.

PR #717, opened 2020-09-25
Eigen over master

Evil-Spirit:
don't try to compile since I can't open and build it with QT anymore.


Evil-Spirit:
Allocator for expressions wasn't merged, but this is the common thing can be implemented by everyone.

PR #725, opened 2020-09-29
WIP: IDF file Linking & Import.

phkahler:
Can read PCB outlines and cutouts, as well as Pin and Mounting holes. Import should be considered deprecated as the sketch is no very useful.

Not ready for merging but I want it out there to get feedback from the community. I hope to add Routing Keepouts as construction lines, and do some cleanups. Components may come later if there is interest.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -175,6 +175,7 @@ set(solvespace_core_SOURCES
     group.cpp
     groupmesh.cpp
     importdxf.cpp
+    importidf.cpp
     mesh.cpp
     modify.cpp
     mouse.cpp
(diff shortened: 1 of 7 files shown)


neuks:
In my workflow, slvs is only used for mechanism simulation of moving parts, which it was originally designed for on its old 2D versions, and i would handle over the more complexed solid modeling jobs to FreeCAD which owns a more capable brep kernel. If slvs wants to be a software package which works for both advanced solid modeling and mechanism simulation, it cannot be done without a complete overhaul of the geometric kernel.

By this post, I mean, after whitequark's leaving, the project needs a new plan or vision for its future functional developments, would it become more advanced in mechanism simulation, or going all the way down to other advanced solid modeling softwares like FreeCAD or CATIA, Solidworks etc.

Issue #652, opened 2020-07-14
Can't cut half sphere out of a shape in NURBS


xzcvczx:
### System information

SolveSpace version: 3.0-c4ca4be9 

Operating system: kiss linux 

### Expected behavior

extrude cube, new sketch on a face, draw closed half circle, revolve, difference 

cube with a half sphere cut out

### Actual behavior

doesn't make the cut

### Additional information

[spherecut.zip](https://github.com/solvespace/solvespace/files/4915837/spherecut.zip)

whitequark:
@phkahler PTAL

phkahler:
@xzcvczx and @whitequark  It works if you check "force NURBS surfaces to triangle mesh" on the lathe group. I tried "Intersection" and that also worked for triangle mesh but not NURBS. So this looks like a typical NURBS issue. Then I decided to use "Revolve" instead of Lathe. Added a construction line perpendicular to the rotation axis (from the middle to the midpoint of the arc) so there would be a handle to drag. I constrained a point to the cube surface to set the angle to 180 degrees. This worked slightly better. NURBS Intersection actually worked correctly, but difference still did not. Triangle mesh worked fine of course. That doesn't help if you want to export STEP though.

This is another nice test case for the NURBS code and will have to wait for the next deep dive into that stuff.

phkahler:
Also for fun I replaced the arc with a rectangle. Lathe worked fine with difference and intersection. Tried the rectangle with Revolve and the NURBS had issues with both boolean operations. There seems to be more than one bug lurking in there.

phkahler:
Update. Since we just merged #673 to master, you will be able to work around this problem with the attached.
[spherecut2.zip](https://github.com/solvespace/solvespace/files/5071280/spherecut2.zip)

It still doesn't like curves coincident with the top of the box but it's fine that the sphere center is on the top. Just keep the arc more than 90 degrees and less than 180.
![spherecut](https://user-images.githubusercontent.com/14852918/90184193-b9bed600-dd82-11ea-8762-f4257cac1c7d.png)

Christoph-D:
For me, lathe difference with a triangle fails in the same way:
![triangle](https://user-images.githubusercontent.com/134562/93716447-e6160100-fb3d-11ea-9b0b-7b05063f271c.png)
[triangle.slvs.zip](https://github.com/solvespace/solvespace/files/5252089/triangle.slvs.zip)

[...]

PR #734, opened 2020-10-12
No zero length normals.

phkahler:
If a surface normal is zero length, try harder by using tangents at different points along the degenerate U or V curve - like at the poles of a sphere.

### DIFF
--- a/src/srf/ratpoly.cpp
+++ b/src/srf/ratpoly.cpp
@@ -340,6 +340,19 @@ void SSurface::TangentsAt(double u, double v, Vector *tu, Vector *tv) const {
            den_u = 0,
            den_v = 0;
 
+    // we don't want zero-length tangents when control points are coincident.
+    if(0 == u) {
+        u += 0.000000001;
+    } else if(1 == u) {
+        u -= 0.000000001;
+    }
+
+    if(0 == v) {
+        v += 0.000000001;
+    } else if(1 == v) {
+        v -= 0.000000001;
+    }
+
     int i, j;
     for(i = 0; i <= degm; i++) {
         for(j = 0; j <= degn; j++) {
@@ -371,9 +384,13 @@ Vector SSurface::NormalAt(Point2d puv) const {
 }
 
 Vector SSurface::NormalAt(double u, double v) const {
-    Vector tu, tv;
+    Vector tu, tv, n;
     TangentsAt(u, v, &tu, &tv);
-    return tu.Cross(tv);
+    n = tu.Cross(tv);
+    if(EXACT(n.Magnitude() == 0)) {
+        dbp("Zero surface normal u,v = %lf, %lf", u, v);    
+    }
+    return n;
 }
 
 void SSurface::ClosestPointTo(Vector p, Point2d *puv, bool mustConverge) {


phkahler:
@jwesthues  This fixes the problem of cutting a hemisphere out of a face, but it introduces some slightly weird shading on cones. Apparently for shading a zero normal is better than this, but for NURBS this is better than zero-length. It is actually correct at the poles of a sphere, which is where the remaining problem was happening in issue #652. Just wondering if you have a better idea how to deal with zero-length normals.

jwesthues:
Without having thought too deeply about this; what's the advantage of what you're doing now with `1 - u` or `1 - v`, vs. just displacing it slightly from the bad point (e.g., if it fails at (u, v) = (0, 0), then try at (0.001, 0.001))?

phkahler:
What I did here gives a correct normal at the pole of a sphere. I was investigating a problem with a sphere. For a cone the best displacement might be at (0.5 0.001) or (0.001 0.5) but that will be similar to this normal and will need to decide between 0.001 or 0.999? There really is no correct normal for the tip of a cone - ideally I think finding the axis of revolution would give the best normal for some definition of best.

jwesthues:
There's no correct normal at the tip of the cone, but there is a correct normal over 100% of the surface area. So it still seems naively like just displacing the test point inwards (like with respect to the trim curves in the (u, v) plane) a bit should always get the desired answer, with either a fixed offset or some kind of heuristic (step until both derivatives have magnitude over some threshold?) to determine that?

I see why your 1 - u approach gets the right answer on the sphere sections, which have a kind of symmetry. It's not obvious to me why that would be correct for arbitrary surfaces, though.

ruevs:
@phkahler I agree with Jonathan.

As you have already explained [here ](https://github.com/solvespace/solvespace/issues/652#issuecomment-696909391) sometimes NURBS patches have a "degenerate" (zero length/to a point) edge/side. In this case the tangents along that edge become undefined(?).

Here is a picture:
NURBSPatchWithDegenerateEdgeCausesZeroLengthNormalsBecauseOfUndefinedTangent
A [normal `n` at a point (u.v)](https://github.com/solvespace/solvespace/blob/d72eba8039c52f5e8d3ab36ec2d8a7598dfd578b/src/srf/ratpoly.cpp#L373) is calculated by [calculating the tangents `tu` and `tv` at that point](https://github.com/solvespace/solvespace/blob/d72eba8039c52f5e8d3ab36ec2d8a7598dfd578b/src/srf/ratpoly.cpp#L335) and taking the cross product.
However when an edge is "degenerate" - as is the (0,1)(1,1) edge in this example part a cone - then its length becomes zero and both `tu` and `tv` become undefined for any point along the (x,1) "degenerate edge".

In my opinion in this case the correct logic is to fix `SSurface::TangentsAt` so that it calculates sensible tangents as @jwesthues suggested:
```
if ((1==v) &&
        (trim curve along (0,1)(1,1) is zero length)) {
    v -= 0.00000000001;
}
if ((0==u) &&
        (trim curve along (0,0)(0,1) is zero length)) {
    u += 0.00000000001;
}
etc.
```

I could not quickly figure out how check if `(trim curve along (0,0)(0,1) is zero length)` so as an experiment I did this stupidity:

```
void SSurface::TangentsAt(double u, double v, Vector *tu, Vector *tv) const {
    Vector num   = Vector::From(0, 0, 0),
           num_u = Vector::From(0, 0, 0),
           num_v = Vector::From(0, 0, 0);
    double den   = 0,
           den_u = 0,
           den_v = 0;

    if(0 == u) {
        u += 0.000000001;
    }
    if(0 == v) {
        v += 0.000000001;
    }
    if(1 == u) {
        u -= 0.000000001;
    }
    if(1 == v) {
        v -= 0.000000001;
    }

    int i, j;
    for(i = 0; i <= degm; i++) {
        for(j = 0; j <= degn; j++) {
            double Bi  = Bernstein(i, degm, u),
```

It is wrong in so many ways (e.g. `==` comparison on a `float`), but if "fixes" both "spherecut.zip" https://github.com/solvespace/solvespace/issues/652#issue-656237557 and "triangle.slvs.zip" https://github.com/solvespace/solvespace/issues/652#issuecomment-695808103

Instead of the above logic it bay be possible to somwhow change the [algorithm itself ](https://github.com/solvespace/solvespace/blob/d72eba8039c52f5e8d3ab36ec2d8a7598dfd578b/src/srf/ratpoly.cpp#L346) to do "the right thing (TM)" in such cases but I am light years away from understanding it :-)

And last  [`SBezier::TangentAt`](https://github.com/solvespace/solvespace/blob/d72eba8039c52f5e8d3ab36ec2d8a7598dfd578b/src/srf/ratpoly.cpp#L57) also uses the same logic so maybe a bezier loop like this:
BezierDiscontinuity
could cause a similar problem is some case... ?

phkahler:
@ruevs  The triangle case from 652 is already fixed on master, it's only when you use the sketch line as the rotation axis that there is a problem and it seem to be due to the zero normal vector at the "pole".

My fix works by using 2 different tangent vector at the pole. Only one of them will go to zero which causes the cross product (normal) to be zero. I think what I did has the normal pointing inward or outward at opposite "ends" of the degenerate curve.

I'm going to try what @jwesthues suggests and just take tangents nearby along the "good" U or V axis.

Another thought I had... For argument lets assume the V direction is the zero-length edge at a given U value. I think a better way to find tangent tv (only as a substitute when the real tv is zero) would be to calculate the derivative to tv with respect to U. that should work and would produce different tv vectors depending on the V coordinate even though it's always at the same point in space. Similar to offsetting along U but without actually doing so. I don't have the ambition to write that at the moment so I'm going to try the nearby u,v point ;-)

phkahler:
I tried this. Both still fail often for the case of a triangle lathed around it's own edge:
```
    if(tu.MagSquared() < tv.MagSquared()) {
        // weighted average of v and 1-v.
        v = (v*31 + (1-v))/32.0;
    } else {
        u = (u*31 + (1-u))/32.0;
    }
    TangentsAt(u, v, &tu, &tv);
    n = tu.Cross(tv);
```
But this version doesn't have strange shading issues, probably the other version points the normal inward sometimes and this doesn't.

phkahler:
I've updated to the @ruevs method which works better even for the spherecut. There are probably places that don't like the tangents to be zero, so fixing it there is probably better.

jwesthues:
Cool. If you want something nicer numerically than a check for exact equality, you could try checking if the magnitude of the vector is less than a threshold, and then incrementing (if u or v < 0.5) or decrementing (otherwise) u and v until that magnitude is greater or until you give up with an error.

phkahler:
@ruevs  it's fine if you think we can do better. The check for weather a tangent is going to be zero would be that the adjacent control point is equal to the one we're evaluating at.

So if i,j = (0,0) we should check if ctrl[0][0].EQUALS(ctrl[1][0])  then the tu tangent is going to be zero so we should offset our v coordinate to some small value. But if ctrl[0][0].EQUALS(ctrl[0][1]) then tv is going to be zero and we should offset u.

At the other extreme we're at u,v = (1,1) we'd check if ctrl[degm][degn].EQUALS(ctrl[debm-1][degn]) to see if tu is zero and set v = 0.999999. And so on. There are a lot of conditions that can probably be somewhat simplified with a bit of cleverness.

Don't take my word for which control points need to match here, after that other PR for the other NURBS issue I'm not sure I can logic today ;-)

jwesthues:
Being lazy, I'd probably just nudge both. The unnecessary nudge won't hurt much.

Another direction is to not generate coincident control points in the first place. I'm not perfectly confident that's always possible, but I suspect that it is (as long as you don't need the trim curve to be a square from (0, 0) to (1, 1), like it is now for all new extruded or lathed surfaces; whenever there's a singularity like the tip of a cone, you'd need to make that correspond to a single point of the trim curve).

ruevs:
Here is what I came up with. Seems to work :-)
https://github.com/solvespace/solvespace/pull/736

phkahler:
Closing as inferior to #736.

PR #735, opened 2020-10-12
Convert revovled line to planes correctly when theta < 0.

phkahler:
Fixes issue 680.

I think the function ideally needs to be rewritten or it may fail on revolves or lathes around an axis that is not in the sketch plane. It also won't work if we allow sections over 90 degrees - things might fall outside the range [0..1] for U or V.

### DIFF
--- a/src/srf/surface.cpp
+++ b/src/srf/surface.cpp
@@ -984,6 +984,10 @@ void SShell::MakeFirstOrderRevolvedSurfaces(Vector pt, Vector axis, int i0) {
 
                 double vm = (ff.Minus(of)).Dot(v);
                 v = v.ScaledBy(vm);
+                if(vm <  0) { // happens when we revolve with negative theta
+                    of = of.Minus(v);
+                    v = v.ScaledBy(-1);
+                }
 
                 srf->degm = 1;
                 srf->degn = 1;


phkahler:
This is bothering me. Either the two lines in the if {} block should be in the reverse order, or the Minus() should be a Plus(). But it works the way it is and does not work when corrected. Or maybe I'm not thinking clearly for some reason.

phkahler:
@ruevs if you could have a look. I'm beside myself as to why this works when it looks like a mistake. This is my last NURBS fix for a while really ;-)

phkahler:
I visualized the surface boundaries and there is nothing wrong with the original code. The "fix" messes them up badly, but somehow convinces the bug in 680 to go away. Making a correct but larger bbox surface causes another surface problem in the same sketch. Not converting the swept lines to 1st order surfaces fixes the problem as well, but we like to have planes that can be selected.

ruevs:
I don't understand what is going on. First I'll have to understand the the whole `MakeFirstOrderRevolvedSurfaces` better.

phkahler:
@ruevs  I've confirmed the problem is not in that function, but still not sure where it is. That function recognizes planes created by lathe or revolve and replaces the surface (not the trim curves) with a degree 1x1 surface (rather than a 1x2 with 6 control points). By doing so it allows us to constrain to the flat surface and also lets the flat surface be merged with other coincident flat surfaces. By not calling that function for revolved surfaces the problem goes away but so do the benefits of calling it.

My fix was actually wrong and messed up the bounding box (flat surface control points). Somehow that prevented what I think is an extraneous edge from appearing later in the surface polygon. I have confirmed that the extra edge is produced by a surface-surface intersection. In fact, modifying MakeFirstOrderRevolvedSurfaces to add 1 percent extra space around the edges causes another instance of a false edge getting created. It seems the edges that result from surface intersections need to be chopped up better.

Issue #680, opened 2020-08-16
second revolve breaks the surface


zzj1213:
As shown on the test file:  the **second** revolve, based on the workplane resulted from the first resolve, breaks the surface.

Solvespave version: **Version 3.0 (3.0~bc4924ae)**,
macOS Catalina Version 10.15.2

![second_revolve_test_1](https://user-images.githubusercontent.com/1044245/90326430-c7f73700-dfba-11ea-83af-56b069b7036e.png)

![second_revolve_test](https://user-images.githubusercontent.com/1044245/90326403-9bdbb600-dfba-11ea-80c0-c93464540d19.png)

[second_revolve_test.slvs.zip](https://github.com/solvespace/solvespace/files/5079984/second_revolve_test.slvs.zip)

zzj1213:
related to this issue: #439 

This consecutive  "revolve" workaround for Extrude along an arbitrary path seems a very important feature.

The path in #439 also has the same issue.

![path](https://user-images.githubusercontent.com/1044245/90326634-405ef780-dfbd-11ea-9119-895a4db6b095.png)

[path.zip](https://github.com/solvespace/solvespace/files/5079994/path.zip)

phkahler:
This looks like a problem with coincident surfaces - the end face of one revolve and the "bottom" face of the next one. Unfortunately this is common. As we learn more about the NURBs kernel some of these issues might get fixed.

If you don't need STEP export this does work if you check "force NURBS to triangle mesh" on the 3rd revolve group and continue construction.

Path extrusion would offer a more reliable way to make this, but it's still a long way off. This is exactly the kind of thing I could see being created as one piece (hence no NURBS problems), but you'd still need a difference or intersection to cut the bottom flat.

zzj1213:
The first end and second begin faces are carbon copy, but the NURBS kernel treats them different.   Is it because of approximation?

This brings up another important feature: **stitch**: filling the gap between NURBS faces when the faces are not a plane. 

After using Solvespace intensively for two weeks, really love it.  But in the meantime,  found out these two major limitations to handle serious jobs:
- path extrusion: revolve can be a workaround in some cases;
- stitch: no workaround for now.

zzj1213:
Ran into this: https://github.com/wsteffe/solvespace_occ 

First tests with **opencascade** bezier curves: 

https://github.com/wsteffe/solvespace_occ/commit/f186b74006feb40814917e6b7b5502b3f56d6a68

[...]

Issue #497, opened 2019-10-23
Nurbs geometry differencing crashes.


neuks:
### System information

SolveSpace version: 3.0~5d78f99

Operating system: Windows7 64bits

### Expected behavior

When differencing a cylinder with the same diameter with the direction right perpendicular and through the axis line, there should be a hole which right seperated the left and right side of the cylinder.

PS: This is an important feature, similar conditions including drilling some perpendicular screw holes which intersects.

### Actual behavior

When differencing a cylinder with the same diameter with the direction right perpendicular and through the axis line, the nurbs geometry solver crashes.

![shot1](https://user-images.githubusercontent.com/12725624/67372051-e2859d00-f5af-11e9-9aca-c549e87375c0.png)

I know there's a solution by activating the 'Force nurbs surface to triangle mesh', but it's not what I expected, as the latter modifications would always be resolved to triangle meshes.

### Additional information

This situation happens with all versions of SolveSpace, as checked on version 2.3~latest.

Great thanks for your work on SolveSpace, and it is very helpful with my research work.

For bugs, please attach a savefile that shows the problematic behavior.
You can attach `.slvs` files by archiving them into a `.zip` first.

neuks:
I had a solution myself, by setting the second diameter to 20.001, then the solver appears normal, maybe this method could be used as a workaround for this problem, as it seems to be difficult to solve.

phkahler:
[issue_497.zip](https://github.com/solvespace/solvespace/files/3764380/issue_497.zip)
I created something similar. Was not able to make it crash, but there were NURBS boolean issues. The attached model works with equal sized cylinders. The trick was to create the plane for the second one at an angle (see the construction lines for that workplane). A circle is actually composed of 4 arcs of 90 degrees. When doing this difference using axially aligned circles, the corner points are at the intersection of 4 surface patches. By rotating one of them the complexity of the surface intersections is reduced a bit. That's my guess as to why it works anyway.

Oh wait.  The success also required the changes in my nurbs branch, so that attached file doesn't work with master. It doesn't crash. There are some missing edges, so this may be of interest to @ruevs who had been doing recent work on booleans.

The only thing in the nurbs branch (PR #473) that seems like it might apply is the one change to keep U,V coordinates in range 0 to 1. That was unexpected for me, so I might want to go looking at how that might even be relevant in this case.

neuks:
[...]

PR #737, opened 2020-10-15
Add vertexes to curve intersection list in addition to surface inters…

phkahler:
…ections.

Sometimes a vertex can be used to split a curve where surface intersections can't. Those unsplit curves can cause boolean failures. This fixes issue #680 

Needs more testing to be sure it doesn't break anything.

@ruevs and @jwesthues  This and the improvements suggested in the code comments should fix some other open issues.

### DIFF
--- a/src/srf/boolean.cpp
+++ b/src/srf/boolean.cpp
@@ -20,6 +20,32 @@ void SShell::MakeFromIntersectionOf(SShell *a, SShell *b) {
     MakeFromBoolean(a, b, SSurface::CombineAs::INTERSECTION);
 }
 
+// We will be inserting existing verticies into curves to split them
+// todo: this is only using the ends of exact curves, and it is only
+// using them to split existing curves, not new intersections.
+// It resolves some issues but we could do better. We will need to
+// reorder things so the surface intersection curves exist prior to
+// splitting any curves at all in order to have their verticies too.
+// Better still would be to get curve/surface intersection to work
+// more reliably at the edges - maybe do curve/curve tests as part
+// of the curve-surface intersection test.
+static void FindVertsOnCurve(List<SInter> *l, const SCurve *curve, SShell *sh) {
+    for(auto sc : sh->curve) {
+        if(!sc.isExact) continue;
+        for(int i=0; i<2; i++) {
+            Vector pt = sc.exact.ctrl[ i==0 ? 0 : sc.exact.deg ];
+            double t;
+            curve->exact.ClosestPointTo(pt, &t, /*must converge=*/ false);
+            double d = pt.Minus(curve->exact.PointAt(t)).Magnitude();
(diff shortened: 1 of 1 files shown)


phkahler:
Like so many other times, I let this bake in my head for a while and think it's not a great implementation. It seems that curves are not split when they intersect the edge of another surface - not just at a vertex, but anywhere (see issue 497). Going to let it sit a while longer and see if something better comes to mind. I'm thinking about curve-curve intersections, which has me thinking about number 171 as well.

Or I could merge it and do something better later? What do people think about that approach? This is a fairly small patch.

jwesthues:
On reflection, I don't understand this either. My high-level idea was that if a curve from shell A intersects a curve or vertex from shell B, then it must also intersect a surface from shell B at that same point (since curves lie on surfaces, and vertices lie on curves). If your change helps, then I think that shows the curve-surface intersection isn't working like I'd intended. That observation would be a good lead for debugging the curve-surface intersection, though.

phkahler:
> If your change helps, then I think that shows the curve-surface intersection isn't working like I'd intended.

Correct. I think in cases like a small box having a coincident edge with a larger box the planes cut the long edge correctly and things work as intended. In the case of #680 there is a curve created by the intersection of two planes, but one of them is larger than the geometry inside it so that curve extends too far. The surfaces that meet at the magic vertex that fixes the problem are all tangent at that point - 2 flat surfaces and 2 revolved arcs. But I've seen other cases where a line should be cut at the edge of a surface and it doesn't happen - I think #497 and #537 are cases where a straight trim curve should be cut on the edge of another surface but doesn't.

One thing I'm considering is moving these kinds of checks so they are more like a part of the curve-surface intersection (test against the trim curves of that surface explicitly). which would be less invasive.

Similarly, I don't understand why the generic surface/surface intersection doesn't work when the intersection is along the edge of a surface, like on lathe seams. If we just move the cutting plane a tiny bit it works. Now we make a copy of the trim curve which can be exact, but that doesn't explain why it didn't work before.

jwesthues:
Yeah, seems strongly like the root problem is in the curve-surface intersection. We expect the Newton-type solver to break around tangencies and such, but it seems like there's also breakage beyond that. I see at least two possible directions:

1. Improve the numerical solver.
2. Add more special cases for stuff we can solve in closed form (cylinder-line, cone-line, plane-circle, cylinder-circle, etc.), or at least numerically with better behavior than the general case.

Option (2) is less elegant, but I suspect it might have the biggest near-term payoff. It's also a speedup.

phkahler:
I'm going to merge this for now because it fixes a few real problems. In the near future I intend to redo it to get better coverage of existing open issue but I don't have the cycles right now.

ruevs:
A further fix for this is here https://github.com/solvespace/solvespace/pull/786

PR #741, opened 2020-10-16
Travis: add windows target

ppd:
Mimic the appveyor cfg without the artifact deployment.
Also upgrade to Visual Studio 2017/v141 toolset.


phkahler:
Thank you @ppd

PR #743, opened 2020-10-17
Add macOS travis config, fix issue in CMakeLists.txt causing macOS build to break

vespakoen:
Fingers crossed ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,36 @@ jobs:
       name: Snap arm64
   include:
     - stage: test
+      name: macOS
+      os: osx
+      osx_image: xcode11.2
+      install: "./.travis/install-macos.sh"
+      script: "./.travis/build-macos.sh"
+      deploy:
+        provider: releases
+        api_key:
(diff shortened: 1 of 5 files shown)


vespakoen:
The macOS build seems flaky, and when a lot of output is written to stdout it sometimes crashes with a "write error".
I'll have to research how to fix this, I had it on my own computer as well, and a second run worked without problems.

At least it seems to be compiling again!

vespakoen:
I will also experiment with https://github.com/kakashidinho/metalangle for Big Sur support (which removed OpenGL)

vespakoen:
Removing `VERBOSE=1` seems to have fixed it, should be all good now

phkahler:
It works except for this:
```
Success! 254 test cases (0 skipped), 876 checks, 13.403s

[100%] Built target test_solvespace

/Users/travis/.travis/functions: line 109: ./.travis/sign-macos.sh: Permission denied

The command "./.travis/build-macos.sh && ./.travis/sign-macos.sh" exited with 126.

```
I'm merging it because it's way better than before ;-)

vespakoen:
I forgot to chmod +x that script when i brought it in again, but even with
that permission it would fail because of missing build secrets.

I will get them over to you as soon as I can (probably monday)

ruevs:
> I will also experiment with https://github.com/kakashidinho/metalangle for Big Sur (the upcoming macOS version) support (which removed OpenGL)

Aplle are... strange. Such a nice foundation (NeXT, the Match/XNU kernel, UNIX/Darwin) and on top of it... a UI where there are still modal dialog boxes with a single OK button that one can only dismiss with a mouse.

For I long time I though they would backtrack on removing OpenGL... Ehhh...

ppd:
Now we have two OSX jobs in the deploy stage. Isn't that one too many?

vespakoen:
@ruevs True... it's hard to keep up with them, especially as an OSS project, but I'll give it a go anyways...

@ppd you are right, I was a bit careless when copying over the config from my last PR, I am fixing it in #748

ruevs:
@vespakoen by the way take a look here:
https://chromium.googlesource.com/angle/angle#level-of-opengl-es-support-via-backing-renderers
It seems the Metal support (to the extent it is ready) has been upstreamed.

vespakoen:
Oh cool! that's good to know!

PR #746, opened 2020-10-17
Avoid zero tangnet vectors on degenerate NURBS edges. Fixes #652

### LINKED ISSUE #652: Can't cut half sphere out of a shape in NURBS

phkahler:
@ruevs This is probably the clearest and cleanest we can get. What do you think?

### DIFF
--- a/src/srf/ratpoly.cpp
+++ b/src/srf/ratpoly.cpp
@@ -332,7 +332,7 @@ Vector SSurface::PointAt(double u, double v) const {
     return num;
 }
 
-void SSurface::TangentsAt(double u, double v, Vector *tu, Vector *tv) const {
+void SSurface::TangentsAt(double u, double v, Vector *tu, Vector *tv, bool retry) const {
     Vector num   = Vector::From(0, 0, 0),
            num_u = Vector::From(0, 0, 0),
            num_v = Vector::From(0, 0, 0);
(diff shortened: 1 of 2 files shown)


ruevs:
Calling recursively looks clean, I may change [mine](https://github.com/ruevs/solvespace/commits/NURBSTangentsNumerical) to do the same. But is there some weird potential "NURBS universe" where we never exit the recursion?

How does ` v+(1-v)*0.000001` work if v=1?

You need to pass a tolerance to Equals, IMHO the default is too big.

Check `tv` first. So far I could not figure out how to construct anything pinched along a u edge. Makes it faster by avoiding a check that may never be true with the current construction tools.

The flag is redundant. See my comment here:
https://github.com/solvespace/solvespace
/pull/736#issuecomment-708195950

"sungularities" = singularities

As for my overall opinion on the "numerical" (vs. logical) approach - did you take a look at this?
https://github.com/solvespace/solvespace/pull/736#issuecomment-709941205

phkahler:
> How does ` v+(1-v)*0.000001` work if v=1? I think you meant ` v+(0.5-v)*LENGTH_EPS`.

Oops. I started out thinking a weighted average so 0.99*v + 0.01*(1-v) and then was thinking something more like what you suggest. Brain just didn't write either one correctly. I prefer (0.5-v). Isn't LENGTH_EPS really small?

Using the flag to avoid infinite recursion just seems like a good idea, especially if you want to use such a tiny movement. Like I said before, it also allows flexibility at the call site but we don't really need that once it's fixed.

phkahler:
You mean switch the order? OK. Does that mean you're ok with "check result and retry" vs the "check for degenerate curve and pre-emptively move u,v" ?

ruevs:
This one is shorter, that is it's advantage. The other one is faster (but I need to measure, this is just based on amount of calculations) and "more correct" (whatever that meas) but more verbose...

The problem is that I have not had time to work on it more...

PR #747, opened 2020-10-18
Add executable permission on sign-macos.sh script

vespakoen:
Follow up on #743


vespakoen:
Closed in favor of #748

PR #748, opened 2020-10-18
Update .travis.yml for macOS build

vespakoen:
This removes the old OSX build, and fixes the repo argument (it was still set to my fork vespakoen/solvespace)

I also used the api_key that was set in the old OSX build, not sure if this will work though...

This PR is dependent on #747


vespakoen:
I just sent the secrets by email, once added to travis this build should succeed

PR #749, opened 2020-10-19
macOS CI fixes

vespakoen:
Those were already in my old PR, somehow forgot the bring them into the last one and basically did the work twice now 😂

### DIFF
--- a/.travis/build-macos.sh
+++ b/.travis/build-macos.sh
@@ -4,6 +4,6 @@ if echo $TRAVIS_TAG | grep ^v; then BUILD_TYPE=RelWithDebInfo; else BUILD_TYPE=D
 
 mkdir build
 cd build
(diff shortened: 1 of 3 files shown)


vespakoen:
This seems to work, see: https://travis-ci.org/github/vespakoen/solvespace/jobs/737010689

Just have to add the secrets to travis-ci.

ppd:
This will run `sign-macos.sh` unconditionally, yes? So we'll fail the macOS deploy job for every PR (which does not see the secret). Maybe we should even skip the whole macOS deploy job for such cases.

We also have two deployments for both macOS jobs; that doesn't seem right either.

vespakoen:
You are correct!

I will remove the deploy on the test stage.

For the deploy stage we can add a conditional "if: tag IS present" I think,
I just left it in there for now so its easier to see if it's working once
the build secrets are added.

PR #750, opened 2020-10-19
Try to build windows and debian on 2 cores

vespakoen:


ppd:
Wouldn't it be better to use something like `nproc` to adjust to the available CPUs?

ppd:
And `msbuild` takes `maxcpucount` or `m` without a value to automatically make use of all CPUs.

vespakoen:
I'll add both suggestions, thanks!

vespakoen:
It doesn't seem to improve the speed a whole lot but it might be worth it to keep anyways.

phkahler:
@vespakoen  We should have everything over at travis now.

ppd:
@vespakoen What do you think about adopting a commit prefix for all CI related changes? Like `Travis: Build windows and debian on max amount of cores`.

vespakoen:
This got added in #751, closing...

@phkahler I just noticed it, thanks! I will make sure that the macOS signing and notarising works (doing that over here: #753)
@ppd That's fine with me. We could also use something like https://www.conventionalcommits.org/en/v1.0.0/ ? maybe that's overkill though...

PR #751, opened 2020-10-19
Remove deploy step for macOS CI

vespakoen:
Removing deploy step from macOS CI `test` stage.
Will also add the `if: tag IS present` condition if the first build succeeds.

### DIFF
--- a/.travis/build-macos.sh
+++ b/.travis/build-macos.sh
@@ -1,9 +1,25 @@
 #!/bin/sh -xe
 
-if echo $TRAVIS_TAG | grep ^v; then BUILD_TYPE=RelWithDebInfo; else BUILD_TYPE=Debug; fi
+if echo $TRAVIS_TAG | grep ^v; then
+    BUILD_TYPE=RelWithDebInfo
+else
+    BUILD_TYPE=Debug
+fi
 
-mkdir build
+mkdir build || true
 cd build
-cmake -DCMAKE_OSX_DEPLOYMENT_TARGET=10.9 -DCMAKE_BUILD_TYPE=$BUILD_TYPE ..
-make -j2
-make test_solvespace
+
+LLVM_PREFIX=$(brew --prefix llvm@9)
+export CC="${LLVM_PREFIX}/bin/clang"
+export CXX="${CC}++"
+export LDFLAGS="-L${LLVM_PREFIX}/lib -Wl,-rpath,${LLVM_PREFIX}/lib" \
+export CFLAGS="-I${LLVM_PREFIX}/include"
+export CPPFLAGS="-I${LLVM_PREFIX}/include"
+
+cmake \
+    -DCMAKE_OSX_DEPLOYMENT_TARGET=10.9 \
+    -DCMAKE_BUILD_TYPE=$BUILD_TYPE .. \
+    -DENABLE_OPENMP=ON
+
+cmake --build . --config $BUILD_TYPE
+make -j$(nproc) test_solvespace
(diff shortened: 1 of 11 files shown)


ppd:
Suggestion: commit message `Travis: remove deploy step from macOS CI test stage`

vespakoen:
Allright! I will squash those commits before merge.

- I also added OpenMP
- Forced it to build with `BUILD_TYPE=Release` to try that out
- Updated the xcode version in the travis build

Let's see what happens, i'll use this branch to try out some things and hopefully land on a nice setup for macOS soon.

src/drawentity.cpp
```diff
@@ -315,9 +315,13 @@ void Entity::ComputeInterpolatingSpline(SBezierList *sbl, bool periodic) const {
             } else {
                 // The wrapping would work, except when n = 1 and everything
                 // wraps to zero...
-                if(i > 0)     bm.A[i][i - 1] = eq.x;
-                /**/          bm.A[i][i]     = eq.y;
-                if(i < (n-1)) bm.A[i][i + 1] = eq.z;
+                if(i > 0) {
```
ruevs, line 318: @vespakoen @phkahler 
These changes (I am not sure if they make sense) do not belong in this commit!

src/solvespace.h
```diff
@@ -513,7 +513,7 @@ class SolveSpaceUI {
     GraphicsWindow              GW;
 
     // The state for undo/redo
-    typedef struct {
+    typedef struct UndoState {
```
ruevs, line 516: Yes, I've been meaning to make a pull request for this annoying warning.
But, again, in my opinion it does not belong in this his commit.

src/mouse.cpp
```diff
@@ -624,24 +624,24 @@ void GraphicsWindow::MouseRightUp(double x, double y) {
         }
         if(gs.withEndpoints > 0) {
             menu->AddItem(_("Select Edge Chain"),
-                          [this]() { MenuEdit(Command::SELECT_CHAIN); });
+                []() { MenuEdit(Command::SELECT_CHAIN); });
```
ruevs, line 627: Same here. This whole file...

ruevs (commented):
Please be more careful what you include in the pull requests.

.travis/build-macos.sh
```diff
@@ -1,9 +1,23 @@
...
+if echo $TRAVIS_TAG | grep ^v; then BUILD_TYPE=Release; else BUILD_TYPE=Debug; fi
 
-mkdir build
+mkdir build || true
 cd build
-cmake -DCMAKE_OSX_DEPLOYMENT_TARGET=10.9 -DCMAKE_BUILD_TYPE=$BUILD_TYPE ..
-make -j2
-make test_solvespace
+
+export BUILD_TYPE=Release
```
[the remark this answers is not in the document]
vespakoen: This was just for checking if the build would work in Release mode, I changed it back to the default (automatic switching between `Debug` and `RelWithDebInfo` based on the presence of `$TRAVIS_TAG`

vespakoen:
@ruevs Points taken, sorry, I will split these up better in the future.

There were a lot of warnings, and I got some LLVM (v11) crashes / parse errors, so I thought maybe removing those warnings would get me further.

.travis/build-macos.sh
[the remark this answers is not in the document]
ruevs: I noticed too late. Sorry.

phkahler:
Wondering if we should unconditionally build with OpenMP now. Maybe the option could disable it - that's sometimes helpful in development. We were not using it before because whitequark didn't want users to need the extra DLL on Windows where we ship with no installer (just an exe). Jonathan is OK with requiring that, as am I. So if we're going to enable it for releases, should we just build with it unless set otherwise?

LTO is different because it takes a lot (3 to 4x)  longer to build and offers only 30% faster code (in the one test I measured). So I still think that has to be for release builds only.

PR #753, opened 2020-10-20
Travis: fixing travis lint problems, re-enable multi core building for macOS, quote variables

vespakoen:
- Fixing travis lint problems
- Re-enable multi core building for macOS
- Quote variables

I was wondering why signing wasn't working, only to find out that build secrets are not exposed to pull request builds (duhhh)
But these changes are probably helpful anyways.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+os: linux
+dist: xenial
 language: c
 git:
   submodules: false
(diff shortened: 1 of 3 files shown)


src/platform/guimac.mm
```diff
@@ -371,6 +371,7 @@ @implementation SSView
 
 - (id)initWithFrame:(NSRect)frameRect {
     NSOpenGLPixelFormatAttribute attrs[] = {
+        NSOpenGLPFADoubleBuffer,
```
ruevs, line 374: Again a source file change in a CI configuration commit...

vespakoen: I was experimenting with this in between builds and I guess I `git add .`'d it

I will clone the repository again for my own experiments and make a separate commit or PR for the change in `guimac.mm`

ruevs: Did you check these as well?
https://github.com/solvespace/solvespace/pull/751#pullrequestreview-512386393

ruevs:
(Sorry for the rant, and thank you for doing this important work.)

I don't mean to be negative, but was it really necessary to have [six commits in four days](https://github.com/solvespace/solvespace/commits?author=vespakoen) just to update the Travis config? Perhaps it would be better to experiment in a branch/pull request and "when it's ready" squash and merge just one commit.

There is a (semi)logical reason, why I think this is good - so far the history on master has been extremely low noise - call me obsessive compulsive, but when I've been away for a while I like to go through the log and review the commits. The last few days are...noise.

phkahler:
@ruevs  You have a point. But at the same time a lot of that has been messing with the CI stuff and I'm not sure how you really test changes to that without running stuff through it. I'm thinking it's just a temporary flurry of activity given the changes taking place. I expect to return to a smoother pace very soon. One thing whitequark did is enforce the same discipline you advocate here, so thanks for that. I agree with much of that philosophy even if I go against it sometimes myself.

ruevs:
> @ruevs You have a point. But at the same time a lot of that has been messing with the CI stuff and I'm not sure how you really test changes to that without running stuff through it.

Without knowing anything about Travis I was silently presuming that there is a way to "run it" on some commit without creating a new pull request. If that is not the case I apologise.

> One thing whitequark did is enforce the same discipline you advocate here, so thanks for that. I agree with much of that philosophy even if I go against it sometimes myself.

There are advantages to both approaches in different circumstances so don't mind me too much ;-)

vespakoen:
@ruevs I totally agree with you, this was partly due too me trying to get the builds green again, and seeing if the build was working on the solvespace repository (with it's api_key and secrets), which as turns out can never work in a pull request because it doesn't get the secrets (which makes total sense).

It was also partly due to my excitement of getting this to work properly, and then figuring out some optimisations later.
In hindsight, I shouldn't have rushed it.

I setup travis-ci on my fork as well, for doing experiments, and there I can test most of it as well.

Anyways, I will be more careful from now on, I don't think it's bad that you brought it up because it's a valid point, and I didn't take it personally ;)

ppd:
@vespakoen I think this PR broke snap deployments. Are you sure with the `skip_cleanup` -> `cleanup` change? We have to slow down a little; there's too much going on too quickly.

vespakoen:
@ppd It indeed looks like the snap started crashing after that PR. stange that it didn't crash during the PR itself, guess this is another case of "no secrets during PR builds".

I guess I have been bitten by this: https://travis-ci.community/t/misleading-dpl-v2-deprecation-warnings-with-v1-config/8153

ppd:
It's a case of artifacts being cleaned up after the build, so there's nothing to upload.

PR #754, opened 2020-10-20
macOS: add NSOpenGLPFADoubleBuffer to NSOpenGLPixelFormatAttribute

vespakoen:
This seems to solve a flickering issue.
Should I be using the `macOS:` prefix?

### DIFF
--- a/src/platform/guimac.mm
+++ b/src/platform/guimac.mm
@@ -371,6 +371,7 @@ MenuBarRef GetOrCreateMainMenu(bool *unique) {
 
 - (id)initWithFrame:(NSRect)frameRect {
     NSOpenGLPixelFormatAttribute attrs[] = {
+        NSOpenGLPFADoubleBuffer,
         NSOpenGLPFAColorSize, 24,
         NSOpenGLPFADepthSize, 24,
         0


phkahler:
@vespakoen  the prefix is good, as is adding a label. There exists a label for MacOS issues and you can also search on it. I don't think we even have any outstanding Mac issues at the moment but feel free to look.  I had noticed the toolbar flickering on your build yesterday and was planning to open an issue - it that the one this fixes? Also while I'm here, are the secrets working on Travis or did you disable a step? I think I may have screwed up when I copied them over.

Thanks!

vespakoen:
I also thought you screwed up adding the secrets, so I re-added them only to find out that travis doesn't add secrets to pull requests builds (which makes a lot of sense).

It should be fine now though, but will only run when pushing tags, you could try it out with something like:

```
git tag -a 3.0.0-alpha.1 -m "3.0.0-alpha.1"
git push --tags
```

vespakoen:
Oh and, yes, that is the issue this is fixing, I noticed some flickering (just rarely) when hovering over UI elements. This seems to have fixed it (at least from my testing so far)

PR #757, opened 2020-10-20
CI: cleanup, move to arm64-graviton2, deploy only on master

ppd:
A few cleanups and proposals:

- Rename Travis job Debian to Ubuntu. I think this makes more sense than what we have right now, but it's not too important.
- arm64-graviton2: From my tests, this will result in way fewer build-fails.
- Run deploy stage only on master. 

Comments & opinions welcome.


phkahler:
To @ruesvs point, can some of these be squashed and just make a multi-line commit message? One summary line, a blank and then the rest. Something like:

CI configuration cleanups

Remove apveyot.yml.  Travis-remane debian stuff.  Next thing, etc...

If something is distinctly different go ahead an make a separate commit. Just try to reduce excess.

And Thanks! this is comming together quickly.

ppd:
Yes, sure I'll squash them in the end. This is just a RFC. Multiple commits are nicer for review, I think.

ruevs:
Yeah I'd leave just "remove appveyor.yml" separate. But after all it's just a short flurry of activity... let it be.

.travis.yml
```diff
@@ -3,6 +3,10 @@ dist: xenial
 language: c
 git:
   submodules: false
+stages:
+  - test
+  - name: deploy
+    if: (NOT type IN (pull_request)) AND (branch = master)
```
vespakoen, line 9: Aren't pull requests never on master anyways?

vespakoen:
Thanks for this!

Is every commit to master released to snap?
If that's the case, we might have to figure out a way to release macOS as well with a build number (amount of commits?) or something?
Right now I am using `if: tag IS present` on the deploy stage for macOS

.travis.yml
ppd: In the case of a PR, branch equals the base branch (=master).

ppd:
Every commit to master is release to the `edge` channel. Tagged commits are release to `edge` and `beta`. As it is common in snap-land to publish nightlies or master builds to `edge`, we decided to do it like that as well when we introduced the snap deployment.

I don't have strong feelings about deploying master builds for other platforms, but I think @phkahler is not fond of the idea.

phkahler:
@ppd  I'm ok with nightly builds. It's just that we had people apparently using builds that were generated from running a PR through CI and had not been merged to master yet. It's no fun to triage an issue report and not being able to locate the hash they provided - good user, good PR complete with commit hash but it didn't exist.

I'm ok with stuff that has actually been merged leaving an executable somewhere for people to try. It's nice to be able to say "oh that should be fixed, try the latest" to people who don't build their own. It just needs to be based on master so we can figure out what version a but report actually happened with.

ppd:
@phkahler The thing with GitHub releases is that you need a tag. We could do a "rolling tag" like `edge` and force replace it before the release. Then the new release should replace the old one.

vespakoen:
I am experimenting with it over here: https://github.com/vespakoen/solvespace/blob/master/.travis.yml and it seems to work.

I am waiting for a build to finish to see if both the Windows and macOS build get added to the same release.

ppd:
But you're creating a new tag for every build, correct? That would absolutely flood the repo and the releases page. I was thinking more of a simple `edge` tag that gets updated to point to current HEAD in master.

phkahler:
To clarify some goals ;-)
1) Pull requests should get a debug build with sanitizers and run the test suite.
2) PRs merged to master can be used for "nightly" builds. These should replace the older one and not accumulate. Ideally built as what? Maybe release with debug info? We shouldn't have to run them through CI if everything goes through a PR first, but if we bypass that and brake something it will affect the next PR, so it will be discovered if a little late. I'm ok with running merges to master through CI, just saying I don't think it's a necessity.
3) Tagged releases should end up on the download page as full release builds. Question: Should these be built with debug info? Have they been before?

Again, even if the changes are a little chaotic it would not be happening nearly as quick if left to me. Thanks!

ppd:
>     1. Pull requests should get a debug build with sanitizers and run the test suite.

Obviously. Test stage yes, deploy stage no.

>     2. PRs merged to master can be used for "nightly" builds. These should replace the older one and not accumulate. Ideally built as what? Maybe release with debug info? 

There should be a floating tag, with only the current builds attached as a release. RelWithDebInfo sounds fine.

> We shouldn't have to run them through CI if everything goes through a PR first

It really can't hurt though :). Ten minutes extra for peace of mind.

>     3. Tagged releases should end up on the download page as full release builds. Question: Should these be built with debug info? Have they been before?

I quote whitequark:

```shell
if echo $TRAVIS_TAG | grep ^v; then BUILD_TYPE=RelWithDebInfo; else BUILD_TYPE=Debug; fi
```
So `v[...]` tagged releases were meant to be RelWithDebInfo, everything else Debug. On the other hand, Release makes a lot of sense if we offer those as end-user oriented releases.

vespakoen:
If I understood the above correctly, this should have everything we want: #764 764

ruevs:
At least on MSVC the EXE of "RelWithDebInfo" is almost the same as "Release" - the difference is that .pdb files are generated and /Ob1 (only expand `inline` functions) vs. /Ob2 (inline expand all suitable functions).
So I think for Windows RelWithDebInfo could be fine even for releases, but then again - anyone qualified to utilize the pdb is probably qualified to compile and debug...

So after all for "official releases" on Windows I would vote to use:
- "Release"
- ENABLE_LTO for sure
- OpenMP - the extra dll discussion... I am split - I love the single EXE, the extra speed is nice...

vespakoen:
What about providing both versions for Windows? People that want to go the extra mile for more speed will have the option to do so, people that want to grab solvespace to quickly draw something up can do so too.

@ruevs I agree that `Release` is probably the better build type for Windows in that case.

phkahler:
> What about providing both versions for Windows? People that want to go the extra mile for more speed will have the option to do so, people that want to grab solvespace to quickly draw something up can do so too.

I think that's a good idea.

ppd:
This has been superseded by @vespakoen's work at #766.

Issue #759, opened 2020-10-21
Performance Regression


phkahler:
### System information

For SovleSpace version see the bad commit listed below.
Operating system: Fedora 32

### Expected behavior

Performance should be similar to before

### Actual behavior

After commit:  https://github.com/solvespace/solvespace/commit/ab10e38d44b26de3ee64b60db0d154278d1ee8ac

The performance test I've been using went from 15 seconds to 193 with a Release + OMP build.

### Additional information

I'm going to try to make the additional improvements suggested in the fix rather than revert it.

For reference I take the attached file and change the number of repeats in the last group from 4 to 40. Wait for a bit and the Generate::DIRTY time is printed in the terminal. BTW it has always given a few bad edges.
[perf_test.zip](https://github.com/solvespace/solvespace/files/5412461/perf_test.zip)

ruevs:
Indeed. After https://github.com/solvespace/solvespace/pull/737 https://github.com/solvespace/solvespace/commit/ab10e38d44b26de3ee64b60db0d154278d1ee8ac the time to switch from "g005" to "g006" increases from 12s. to 45s. on a release build with ENABLE_LTO, on this "little model" [PrismConeNURBSNormalsTangents300.zip](https://github.com/solvespace/solvespace/files/5414683/PrismConeNURBSNormalsTangents300.zip) I've been using (don't even try it in debug mode ;-) ).

chord tolerance 0.5
max segments 40

On the other hand your model (as is) is pretty fast. I presume you increase the "repeat" times to much more than 4 in the rotate group (g007)? With 20 repeats switching from g006 to g007 takes 8s. and 40s. respectively.

phkahler:
> I presume you increase the "repeat" times to much more than 4 in the rotate group (g007)?

Yes. My test is to change it from 4 to 40.  Without the "fix" 15.5 seconds for me. With the fix 193s. Not sure why I never added OMP to SShell::CopyCurvesSplitAgainst() but that's easy and drops the version before the regression to 10 seconds.

I have not mentioned LTO, but that takes the original 15s down to 11, so combining that with OMP in that function should get it down around 7. Not bad for something that started around 2 minutes. But the regression.... I can't justify that performance penalty for a so-so NURBS fix and I really want the fix, so today I will be busy ;-)

PR #762, opened 2020-10-21
Performance optimization of the Vector class

ruevs:
Profiling with MSVC 2019 showed that many of the Vector methods are on a critical path (not surprising). They are changed to be inline and unnecessary temporaries are removed.

On the example below generate times decreased from 102s. to 64s.
At the same time the executable size shrank from 5569536 to 5150208 bytes in release mode (with global optimizations).

This should not stop us from working on optimizing inner loops e.g. https://github.com/solvespace/solvespace/issues/759 .

[Test model](https://github.com/solvespace/solvespace/files/5414683/PrismConeNURBSNormalsTangents300.zip)

Profiling results:
[20201021_Profiling_Before.txt](https://github.com/solvespace/solvespace/files/5416268/20201021_Profiling_Before.txt)
[20201021_ProfilingAfterVectorInline1.txt](https://github.com/solvespace/solvespace/files/5416269/20201021_ProfilingAfterVectorInline1.txt)
[20201021_ProfilingAfterVectorInline2.txt](https://github.com/solvespace/solvespace/files/5416270/20201021_ProfilingAfterVectorInline2.txt)
[20201021_ProfilingAfterVectorInline3.txt](https://github.com/solvespace/solvespace/files/5416267/20201021_ProfilingAfterVectorInline3.txt)

### DIFF
--- a/src/dsc.h
+++ b/src/dsc.h
@@ -155,6 +155,52 @@ inline bool Vector::Equals(Vector v, double tol) const {
     return dv.MagSquared() < tol*tol;
 }
 
+inline Vector Vector::From(double x, double y, double z) {
+    return {x, y, z};
+}
+
+inline Vector Vector::Plus(Vector b) const {
+    return {x + b.x, y + b.y, z + b.z};
+}
+
+inline Vector Vector::Minus(Vector b) const {
+    return {x - b.x, y - b.y, z - b.z};
+}
+
+inline Vector Vector::Negated() const {
+    return {-x, -y, -z};
+}
+
+inline Vector Vector::Cross(Vector b) const {
+    return {-(z * b.y) + (y * b.z), (z * b.x) - (x * b.z), -(y * b.x) + (x * b.y)};
+}
+
+inline double Vector::Dot(Vector b) const {
+    return (x * b.x + y * b.y + z * b.z);
+}
+
+inline double Vector::MagSquared() const {
+    return x * x + y * y + z * z;
+}
+
+inline double Vector::Magnitude() const {
+    return sqrt(x * x + y * y + z * z);
+}
+
+inline Vector Vector::ScaledBy(const double v) const {
+    return {x * v, y * v, z * v};
+}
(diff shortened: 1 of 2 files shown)


ruevs:
Very interesting. GCC complained about more of the `inline` functions than MSVC...
XCode on the other hand was fine.

I would have tried with MinGW GCC but the "mimalloc" library we started using breaks both GCC and clang on windows since it uses bcrypt.h> <winternl.h> header that is included only in the Microsoft SDK :-(

I'll move the ones the Travis build complained about in the header. Tomorrow...

phkahler:
Also, please don't remove  Negated or MakeMaxMin   I was actually looking for MakeMaxMin earlier but didn't notice it. I'll be using it soon.

Also, I'm not familiar with that syntax:    return { a,b,c }    has that been possible for a long time or more recent? Either way I think I like it.

phkahler:
Never mind, I see you moved them to the header file.

ruevs:
> Also, I'm not familiar with that syntax: return { a,b,c } has that been possible for a long time or more recent? Either way I think I like it.

It's C++11 - returning list initialized objects directly is possible:
https://en.cppreference.com/w/cpp/language/list%20initialization
>8) in a return statement with braced-init-list used as the return expression and list-initialization initializes the returned object

I checked the generated assembler and is makes it more compact (event with full optimizations) by not asking the compiler to create a local variable on the stack. When inlined I hope it helps even more.

phkahler:
@ruevs  this has little effect on that sketch I've been using as a benchmark (which is down to 8s on LTO master) but it does have a very perceptible improvement in the smoothness of dragging shapes with compound curvature like resizing a torus or dragging control points of the curved surface from issue 315. Different sketches, different bottlenecks. Nice improvement here!

ruevs:
The credit goes entirely to MSVC 2019 which makes it extremely easy to profile. These were just the "low hanging fruit" :-)

Issue #452, opened 2019-07-29
Exported STEP files do not include the colours


nabijaczleweli:
### System information

SolveSpace version: 3.0~5df53fc5

Operating system: Windows 10

### Expected behavior

Models exported into STEP, then imported into KiCad should retain the original colouring specified in the model.

### Actual behavior

Having disabled forcing surfaces to mesh and exported [this file](https://github.com/nabijaczleweli/kicad-lib/blob/de913c25aa840270c0a87a048928c60ae74da781/models/SOT110-1/assembly.slvs) into the STEP format, yielding [this file](https://github.com/nabijaczleweli/kicad-lib/blob/de913c25aa840270c0a87a048928c60ae74da781/models/SOT110-1/assembly.step), and pointing Pcbnew at it yielded the following 3D render:

![render with single-tone grey model](https://pbs.twimg.com/media/EApEf58XUAUza-u.png:large)

### Additional information

I don't know if this is a Solvespace bug, a KiCad bug, or, indeed, a bug at all, because I don't understand the ecosystem, so I'm opening this issue after being [nudged](https://twitter.com/app4soft/status/1155944830951800832) to, but apologies in advance if this is out of scope.

whitequark:
I don't think the current STEP exporter exports colors at all.

nabijaczleweli:
So, FWIW, I tried to add an implementation p.a. to some STEP files I found in KiCad and stole off the internet, and landed at the following diff, which by all accounts I could find and understand *should* work, but Pcbnew's 3D viewer segfaults when it tries to load it and ABViewer 14 shows naught.

```diff
diff --git a/src/exportstep.cpp b/src/exportstep.cpp
index 970c925..26ffe72 100644
--- a/src/exportstep.cpp
+++ b/src/exportstep.cpp
@@ -278,6 +278,39 @@ void StepFileWriter::ExportSurface(SSurface *ss, SBezierList *sbl) {
 
         id++;
         listOfLoops.Clear();
+
+        fprintf(f, "#%d=PRESENTATION_LAYER_ASSIGNMENT('','',(#%d));\n", id, id + 2);
+        id++;
+        fprintf(f, "#%d=MECHANICAL_DESIGN_GEOMETRIC_PRESENTATION_REPRESENTATION('',(#%d),#168);\n", id, id + 1);
+        id++;
+        fprintf(f, "#%d=STYLED_ITEM('',(#%d),#%d);\n", id, id + 1, advFaceId);
+        id++;
[...]

PR #763, opened 2020-10-21
STEP Export: include colors and alpha

ruevs:
The implementation may be sub-optimal, since the colour and alpha is defined for each NURBS surface instead of on group level, but the STEP export currently does not represent group structure at all and I am not familiar with the format in order to change this.

Fixes: https://github.com/solvespace/solvespace/issues/452

### DIFF
--- a/src/exportstep.cpp
+++ b/src/exportstep.cpp
@@ -273,9 +273,36 @@ void StepFileWriter::ExportSurface(SSurface *ss, SBezierList *sbl) {
         }
 
         fprintf(f, "),#%d,.T.);\n", srfid);
-        fprintf(f, "\n");
         advancedFaces.Add(&advFaceId);
 
+        // Export the surface color and transparency
+        // https://www.cax-if.org/documents/rec_prac_styling_org_v16.pdf sections 4.4.2 4.2.4 etc.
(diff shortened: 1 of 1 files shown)


phkahler:
@ruevs Have you tried importing the step file into anything to verify this works? I spent a lot of time trying to get colors in Step and failed (see the mess on my color branch). There are many ways to represent things in Step so maybe this simpler way is actually better.

In particular, can the colors be read by KiCad? That was the goal and would be awesome!

ruevs:
Yes. The screen shots are from this program
https://viewer.kisters.de/en/products/3dviewstation-desktop-version.html
which seems quite professional.

ruevs:
@phkahler  I'll look at your branch and try to figure out what is wrong with it. Now that I (kind of) know what works ;-)

phkahler:
> @phkahler I'll look at your branch and try to figure out what is wrong with it. Now that I (kind of) know what works ;-)

I rearanged a LOT just to make things more like FreeCAD output.

I'll merge this. I think I can find someone to verify our files are still acceptable to a Parasolid based app and maybe the autodesk kernel (forget the name).

Thanks!

PR #764, opened 2020-10-21
Travis: release macOS and Windows edge builds

vespakoen:
This will release every commit on the master branch to the "edge" tag, it also:

- Fixes windows build type not being picked up
- Revert `cleanup: false` back to `skip_cleanup: true` (to fix snap releases)
- Waits 10 seconds to get notarize_uuid for macOS since it would sometimes fail
- Build test stages in `Debug` mode and deploy stages in `RelWithDebInfo` mode

### DIFF
--- a/.travis/build-windows.sh
+++ b/.travis/build-windows.sh
@@ -3,14 +3,23 @@
 MSBUILD_PATH="c:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\MSBuild\15.0\Bin"
 export PATH=$MSBUILD_PATH:$PATH
 
-if echo $TRAVIS_TAG | grep ^v; then BUILD_TYPE=RelWithDebInfo; else BUILD_TYPE=Debug; fi
-
 mkdir build
 cd build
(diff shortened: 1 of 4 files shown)


.travis/build-windows.sh
```diff
@@ -3,14 +3,23 @@
...
 cd build
-cmake .. -G "Visual Studio 15 2017 Win64" -DCMAKE_BUILD_TYPE=$BUILD_TYPE
 
-MSBuild.exe "src/solvespace.vcxproj" -maxcpucount
-MSBuild.exe "src/solvespace-cli.vcxproj" -maxcpucount
-MSBuild.exe "test/solvespace-testsuite.vcxproj" -maxcpucount
+if [ "$1" == "release" ]; then
+    BUILD_TYPE=RelWithDebInfo
+    cmake \
+        -G "Visual Studio 15 2017 Win64" \
```
ruevs, line 12: I think we should be building x86 (32bit) executables for Windows.
SolveSpace will become unusably slow long before consuming 2GB of RAM, which is the limit for 32bit Win32 processes.

PR #765, opened 2020-10-21
Vertex split

phkahler:
Two commits. One fixes the huge performance regression and is only a little slower than prior to the regression. 18.5 seconds compared to 193 seconds. Prior to the NURBS fix that same test took a bit over 15 seconds. That seems reasonable for adding more code in order to fix an issue.  The second commit here adds OpenMP parallelism to the same part of the code and bring the test down to 11.x seconds.

### DIFF
--- a/src/srf/boolean.cpp
+++ b/src/srf/boolean.cpp
@@ -20,18 +20,36 @@ void SShell::MakeFromIntersectionOf(SShell *a, SShell *b) {
     MakeFromBoolean(a, b, SSurface::CombineAs::INTERSECTION);
 }
 
-// We will be inserting existing verticies into curves to split them
-// todo: this is only using the ends of exact curves, and it is only
-// using them to split existing curves, not new intersections.
-// It resolves some issues but we could do better. We will need to
-// reorder things so the surface intersection curves exist prior to
-// splitting any curves at all in order to have their verticies too.
-// Better still would be to get curve/surface intersection to work
-// more reliably at the edges - maybe do curve/curve tests as part
-// of the curve-surface intersection test.
+void SCurve::GetAxisAlignedBounding(Vector *ptMax, Vector *ptMin) const {
+    *ptMax = {VERY_NEGATIVE, VERY_NEGATIVE, VERY_NEGATIVE};
+    *ptMin = {VERY_POSITIVE, VERY_POSITIVE, VERY_POSITIVE};
+
+    for(int i = 0; i <= exact.deg; i++) {
+        exact.ctrl[i].MakeMaxMin(ptMax, ptMin);
+    }
+}
+
+// We will be inserting other curve verticies into our curves to split them.
+// This is helpful when curved surfaces become tangent along a trim and the
+// usual tests for curve-surface intersection don't split the curve at a vertex.
+// This is faster than the previous version that split at surface corners and
+// handles more buggy cases. It's not clear this is the best way but it works ok.
 static void FindVertsOnCurve(List<SInter> *l, const SCurve *curve, SShell *sh) {
+
+    Vector amax, amin;
+    curve->GetAxisAlignedBounding(&amax, &amin);
+
     for(auto sc : sh->curve) {
         if(!sc.isExact) continue;
+        
+        Vector cmax, cmin;
+        sc.GetAxisAlignedBounding(&cmax, &cmin);
+
+        if(Vector::BoundingBoxesDisjoint(amax, amin, cmax, cmin)) {
+            // They cannot possibly intersect, no curves to generate
+            continue;
+        }
+        
         for(int i=0; i<2; i++) {
             Vector pt = sc.exact.ctrl[ i==0 ? 0 : sc.exact.deg ];
             double t;
@@ -63,11 +81,13 @@ SCurve SCurve::MakeCopySplitAgainst(SShell *agnstA, SShell *agnstB,
 
     // First find any vertex that lies on our curve.
     List<SInter> vertpts = {};
-    if(agnstA)
-        FindVertsOnCurve(&vertpts, this, agnstA);
-    if(agnstB)
-        FindVertsOnCurve(&vertpts, this, agnstB);
-
+    if(isExact) {
+        if(agnstA)
+            FindVertsOnCurve(&vertpts, this, agnstA);
+        if(agnstB)
+            FindVertsOnCurve(&vertpts, this, agnstB);
+    }
+    
     const SCurvePt *p = pts.First();
     ssassert(p != NULL, "Cannot split an empty curve");
     SCurvePt prev = *p;
--- a/src/srf/surface.h
+++ b/src/srf/surface.h
@@ -219,6 +219,7 @@ public:
     SSurface *GetSurfaceB(SShell *a, SShell *b) const;
 
     void Clear();
+    void GetAxisAlignedBounding(Vector *ptMax, Vector *ptMin) const;
 };
 
 // A segment of a curve by which a surface is trimmed: indicates which curve,


phkahler:
Going through all of my NURBS regression tests, this fails with a file named wrong2.slvs

I'll try more variations. Going over every curve seemed to be the problem. Still appreciate feedback, this is looking kind of ulgy IMHO but still somewhat contained.

src/srf/boolean.cpp
```diff
@@ -20,20 +20,23 @@ void SShell::MakeFromIntersectionOf(SShell *a, SShell *b) {
...
-            Vector pt = sc.exact.ctrl[ i==0 ? 0 : sc.exact.deg ];
+// We will be inserting existing surface verticies into curves to split them.
+// Using surface verts is faster than curve verts. Neither may be optimal but
+// doing this resolves some issues where curves are not split.
+static void FindVertsOnCurve(List<SInter> *l, const SCurve *curve, SShell *sh,
+                            Vector amin, Vector amax) {
+
+    for(auto srf : sh->surface) {
+        Vector bmax, bmin;
+        srf.GetAxisAlignedBounding(&bmax, &bmin);
```
ruevs, line 31: SolveSpace spends most of it's time in these two anyway ;-) Just looked at them today.

src/srf/boolean.cpp
```diff
@@ -63,11 +66,24 @@ SCurve SCurve::MakeCopySplitAgainst(SShell *agnstA, SShell *agnstB,
...
-        FindVertsOnCurve(&vertpts, this, agnstA);
-    if(agnstB)
-        FindVertsOnCurve(&vertpts, this, agnstB);
-
+    if(isExact) {
+        // get a bounding box for our curve
+        Vector cmax = Vector::From(-1e10, -1e10, -1e10);
+        Vector cmin = Vector::From(1e10, 1e10, 1e10);
+        for(int i=0; i<=exact.deg; i++) {
+            cmax.x = max(cmax.x, exact.ctrl[i].x);
```
ruevs, line 74: Just today I looked at very similar code somewhere else... There is function to do this GetMaxMin...something I inlined it...

src/srf/boolean.cpp
```diff
@@ -63,11 +66,24 @@ SCurve SCurve::MakeCopySplitAgainst(SShell *agnstA, SShell *agnstB,
...
     // First find any vertex that lies on our curve.
     List<SInter> vertpts = {};
-    if(agnstA)
-        FindVertsOnCurve(&vertpts, this, agnstA);
-    if(agnstB)
-        FindVertsOnCurve(&vertpts, this, agnstB);
-
+    if(isExact) {
+        // get a bounding box for our curve
+        Vector cmax = Vector::From(-1e10, -1e10, -1e10);
```
ruevs, line 71: There are defines for "very big" and "very small" don't use literals in the code apart from 0 and 1

src/srf/boolean.cpp
```diff
@@ -20,20 +20,23 @@ void SShell::MakeFromIntersectionOf(SShell *a, SShell *b) {
...
+// Using surface verts is faster than curve verts. Neither may be optimal but
+// doing this resolves some issues where curves are not split.
+static void FindVertsOnCurve(List<SInter> *l, const SCurve *curve, SShell *sh,
+                            Vector amin, Vector amax) {
+
+    for(auto srf : sh->surface) {
+        Vector bmax, bmin;
+        srf.GetAxisAlignedBounding(&bmax, &bmin);
+
+        if(Vector::BoundingBoxesDisjoint(amax, amin, bmax, bmin)) {
```
ruevs, line 33: Yes, this should bring the profiling to "before https://github.com/solvespace/solvespace/commit/ab10e38d44b26de3ee64b60db0d154278d1ee8ac state" where most of the time was spent on bounding boxes instead of on other vector:: methods.

src/srf/boolean.cpp
```diff
@@ -20,20 +20,23 @@ void SShell::MakeFromIntersectionOf(SShell *a, SShell *b) {
     MakeFromBoolean(a, b, SSurface::CombineAs::INTERSECTION);
 }
 
-// We will be inserting existing verticies into curves to split them
-// todo: this is only using the ends of exact curves, and it is only
-// using them to split existing curves, not new intersections.
-// It resolves some issues but we could do better. We will need to
-// reorder things so the surface intersection curves exist prior to
-// splitting any curves at all in order to have their verticies too.
-// Better still would be to get curve/surface intersection to work
```
ruevs, line 29: Leave some more of the explanation in. Otherwise it will get buried in the history and it is important.

ruevs (changes requested):

ruevs:
I need to debug this... not enough stack in my brain to think about it "on paper" (the phone screen) :-)

phkahler:
@ruevs  If you look at a simplified test using triangle mesh:
unmerged
I think what's happening but not certain:  The two surfaces from the colinear sketch lines get merged into one surface but I don't think the lines get merged, so it has 6 trim curves. Then when it does union with the large box, the long edge is split at the perpendicular surfaces of the small box, but not at the point where the 2 shorter trim curves meet, so then it fails to assemble everything correctly. Splitting a line at every curves end vertex caused the additional split, but since the 2 coincident rectangles merged into one, splitting at that surface corners is not enough. have not confirmed but it seems plausible.

I suspect going over all the curves was killing cache performance in addition to that being a complex test (nearest point to curve). Every curve was being checked against every curve and then every surface in the other shell. Maybe I'll try this surface bounding box test and then check the ends of that surface trim curves instead of the surface corners.

phkahler:
Updated with MakeMaxMin, changed the comment, pulled the OMP out into a different PR.

src/srf/boolean.cpp
```diff
@@ -20,20 +20,25 @@ void SShell::MakeFromIntersectionOf(SShell *a, SShell *b) {
...
+        Vector bmax, bmin;
+        srf.GetAxisAlignedBounding(&bmax, &bmin);
+
+        if(Vector::BoundingBoxesDisjoint(amax, amin, bmax, bmin)) {
+            // They cannot possibly intersect, no curves to generate
+            continue;
+        }
+        
+        for(int i=0; i<4; i++) {
+            Vector pt = srf.ctrl[ (i&1)==0 ? 0 : srf.degm ][ (i&2)==0 ? 0 : srf.degn];
```
ruevs, line 41: So you are only checking whether the four corners of the surface lie on the curve. Not all control points.
This sounds logical for what you are trying to do, but the function should be called "FindSrfCornersOnCurve" or something.
And the comment above should be explicit about it.

src/srf/boolean.cpp
```diff
@@ -20,20 +20,25 @@ void SShell::MakeFromIntersectionOf(SShell *a, SShell *b) {
...
-static void FindVertsOnCurve(List<SInter> *l, const SCurve *curve, SShell *sh) {
-    for(auto sc : sh->curve) {
-        if(!sc.isExact) continue;
-        for(int i=0; i<2; i++) {
-            Vector pt = sc.exact.ctrl[ i==0 ? 0 : sc.exact.deg ];
+// We will be inserting existing surface verticies into curves to split them.
+// This is helpful when curved surfaces become tangent along a trim and the
+// usual tests for curve-surface intersection don't split spllit the curve.
+// Another option might be to split at the trim curve verticies which can
+// handle more a few more cases. It's not clear if either is the best approach.
```
ruevs, line 27: Typos
"split spllit"
"handle more a few more cases."

ruevs (changes requested):
OK now I understand thoroughly what you are doing. And after you visual explanation above I think you are doing what you intend to with this code. And now I understand why it helps.

So apart from the few comments in this review I think it is fine to merge this.

Another important point I want to make. Jonathan's logic that, if this is needed, then surface curve intersection is not doing it's job still stands. I tried to dig a little, but did not get far. However I did look at this:

https://github.com/solvespace/solvespace/blob/0761339ec96a5c5e8936c2cc8fad8e04b757062d/src/srf/surface.cpp#L203

And I think somewhere we may have a bug in an equivalent (hypothetical) "CurveEntirelyOutsideBbox" logic. But I did not find it and am out of time for now.

phkahler:
@ruevs  I changed it back to using curve end verticies. It's faster and fixes wrong2.slvs as well. Turns out the performance problem was solved by using bounding boxes prior to doing math like ClosestPointTo().

Should it stay two commits or should I squash them?

ruevs:
@phkahler I apologize for the impertinence but I pushed a small cosmetic improvement to the code. NFC.

The performance is great.

- [failed.lathe.slvs.zip](https://github.com/solvespace/solvespace/files/408873/failed.lathe.slvs.zip) works
- [adapter-plate.tar.gz](https://github.com/solvespace/solvespace/files/4054379/adapter-plate.tar.gz) works
- [wrong.zip](https://github.com/solvespace/solvespace/files/5112393/wrong.zip) still broken when you shorten the big extrusion - as before
- [wrong2.zip](https://github.com/solvespace/solvespace/files/5145107/wrong2.zip) works
- [coincident4.zip](https://github.com/solvespace/solvespace/files/5357050/coincident4.zip) works

phkahler:
> @phkahler I apologize for the impertinence but I pushed a small cosmetic improvement to the code.

No problem at all. BTW how do you push changes to a PR like that?

Going to merge now ;-)

ruevs:
I think that if you check the box "allow maintainers to change" or something when creating the pull request then it gives rights to them to push to you branch.

So I have a remote for you:
phkahler        https://github.com/phkahler/solvespace

So:
`git fetch --all`
`git checkout phkahler vertex-split`
    change stuff
`git add.... whatever`
`git commit --amend`
`git push --force`

PR #766, opened 2020-10-22
Travis: fix release notes, add Windows OpenMP

vespakoen:
We almost have edge releases for macOS and Windows!
Unfortunately, using the commit message for the `body:` parameter in the release breaks on newlines.
This PR uses the "edge" release module that fixes that with the `release_notes:` property [0].

I also added an extra windows deploy step for building without OpenMP.

[0] See https://github.com/travis-ci/dpl/pull/1069


ppd:
Haven't looked at it in detail, but: does this work on your fork? I.e. does it create an `edge` tag and upload the release to it?
I see no pushing of the tags after `git tag $TRAVIS_TAG`. Does the Travis deploy provider do that automatically?

.travis.yml
```diff
@@ -45,12 +46,12 @@ jobs:
       install: .travis/install-debian.sh
       script: .travis/build-debian.sh
     - stage: test
-      name: "Windows Visual Studio 2017 test"
+      name: "Windows test"
```
ppd, line 49: Why do we annotate jobs now with their stages? That's a little redundant, isn't it?

vespakoen:
Yes it works on my fork. The tag is created by the deploy provider indeed.

I was thinking of adding your changed from #757 in this PR as well, since it might be annoying to merge it otherwise.

And I also plan to add the cmake_generator_platform x86 flag to this PR

.travis.yml
```diff
@@ -64,14 +65,37 @@ jobs:
...
+    - stage: deploy
+      name: "Windows deploy with OpenMP"
+      os: windows
+      install: .travis/install-windows.sh
+      script: .travis/build-windows.sh release openmp
+      before_deploy:
+        - git config --local user.name "solvespace-cd"
+        - git config --local user.email "no-reply@solvespace.com"
+        - export TRAVIS_TAG=${TRAVIS_TAG:-edge}
+        - git tag $TRAVIS_TAG
```
ppd, line 85: Do we not need to push this? I can't see an `edge` tag on your fork. Have you tested this?

.travis/build-windows.sh
```diff
@@ -7,11 +7,16 @@ mkdir build
 cd build
 
 if [ "$1" == "release" ]; then
+    if [ "$2" == "openmp" ]; then
```
ppd, line 10: Isn't == a bashism? sh is bash on Travis, so it doesn't really matter.

.travis.yml
vespakoen: True, I updated it, will force push it on here

.travis.yml
vespakoen: Travis pushes the tag automatically, this piece of code is straight from their documentation

.travis/build-windows.sh
vespakoen: Updated it to sh syntax, and explicitly use bash in sign-macos.sh, see latest push

ppd:
> I was thinking of adding your changes from #757 in this PR as well, since it might be annoying to merge it otherwise.

You can cherry-pick the arm64 one and maybe the Debian → Ubuntu thingy.

Re the `(NOT type IN (pull_request)) AND (branch = master)``part: that only applies to the macOS deploy job right now. I think we wanted to skip deploy for everything not master. So you could pick that too.

vespakoen:
See the latest force push, it adds everything from your PR, it looks like you had a cleaner way to add the condition `(NOT type IN (pull_request)) AND (branch = master)` in one place only though. I will add that as well.

ppd:
While we're at it: isn't it advisable to run tests with OpenMP too? @phkahler

vespakoen:
I also think it's probably a good idea to build something closer to the "real thing" during test stage.

Only some issue building for Win32 now left...

.travis/build-windows.sh
```diff
@@ -6,20 +6,29 @@ export PATH=$MSBUILD_PATH:$PATH
...
+        -G "Visual Studio 15 2017" \
+        -DCMAKE_BUILD_TYPE="${BUILD_TYPE}" \
+        -DCMAKE_GENERATOR_PLATFORM="Win32" ..
 fi
 
 cmake --build . --config "${BUILD_TYPE}" -- -maxcpucount
 
 bin/$BUILD_TYPE/solvespace-testsuite.exe
+
+mv bin/$BUILD_TYPE/solvespace.exe bin/$BUILD_TYPE/solvespace-openmp.exe
```
ruevs, line 34: Should this be done only `if [ "$2" = "openmp" ];` ?

vespakoen: Good catch!

phkahler:
> While we're at it: isn't it advisable to run tests with OpenMP too? @phkahler

Absolutely. One of the OMP changes worked fine on my machine but had memory leaks caught by ASAN.

vespakoen:
I enabled OpenMP during test.

Feedback is appreciated!

ppd:
Does `ENABLE_SANITIZERS=ON` make sense for macOS and Windows tests?

vespakoen:
`ENABLE_SANITIZERS` was only enabled in `build-ubuntu.sh`, I have disabled it there.
I also got rid of the `allow_failures`.

ppd:
Re `ENABLE_SANITIZERS`: https://github.com/solvespace/solvespace/pull/757#issuecomment-713684583

vespakoen:
Ah, so it's the other way around, will update...

ppd:
> CMake Error at CMakeLists.txt:181 (message):
> 
>   Sanitizers are only available when using GCC or Clang

I think that answers it for the Windows build

vespakoen:
Disabled them for windows...

.travis.yml
```diff
@@ -64,14 +65,37 @@ jobs:
...
+        release_notes: $TRAVIS_COMMIT_MESSAGE
         file: build/bin/RelWithDebInfo/solvespace.exe
         on:
           repo: solvespace/solvespace
+    - stage: deploy
+      name: "Windows with OpenMP"
+      os: windows
+      install: .travis/install-windows.sh
+      script: .travis/build-windows.sh release openmp
+      before_deploy:
```
ppd, line 81: Should we de-duplicate this before_deploy script and put it in a separate file? Or as a YAML anchor?

vespakoen: I think a bit of duplication is not a big deal, I for one have trouble understanding the YAML anchor from Snap.

vespakoen:
Not sure why the macOS build failed, something with this line: https://github.com/solvespace/cairo/blob/master/src/cairo-image-compositor.c#L133

I restarted it to see if it happens again...
I thought the ENABLE_SANITIZERS is only adding more warnings and doesn't influence the compilation?

ppd:
IIRC clang sanitizers do what they do at runtime and just exit with non-zero exit code.

vespakoen:
I see, so I guess we are getting our first "fruits" from our labour, it found a potential issue =)

ppd:
```
/Users/travis/build/solvespace/solvespace/extlib/cairo/src/cairo-image-compositor.c:133:34: runtime error: left shift of 255 by 24 places cannot be represented in type 'int'
```
Yes. That's UBSan in action.

vespakoen:
The latest master of cairo still has the same code in there, is the problem in the way we use cairo perhaps?

vespakoen:
```
# set(SANITIZE_FLAGS "${SANITIZE_FLAGS} -fno-sanitize-recover=address,undefined")
set(SANITIZE_FLAGS "${SANITIZE_FLAGS} -fsanitize=address,undefined")
```

This should make it continue hopefully, so we can deal with those issues on our time, get builds out in the meantime, and re-enable the "crash" mode once we have solved the current issues. What do you think?

vespakoen:
That made it green again, and now it shows all of them:

```
/Users/travis/build/solvespace/solvespace/extlib/cairo/src/cairo-image-compositor.c:133:34: runtime error: left shift of 255 by 24 places cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/travis/build/solvespace/solvespace/extlib/cairo/src/cairo-image-compositor.c:133:34 in 
/Users/travis/build/solvespace/solvespace/extlib/cairo/src/cairo-fixed-private.h:64:14: runtime error: left shift of negative value -8388608
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/travis/build/solvespace/solvespace/extlib/cairo/src/cairo-fixed-private.h:64:14 in 
/Users/travis/build/solvespace/solvespace/extlib/cairo/src/cairo-fixed-private.h:64:14: runtime error: left shift of negative value -31
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/travis/build/solvespace/solvespace/extlib/cairo/src/cairo-fixed-private.h:64:14 in 
```

I think we are ready to merge this one?

ppd:
> The latest master of cairo still has the same code in there, is the problem in the way we use cairo perhaps?

Well, the intermediate result type of a short or char shifted by an int is an int, which is 32-bit here and obviously can't fit (with defined behaviour) 32 bit of unsigned data. So the complaint has some merit.

vespakoen:
I agree, but don't think its easy to set it up that way. The reality is, this issue was already in the codebase since a long time, it just surfaced right now.
Maybe we can create a separate issue for solving those errors and turning on the strict mode again, then we can at least have edge builds for macOS and Windows now and gather some feedback.

ppd:
Yes. We'd need to set flags for each target instead of globally (something like `target_compile_options()`). Let's leave the decision between warnings instead of exits or no sanitizers for macOS to @phkahler.

phkahler:
> Yes. We'd need to set flags for each target instead of globally (something like `target_compile_options()`). Let's leave the decision between warnings instead of exits or no sanitizers for macOS to @phkahler.

Ooof.  Is this just a macOS thing? Otherwise, any idea why it just showed up? IMHO you can skip UBSAN if it's just on macOS - I would hope that triggers the same on all platforms for code that is ours.

> Also: thanks a lot for your patience and efforts @vespakoen. It's easy to criticize, but often quite a bit harder to implement something; so don't take the nagging personally.

This is really important. Every small thing makes SolveSpace better, and hopefully us too ;-)  I've learned quite a bit working on this thing and I'm still rough around the edges. IMHO it's coming together really well. Thanks guys!

ppd:
> Ooof. Is this just a macOS thing

On Ubuntu we don't build cairo, and on Windows we don't use any sanitizers.

> IMHO you can skip UBSAN if it's just on macOS

As in: skip ubsan on macOS, but run it on Ubuntu? Or just don't bother with any sanitizers on macOS?

phkahler:
@ppd feel free to merge these once you feel they're ready.

phkahler:
> On Ubuntu we don't build cairo

I thought cairo was how solvespace-cli rendered images for the test suite. Am I mistaken?

ppd:
> I thought cairo was how solvespace-cli rendered images for the test suite. Am I mistaken?

We use it, but we only link to it. On other platforms we build it.

ppd:
@vespakoen I think you can merge this after that last round of CI. Maybe the commit message's first line could be a bit more generic. After all, this PR grew to squash quite some gremlins.

vespakoen:
Interesting... Looks like it found something in the C++ code now?

We are pretty much forced to use llvm v9 for macOS since the later ones crash during compilation (and it's the only thing where OpenMP seems to work, so we will have to look for another alternative I guess.

I think the "warning" mode we had before is probably the best, we just have to take action on the issues we find (fix or report them).

ppd:
Yes, but not ours ;). I'll test the suppression file for ubsan later and hope we can achieve a less "big hammer" solution. Otherwise setting ubsan to warn on macOS only is also possible.

vespakoen:
From UBSan: `Check groups (like undefined) can’t be used in suppressions file, only fine-grained checks are supported.`
I tried blacklisting the `vector` file only, but it cannot find it somehow...

I will use the warning mode for macOS.

PR #768, opened 2020-10-22
Add OpenMP parallel for to SShell::CopyCurvesSplitAgainst

phkahler:
Makes boolean operations a bit faster.

### DIFF
--- a/src/srf/boolean.cpp
+++ b/src/srf/boolean.cpp
@@ -181,16 +181,19 @@ SCurve SCurve::MakeCopySplitAgainst(SShell *agnstA, SShell *agnstB,
 }
 
 void SShell::CopyCurvesSplitAgainst(bool opA, SShell *agnst, SShell *into) {
-    SCurve *sc;
-    for(sc = curve.First(); sc; sc = curve.NextAfter(sc)) {
+#pragma omp parallel for
+    for(int i=0; i<curve.n; i++) {
+        SCurve *sc = &curve[i];
         SCurve scn = sc->MakeCopySplitAgainst(agnst, NULL,
                                 surface.FindById(sc->surfA),
                                 surface.FindById(sc->surfB));
         scn.source = opA ? SCurve::Source::A : SCurve::Source::B;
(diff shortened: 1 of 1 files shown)


ruevs:
Wasn't this the one that deformed surfaces due to race conditions some time ago? I don't remember, so if it's not just ignore me ;-)

src/srf/boolean.cpp
```diff
@@ -181,16 +181,19 @@ SCurve SCurve::MakeCopySplitAgainst(SShell *agnstA, SShell *agnstB,
 }
 
 void SShell::CopyCurvesSplitAgainst(bool opA, SShell *agnst, SShell *into) {
-    SCurve *sc;
-    for(sc = curve.First(); sc; sc = curve.NextAfter(sc)) {
+#pragma omp parallel for
+    for(int i=0; i<curve.n; i++) {
```
ruevs, line 185: 'for(auto sc = curve.First(); sc; sc = curve.NextAfter(sc)) {'
or
'for(SCurve sc = curve.First(); sc; sc = curve.NextAfter(sc)) {'
not possible?

ruevs: Hmm probably not I guess OpenMP won't be able to handle the iterators...

phkahler:
> Wasn't this the one that deformed surfaces due to race conditions some time ago?

Nope. That was caused by recursive calls to SSurface::SplitInHalf() from multiple threads. That function actually modified the weights of surface control points and then put them back when it was done. Now it makes a temporary copy of the surface, modifies it and doesn't bother putting it back before discarding.

Thanks for the reminder, it took me a few minutes to remember that one.

src/srf/boolean.cpp
phkahler: @ruevs  yeah, I think OMP version 5.0 allows some iterators, but MSVC is still stuck at version 2.0.  BTW that really bothers me because OMP is a great way to do the most trivial kind of parallel processing.

PR #771, opened 2020-10-23
CMake: use sanitizer flags for internal targets only

ppd:
Previously sanitizer flags were set unconditionally for all code, including that of external libraries.
Set them only for targets in src/, tests/ and exposed/.

Unfortunately, the linker equivalent to add_compile_options,
add_link_options, is only available for CMake version >= 3.13.
So add the sanitizer flags manually to each target's linker options.

Should help with #766


vespakoen:
Awesome! Looks good to me.

Let's merge this in first, then I'll bring #766 up to date so it can cleanly merge with master

Issue #782, opened 2020-10-26
Property Browser scroll bar returns to top


kk6mrp:
### System information

SolveSpace version: Edge 3.0~3694c9b3

Operating system: Windows 10

### Expected behavior

Scroll bar should stay in the position it was moved to.

### Actual behavior

Scroll bar and window returns to top of screen.

### Additional information

![property browser](https://user-images.githubusercontent.com/9358855/97201481-be830b80-176f-11eb-8f36-4f744650ef9a.gif)

ruevs:
It stopped working here:

f324477dd06c15022670a4486bf83be3abbf48b5 is the first bad commit
commit f324477dd06c15022670a4486bf83be3abbf48b5
Author: whitequark
Date:   Thu Jul 12 19:29:44 2018 +0000

    Implement a platform abstraction for windows.

ruevs:
@phkahler @ppd @vespakoen this is fixed here https://github.com/solvespace/solvespace/pull/783 please test on Linux and macOS since this should affect all platforms.

phkahler:
@ruevs  I understand the need to test the fix on all platforms, but why did you tag the issue with the non-windows platforms?

ruevs:
> @ruevs I understand the need to test the fix on all platforms, but why did you tag the issue with the non-windows platforms?

From reading the code it appears to me that the problem should have affected all platforms. If moving the scrollbar with the mouse works on Linux and macOS before https://github.com/solvespace/solvespace/pull/783 then for sure remove the tags.

kk6mrp:
This does fix the issue for me

PR #783, opened 2020-10-27
UI: Fix the Property Browser window scrollbar

ruevs:
- Scrolling with the scrollbar now works properly.

- Do not scroll with the mouse wheel while an edit field is active in the property browser.

Fixes: https://github.com/solvespace/solvespace/issues/782

### DIFF
--- a/src/textwin.cpp
+++ b/src/textwin.cpp
@@ -252,8 +252,8 @@ void TextWindow::Init() {
                     MouseLeave();
                     return true;
                 } else if(event.type == MouseEvent::Type::SCROLL_VERT) {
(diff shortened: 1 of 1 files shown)


ruevs:
@phkahler @ppd @vespakoen please test on Linux and macOS since this should affect all platforms.

phkahler:
Seems to have no negative effect on Linux.

ppd:
Fine on snap/Linux.

phkahler:
I'll just merge this. I don't build on macOS but I can vertify there when it gets to edge.

PR #784, opened 2020-10-27
Minor fixes - max undo-redo  and some defines

phkahler:


src/srf/surface.cpp
```diff
@@ -507,9 +507,9 @@ void SShell::MakeFromExtrusionOf(SBezierLoopSet *sbls, Vector t0, Vector t1, Rgb
     Vector n = sbls->normal.ScaledBy(-1);
     Vector u = n.Normal(0), v = n.Normal(1);
     Vector orig = sbls->point;
-    double umax = 1e-10, umin = 1e10;
+    double umax = VERY_NEGATIVE, umin = VERY_POSITIVE;
```
ruevs, line 510: Good catch! I wonder what this broke?

ruevs: Probably nothing? Aren't u and v usually 0-1?

ruevs (approved):

src/srf/surface.cpp
phkahler: @ruevs  That code is finding the bounding box of the sketch which is used to create the top and bottom surfaces of an extrusion. In some cases that error would result in the bounding box being larger than needed. The new faces are created based on the bounding box so that U ~ [0,1] and V ~ [0,1]. Having a larger bounding box might cause lower performance, but probably wouldn't affect anything else. It might affect chord tolerance which could end up triggering (or fixing) other issues that are sensitive to CT.

PR #786, opened 2020-10-28
Don't do numeric surface intersections if an exact curve has been cop…

phkahler:
…ied. We don't want 2 overlapping but different sets of PWLs.

### DIFF
--- a/src/srf/surfinter.cpp
+++ b/src/srf/surfinter.cpp
@@ -331,6 +331,7 @@ void SSurface::IntersectAgainst(SSurface *b, SShell *agnstA, SShell *agnstB,
                 sext = this;
                 shext = agnstA;
             }
+            bool foundExact = false;
             SCurve *sc;
             for(sc = shext->curve.First(); sc; sc = shext->curve.NextAfter(sc)) {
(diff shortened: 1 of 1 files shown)


src/srf/surfinter.cpp
```diff
@@ -341,8 +342,14 @@ void SSurface::IntersectAgainst(SSurface *b, SShell *agnstA, SShell *agnstB,
                 if(splane->ContainsPlaneCurve(sc)) {
                     SBezier bezier = sc->exact;
                     AddExactIntersectionCurve(&bezier, b, agnstA, agnstB, into);
+                    foundExact = true;
```
ruevs, line 345: Could we add more than one curve here?
If we find at least one then we skip PWL?
What if directly `return` from  here after adding the first one? This changes the current logic, but which is the correct one?

phkahler: I tried putting a return there and it fixes the immediate problem. Then I realized if for some reason there were more than one exact trim curve coincident with the plane we want them all, so i added the flag. This function needs to find all the intersections of two surfaces.

Issue #776, opened 2020-10-24
Snap Selection to Grid is broken


phkahler:
### System information

SolveSpace version: 3.0~70350715

Operating system: Fedora 32

### Expected behavior

Turn on the grid. Draw some line segments. Select a point. Constrain to grid.
The point should move to either a line or vertex on the grid (I don't know which was intended).

### Actual behavior

Nothing happens. The point remains off the grid.

phkahler:
To clarify a bit: 
With a single line segment the endpoints can be snapped to grid, but they are not constrained and can still be dragged. Is that expected?
With multiple segments, selecting one of the non-endpoints and snapping to grid does nothing.

Going back to an old version (old enough to not show its version number) the snapped points are still movable. The non-endpoints can also be snapped, so that part seems to be the only regression.

vespakoen:
When you also select a segment next to the point it also seems to work.
When selecting just a point (in the middle of a line) I can also see it snapping but then returning back to its original position (in a split second).

phkahler:
When there are 2 segments meeting at a point there are actually 2 points there with a coincident constraint added. It's probably moving one of them to the nearest grid point but not the other, and then the solver enforces the constraint and puts it back?  Wonder when that got broken.

ruevs:
Too late to do one more bisect today... tomorrow maybe :-)

ruevs:
The regression was introduced here:

25631d4fb214239a1b46f8cd6248a3796dd2a5d4 is the first bad commit
Author: @Evil-Spirit 
Date:   Thu Jul 21 23:58:18 2016 +0600

    Choose entities to select in a way appropriate for the operation.

ruevs:
I can confirm https://github.com/solvespace/solvespace/pull/787 fixes this for me.

PR #787, opened 2020-10-30
Use zOrder for selections within the same group

### LINKED ISSUE #521: Z order for selecting coplanar entities is inverted
### System information

- SolveSpace version: `3.0~690f87c`
- Operating system: Debian 9.x stretch (x86_64)

### Expected behavior

If try to select one of co-planar entities, SolveSpace should prioritize to select nearest entity (according display Z axis).

### Actual behavior

When tried to select one of co-planar entities, SolveSpace selected remote in space entity (according display Z axis), instead of selecting nearest entity.

pic.1

pic.2

### Additional information

For bugs, please attach a savefile that shows the problematic behavior.
You can attach `.slvs` files by archiving them into a `.zip` first.

phkahler:
Fixes #521

### DIFF
--- a/src/draw.cpp
+++ b/src/draw.cpp
@@ -352,7 +352,7 @@ GraphicsWindow::Selection GraphicsWindow::ChooseFromHoverToSelect() {
 
         Group *g = SK.GetGroup(hg);
         if(g->order > activeGroup->order) continue;
-        if(bestOrder != -1 && (bestOrder >= g->order || bestZIndex > hov.zIndex)) continue;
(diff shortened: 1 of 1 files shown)


phkahler:
As a side effect this seems to fix #776 

The old problem of dragging a small circle diameter constraint out of the circle remains fixed although the place where the constraint highlights is a little less intuitive. When it does highlight it drags fine. I'd rather that aspect not change with this PR but overall I think this is a huge win due to fixing 521.

PR #789, opened 2020-11-01
Update README.md

nii236:
After spending a weekend trying to get the code to compile from Windows for Windows, I ended up going with WSL and after some time managed to get it compiling. 

I've updated the instructions to help anyone else in my situation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,8 +104,9 @@ by passing the `-DENABLE_GUI=OFF` flag to the cmake invocation.
 
 ### Building for Windows
 
-You will need the usual build tools, CMake, a Windows cross-compiler, and flatc.
-On a Debian derivative (e.g. Ubuntu) these can be installed with:
+Ubuntu will require 20.04 or above. Cross-compiling with WSL is also confirmed to work.
+
+You will need the usual build tools, CMake, a Windows cross-compiler, and flatc. On a Debian derivative (e.g. Ubuntu) these can be installed with:
 
     apt-get install git build-essential cmake mingw-w64 libflatbuffers-dev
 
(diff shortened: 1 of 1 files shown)


phkahler:
Seems slightly Ubuntu-centric, are you sure about 20.04 as requirement?

Also why remove the 32bit instructions and only tell people to build 64bit? Our released Windows builds are actually 32bit.

nii236:
Ubuntu 18.04 does not have libflatbuffers-dev available in apt, and the PPAs available weren't useful either. Instructions only worked for Ubuntu 20.04. I lost a weekend messing around with compilation instructions so I wanted to help save time for future people who may be entering this repo fresh.

32 bit compilation failed due to issues in:
```
./solvespace/extlib/angle/src/libANGLE/renderer/d3d/loadimageSSE2.cpp:50:48
```
The specific error was:
```
/usr/lib/gcc/i686-w64-mingw32/9.3-win32/include/emmintrin.h:1004:1: error: inlining failed in call to always_inline ‘__m128i _mm_unpacklo_epi8(__m128i, __m128i)’: target specific option mismatch
```

Compiling for 64bit worked without incident.

kk6mrp:
Shouldn't the 32 bit commands be left for other distributions?

nii236:
@kk6mrp then it needs to be clarified which ones work and which don't.

phkahler:
@kk6mrp  The best way to build for Windows is to build ON windows, which has its own instructions. Cross compiling can be useful but we dont need to cover every case here, just enough to help people do it successfully.

kk6mrp:
That is true but it seems a little more confusing now than before, maybe that is just me though.

Issue #795, opened 2020-11-12
RFC: Should we remove Q3DO export?


ruevs:
Should we remove [Q3DO](https://github.com/q3k/q3d/) export and with it rip out flatbuffers (which is a PITA as far as compiling in "strange" environments goes)?

https://github.com/solvespace/solvespace/pull/794#issuecomment-726353146

It seems that [Q3D](https://github.com/q3k/q3d/) export [was added](https://github.com/solvespace/solvespace/pull/384) as a way to get colorful 3D meshes into Horizon EDA https://github.com/horizon-eda/horizon/pull/236. However that pull request was not merged.

In the meantime SolveSpace now exports Step files with color that Horison displays correctly :-)
https://github.com/solvespace/solvespace/pull/767
https://github.com/solvespace/solvespace/pull/763
https://github.com/solvespace/solvespace/issues/452

So it appears to me that SolveSpace is the only program (in this universe) that exports [Q3D](https://github.com/q3k/q3d/) files and nothing imports them. On the other hand I do like this otherwise unused format - aesthetically...

 @q3k, @carrotindustries, @jwesthues, @whitequark (sorry for pinging you) what do you think?

From [the PR that added Q3D](https://github.com/solvespace/solvespace/pull/384) I know that Whitequark had plans to use flatbuffers as a basis for a future "new .slvs format" but I do not know any details - and they can always be added back if/when this happens.

ruevs:
https://freenode.irclog.whitequark.org/solvespace/2019-02-24

phkahler:
**Should we remove Q3DO export?**
Yes we should.  Q3D is apparently of no use to anyone and was invented to solve a problem that no longer exists.

ppd:
Code removed is code debugged ;-)

ruevs:
I'll prepare the change. Let it sit for a few days in case someone complains (I'll ask in the forum too) - if no one does I'll make a pull request.

ruevs:
The removal https://github.com/ruevs/solvespace/commit/491108ff3ecd6ee9ce7b82684dd7caa723fbabd6
A post in the forum https://solvespace.com/forum.pl?action=viewthread&parent=3256

q3k:
Go ahead. If maintaining this is too much effort, absolutely do rip it out.

I’m likely the only user of this, and I don’t mind maintaining my own best effort fork instead.

phkahler:
[...]

PR #803, opened 2020-11-19
 Remove Q3DO export and flatbuffers

ruevs:
It was added in 3a3a275 as a potential way to export colorful meshes
to Horizon EDA but ended up being supported only by SolveSpace. Since no
software can consume the exported q3do files the feature is superfluous.

See https://github.com/solvespace/solvespace/issues/795 for details.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -20,12 +20,6 @@
 [submodule "extlib/angle"]
 	path = extlib/angle
 	url = https://github.com/solvespace/angle
-[submodule "extlib/flatbuffers"]
-	path = extlib/flatbuffers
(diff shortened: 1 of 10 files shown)


phkahler:
Looks good from here. I'm not sure you needed to delete the translations, but OK.

ruevs:
The removed strings  get saved as comments at the bottom (see the German one) - a really nice behaviour!

I will remove the update from this pull request. After your translation update a few days back I started to look at the translations and noticed that solvespave.cpp needs to be added - it contains 38 strings that need translations. So I'll do it as a separate pull request.

Issue #173, opened 2017-01-28
select underlying line segement


Leichtbier:
### System information

SolveSpace version: 2.3~7c1ca460

Operating system: Win10

I'm not able to select the underlying line segement if two line segemnts are upon each other.
I can not hide the first selection to select the underlying line segement. I'm not able to select both to choose. The only way is to search the list "requests in group". Is there are an other way to select an underlying line segement?

wpwrak:
Now that we can select such things from the text window, perhaps a selection of multiple items could could make the items in question also selectable in the text window, with the option of de-selecting ? That way, an ambiguous selection could just select everything applicable, and the user could get rid of the undesired bits.

Another way for handling ambiguous selections that works quite nicely (but would require some changes to the selection logic if applied to SolveSpace) is something I've implemented in fped:
- first, make a list of the candidates, sorted roughly by the difficulty of catching the respective item,
- tentatively select the one that's hardest to catch,
- if clicking again on the same selection, pick the item after the one currently selected,
- and so on.

whitequark:
We even should already have the list of entities under cursor.

Evil-Spirit:
This even was already implemented as "select by order"

whitequark:
I'm not sure if that's the best UI really, compared to e.g. listing all entities in the text window. I'll need to think about it.

Evil-Spirit:
https://github.com/Evil-Spirit/solvespace/commit/a7be01991d4d30556d1442d5d3c5da13b67bcbe9

phkahler:
@Leichtbier and  @wpwrak  There is a pending pull request to give preference to non-construction lines when selecting entities ( #810 ). What do you think of that? Would that be better than having the current behaviour which seems more random?

phkahler:
FYI It is now possible to pick coincident lines by slightly rotating the sketch. For depth comparison each line is considered to be at the depth of its midpoint. That means if you rotate the sketch a little bit it will select the "closest" line when they overlap.

phkahler:
Current master also has the exploded view for 2d sketches, which separates all lines by an adjustable interval perpendicular to the sketch plane.

phkahler:
I'm going to close this, as the exploded view should cover it. There have been other improvements to selection since this was opened as well, and there has been no feedback on any of that.

PR #810, opened 2020-11-21
Move z-index of construction segments behind normal segments

vespakoen:
Right now, when drawing a segment on a construction segment, the segment becomes invisible.

I think it makes more sense that construction lines stay in the background.

This change seems to do the trick.


phkahler:
This scares me because the picking code is so complicated. At the same time I'm happy someone else is having a look at it. I'd like to try a bunch of sketches with this before merging.

There is at least one open issue this might help with - see #173.  I've found that rotating the view slightly and fiddling with the exact mouse location can sometimes allow me to pick a different line, but this will make that impossible. So is it really the best solution?

phkahler:
@vespakoen  The CI is failing with:
test/constraint/angle/test.cpp:69: FAILED: render doesn't match reference; 12 (0.00%) pixels differ

Looks like only one of the test images needs to be updated to match. To quote whitequark on fixing this:
"When you run the tests locally, for failing tests, you get an actual result next to the reference result. Rename the former to the latter, done."  from discussion on #627 talking about test suite images. IIRC the result images might be automatically deleted if the test passes? If that's the case, delete the reference image and rename the result to the reference - after comparing them to be sure nothing is actually broken.

phkahler:
I'm not sure why the image is different. The change affected more than selection.

phkahler:
Oh, it's broken on master I think.

phkahler:
@vespakoen never mind my babbling. My PR was failing a different test, and it should have. To fix the CI issue with this PR, do the following:

1) build your branch as usual
2) run build/bin/solvespace-testsuite
3) notice that it fails on constraint/angle/skew-render
4) open the folder:  test/constraint/angle
5) find files  "skew.out.png" and "skew.out".
6) delete "skew.out" and rename "skew.out.png" to "skew.out"
7) commit the change to your branch and push to you GH.

I have no idea which 12 pixels are different in those images or why this change causes it ;-)

vespakoen:
For reference:

skew.out.png
skew out

skew.png
skew

skew.diff.png
skew diff

Looks like the segment is now underneath the point at X0 Y0.

I have updated the image so the test will pass.

phkahler:
This PR is really really nice. Not only does it make selection better, it improves rendering when you do rounded corners for example. Thanks!

Issue #806, opened 2020-11-20
Mouse-wheel zoom is affected by window position (Windows)


notwa:
### System information

SolveSpace version: 3.0~427a29ab

Operating system: Windows 10.0.18363

### Expected behavior

With the mouse cursor hovering over the main window of solvespace, scrolling the mouse-wheel should zoom in and out with the mouse cursor as its origin. Entities near the mouse cursor will typically remain visible when zooming.

### Actual behavior

Zooming with the mouse-wheel appears to be relative to the top-left corner of the active monitor, ignoring the actual window position. For example, if I have solvespace taking up a quarter of my screen and floating in the bottom right corner, then zooming in will result in all the entities flying out of sight to the top left. This is somewhat mitigated by using solvespace with its main window maximized, but it's still slightly incorrect as it doesn't seem to consider its window decorations offsetting the y coordinate.

### Additional information

This is with (default) 100% DPI scaling in Windows, in case that affects anything. Also, zooming with the `-` and `+` keys works fine.

ruevs:
Thank you for pointing this out! It's been bugging me (subconsciously) for a while, but since I always work full screen and thus it is only slightly bad on the Y axis I just adapted.

Now I fixed it - see the pull request.

This is in fact a regression - it works in 2.3 and older. It would be interesting to bisect and see where it happened.

@phkahler, @vespakoen Does this work properly on Linux/GTK and Mac?

The points where the scroll wheel event is sent to the (platform independent) mouse.cpp:
https://github.com/solvespace/solvespace/blob/942bf3f35409ac58b3f2bc90144dea67a96f6200/src/platform/guigtk.cpp#L548
https://github.com/solvespace/solvespace/blob/942bf3f35409ac58b3f2bc90144dea67a96f6200/src/platform/guimac.mm#L556

PR #815, opened 2020-11-22
Win32: Mouse wheel zooming always remains properly centered

ruevs:
On scroll wheel events convert the mouse coordinates from screen to client
area so that scroll wheel zooming remains centered irrespective of the
window position.

Fixes https://github.com/solvespace/solvespace/issues/806


phkahler:
Thanks @ruevs!

Issue #819, opened 2020-11-23
[3.0rc1] adding a line segment results in status "err" for "degree of freedom"


fe60:
### System information
SolveSpace version: 3.0.rc1
Operating system: Windows 10 (2004)

### Bug
After adding a line segment, the status shown for **"degree of freedom" switches from "ok" to "err"**.
Remove the line segment and the status switches back to "ok".
The problem does not occur with SolveSpace 2.3.
![dof_err](https://user-images.githubusercontent.com/31078918/99941022-9afeb100-2d6d-11eb-9a59-816f30ac6f23.png)

### Additional information
slvs-file shown in the screenshot
[dof_err.zip](https://github.com/solvespace/solvespace/files/5581666/dof_err.zip)

ruevs:
The "err" indicator is a new feature. It is intended to indicate that the sketch would not result in a valid solid if extruded for example. The feature was implemented here c2c26e95adb and last changed here df6777aaf39.

There is a small problem with it, that at the end of adding a closed polygon the status remains on "err" until `Esc` is pressed or a left click anywhere.

ruevs:
I added a description in the [changelog](https://github.com/solvespace/solvespace/commit/8dcb2db6e6b9539e82b7ae8b0fe967914b5f90fc).

fe60:
@ruevs  I use SolveSpace for 2d modeling only. Hence I have deactivated the option "check sketch for closed contour".
From my point of view this should deactivate the newly introduced warnings and errors in the list of groups too.

ruevs:
@fe60 this makes sense to me and is easy to implement. I created a pull request. Lets see what the others and @wpwrak (who implemented the feature) think.

phkahler:
@ruevs  Is the check enabled by default on a fresh install?  I think these new options should probably be enabled by default - things like automatic constraints and "edit newly added dimensions". If not enabled, new users and even people upgrading may never notice them or realize what the option does. Existing users who don't want the new features will get them and perhaps complain about new behaviour at which point we can say - turn it off here. Does that make sense?

ruevs:
It makes perfect sense.
It is turned on by default:
https://github.com/solvespace/solvespace/blob/f2850246fa4f56c9ef080d620d9b7266764dd273/src/solvespace.cpp#L82

PR #823, opened 2020-11-23
UI: Display "err" in the property browser only if "check sketch for closed contour"

ruevs:
... is enabled.

"err" was first introduced in c2c26e95adb to indicate sketches that may cause
problems in the subsequent 3D groups. But is makes sense not to display the
error if the "check sketch for closed contour" option is turned off. The user
obviously does not want to be warned.

Based on a suggestion in https://github.com/solvespace/solvespace/issues/819

### DIFF
--- a/src/textscreens.cpp
+++ b/src/textscreens.cpp
@@ -143,7 +143,7 @@ void TextWindow::ShowListOfGroups() {
                // Link to the errors, if a problem occurred while solving
                ok ? (warn ? 'm' : (dof > 0 ? 'i' : 's')) : 'x',
                g->h.v, (&TextWindow::ScreenHowGroupSolved),
-               ok ? (warn ? "err" : sdof) : "",
+               ok ? ((warn && SS.checkClosedContour) ? "err" : sdof) : "",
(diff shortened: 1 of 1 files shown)


phkahler:
@ruevs what would you think of adding this:
https://github.com/phkahler/solvespace/commit/b82051db42d74a51dee252e8f5cc808acabcb6f2
I have a hard time reading red on black, so this helps quite a bit. I don't want it to look super bright for normal people though.

phkahler (approved):

vespakoen (approved):
LGTM

ruevs:
> 
> 
> @ruevs what would you think of adding this:
> [phkahler@b82051d](https://github.com/phkahler/solvespace/commit/b82051db42d74a51dee252e8f5cc808acabcb6f2)
> I have a hard time reading red on black, so this helps quite a bit. I don't want it to look super bright for normal people though.

I would say no - it is "annoying enough" to make people think what is going on. It will be too obtrusive if brighter. And no one reads it after the first time.

But my opinion on UI design should not be taken too seriously :-)

PR #824, opened 2020-11-23
CI: Replace Travis with GitHub Actions 

ppd:
This is a straightforward port of what we currently have on Travis. Comments welcome.

## What it does
Two workflows:
1. `Test` - Run test suite for all pushes except those to master on Windows, macOS, Ubuntu
2. `CD` - Run tests (like Test), do release builds, update `edge`-release on GitHub, upload those assets there, build amd64 & arm64 snap and deploy to store

`CD` runs for pushes to master and new releases. If a new release is the trigger, it will upload the release assets there instead of to the rolling `edge`-release.

## A few thoughts in no particular order

- I thought about chaining workflows like Test → CD, but I don't think the experience is very nice. With the monolithic workflows, there's more duplication, but the results are displayed in a clear and obvious way.
- The new release workflow works by manually creating a release on GitHub instead of having the CI "magically" promote a tag to release. This is more explicit and clearer too, I think.
- There're two workflows, one for Test + Deployment on `master` and for releases, the other one for testing PRs, branches and such. That creates a little duplication, but the workflows remain simpler.
-  **UX regression**: Cannot re-build individual jobs any more (compared to Travis)
- **Performance regression**: arm64 builds are only possible via QEMU on amd64, which leads to longer build-times. Maybe there'll be hosted arm64 runners in the near future with Apple's recent switch to that platform, which would solve that problem.
- The experience with GH actions is rather nice. Using versioned scripts from upstream for job steps is a good idea. Everything feels more explicit, too. There's less arcane magic hidden underneath compared to Travis.

## Still missing, but trivial:
- [x] snap store token and all those macOS signing secrets
- [x] move scripts to `.github/scripts` folder

## Example
- CD run: https://github.com/ppd/solvespace/runs/1443503136
- Test run: https://github.com/ppd/solvespace/actions/runs/379477027
- Edge release: https://github.com/ppd/solvespace/releases/tag/edge


.github/workflows/cd.yml
```diff
@@ -0,0 +1,235 @@
...
+        run: .travis/install-ubuntu.sh
+      - name: Build & Test
+        run: .travis/build-ubuntu.sh
+
+  test-windows:
+    runs-on: windows-2019
+    name: Test Windows
+    steps:
+    - uses: actions/checkout@v2
+    - name: Add msbuild to PATH
```
ppd, line 27: unnecessary

vespakoen:
Hey, this looks awesome!

I tried to create a new certificate but I have reached the maximum limit of certificates, there isn't a way to remove them myself anymore and apple tells me to contact them to revoke any unused certificates.

I'll see if I can re-export the certificate from a backup and add them to Github.

ppd:
Excellent. The secrets need to be added to GitHub and then must be explicitly passed as env variables. That's still missing.

I'm also thinking about a "rebuild snaps for specific tag, please" workflow. The reason being that sometimes deb packages we include in the snap are updated to fix vulnerabilities, and we'll get nagged by the Snap Store to re-build our stable release to ship those fixes.

vespakoen:
@ppd It is possible to restart a job from Github's UI, it will however take exactly the same parameters / code as the initial run.

Passing environment variables goes like this:

```yaml
        env:
          MACOS_CERTIFICATE_P12: ${{ secrets.MACOS_CERTIFICATE_P12 }}
          MACOS_CERTIFICATE_PASSWORD: ${{ secrets.MACOS_CERTIFICATE_PASSWORD }}
```

Can you send an e-mail to koen at schmeets.de. then I can send you the secrets for the macos build.

ppd:
> @ppd It is possible to restart a job from Github's UI, it will however take exactly the same parameters / code as the initial run.

Are you sure? I can only see the option to re-start the whole **workflow**, i.e. all jobs in a workflow.

> Can you send an e-mail to koen at schmeets.de. then I can send you the secrets for the macos build.

So I'll try those on my fork's master branch and later we can use the very same secrets for `solvespace/solvespace` if we decide to go with GitHub actions, correct?

vespakoen:
Ah I see, yeah that's true, you can re-start the whole workflow only:

Screenshot 2020-11-23 at 21 55 45

and yes, the secrets can be shared on multiple repositories

Symbian9:
> UX regression: Cannot re-build individual jobs any more (compared to Travis)

Is it possible to restart build (for at least once more if it failed on first run) during compiling directly in same job? (maybe, need add some additional script to build workflow)

ppd:
@Symbian9 There's no easy option to restart the whole job from a clean state. You could hack together some retry logic, but I'd rather wait for the inevitable introduction of that feature. It's mentioned as being on the backlog of the GHA team somewhere  on their forums.

ppd:
Sorry for the delay. This is now ready for final review. I tried to write a comprehensive commit message, so yell if it doesn't make sense. @vespakoen Since you have given the most input into CI-related issues so far, I'd leave it up to you to merge it if you feel it's ready.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,239 @@
...
+        echo "::set-output name=upload_url::$upload_url"
+        echo "Upload URL: $upload_url"
+    - name: Upload solvespace.exe
+      uses: actions/upload-release-asset@v1
+      continue-on-error: true
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        upload_url: ${{ steps.get_upload_url.outputs.upload_url }}
+        asset_path: solvespace.exe/solvespace.exe
```
vespakoen, line 217: Do these asset_paths make sense? they look different to what we had before, so not sure if it's a mistake or not...

.github/workflows/cd.yml
```diff
@@ -0,0 +1,239 @@
...
+      with:
+        tag_name: edge
+        release_name: Edge
+        prerelease: true
+        body: ${{ github.event.head_commit.message }}
+
+  upload-release-assets:
+    name: Upload Release Assets
+    needs: [update_edge_release]
+    if: always()
```
vespakoen, line 191: Should this always() be here?

.github/workflows/cd.yml
vespakoen: Ah, upload/download-artifact, guess this is fine...

vespakoen:
This looks good to me, but there is not way to check for me if the build secrets got added to this repository already, so I'll leave the merging to the person that will add them.

.github/workflows/cd.yml
ppd: This makes the filenames look nicer in the artifacts overview. We could do `win-release/solvespace.exe`, but then it shows as `win-release`.

.github/workflows/cd.yml
ppd:
```yml
update_edge_release:
 if: always() && github.event_name == 'push'. 
```
So it doesn't run for release events. If update_edge_release were skipped, update-release-assets would be too unless it uses `always()`.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,239 @@
...
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        tag_name: edge
+        release_name: Edge
+        prerelease: true
+        body: ${{ github.event.head_commit.message }}
+
+  upload-release-assets:
+    name: Upload Release Assets
+    needs: [update_edge_release]
```
ppd, line 190: In my tests, this inherited the `needs` from `update_edge_release`, even if that job got skipped. But perhaps it would be more clear to expand it to `needs: [build-release-windows, build-release-windows-openmp, build-release-macos update_edge_release]`.

vespakoen: Yes I think that makes sense.
Nitpick: I'd also change the `_` to `-` (or the other way around) to make it more consistent with the rest

ppd: Yeah. Dashes are not legal in property dot notation, but I need to reference this job later: `needs.update_edge_release.outputs.upload_url`. So underscores it is then.

ppd:
Changes:
- Use snake case for job ids consistently
- Switch Travis badge to GHA badge

One remaining question: The default build asset retention period is 90 days. Do we need them that long? We might hit our limit of 500 MB early if we have more than floor(500/~22) = 22 builds from master in 90 days.

Symbian9:
> The default build asset retention period is 90 days. Do we need them that long? We might hit our limit of 500 MB early if we have more than floor(500/~22) = 22 builds from master in 90 days.

Is it possible upload artifacts as `continuous` builds asset & then remove them from storing in GHA storage if job is succes?

- https://github.com/aferrero2707/gimp-appimage/releases/tag/continuous

ruevs:
> Is it possible upload artifacts as `continuous` builds asset & then remove them from storing in GHA storage if job is succes?
> 
>     * https://github.com/aferrero2707/gimp-appimage/releases/tag/continuous

In my opinion we should not (even if it is possible). No one needs hundreds of builds downloadable by anyone at any time (and taking hundreds of megabytes on some server somewhere). Makes bug reports rather more complicated because most people don't even follow the simple rule to give the SHA1 of the commit from the Help | About dialog.

ppd:
The build artifacts are more of a side effect of sharing data between jobs (build job -> upload to release job). A short retention period of a few days is probably a good idea as we can still access release builds from a few commits ago for some time.

@Symbian9 it's possible, sure. At the moment we update the edge tag & release to point at HEAD on each build. This removes old assets though. You'd rather accumulate every build? I think I'm in @ruevs's camp here. The edge builds are good simple and good enough for quick testing. If the need arises, we can still provide someone with old builds from the previous workflow runs for some crude bisecting.

Symbian9:
> You'd rather accumulate every build?

At least I'm trying to keep few older builds. But I may agree on upload only latest builds (just by removing previous).

The only thing that should not happen: if builds fails, previous successful builds should NOT be removed until next successful builds would replace it.

phkahler:
We don't need any build artifacts from PRs, only from merges to master which presumably would work. And then only the most recent one as Edge. There is no value in storing more than the latest master IMO. Any prior builds from master can be recreated from the history as needed.

ppd:
`actions/create-release` started creating the edge release in draft mode, which is not what we want. I'm investigating this, and after that we can merge. @phkahler Could you please add the secrets to the repo? See https://github.com/solvespace/solvespace/pull/824#issuecomment-735252401

ppd:
This is ready. Just waiting for the secrets.

phkahler:
I'll get on it tonight. Haven't been able to sit at the computer since last week.

phkahler:
Added the 5 MacOS secrets to the repo but I don't have the snapstore login AFAIK.  Also notice this on the secrets page:
"Secrets are not passed to workflows that are triggered by a pull request from a fork."

But that should be ok - we don't need them until a merge and build for edge or release correct?

ppd:
> But that should be ok - we don't need them until a merge and build for edge or release correct?

The secrets are only needed for the __CD__ workflow, which runs for master commits or releases. So you are indeed correct.
Pull requests are handled by the __Test__ workflow, which does __not__ need any deploy secrets; it only runs tests.

> but I don't have the snapstore login AFAIK

If you have snapcraft installed, you can generate one like so:
```shell
snapcraft export-login --snaps=solvespace --acls package_access,package_push,package_update,package_release --channels edge,beta snapstore-login.txt
```
If not, I can generate one for you when I'm back at my PC.

vespakoen:
Since Travis is not building our stuff anymore, I suggest we merge this so we can have some checks again.
If the snapcraft secrets are not added yet, let's add another issue for it.

ppd:
I agree. It's only the snapstore token anyway.

vespakoen:
So here goes nothing ;)

https://github.com/solvespace/solvespace/actions/runs/408774587

ppd:
Wow. That run overview is fantastic. Is that new?

ruevs:
And it works!

ruevs:
@ppd @phkahler @vespakoen Did you guys notice that forks now get CI and an edge release? E.g.
https://github.com/ruevs/solvespace/releases
I don't mind, (and don't understand anything about the CI configuration) just pointing it out in case it was not intentional.

ppd:
@ruevs We should restrict that to the main repo. Noted, thanks!

ruevs:
@ppd @phkahler Just noticed this:
> Ubuntu-latest workflows will use Ubuntu-20.04 soon. For more details, see https://github.com/actions/virtual-environments/issues/1816

here: https://github.com/solvespace/solvespace/actions/runs/474694562 in the "Annotations" section and I thought to let you know in case we care.

PR #829, opened 2020-11-25
Win32: Remove sscheck on IsWindowVisible and ShowWindow

ruevs:
The return value of these functions is not an error code and according to
the Win32 API documentation they can not affect `GetLastError`.

Calling sscheck on them normally does not fail since it does
SetLastError(0) before running the checked expression and only then
GetLastError(). However in issue 817 a user discovered that when running
"DisplayFusion" software GetLastError does return an error and SolveSpace
closes.

So while not a strict bug-fix this is a "correctness improvement" for
SolveSpace and works around a possible bug in DisplayFusion.

Similarly the return value of Reg*** functions is now compared to
ERROR_SUCCESS which is zero. Before the sschecks were strictly wrong but
did not cause problems for the same reason as above.

### DIFF
--- a/src/platform/guiwin.cpp
+++ b/src/platform/guiwin.cpp
@@ -183,7 +183,7 @@ public:
 
     HKEY GetKey() {
         if(hKey == NULL) {
-            sscheck(RegCreateKeyExW(HKEY_CURRENT_USER, L"Software\\SolveSpace", 0, NULL, 0,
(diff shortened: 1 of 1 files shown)


phkahler:
I find it strange that the old version was doing sscheck on things like IsVisible that were clearly meant to return either true or false.

phkahler (approved):

ruevs:
https://www.displayfusion.com/Discussions/View/displayfusion-can-crash-solvespace/?ID=0b3c924c-7594-4157-bd4b-17af480401f6

PR #830, opened 2020-11-27
Swap vertical and horizontal constraints when rotating 90/270 degrees

vespakoen:
I noticed, when paste-rotating my sketch at 90 degrees, that the vertical and horizontal constraints didn't get updated.
This fixes it.

Should we also take multiples of "X" in: `90+(X*180)`? and negative numbers? (sorry, couldn't explain it better).

### DIFF
--- a/src/clipboard.cpp
+++ b/src/clipboard.cpp
@@ -245,6 +245,7 @@ void GraphicsWindow::PasteClipboard(Vector trans, double theta, double scale) {
         c.reference = cc->reference;
         c.disp = cc->disp;
         c.comment = cc->comment;
+        bool removeConstraint = false;
         switch(c.type) {
             case Constraint::Type::COMMENT:
                 c.disp.offset = c.disp.offset.Plus(trans);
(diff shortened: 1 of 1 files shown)


phkahler:
I think you may want to use Equals() rather than an exact floating point comparison. Search for it to see usage.

vespakoen:
Equals only seems to be for Vectors, I looked at it's implementation and it uses LENGTH_EPS, does this make sense to use for the Theta though?

I updated the code to also remove vertical and horizontal constraints when not rotating 90, 180, 270, 360 etc degrees, otherwise the drawing would jump around.

I also used fmod to make the automatic swapping of vertical and horizontal constraints more generic.

vespakoen:
Just found in the import code (importidf.cpp) that it uses 0.01 for the theta. will look into this some more...

ruevs:
Maybe [`EXACT`](https://github.com/solvespace/solvespace/blob/9390ab02d5e1d921cfaa4e4948082a818d0e8246/src/solvespace.h#L46) is what @phkahler had in mind. It is a macro that disables compiler  warnings for using `==` on floating poits - which normally is stupid.
So 

`if (fmod(theta + (PI/2), PI) == 0)`
Becomes
`if (EXACT(fmod(theta + (PI/2), PI) == 0))`

ruevs:
Alternatively:
`if (fmod(theta + (PI/2), PI) < LENGTH_EPS)`

an `fabs` may be needed - I'm not sure... writing from the phone...

vespakoen:
Negative numbers should also work with fmod, so I think the fabs in unnecessary

phkahler:
I just meant there should be some tolerance around the angle. What if someone types 3.14159/2 or if FP comparison isnt quite exact for whatever reason.

I dont remember if we have an FP version of wrap(). Cant look right now...

vespakoen:
Actually, the rotation is provided in degrees by the user, if someone types 90.00001, that still means we don't want to swap the V/H constraints, they should be removed because they would not be perfectly horizontal or vertical anymore.

I think the EXACT makes the most sense here.

ruevs:
In my opinion this you should have squashed this PR into one commit.

vespakoen:
There is a option in the repository settings to disable merge commits / only allow squash which might be helpful to enforce this for future PR's

ruevs:
@vespakoen no, because a PR may contain multiple commits that make sense to be separate. In fact I think most (all) PR-s should be merged with rebase.

One more note - you made the swap-vertical-horizontal-constraints on the solvespace/solvespace (https://github.com/solvespace/solvespace/tree/swap-vertical-horizontal-constraints) upstream repository instead of your own fork. I am not sure (I don't have much experience with contributing to open source) but think that this is considered bad form.

vespakoen:
Ah shoot, thanks for letting me know, this was not my intention. I should really be more careful with my remotes, my "muscle memory" is used to working with the main repository directly, which probably caused this issue...

PR #833, opened 2020-12-01
Clipboard fixes

vespakoen:
I found some more oddities with paste transformed.

I am hoping I can solve them and am planning to add some new tests for these as well.

~~The first thing I noticed is that rotating happens in counter clockwise direction (at least in the workplane that is opened by default when starting a new sketch)~~ stupid me, scrap that.

The second thing I noticed is that when flipping an arc with tangent constraints, the arc gets inverted, removing the constraints fixes this, but there should be a better way to actually invert the arc which I haven't found yet.

Here is a little sketch to try out.

[rotate-and-flip-me.slvs.zip](https://github.com/solvespace/solvespace/files/5621762/rotate-and-flip-me.slvs.zip)

### DIFF
--- a/src/clipboard.cpp
+++ b/src/clipboard.cpp
@@ -245,18 +245,41 @@ void GraphicsWindow::PasteClipboard(Vector trans, double theta, double scale) {
         c.reference = cc->reference;
         c.disp = cc->disp;
         c.comment = cc->comment;
-        bool removeConstraint = false;
+        bool dontAddConstraint = false;
         switch(c.type) {
             case Constraint::Type::COMMENT:
                 c.disp.offset = c.disp.offset.Plus(trans);
                 break;
-
             case Constraint::Type::PT_PT_DISTANCE:
             case Constraint::Type::PT_LINE_DISTANCE:
             case Constraint::Type::PROJ_PT_DISTANCE:
             case Constraint::Type::DIAMETER:
                 c.valA *= fabs(scale);
                 break;
+            case Constraint::Type::ARC_LINE_TANGENT: {
+                Entity *line = SK.GetEntity(c.entityB),
+                       *arc  = SK.GetEntity(c.entityA);
+                if(line->type == Entity::Type::ARC_OF_CIRCLE) {
+                    swap(line, arc);
+                }
+                Constraint::ConstrainArcLineTangent(&c, line, arc);
+                break;
+            }
+            case Constraint::Type::CUBIC_LINE_TANGENT: {
+                Entity *line  = SK.GetEntity(c.entityB),
+                       *cubic = SK.GetEntity(c.entityA);
+                if(line->type == Entity::Type::CUBIC) {
+                    swap(line, cubic);
+                }
+                Constraint::ConstrainCubicLineTangent(&c, line, cubic);
+                break;
+            }
+            case Constraint::Type::CURVE_CURVE_TANGENT: {
+                Entity *eA = SK.GetEntity(c.entityA),
+                       *eB = SK.GetEntity(c.entityB);
+                Constraint::ConstrainCurveCurveTangent(&c, eA, eB);
+                break;
+            }
             case Constraint::Type::HORIZONTAL:
             case Constraint::Type::VERTICAL:
                 // When rotating 90 or 270 degrees, swap the vertical / horizontal constaints
@@ -267,13 +290,13 @@ void GraphicsWindow::PasteClipboard(Vector trans, double theta, double scale) {
                         c.type = Constraint::Type::HORIZONTAL;
                     }
                 } else if (fmod(theta, PI/2) != 0) {
-                    removeConstraint = true;
+                    dontAddConstraint = true;
                 }
                 break;
             default:
                 break;
         }
-        if (!removeConstraint) {
+        if (!dontAddConstraint) {
             hConstraint hc = Constraint::AddConstraint(&c, /*rememberForUndo=*/false);
             if(c.type == Constraint::Type::COMMENT) {
                 MakeSelected(hc);
--- a/src/constraint.cpp
+++ b/src/constraint.cpp
@@ -127,6 +127,66 @@ hConstraint Constraint::ConstrainCoincident(hEntity ptA, hEntity ptB) {
         Entity::NO_ENTITY, Entity::NO_ENTITY, /*other=*/false, /*other2=*/false);
 }
 
+void Constraint::ConstrainArcLineTangent(Constraint *c, Entity *line, Entity *arc) {
+    Vector l0 = SK.GetEntity(line->point[0])->PointGetNum(),
+           l1 = SK.GetEntity(line->point[1])->PointGetNum();
+    Vector a1 = SK.GetEntity(arc->point[1])->PointGetNum(),
+           a2 = SK.GetEntity(arc->point[2])->PointGetNum();
+    if(l0.Equals(a1) || l1.Equals(a1)) {
+        c->other = false;
+    } else if(l0.Equals(a2) || l1.Equals(a2)) {
+        c->other = true;
+    } else {
+        Error(_("The tangent arc and line segment must share an "
+                "endpoint. Constrain them with Constrain -> "
+                "On Point before constraining tangent."));
+        return;
+    }
+}
+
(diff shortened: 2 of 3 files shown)


ruevs:
In physics (and math) positive angles go counter-clockwise from the X+ axis. Please do not change that ;-)
https://www.physicsforums.com/threads/why-do-trigonometry-angles-go-counter-clockwise.554994/

src/clipboard.cpp
```diff
@@ -210,7 +210,7 @@ void GraphicsWindow::PasteClipboard(Vector trans, double theta, double scale) {
             pt = pt.ScaledBy(fabs(scale));
             pt = pt.ScaleOutOfCsys(u, v, Vector::From(0, 0, 0));
             pt = pt.Plus(p);
-            pt = pt.RotatedAbout(n, theta);
+            pt = pt.RotatedAbout(n, -theta);
```
ruevs, line 213: All the (dead) mathematicians and physicists are spinning in their graves ;-)

ruevs (changes requested):

ruevs:
The arc one is tough - it's been discussed before, but I can not find where right now...

src/clipboard.cpp
vespakoen: Oops, sorry for that, I never noticed it was the other way around. I didn't do too much CAD in my life. Will revert.

phkahler:
The trick with arcs is that the end points need to be swapped. You can see code for that in the paste-transformed. The reason is that arcs are defined as counter-clockwise from the start point to the end point. This also why arcs are not created for Revolve extrusions - they were either correct or went the wrong way around depending which way you dragged it.

For the paste you may run into other issues if you swap them as there may be constraints or something that won't be right.  An alternative to swapping the end points is to flip the normal.  Arcs have a normal and that needs to be copied with a flip. This is the same issue with flipping text entites which I haven't had enough time to look into. Maybe you or I will get them both fixed at once.

BTW what does it mean to flip a normal? It's either scale it by -1 or a member function to flip it... I think... maybe.  :-)

vespakoen:
I found out that the `other` and `other2` properties influence the arc tangent position.

**ARC_LINE_TANGENT** only uses the `other` property as seen here:
https://github.com/solvespace/solvespace/blob/b316a8863bfe72f7acebe7124bef10830675157f/src/constraint.cpp#L649-L658
Swapping that when scale < 0 seems to do the trick.

**CURVE_CURVE_TANGENT** uses both the `other` and `other2` properties as seen: https://github.com/solvespace/solvespace/blob/b316a8863bfe72f7acebe7124bef10830675157f/src/constraint.cpp#L672-L686

Swapping those, and the entities also seems to work, but i'm not sure if it's the right / cleanest solution.

vespakoen:
My limited? test: Screenshot 2020-12-02 at 09 08 20

vespakoen:
Another improvement? (at least to most of my usecases) I added:

**EQUAL_RADIUS**,  **EQUAL_LINE_ARC_LEN** and **EQUAL_LENGTH_LINES** now will link to the "original"

**DIAMETER** gets changed into a **EQUAL_RADIUS** constraint with the "original"

So when I paste things now, it will all reference the original copy. I guess this is what we mostly want?
Let me know if I should put this behind a flag somewhere.

Screenshot 2020-12-02 at 10 30 13

vespakoen:
In hindsight I shouldn't have added the last changes to this PR, it's not working 100% correctly and it is probably out of scope of what this PR was trying to achieve in the first place (fixing bugs in the current implementation).

Will update this one soon.

phkahler:
I'm for fixing bugs for 3.0.

After that i was wondering if copy-paste should use the entity copy-as stuff (at the end of group.cpp) so that all the copies will change if the original is changed. This is how copies happen for new groups - like the 2 ends of an extrude. The copies are created by applying one of several transformations to the original. All copies use the same transform parameters so they move together. Maybe that's worth opening a separate issue to discuss?

@vespakoen  did you flip the arc normals or fix the tangent constraints? Also, drop a text entity in there ;-)

ruevs:
Copy-paste in my opinion should leave the copies completely independent - and thus independently editable. That is how copy-paste works "everywhere".

On the other hand step translating does almost what you describe. Perhaps something like "clone multiple" which makes N clones but allows them to be moved freeley?

ruevs:
At any rate all of this should be post 3.0

vespakoen:
Reverted the last change, now it only fixes the H / V constraints and tangent problems.

vespakoen:
(So no more dependencies between copies)

It simply:
- Fixes the direction of `CURVE_CURVE_TANGENT`, `CUBIC_LINE_TANGENT`, and `ARC_LINE_TANGENT` when mirroring
- Removes `HORIZONTAL` and `VERTICAL` constraints when NOT rotating 90 degrees or any multiple of it.

@ruevs I think this can safely be included in 3.0, it's a pretty annoying bug and I think is worth including the fix asap.

src/clipboard.cpp
```diff
@@ -267,13 +281,13 @@ void GraphicsWindow::PasteClipboard(Vector trans, double theta, double scale) {
                         c.type = Constraint::Type::HORIZONTAL;
                     }
                 } else if (fmod(theta, PI/2) != 0) {
```
ruevs, line 283:
```C++
} else if (EXACT(fmod(theta, PI/2) != 0)) {
```
I missed this last time

ruevs (changes requested):
This does not quite work. It moves the constraints to the "other"/wrong end of the curves sometimes.

ruevs:
This does not quite work. It moves the constraints to the "other"/wrong end of the curves sometimes.
[TangentCurvesPateTransformed.zip](https://github.com/solvespace/solvespace/files/5668905/TangentCurvesPateTransformed.zip)
TangentCurvesPateTransformed

ruevs:
TangentCurvesPateTransformed2

ruevs:
Clipboard02

vespakoen:
Thanks testing! I'll take another look at this.

vespakoen:
Think I fixed it, works with both of our tests now.

ruevs:
Hmm not quite. Lines tangent to beziers still get "swapped" incorrectly.
[pr833_TangentCurvesLinesPasteTransformedBug.zip](https://github.com/solvespace/solvespace/files/5689848/pr833_TangentCurvesLinesPasteTransformedBug.zip)

ruevs (changes requested):
Still not correct for two lines tangent at the ends of a bezier.

vespakoen:
I guess I found the correct solution, but this duplicates code from constraint.cpp.
There are also calls to "swap" in constraint.cpp of which I am not sure what exactly they do and if those should be taken into account.

I should probably move these into a function that can be called from the clipboard.cpp as well.

vespakoen:
Just moved the duplicated stuff into new functions, seems to work good now, let me know if this is the right way / place of things.

phkahler:
@vespakoen  swap(a,b) literally swaps the "values" two variables of the same type. By values I mean all members, so Vectors will exchange x,y, and z values. For example, line/arc tangent constraint appears to be unsure of which parameter is the arc because they are the same type? So it checks and if the one whose variable name is line happens to be the arc, it swaps line and arc. Why? because one has 2 points and one has 3 plus a normal, so it's important to have them the right way when making the constraint equations.

An example would be mirroring an arc. The entity contains handles to the 2 end points, so the mirror image should mirror all points and then in the arc entity it should swap the handles for the 2 end points to ensure it goes the right way around. Either that or flip the normal.

vespakoen:
@ruevs @phkahler What do you think of the current implementation?
I moved the code that is duplicated into static functions on the Constraint class, I wonder if this is the right place for it though.

phkahler:
I think the implementation is fine. You have to try all 3 kinds of constraints you modified code for, both pasted and regularly applied to existing entities. Then I'm happy with it.

vespakoen:
I just did a whole bunch of testing and actually found a problem with my code, but that is now fixed and it is working perfectly now.

I have tested all the changed constraints, in both "directions" of selecting, and also mirrored and pasted, then removed and re-applied constraints.

All is working fine.

ruevs:
OK now it works for all test I tried. And it looks good. I will do an extra small clean up.

PR #843, opened 2020-12-08
Benchmarks in CI

vespakoen:


phkahler:
What is this?
It looks like you're making OpenMP unconditional for release - we do with and without for windows.
Don't really want to run benchmarks on the CI system do we?

vespakoen:
I just added a Debug / Release mode to the `build-ubuntu` script, nothing OpenMP related got changed.
It also looks like my editor formatted the whole .yml file...

I think it would actually be pretty cool to run benchmarks in CI, it allows us to see which PR's speed things up or slow things down.
Right now it doesn't store the results anywhere so it's not that valuable yet because results cannot easily be compared.

phkahler:
We have no idea what hardware the CI is running on, if it's shared, weather it's the same as last time, etc... Benchmarking should always be done locally.

vespakoen:
Allright, closing this.

PR #856, opened 2020-12-14
WIP: Solver patches

phkahler:
This is just 3 of the Evil Spirit solver patches with manual tweaking to get them to build against master.


phkahler:
These versions can be individually applied and compiled, which I can't do with the originals. Also no memory leaks so far. No big performance wins that I noticed either...

Evil-Spirit:
For testing SolveBySubstitution you just need to use drawing with HV, pt-pt-coincident only constraints (like drawing plan of builing).

PR #860, opened 2020-12-16
Fix selection bugs

vespakoen:
I noticed that sometimes normals would get selected even when I didn't select them.
So I added some code to visualise the bounding boxes, which showed me this:

Screenshot 2020-12-16 at 17 21 16

Notice the little yellow dots around the center of the drawing, these are the normals of the arcs.

I then also noticed that when in "3d view", the selection didn't work sometimes, this was caused by the selection marquee only having a "depth" of 1 ... "unit" ?

This PR fixes both of these things, I wonder though, if for the BBox we should add a "FlatOverlap / Overlap2d" check that ignores the Z value, right now I am using the VERY_NEGATIVE and VERY_POSITIVE values for the Z range which is not optimal.

Here is the code for drawing the BBoxes by the way if anyone is interested:

```
+        // Draw the bboxes
+        Entity *e;
+        for(e = SK.entity.First(); e; e = SK.entity.NextAfter(e)) {
+            if(e->group != SS.GW.activeGroup) continue;
+            if(e->IsFace() || e->IsDistance()) continue;
+            if(!e->IsVisible()) continue;
+            bool entityHasBBox;
+            BBox entityBBox = e->GetOrGenerateScreenBBox(&entityHasBBox);
+            if (entityHasBBox) {
+                uiCanvas.DrawRect((int)(entityBBox.minp.x - 1) + (int)camera.width / 2,
+                    (int)(entityBBox.maxp.x + 1) + (int)camera.width / 2,
+                    (int)(entityBBox.minp.y - 1) + (int)camera.height / 2,
+                    (int)(entityBBox.maxp.y + 1) + (int)camera.height / 2,
+                    /*fillColor=*/Style::Color(Style::HOVERED).WithAlpha(25),
+                    /*outlineColor=*/Style::Color(Style::HOVERED));
+            }
+        }
```

I added that code right underneath the following code in draw.cpp:

```
    // If a marquee selection is in progress, then draw the selection
    // rectangle, as an outline and a transparent fill.
    if(pending.operation == Pending::DRAGGING_MARQUEE) {
        Point2d begin = ProjectPoint(orig.marqueePoint);
        uiCanvas.DrawRect((int)orig.mouse.x + (int)camera.width / 2,
                          (int)begin.x + (int)camera.width / 2,
                          (int)orig.mouse.y + (int)camera.height / 2,
                          (int)begin.y + (int)camera.height / 2,
                          /*fillColor=*/Style::Color(Style::HOVERED).WithAlpha(25),
                          /*outlineColor=*/Style::Color(Style::HOVERED));
```

I wonder if the Z range was intentional to not select "all the way through". However, If that was the case I guess it would be better to only select visible things in stead of only selecting things around Z=0

### DIFF
--- a/src/draw.cpp
+++ b/src/draw.cpp
@@ -207,8 +207,8 @@ void GraphicsWindow::MakeSelected(Selection *stog) {
 //-----------------------------------------------------------------------------
 void GraphicsWindow::SelectByMarquee() {
     Point2d marqueePoint = ProjectPoint(orig.marqueePoint);
-    BBox marqueeBBox = BBox::From(Vector::From(marqueePoint.x, marqueePoint.y, -1),
-                                  Vector::From(orig.mouse.x,   orig.mouse.y,    1));
+    BBox marqueeBBox = BBox::From(Vector::From(marqueePoint.x, marqueePoint.y, VERY_NEGATIVE),
+                                  Vector::From(orig.mouse.x,   orig.mouse.y,   VERY_POSITIVE));
 
     Entity *e;
     for(e = SK.entity.First(); e; e = SK.entity.NextAfter(e)) {
--- a/src/drawentity.cpp
+++ b/src/drawentity.cpp
@@ -64,7 +64,7 @@ BBox Entity::GetOrGenerateScreenBBox(bool *hasBBox) {
         Vector proj = SS.GW.ProjectPoint3(PointGetNum());
         screenBBox = BBox::From(proj, proj);
     } else if(IsNormal()) {
-        Vector proj = SK.GetEntity(point[0])->PointGetNum();
+        Vector proj = SS.GW.ProjectPoint3(SK.GetEntity(point[0])->PointGetNum());
         screenBBox = BBox::From(proj, proj);
     } else if(!sbl->l.IsEmpty()) {
         Vector first = SS.GW.ProjectPoint3(sbl->l[0].ctrl[0]);
--- a/src/drawentity.cpp
+++ b/src/drawentity.cpp
@@ -70,7 +70,7 @@ BBox Entity::GetOrGenerateScreenBBox(bool *hasBBox) {
         Vector first = SS.GW.ProjectPoint3(sbl->l[0].ctrl[0]);
         screenBBox = BBox::From(first, first);
         for(auto &sb : sbl->l) {
-            for(int i = 0; i < sb.deg; ++i) { screenBBox.Include(SS.GW.ProjectPoint3(sb.ctrl[i])); }
+            for(int i = 0; i <= sb.deg; ++i) { screenBBox.Include(SS.GW.ProjectPoint3(sb.ctrl[i])); }
         }
     } else
         ssassert(false, "Expected entity to be a point or have beziers");


phkahler:
Those bounding boxes don't make sense to me. They don't seem to bound anything properly? It's also not clear what the bug was that this fixes. You could have opened an issue to report and discuss the bug :-)

vespakoen:
Those bounding boxes are "debug information", they are not added in this PR, but they were used in the process of fixing the bug.

The actual bug is that the bounding box for normals, which the marquee selection box is compared with, are in the wrong place, see:

Screenshot 2020-12-17 at 15 56 06

Those "dots" are the bounding boxes for the normals of the arcs in other places of the drawing.
Those dots were missing the "ProjectPoint3" call:

```diff
-        Vector proj = SK.GetEntity(point[0])->PointGetNum();
+        Vector proj = SS.GW.ProjectPoint3(SK.GetEntity(point[0])->PointGetNum());
 ```

The other problem this PR fixes, is that selecting in 3D didn't work, because the "selection marquee bbox" only works around Z=0, or actually only on the plane perpendicular to the camera, crossing the origin point (if that makes sense).

**Comment by vespakoen** [comment]
Looks like cubic splines, arcs, images and segments are not working as intented:

Screenshot 2020-12-17 at 16 45 02

**Comment by phkahler** [comment]
Are all the correct points take into account for the bounding boxes? Remember that curves have control points. The left-most arc in that image looks like the bounding box is taking 2 control points (one for each ~90 degrees) into account but not the one end point that is outside the box.  Based on the text entity it looks like these are determined after an entity is converted to splines which is good. No need to include the center point of an arc or the corners of the text.

**Comment by vespakoen** [comment]
I found the problem, see: https://github.com/solvespace/solvespace/pull/860/files
The loop that includes all the bezier points into the BBox didn't run all the way to the end.

I also added a bounding box to images.

The last thing I noticed is that the BBox of arcs are a bit bigger than they should be, also, when drawing a segment at 45 degrees, the BBox get's pretty big, maybe we should spend a little more resources here and implement more accurate bboxes & bbox hit testing?

Screenshot 2020-12-17 at 16 58 55

**Comment by phkahler** [comment]
Good catch.  Diagonal segments will be problematic. These are axis-aligned BBoxes, so that's going to be a given. As for arcs and splines, the convex hull of the control points is guaranteed to contain the curve but that isn't as tight as you can get.  It is conceivable to take the derivatives of the spline polynomials and solve for the T parameter where the x,y, or z derivative is zero. That will give the locations of the extremes in x y z. We'd need to take all 3 derivatives separately because each coordinate has its own equation in T. The good news is we only support 3rd degree so the derivative is never more than 2nd degree. I had considered writing a BBox function in ratpoly that does this but there are a million other things too ;-) You also get into the cost/benefit of more time calculating a bbox vs the time it saves by using it in intersection tests.

BTW does this mean that hovering over the end of that arc would not have highlighted it without the fix?

**Comment by vespakoen** [comment]
Thanks for the info, so I guess the simple BBox is fine for now =)

And correct,

- Selecting a box around the "dead zones" did not select anything
- Selecting could randomly include normals of arcs somewhere totally else
- Selecting segments using a marquee didn't work (although it oddly selects a segment sometimes when a point is included in the selection?)

**Comment by ruevs** [comment]
Yeah - selection has always been "funny". But please give me a few days to look at this carefully on a computer.

File: src/drawentity.cpp [hunk 1]
```diff
@@ -54,7 +54,7 @@ BBox Entity::GetOrGenerateScreenBBox(bool *hasBBox) {
     SBezierList *sbl = GetOrGenerateBezierCurves();
 
     // We don't bother with bounding boxes for workplanes, etc.
-    *hasBBox = (IsPoint() || IsNormal() || !sbl->l.IsEmpty());
+    *hasBBox = (IsPoint() || IsNormal() || type == Type::IMAGE || !sbl->l.IsEmpty());
```
ruevs, line 57: Again "no images in selection" is intentional in my opinion.

src/drawentity.cpp
```diff
@@ -64,13 +64,22 @@ BBox Entity::GetOrGenerateScreenBBox(bool *hasBBox) {
         Vector proj = SS.GW.ProjectPoint3(PointGetNum());
         screenBBox = BBox::From(proj, proj);
     } else if(IsNormal()) {
-        Vector proj = SK.GetEntity(point[0])->PointGetNum();
+        Vector proj = SS.GW.ProjectPoint3(SK.GetEntity(point[0])->PointGetNum());
```
ruevs, line 67: OK this is also absolutely a bug an can be included with the `i <= sb.deg;` below.

src/draw.cpp
```diff
@@ -207,8 +207,8 @@ void GraphicsWindow::MakeSelected(Selection *stog) {
 //-----------------------------------------------------------------------------
 void GraphicsWindow::SelectByMarquee() {
     Point2d marqueePoint = ProjectPoint(orig.marqueePoint);
-    BBox marqueeBBox = BBox::From(Vector::From(marqueePoint.x, marqueePoint.y, -1),
-                                  Vector::From(orig.mouse.x,   orig.mouse.y,    1));
+    BBox marqueeBBox = BBox::From(Vector::From(marqueePoint.x, marqueePoint.y, VERY_NEGATIVE),
+                                  Vector::From(orig.mouse.x,   orig.mouse.y,   VERY_POSITIVE));
```
ruevs, line 211: Definitely fixes the marquee selection when the view is rotated. However I have to understand the code much better in order to say what the Z depth of the `marqueeBBox` should be.

Perhaps separate this change as a different commit? e.g.

> Fix marquee selection when the view is rotated away from the working plane.

ruevs (changes requested):
@vespakoen
- Split in two commits (see the comments in the code).
- Remove the image selection (in my opinion).

vespakoen:
I split this PR up into 3 commits, and removed the IMAGE selection.

ruevs (approved):
Looks good. Works good. I'll merge this.

PR #861, opened 2020-12-16
Fix snap to grid not working in some situations

vespakoen:
I found that snap to grid was still not working in some cases, looking at the code, I think this line:

```
SS.GW.ClearPending()
```

Got added unintentionally.

Also, `SS.GW.ClearSelection();` already seems to call `SS.GW.Invalidate();`, so I removed the `SS.GW.Invalidate();` call as well.

After this change snap to grid seems to be working fine.

I wonder if the pending list should still be cleared though, perhaps after the call to ClearSelection.

Will look into this.

To test the broken snap to grid, turn the grid on, draw a rectangle, try to snap each corner, some (might) fail.

### DIFF
--- a/src/graphicswin.cpp
+++ b/src/graphicswin.cpp
@@ -1175,10 +1175,7 @@ void GraphicsWindow::MenuEdit(Command id) {
             }
             // Regenerate, with these points marked as dragged so that they
             // get placed as close as possible to our snap grid.
-            SS.GW.ClearPending();
-
             SS.GW.ClearSelection();
-            SS.GW.Invalidate();
             break;
         }
 


vespakoen:
I just tried out to move `ClearPending` after the `ClearSelection` call, but then the snapping breaks again, I guess this happens because the solving happens asynchronously, and the ClearPending does not.

However, the points do seem to be "cleared" when I drag something after snapping, it doesn't affect the snapped points.
I believe ClearPending gets called when you start dragging points.

I cannot get it to behave strangely, so I think this PR is correct.

src/graphicswin.cpp
```diff
@@ -1175,10 +1175,7 @@ void GraphicsWindow::MenuEdit(Command id) {
             }
             // Regenerate, with these points marked as dragged so that they
             // get placed as close as possible to our snap grid.
-            SS.GW.ClearPending();
```
ruevs, line 1178: Wow! This is code from 11 years ago :-)
https://github.com/solvespace/solvespace/blame/094eff755b035bab2d5eb8037f5264b0033f3ec5/src/graphicswin.cpp#L1178

src/graphicswin.cpp
```diff
@@ -1175,10 +1175,7 @@ void GraphicsWindow::MenuEdit(Command id) {
...
-
             SS.GW.ClearSelection();
-            SS.GW.Invalidate();
```
ruevs, line 1181: And this one is from three years ago.

ruevs (commented):
It is always nice to see old redundant/wrong code go (I have not debugged or understood this so I am not confirming that the code was redundant/wrong ;-) )

Issue #834, opened 2020-12-01
Add option to "Do not shadow inactive groups"


Symbian9:
### System information

- SolveSpace version: `3.0~` https://github.com/solvespace/solvespace/commit/9390ab02d5e1d921cfaa4e4948082a818d0e8246
- Operating system: all supported

### Expected behavior

Add option to set *"Do not shadow inactive groups"* which would prevent "shadowing" of all solids in model/assembly.

### Actual behavior

Actually adding "Sketch in 3D" or "Sketch in Workplane" new group just after `lathe`/`extrude`/`translate`/`rotate` groups, applies `s003-#def-inactive-grp` color style to all previous solids/entities (e.g. all solids are "shadowed").

ruevs:
All 3D groups are "shadowed" when a 2D ("sketch-in-plane") group is active. When a 3D group is active the inactive 3D groups are drawn "normally". In my opinion this makes perfect sense. What is the use case for such an option?

Symbian9:
> What is the use case for such an option?

I use adding *"New Group > Sketch in 3D"* + enabling *"allow redundant constraints"* to draw measurements in 3D just next after solid group.

Video (showcase) — https://www.youtube.com/watch?v=1uZgUpjTg18

Screenshot:

![Measurements in 3D](https://i.imgur.com/GnVSiis.png)

Just for now (as a workaround; "wrong way") I should tune each time `s003-#def-inactive-grp` colors style (colors, lines, stipples, etc.) to be the same as last solid group colors; BUT it still not work for multi-color parts as "shadowing" fully "broke" multi-color parts for this usage (as it just set all colors to grey with default `s003-#def-inactive-grp`) 

> In my opinion this makes perfect sense.

I request just add *"Do not shadow inactive groups"* as *togglable option* (not replace existing behavior, but make it customizable instead) because it would be really useful for producing such types of drawings (exporting as 2D View) —  as for me this is fully ***must have*** feature for *SolveSpace* `3.0`.

phkahler:
@Symbian9  I was wondering how you made those drawings, they look very good. To clarify, after creating the entire object you add a new "sketch in 3d" group that only contains the dimensions, and you allow redundant constraints so they don't appear as reference. Having the constraints in their own group allows you to turn off all entities in previous groups which results in consistent and clean lines, while showing only dimensions from the last group. Correct? I realize you already said this using different (and fewer) words.

[...]

PR #867, opened 2020-12-21
Add menu checkbox Dim Solid for Sketch Groups. Makes shadowing the sol…

phkahler:
…id optional for sketch in 2d/3d groups. Handy for making dimensioned drawings by putting dims in their own group. #834

This works but the change does not take effect until we select a group or other operations. I'm not sure how to trigger the needed redrawing when the checkbox option changes. Help?


phkahler:
Got it.  Invalidate() seems to need True passed to it in order to trigger the full redraw with the option changed. Not sure why other View menu changes don't need that.

PR #869, opened 2020-12-22
Add OpenMP debug information to conf screen

vespakoen:
This PR shows if OpenMP is enabled and shows the maximum number of threads.

It also helps CMake to find the correct libomp (from homebrew) on macOS, by adding the homebrew path to `CMAKE_PREFIX_PATH` (without this, `OpenMP_CXX_INCLUDE_DIRS` is incorrect and causes other bugs).


src/confscreen.cpp
```diff
@@ -5,6 +5,7 @@
 // Copyright 2008-2013 Jonathan Westhues.
 //-----------------------------------------------------------------------------
 #include "solvespace.h"
+#include <omp.h>
```
ruevs, line 8:
```suggestion
 #if defined(_OPENMP)
    #include <omp.h>
 #endif
```

CMakeLists.txt
```diff
@@ -115,11 +115,14 @@ if(ENABLE_LTO)
 endif()
 
 if(ENABLE_OPENMP)
-    find_package( OpenMP REQUIRED )
-    include(FindOpenMP)
```
ruevs, line 119: Is this not needed? (I do not understand CMake much)

vespakoen: No, this looks like a relic from the past.
'find_package' will do the same.

CMakeLists.txt
```diff
@@ -115,11 +115,14 @@ if(ENABLE_LTO)
...
-    if(OPENMP_FOUND)
+    if(APPLE)
+        set(CMAKE_PREFIX_PATH "/usr/local/opt/libomp;${CMAKE_PREFIX_PATH}")
```
ruevs, line 119: (I may be talking totally stupid things)
Shouldn't `OpenMP_CXX_INCLUDE_DIR` be used as an _input_ to provide this path?
Described at the bottom of this page:
https://cmake.org/cmake/help/latest/module/FindOpenMP.html

vespakoen: No, but it looks like we can use OpenMP_DIR instead, I will test that out.

See https://cmake.org/cmake/help/latest/command/find_package.html#id5

vespakoen: Just tried with the `OpenMP_DIR` variable and that didn't work, looking into the docs some more, those (and also the HINTS and PATHS options to find_package) are for specifying the location for the FindOpenMP.cmake file, so I think my current solution is correct.
I will jump on IRC tomorrow and ask some people in the cmake channel.

phkahler: I don't think you need the #pragma omp single or the curly braces. The code should be running single threaded in there anyway. The #if of course is needed because that function won't exist in a non-OMP build.

vespakoen: @ruevs Looks like I dismissed your answer to quickly, I found it at the bottom of that page now too =)

And adding
```
set(OpenMP_CXX_INCLUDE_DIR "/usr/local/opt/libomp/include")
set(OpenMP_C_INCLUDE_DIR "/usr/local/opt/libomp/include")
 ```
 Seems to be enough to make thing work correctly.

**Comment by vespakoen** [comment]
Updated with all suggestions, crazy how "wrong" a couple of lines of code can be =)
Thanks for all the feedback.

File: CMakeLists.txt [hunk 4]
```diff
@@ -115,11 +115,15 @@ if(ENABLE_LTO)
...
+    if(APPLE)
+        set(OpenMP_CXX_INCLUDE_DIR "/usr/local/opt/libomp/include")
+        set(OpenMP_C_INCLUDE_DIR "/usr/local/opt/libomp/include")
+    endif()
+    find_package(OpenMP REQUIRED)
+    if(OpenMP_FOUND)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
-        message(STATUS "found OpenMP, compiling with flags: " ${OpenMP_CXX_FLAGS} )
+        message(STATUS "found OpenMP, compiling with flags: " ${OpenMP_CXX_FLAGS})
+        include_directories(${OpenMP_CXX_INCLUDE_DIRS})
```
ruevs, line 126: Let me ask one more (potentially stupid) question. Does it work without this change on macOS after the change on line 119-120?

vespakoen: The `include_directories` is necessary to be able to use the `omp_get_max_threads()` function.
Since we don't directory call any functions from OpenMP, we didn't need the include path to be added, but for this change it is needed. So yes, it works without this change as well, it is only needed to display the number of threads.

However, I was just looking at the master branch again and debugging the OpenMP variables, and found it actually all to be correct, so that would mean that the `OpenMP_CXX_INCLUDE_DIR` is unnecessary.

I wonder how it had trouble finding the omp.h file though, I might not have cleared the CMake cache before a build or bricked OpenMP by recompiling it for am64.

In any case, I think we can do without the whole
```
set(OpenMP_CXX_INCLUDE_DIR "/usr/local/opt/libomp/include")
set(OpenMP_C_INCLUDE_DIR "/usr/local/opt/libomp/include")
```

Stuff, and just keep
```
include_directories(${OpenMP_CXX_INCLUDE_DIRS})
```
Around in the CMakeLists.txt

vespakoen: Ahaaa... So I just reverted that change and get png compilation errors because (I THINK) it is trying to use the one from homebrew...
I guess that is what this whole thing was about, and it brings back some of my sanity hehe.

Anyways, I will look into this another time, I am off for a while during Christmas.

Happy holidays everyone ;)

ruevs: Sounds good.

vespakoen: I moved the `OpenMP_CXX_INCLUDE_DIRS` to the end of the `include_directories` call in `src/CMakeLists.txt` and added a comment that it should be last for macOS.

The joy of cross platform development I guess ;)

vespakoen: And looking at it now, it might not even need to be last, the include_directories probably just shouldn't be before adding the vendored libraries stuff...
I guess I really need a break from programming for a while ;)

CMakeLists.txt
```diff
@@ -115,11 +115,10 @@ if(ENABLE_LTO)
...
 
 if(ENABLE_OPENMP)
-    find_package( OpenMP REQUIRED )
-    include(FindOpenMP)
-    if(OPENMP_FOUND)
+    find_package(OpenMP REQUIRED)
+    if(OpenMP_FOUND)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
-        message(STATUS "found OpenMP, compiling with flags: " ${OpenMP_CXX_FLAGS} )
+        message(STATUS "found OpenMP, compiling with flags: " ${OpenMP_CXX_FLAGS})
```
ruevs, line 121: white space only

CMakeLists.txt
```diff
@@ -115,11 +115,10 @@ if(ENABLE_LTO)
 endif()
 
 if(ENABLE_OPENMP)
-    find_package( OpenMP REQUIRED )
-    include(FindOpenMP)
-    if(OPENMP_FOUND)
+    find_package(OpenMP REQUIRED)
+    if(OpenMP_FOUND)
```
ruevs, line 119: This can be all caps - as already discussed

CMakeLists.txt
```diff
@@ -115,11 +115,10 @@ if(ENABLE_LTO)
 endif()
 
 if(ENABLE_OPENMP)
-    find_package( OpenMP REQUIRED )
-    include(FindOpenMP)
```
ruevs, line 119: So this removal is the only change left.

vespakoen:
Reverted the whitespace, casing and comment. What a ride :)

Issue #835, opened 2020-12-01
Values smaller than the number of significant digits displayed as 0 in the property browser (was "OSX 3.0RC1 Export 2D view issue with extra lines in non-closed contour and "export these objects" line style option")


chamnit:
### System information

SolveSpace version: v3.0 3.0~bcb8cd2c

Operating system: OSX Catalina 10.15.7

### Expected behavior

"Export 2d view..." of sketch should have included all sketched lines or any lines that are designated with the user line style with the 'export these objects' option checked and enabled. Sketch composed of lines that are not closed contours. Trying to draw cut lines for a laser cutter to follow. 

### Actual behavior

Exports only the 'closed contour' part of the sketch. Seems to ignore the non-closed contour lines. Moving non-closed contour lines to a user-defined line style with 'export these objects' enable also has no effect. Still missing in output file.

Tested with Windows version (same release, Win7) and the test SLVS file exported as expected with these non-closed contour lines in the PDF output.

### Additional information

Simple sketch of a trapezoid with an additional line through the middle. The middle line does not appear in the exported 2d view file on the OSX version of SolveSpace.

[test.slvs.zip](https://github.com/solvespace/solvespace/files/5625194/test.slvs.zip)

ruevs:
Indeed works fine on Windows.
[untitled.pdf](https://github.com/solvespace/solvespace/files/5625490/untitled.pdf)

phkahler:
Related or same as #682

chamnit:
@phkahler : The issue is related, but not the same. This example file I uploaded does not use the "export these objects" option. It's a standard sketch of a trapezoid and an additional line going through it, all in the same active group. The 'export 2d view' tool does not draw this additional line. 

If I try to move this additional line to a line-style with the "export these objects" option enable, there is no change. The additional line stills fails to be drawn in the export.

I updated the issue title to denote there is a problem with both.

chamnit:
For whatever reason, this test file works fine on my MacBook Air (same OS X and SolveSpace versions) but not on my Mac Mini. I will investigate further to see if there are any obvious differences between the configurations on the two machines.

chamnit:
Didn't find much different, except the renderer. The MacBook Air was running the Intel HD Graphics 5000 OpenGL engine. The Mac Mini wasn't (Intel UHD Graphics 630). The latter is the one that doesn't work properly. 

[...]

PR #884, opened 2021-01-09
WIP: Don't show small values as 0.0

phkahler:
Looking for feedback on this one. It relates to issue #835. I don't know if "0..." is a good thing. We could also figure out the minimum number of digits to display a non-zero value and override the setting (I think that's preferable).

Thoughts?

### DIFF
--- a/src/solvespace.cpp
+++ b/src/solvespace.cpp
@@ -331,7 +331,13 @@ const char *SolveSpaceUI::UnitName() {
 
 std::string SolveSpaceUI::MmToString(double v) {
     v /= MmPerUnit();
-    return ssprintf("%.*f", UnitDigitsAfterDecimal(), v);
+    int digits = UnitDigitsAfterDecimal();
+    double minimum = 0.5 * pow(10,-digits);
+    while ((v < minimum) && (v > LENGTH_EPS)) {
+        digits++;
+        minimum *= 0.1;
+    }
+    return ssprintf("%.*f", digits, v);
 }
 static const char *DimToString(int dim) {
     switch(dim) {


src/solvespace.cpp
```diff
@@ -331,7 +331,13 @@ const char *SolveSpaceUI::UnitName() {
 
 std::string SolveSpaceUI::MmToString(double v) {
     v /= MmPerUnit();
-    return ssprintf("%.*f", UnitDigitsAfterDecimal(), v);
+    int digits = UnitDigitsAfterDecimal();
+    double minimum = 0.5 * pow(10,-digits);
```
ruevs, line 335: How about

```suggestion
    if(v < pow(10, -digits)) {
        return ssprintf("%g", v);
    }
    return ssprintf("%.*f", digits, v);
```

In this way if `0.000` would have been displayed (or any rounding in the most significant digit) the value is instead displayed in scientific notation (e.g, `7.23e-08`).

Editing becomes very inconvenient (since the edit box shows the %g format, but a normal number is expected) and the used is forced to
- live with it
- increase the value
- Increase `digits after decimal point to show` (display more significant digits)

ruevs (changes requested):

src/solvespace.cpp
phkahler: I don't really want scientific notation to start appearing, but the input boxes seem to support it. Also, I used the 0.5 because 0.006 would display as 0.01 for 2 digits while 0.005 would be 0.00. Is this PR better than what we have? Does it solve the problem of showing 0 when the value is not 0? And given recent changes, is it correct?  ;-)

ruevs: I understood your logic behind 0.5 (the rounding).
My proposal of just `pow(10, -digits)` is because with the 0.5 (and the current master logic) if one types:
0.0006234
one would get
0.001
displayed. And the same when editing. So they would loose the extra significant digits _visually_ while interrnally they are still there. This is misleading.

With `pow(10, -digits)` they would get the scientific notation with anything that does not have a non-zero digit in the first N (whatever N is set to).

As for the overall logic:
Yours:
- is nice since it shows the complete number always.
- It is "confusing" because it ignores the number of significant digits the user has set when a number is too small.

Mine:
- is nice, because it respects the users choice of number of significant digits and does something obviously "weird" (the scientific notation) when it can not. This would (hopefully) force the user to think what is going on.
- It is annoying, because not all edit fields support scientific notation and the user will have to re-type the value. For example the ones that use `Expr::From` do not (e.g.  `cutter radius offset` https://github.com/solvespace/solvespace/blob/bd41485ebecce360baa729567273714caac1df6c/src/confscreen.cpp#L497)

So which one is better? Is either of them is better than the current situation? I am not sure...

> And given recent changes, is it correct? ;-)

What recent changes?

phkahler: Mine and your changes that needed to be fixed after merging ;-)

Both of these solve the issue of displaying zero when a value is not zero, so I think it's an improvement.

PR #889, opened 2021-01-15
Update messages

rpavlik:
I have no idea if I did this right: I just ran the "translate solvespace" target and committed the changes. Somehow, I've managed to not have to directly interact with gettext before, so hopefully I did it right.

There's some contact emails in there that should probably be updated: do we have a mailing list?

### DIFF
--- a/res/locales/de_DE.po
+++ b/res/locales/de_DE.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: SolveSpace 3.0\n"
 "Report-Msgid-Bugs-To: whitequark@whitequark.org\n"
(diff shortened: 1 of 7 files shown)


phkahler:
I was not aware of that method. It looks like it worked - there are new strings in the .pot file.  Next up, running poedit.

rpavlik:
@ruevs thumbs down emoji? Is that "we don't have a mailing list?" or did I mess it up?

ruevs:
@rpavlik sorry - wrong button - it was meant to be a thumbs up :-)

You did it right (as far as I have figured it out) I also ran the translate terget the one time I did this and it seems to be the right thing to do.

Then PoEdit for the translators to fill in translations without needing to have a build environment.

There is no mailing list.

PR #891, opened 2021-01-16
Update EN translation .po file

phkahler:


phkahler:
Not sure why the commit only shows a date update. This fixed several missing string warnings for me.

phkahler:
I was being dumb here. My branch wasn't up to date so the diff just showed irrelevant stuff. Move alone.

ruevs:
Since March 2020 many people, mostly "move alone" ;-) :-)

Issue #890, opened 2021-01-16
Call for Translators


phkahler:
We are very close to releasing SolveSpace version 3.0 (final) but we'd like the language translations updated. I'm sure it's no fun to have Chinese menus with a stray English item thrown in the middle for example.

There are several new "strings" in SolveSpace that currently have no translation to other languages. We need speakers of those languages to update the .po file for each language (using poedit for example). Languages in need of update are:

  * [x] de-DE,0407,Deutsch
  * [x] en-US,0409,English (US)
  * [ ] fr-FR,040C,Français
  * [x] ru-RU,0419,Русский
  * [x] tr-TR,041F,Türkçe
  * [x] uk-UA,0422,Українська
  * [ ] zh-CN,0804,简体中文

Other languages can be added as well, but we'd like translations as near-complete as possible, which can take a lot of time.
The master .pot file is up to date in:  /res  folder
Languages  .po files can be found in the /res/locales folder. Pull requests welcome.

Code should be taken from the master branch, the latest "edge" release, or the RC2 release.

phkahler:
While anyone can do this, I'd like to ping the following people who have done things for us before.

@lomatus  if you're still around, could you update CN?
@rurban  any chance you're available?
@Symbian9 for UA?
@Evil-Spirit for RU?

Symbian9:
I'm In for update `uk_UA`.

rurban:
Yes

rurban:
I am ready with de_DE, but got a question:
What does the new dimSolidModel menu option mean?
"Dim Solid for Sketch Groups"
I dont see it from the code. 
I only see groupmesh.cpp: DrawMesh fillFront.color = DIM_SOLID. So it's a default color.

The src and translation is a bit rough there. 
Does it mean diminuated or related to a Dimension? Or something else?

The git commit says: "Makes shadowing the solid optional for sketch in 2d/3d groups. Handy for making dimensioned drawings by putting dims in their own group. #834"
The menu item is not really understandable to me yet.

ruevs:
The oprion is on the view menu. By default it is on. This is the original behaviour.

If it is switched off then solids from previous groups bill not be "dimmed" (as ln lightness, darkened). Instead they will remain the color that is assigned to them and displayed when their group is active.

ruevs:
The style that is forced on "dimmed" models is s00d-#def-dim-solid
[...]

PR #894, opened 2021-01-17
Translations: update de_DE for 3.0

rurban:
See #890
make -C src translate_solvespace to update all line numbers


phkahler:
Thank you!

Issue #857, opened 2020-12-14
Vector source for UI icons


Symbian9:
### System information

SolveSpace version: 

Operating system: 

### Expected behavior

Create vector source for UI icons (vector icons set), which could be rendered to raster in any size.

There are two ways:

1. Using *Inkscape* app (rendering could be automatized via Inkscape CLI):
1.1. "Master page" way — all icons should be placed inside single SVG document, where icons are placed as tile/table map:
1.1.1. See *Inkscape* icons & cursors sets - https://inkscape.org/gallery/=inkscape-gui-icon-set/
1.1.2. 
1.2. "Layers" way — all icons should be placed inside single SVG, but instead of "Master page" way, each icon is a layer on top of foreground template.
1.3. "1 Icon = 1 SVG" way — each icon should be created as separate SVG file.
2. Using *SolveSpace* itself — as SolveSpace itself could render via `solvespace-cli`, it would be possible to re-create all icons as SLVS files & render it directly to PNG (even to SVG).

I guess use of *Inkscape* would be the best (as it de-facto common way to create icons used by many developer),

- Creating a new multicolor icon theme - https://wiki.inkscape.org/wiki/index.php/Creating_a_new_multicolor_icon_theme
- How To Design & Prepare App Icons in Inkscape (video) - https://www.youtube.com/watch?v=r2Kv61cd2P4

BUT *SolveSpace* could be really used too for this task, thanks to "*ambient lighting* `1.0`" feature already implemented (I itself may try create icons using SolveSpace only).

> JFTR, For the feature SolveSpace should support vector icons in SVG format for scalable UI/HiDPI support.

### Actual behavior

Actually UI icons are available in PNG (raster) form & has some issues with *antialiasing* & line styles (width, etc.).

Most of icons created "by hands" manually, and now some of them looks "non-professional"

- https://github.com/solvespace/solvespace/tree/master/res/icons

Need clean up & unify style of dots, lines, color palette (limit it to not more than 4-16 colors), etc.

### Additional information

> **NOTE:** Here would be collected links to useful tutorials. More links to be added later!

In the feature, it could all icons could be tweaked in easy way to be ready for Light & Dark (or any other) color theme.

- https://blenderartists.org/t/new-icons-for-blender-2-8/1112701
- https://blender.stackexchange.com/questions/124614/how-to-edit-2-8-ui-icons

phkahler:
[...]

Issue #892, opened 2021-01-16
Update toolbar icons


phkahler:
Looking for some of the new icon bitmaps from the discussion in #857 since we don't have vector icons yet. I know @Symbian9  and @jkrei0 have some newer version. Can we somehow pick the best of what's been created so far and get them in?

In particular I'd like the isometric-view icon to actually be isometric and I know there were some in the thread ;-)  We also merged a new Revolve and Helix icon from that discussion, so this request is purely to update the image files with no code changes.

phkahler:
So @ruevs made a pull request #897 with more icons pulled from the previous discussion. I made some comments in the linked PR. Question now is: Should I merge that PR as a new starting point? There are before and after images in the PR.

What's a good PNG editor on Linux? I want to fix those red X myself ;-)   240, 32, 32 is better with no transparent.

ruevs:
![in3d](https://user-images.githubusercontent.com/15338069/104856109-5892c400-5919-11eb-82e3-6c0c71f35c06.png) Working on an "isometric" icon. But AN isometric view can look like the current icon. Just not our `F3` 120° one.
For example open the attached file and press `W` ;-) Then press `F3`...
[IsometricViewVsOurIcon.zip](https://github.com/solvespace/solvespace/files/5826848/IsometricViewVsOurIcon.zip)

I agree with most of what @phkahler [said in the pull request](https://github.com/solvespace/solvespace/pull/897#issuecomment-761880092) and will try to fix the offending ones.

I think you should not merge the pull request. Let us make the icons "mostly pleasing to most people" first :-D

A good PNG editor for pixel peeping work is IrfanView - run it under Wine if you must ;-)

ruevs:
I updated the icons again. Please see the pull request for details. #897

PR #897, opened 2021-01-17
UI: Non anti aliased icons for the graphics and text windows

ruevs:
Created by jkrei0 here https://github.com/jkrei0/solvespace

Optimized for size.

Merged all but `shaded.png` since in my opinion the current one is better.

Discussion here https://github.com/solvespace/solvespace/issues/857

### Update 2021-03-03
I updated many of the icons to be much closer to the original ones. I also replaced the `Nearest isometric view` icon with one that looks like our isometric view. `My New` below are the latest icons.

Toolbar:
| Old | My New | jkrei0 | 
| ------- | ------ | ------ |
| ToolbarOldIcons2 | ToolbarNewMyIconsIn3D | ToolbarNewIcons2 |

| Text windoiw |   |
| --- | --- |
| Old | TextWindowsrOldIcons2 |
| My New | TextWindowsrNewMyIcons |
| jkrei0 | TextWindowsrNewIcons2 |


ruevs:
https://www.quora.com/Are-fonts-that-arent-anti-aliased-called-aliased-or-non-anti-aliased-or-something-else
Yes, I was thinking about this while writing the commit message :-D :-D

phkahler:
There are actually a lot of these I don't like:
Circle and Arc look smaller.
The A in text looks IMO worse. The old one should just be brighter if it's not at 255 yet.
Toggle Construction looks smaller. An I imagining these as smaller?
Split - Not sure if there should be a point where they intersect. Leaning toward the new one is better.
Constrain point on line should have a point at the intersection. Or maybe the purple square should show up more?

The rest are good. I know there are more on the way too, so we'll get another pass at it.

One thing I dislike about the old ones is there are so many points. Individually I like the old icons, but as a collection all those green points look like clutter. These new one are better in that way.

The one that really really annoys me is the isometric view icon is not isometric. I know @Symbian9 has a new version of those two.

ruevs:
in3d
Working on it. But AN isometric view can look like the current icon. Just not our `F3` 120° one.
For example open the attached file and press `W` ;-) Then press `F3`...
[IsometricViewVsOurIcon.zip](https://github.com/solvespace/solvespace/files/5826848/IsometricViewVsOurIcon.zip)

phkahler:
> Do you want a pull request with just that icon?
> Shall we move the discussion back to #857 ?

Let's take it to #892

Symbian9:
> There are actually a lot of these I don't like:

Yep, I see it too.

ruevs:
@phkahler @jkrei0 @Symbian9 I updated the icons again. Please see above in the pull request description and let me know what you think.

robnee:
I like the new isometric view icon.  I think it's easier to deduce its function and remember.

Symbian9:
@ruevs, `../graphics-window/pointonx.png` has 3 black 'dummy' pixels which should be removed:

pic.1

phkahler:
Only 2 comments. The stray black pixels pointed out above. And also, did you change the color of the red Xs to ( 240, 32, 32 )?

I'm not sure how to apply this to a local branch since I can't change the commits to patches and save them ;-)

ruevs:
> @ruevs, some icons still needs cleanup:

Thanks for checking. Fixed. I also reviewed all others for transparency problems.

ruevs:
> And also, did you change the color of the red Xs to ( 240, 32, 32 )?

@phkahler if you mean the X-es that cross out "disabled" icons in the text window - they are not bitmaps, they are drawn here:
https://github.com/solvespace/solvespace/blob/8c101d5f281ba670a3af2663302fe3dc7fb15283/src/textwin.cpp#L61
and they also have a transparency of 150, which can also be adjusted. If you want I'll make a separate pull request for this.,

ruevs:
> I'm not sure how to apply this to a local branch since I can't change the commits to patches and save them ;-)

`git remote add ruevs https://github.com/ruevs/solvespace.git`
`git checkout ruevs/IconsFromjkrei0`

or

git rebase? https://git-scm.com/docs/git-rebase

PR #899, opened 2021-01-18
Update build instructon. Add use of OPEN_MP to linux and macOS build …

phkahler:
…instructions, and a mention of ENABLE_LTO.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,13 +94,15 @@ After that, build SolveSpace as following:
 
     mkdir build
     cd build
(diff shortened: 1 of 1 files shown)


ruevs (approved):
Fine with me.

vespakoen:
Nitpick: Maybe use "ON" in stead of "Yes", that seems to be what solvespace uses the most, also only all uppercase "YES" is documented as a valid value over here: https://cmake.org/cmake/help/latest/command/if.html

Issue #720, opened 2020-09-27
When drawing segment, viewport is not redrawing while mouse movement


Evil-Spirit:
### System information

SolveSpace version: 0a061b6f9ea8ea10414e04a6a87c14c6b12ad04b
Operating system: Windows 7

### Expected behavior

Viewport should be redrawn every mouse move event

### Actual behavior

While movement, viewport isn't redrawn so we can't see the actual entity position

### Additional information

Click  line segment button then click first point and move mouse. Second segment point stays in the previous position, but it should follow mouse cursor.
It seems, actual only for segment, but need to be tested more

phkahler:
This issue needs confirmation, and then it should be traced to a specific commit because it would be a regression.

ruevs:
I can not reproduce it on d72eba8039c52f5e8d3ab36ec2d8a7598dfd578b Windows 10

phkahler:
> I can not reproduce it on [d72eba8](https://github.com/solvespace/solvespace/commit/d72eba8039c52f5e8d3ab36ec2d8a7598dfd578b) Windows 10

@Evil-Spirit  What version of Windows are you on? How to reproduce it?

phkahler:
Is it maybe a problem when holding down the mouse button while dragging?

kk6mrp:
I have one of the AppVeyor CI builds from before and drawing segments do fail to refresh immediately when moving the cursor fast, but when moving slow it refreshed.

ruevs:
Can you zip and upload this version here. Also what version does it show in Help | About ?

kk6mrp:
3.0~12b7c8be

[solvespace.zip](https://github.com/solvespace/solvespace/files/5431337/solvespace.zip)

Evil-Spirit:
@kk6mrp 
You have exactly the same issue.

Evil-Spirit:
@phkahler I use Windows 7

kk6mrp:
I can't reproduce the issue on the latest Edge release on Windows 10.

ruevs:
@kk6mrp yes, I can reproduce the problem with your version and it is awful. However the commit seems harmless to me 12b7c8be strange...
If you try with the current edge/master (https://github.com/solvespace/solvespace/releases) how does it work for you (it is not a debug build)?

phkahler:
@Evil-Spirit   Please try with a build from edge and report back: https://github.com/solvespace/solvespace/releases

ruevs:
[...]

PR #908, opened 2021-01-23
rate-limit WM_MOUSEMOVE messages so event loop can keep up

robnee:
This fixes Issue #720 by rate limiting the WM_MOUSEMOVE messages to no faster than one per 20ms.  On my machine this allowed the event loop to keep up and be able to process WM_PAINT messages.  I am not sure if this is the best or even simplest solution.  This should be reviewed by a Windows SS Dev.


ruevs:
I think think this just hides the real problem.
https://github.com/solvespace/solvespace/issues/720#issuecomment-768185574

PR #912, opened 2021-01-28
Add ENABLE_BENCH and ENABLE_DEMO cmake options

vespakoen:
I wonder what you think about this, 99% of the time I just want to build the GUI, maybe these are better defaults to use?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,7 +49,7 @@ set(ENABLE_GUI        ON CACHE BOOL
     "Whether the graphical interface is enabled")
 set(ENABLE_CLI        ON CACHE BOOL
     "Whether the command line interface is enabled")
-set(ENABLE_TESTS      ON  CACHE BOOL
+set(ENABLE_TESTS      ON CACHE BOOL
     "Whether the test suite will be built and run")
 set(ENABLE_COVERAGE   OFF CACHE BOOL
     "Whether code coverage information will be collected")
@@ -57,8 +57,12 @@ set(ENABLE_SANITIZERS OFF CACHE BOOL
     "Whether to enable Clang's AddressSanitizer and UndefinedBehaviorSanitizer")
 set(ENABLE_OPENMP     OFF CACHE BOOL
     "Whether geometric operations will be parallelized using OpenMP")
-set(ENABLE_LTO     OFF CACHE BOOL
+set(ENABLE_LTO        OFF CACHE BOOL
     "Whether interprocedural (global) optimizations are enabled")
+set(ENABLE_BENCH      ON CACHE BOOL
+    "Whether to build the benchmarks")
+set(ENABLE_DEMO       ON CACHE BOOL
+    "Whether to build the C solver demo")
 
 set(OPENGL 3 CACHE STRING "OpenGL version to use (one of: 1 3)")
 
@@ -386,12 +390,18 @@ endif()
 
 add_subdirectory(res)
 add_subdirectory(src)
-add_subdirectory(exposed)
+if(ENABLE_DEMO)
+    add_subdirectory(exposed)
+endif()
 if(ENABLE_TESTS)
     add_subdirectory(test)
 endif()
-if(CMAKE_BUILD_TYPE STREQUAL "Release" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
-    add_subdirectory(bench)
+if(ENABLE_BENCH)
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+        message(WARNING "Benchmarks disabled in debug builds.")
+    else()
+        add_subdirectory(bench)
+    endif()
 else()
-    message(STATUS "Benchmarking disabled in debug builds.")
+    message(STATUS "Benchmarks disabled")
 endif()


CMakeLists.txt
```diff
@@ -49,16 +49,20 @@ set(ENABLE_GUI        ON CACHE BOOL
...
 set(ENABLE_COVERAGE   OFF CACHE BOOL
     "Whether code coverage information will be collected")
 set(ENABLE_SANITIZERS OFF CACHE BOOL
     "Whether to enable Clang's AddressSanitizer and UndefinedBehaviorSanitizer")
 set(ENABLE_OPENMP     OFF CACHE BOOL
     "Whether geometric operations will be parallelized using OpenMP")
-set(ENABLE_LTO     OFF CACHE BOOL
+set(ENABLE_LTO        OFF CACHE BOOL
     "Whether interprocedural (global) optimizations are enabled")
+set(ENABLE_BENCH      OFF CACHE BOOL
```
ruevs, line 62: OK but ON by default.

CMakeLists.txt
```diff
@@ -49,16 +49,20 @@ set(ENABLE_GUI        ON CACHE BOOL
...
+    "Whether to build the benchmarks")
+set(ENABLE_DEMO       OFF CACHE BOOL
```
ruevs, line 64: OK but ON by default.

CMakeLists.txt
```diff
@@ -386,11 +390,13 @@ endif()
 
 add_subdirectory(res)
 add_subdirectory(src)
-add_subdirectory(exposed)
+if(ENABLE_DEMO)
```
ruevs, line 393: OK

CMakeLists.txt
```diff
@@ -386,11 +390,13 @@ endif()
...
+    add_subdirectory(exposed)
+endif()
 if(ENABLE_TESTS)
```
ruevs, line 396: OK

CMakeLists.txt
```diff
@@ -386,11 +390,13 @@ endif()
...
     add_subdirectory(test)
 endif()
-if(CMAKE_BUILD_TYPE STREQUAL "Release" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
+if(ENABLE_BENCH)
```
ruevs, line 399: Maybe. But maybe keep the original conditions too do we want someone running benchmarks in debug mode and complaining?

ruevs (commented):
Overall I like the options you added, but I do not like turning off things by default.
If they are not compiled by default things are at risk of bit-rot.

On the other hand turning them off in the CI scripts is good.
Add developers who know what they are doing know how to turn then off.

CMakeLists.txt
vespakoen: We could do something like:

```cmake
if(ENABLE_BENCH)
  if(CMAKE_BUILD_TYPE STREQUAL "Debug")
    message(FATAL_ERROR "Benchmarks cannot be enabled in Debug mode, please set ENABLE_BENCH=OFF or use another CMAKE_BUILD_TYPE value")
  endif()
  ...
```
?

vespakoen: Thinking about it more I guess a warning makes more sense:

```cmake
if(ENABLE_BENCH)
    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
        message(WARNING "Benchmarks were automatically disabled because you are building in Debug mode, please use another CMAKE_BUILD_TYPE to enable benchmarks. disabling benchmarks...")
    endif()
    ...
 ```

**Comment by vespakoen** [comment]
I turned everything back ON by default, and show a warning when ENABLE_BENCH is ON and CMAKE_BUILD_TYPE is Release.

File: CMakeLists.txt [hunk 6]
```diff
@@ -386,12 +390,18 @@ endif()
...
-if(CMAKE_BUILD_TYPE STREQUAL "Release" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
-    add_subdirectory(bench)
+if(ENABLE_BENCH)
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+        message(WARNING "Benchmarks were automatically disabled because you are building in Debug mode, please use another CMAKE_BUILD_TYPE to enable benchmarks. disabling benchmarks...")
+    else()
+        add_subdirectory(bench)
+    endif()
 else()
     message(STATUS "Benchmarking disabled in debug builds.")
```
ruevs, line 406: This should become "Benchmarks disabled" and I'm happy ;-)

ruevs (changes requested):

phkahler:
> Overall I like the options you added, but I do not like turning off things by default.
> If they are not compiled by default things are at risk of bit-rot.

Completely agree with this. Having them build by default may also make people aware that they exist too.

> On the other hand turning them off in the CI scripts is good.
> And developers who know what they are doing know how to turn then off.

Not sure I agree. It doesn't hurt to run everything through CI and doesn't add much build time.

vespakoen:
> Not sure I agree. It doesn't hurt to run everything through CI and doesn't add much build time.

I was thinking the same, I think it's good to build (and maybe also execute) as much as we can in CI to catch issues early.

CMakeLists.txt
ruevs: I meant that the text in lines 401 and 406 should be swapped.

ruevs:
I still think that the text in lines 401 and 406 should be swapped.

You convinced me that everything should be built in CI :-)

vespakoen:
Done ;)

ruevs (approved):

phkahler:
Couldn't you just type "make solvespace"?

vespakoen:
@phkahler oh damnit, you are so right! makes me wonder if this PR is worth it at all...

PR #913, opened 2021-01-28
Apple M1 arm64 support simplified

### LINKED ISSUE #893: armel build issue in mimalloc
### System information

- **SolveSpace version:** 3.0rc1
- **Operating system:**  Debian testing

https://lists.debian.org/debian-science/2021/01/msg00034.html

vespakoen:
This is another attempt at #851

1) I had to use the `dev` branch from mimalloc, and apply this patch on top of it https://github.com/microsoft/mimalloc/pull/346
Collecting all the patches for M1 support seems to be unreasonable because they are so many, I think we will just have to wait for the dev branch to merge into stable and use (the latest commit on) the dev for the time being.

2) OpenMP is linked dynamically again, I am now simply downloading the compiled libraries from where homebrew stores them for our CI, so we don't have to maintain a OpenMP fork / add a new submodule.
Using `brew install libomp` won't work when compiling arm64 on x86_64 hardware (the github CI) since homebrew determines what to download based on the hardware.

3) Instead of forking libpng, I included the patches that are needed to make it work on arm64 in this PR that are applied at build time.

5) I improved the way libomp is linked to the executables, it does not assume libomp to be in a fixed path anymore but instead uses a relative path `${OpenMP_CXX_INCLUDE_DIRS}/../lib/libomp.dylib`, which is still not great, but the best I could come up with.

4) I am using `sysctl -n hw.logicalcpu` in stead of `nproc` which isn't supported on macOS (I guess Github CI added an alias for it)

So now we basically have to wait for mimalloc to ~~accept the atmel patch (https://github.com/microsoft/mimalloc/pull/350) and the macos arm64 patch (https://github.com/microsoft/mimalloc/pull/346), and~~ merge dev into master.

And for libpng we need these 2 patches https://github.com/glennrp/libpng/pull/354 and https://github.com/glennrp/libpng/pull/359 (the last one (359) is not critical but nice for being able to build with XCode / debug / profile on macOS)

For now I added those patches into this PR, so we can remove them once they are merged upstream.
Note that I added them to the CI script, ideally I would add them to our cmake but since it needs to modify the CMakeLists.txt file of a sub-cmake-project, I couldn't get that to work.

This PR also fixes #893 because it uses the `dev` branch of mimalloc.

### DIFF
--- a/.github/scripts/build-macos.sh
+++ b/.github/scripts/build-macos.sh
@@ -1,27 +1,52 @@
 #!/bin/sh -xe
 
-mkdir build || true
-cd build
-
-OSX_TARGET="10.9"
-
+ENABLE_SANITIZERS="OFF"
 if [ "$1" = "release" ]; then
-    BUILD_TYPE=RelWithDebInfo
-    cmake \
-        -DCMAKE_OSX_DEPLOYMENT_TARGET="${OSX_TARGET}" \
-        -DCMAKE_BUILD_TYPE="${BUILD_TYPE}" \
-        -DENABLE_OPENMP="ON" \
-        -DENABLE_LTO="ON" \
-        ..
+    BUILD_TYPE="RelWithDebInfo"
+    ENABLE_LTO="ON"
 else
-    BUILD_TYPE=Debug
-    cmake \
-        -DCMAKE_OSX_DEPLOYMENT_TARGET="${OSX_TARGET}" \
-        -DCMAKE_BUILD_TYPE="${BUILD_TYPE}" \
-        -DENABLE_OPENMP="ON" \
-        -DENABLE_SANITIZERS="ON" \
-        ..
+    BUILD_TYPE="Debug"
+    ENABLE_LTO="OFF"
 fi
 
-cmake --build . --config "${BUILD_TYPE}" -- -j$(nproc)
-make -j$(nproc) test_solvespace
+# this is an option for our Github CI only, since it doesn't have a macos arm64 image yet
+CMAKE_GENERATOR="Unix Makefiles"
+CMAKE_PREFIX_PATH=""
+if [ "$2" = "arm64" ]; then
+    OSX_ARCHITECTURE="arm64"
+    CMAKE_PREFIX_PATH="/tmp/libomp-arm64/libomp/11.0.1"
+    git apply cmake/libpng-macos-arm64.patch || echo "Could not apply patch, probably already patched..."
+    mkdir build-arm64 || true
+    cd build-arm64
+elif [ "$2" = "x86_64" ]; then
+    OSX_ARCHITECTURE="x86_64"
+    CMAKE_PREFIX_PATH="/tmp/libomp-x86_64/libomp/11.0.1"
+    mkdir build || true
+    cd build
+else
+    mkdir build || true
+    cd build
+fi
+
+if [ "$3" = "xcode" ]; then
+    CMAKE_GENERATOR="Xcode"
+fi
+
+cmake \
+    -G "${CMAKE_GENERATOR}" \
+    -D CMAKE_PREFIX_PATH="${CMAKE_PREFIX_PATH}" \
+    -D CMAKE_OSX_ARCHITECTURES="${OSX_ARCHITECTURE}" \
+    -D CMAKE_BUILD_TYPE="${BUILD_TYPE}" \
+    -D ENABLE_OPENMP="ON" \
+    -D ENABLE_SANITIZERS="${ENABLE_SANITIZERS}" \
+    -D ENABLE_LTO="${ENABLE_LTO}" \
+    ..
+
+if [ "$3" = "xcode" ]; then
+    open solvespace.xcodeproj
+else
+    cmake --build . --config "${BUILD_TYPE}" -j$(sysctl -n hw.logicalcpu)
+    if [ $(uname -m) = "$2" ]; then
+        make -j$(sysctl -n hw.logicalcpu) test_solvespace
+    fi
+fi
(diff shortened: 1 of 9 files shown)


phkahler:
What is the macOS situation with this? (assuming the update mimalloc)  Are you able to cleanly build for M1 using the regular build instructions? Can we get to a fat binary with x64 and M1?  From a practical point of view the Intel version runs fine on M1, but there is a bit of a hype wave to be caught with early M1 support out of the box.

Asking @rpavlik  and @vespakoen  what do you want to do about mimalloc?
I'm thinking it's 3.0 final time, would it be good to wait and bump the mimalloc version fist? That would give Debian and M1 users the ability to grab 3.0 final and use the standard instructions right?

We used to get 750-800 views of the 2.3 download every 2 weeks. Between 2.3 and RC2 we are currently getting 1150 in two weeks. This seems to be an important milestone in the life of SolveSpace, and I think we'll be a while before the next one so it's good to get last minute details right.

rpavlik:
If there's not a lot of risk involved with the update, then yes, bump to a patched mimalloc. If the change is large, I'd feel better with just applying a patch to ours for 3.0 since we know it's working well.

ruevs:
I'll try to take a look in the next 3-4 days

vespakoen:
I hope we can include this in the 3.0 release, and using mimalloc's dev branch seems to make it work.

There is one problem with this PR though, and that is when building in debug mode, SolveSpace crashes on startup with the following error:

```
mimalloc: assertion failed: at "./extlib/mimalloc/src/heap.c":188, mi_heap_get_backing
  assertion: "bheap->thread_id == _mi_thread_id()"
```

Coming from here: https://github.com/microsoft/mimalloc/blob/master/src/heap.c#L188

In release mode, I think this error is ignored and SolveSpace runs just fine.
This does not seem to be caused by OpenMP, because it also crashes when building without OpenMP support. 

I am not sure why this is happening, and I will have to bisect to see where it got introduced.
Anyways, the release build is working fine and I think we should include it in 3.0.

vespakoen:
Updated mimalloc to latest master, and added patch that fixes a crash when running in Debug mode.

vespakoen:
I separated the mimalloc update into a separate commit as suggested here: https://github.com/solvespace/solvespace/issues/893#issuecomment-771485769

phkahler:
Not sure if this helps:
https://brew.sh/2021/02/05/homebrew-3.0.0/

phkahler:
@vespakoen  Solvespace is now using mimalloc v2.0.0.  Not sure which version your commit was using, can you have a look?

vespakoen:
If I remember correctly, mimalloc v2.0.0 will make macOS arm64 support work in Release (and RelWithDebInfo) builds, but crashes in Debug builds. This is not a very big deal (I hope) and alternatively we can use the dev branch for macOS only (put it in the readme for example).

However, we can also pin the submodule to the mimalloc dev branch at a specific commit that we know works well, which would be my preference.

rpavlik:
Yeah I'd be fine with checking out the discussion commit that's at the tip of dev right now since it's just m1 fixes on top of 2.0.0

vespakoen:
Oh actually, it looks like I added a .patch file for the changes that we are missing from the dev branch, so the v2.0.0 should be fine! I can also get rid of the patch and just point it to mimalloc's dev branch's latest commit.

ruevs:
Pointing to the correct commit in the dev branch is better. One less file in our repository. Less logic in the build.

rpavlik:
I'll do a PR and you can test it.

vespakoen:
Oh crazy I just pushed a change that removes the mimalloc patch and checks out the latest commit from the dev branch in the same second you commented! I'm testing it out right now...

rpavlik:
oh awesome I'll leave you to it then :)

vespakoen:
Just verified that the latest dev branch from mimalloc is working fine on arm64 (in Debug and Release builds).

I also turned off some targets that we didn't use, these are the remaining ones now:

Screenshot 2021-02-07 at 17 43 33

There used to be a ton of test / demo stuff in there from libpng and pixman that were compiled but never ran, and made it hard to find solvespace / solvespace-core when using XCode.

It would be cool if we could do a RC for this PR so we can test it again on  "uninfected" macs, and to test if the signing / notarization stuff still works. (I have SIP disabled on my macs).

vespakoen:
This PR now also fixes debugging with XCode, somehow the `-ffile-prefix-map` flag broke it (although the value looked correct).

rpavlik:
So I don't know why, but it's showing here that it has a merge conflict with mimalloc.

vespakoen:
I guess git doesn't like rebasing submodules, I will try to rebase it locally.

vespakoen:
Rebased with master and split up into 2 commits.

phkahler:
@vespakoen you may want to try the latest dev_slice branch of mimalloc. All the fixes we want are in there and they about to make a new release.

vespakoen:
I just tried it out (mimalloc's dev-slice branch, 1b22da3c283ec02987ed1051a6b9d4358b4dd52a) and worked fine!

phkahler:
@vespakoen  Can you try with latest master, I bumped our mimalloc version. Let me know if this is ready to merge.

Thanks!

vespakoen:
Just rebased it with latest master, compiles and works fine for me (on the m1 macbook air).

phkahler:
@vespakoen  I think since I merged this and pulled from upstream master, my git status indicates some uncommitted changes to libpng. I didn't make a change there, and can't get the status to clear. Any ideas?

rpavlik:
This pr at least used to apply a patch to libpng.

vespakoen:
Ah yeah that is true, it applies a patch to libpng, I can make a new PR to only apply this patch when on Apple, or we can fork libpng for the time being, whatever works... glad this got merged though!! thanks

PR #915, opened 2021-01-29
do not Invalidate() dragging new points to prevent Windows refresh glitches

robnee:
@ruevs @phkahler   I don't think this Invalidate() is necessary on Windows and also fixes Issue #720.  Since this is in the SS core vs. platform can you please both review and comment on impact on Linux and Mac?

### DIFF
--- a/src/mouse.cpp
+++ b/src/mouse.cpp
@@ -305,7 +305,6 @@ void GraphicsWindow::MouseMoved(double x, double y, bool leftDown,
             HitTestMakeSelection(mp);
             SS.MarkGroupDirtyByEntity(pending.point);
             orig.mouse = mp;
-            Invalidate();
             break;
 
         case Pending::DRAGGING_POINTS:


robnee:
This undoes a change introduced [here](https://github.com/solvespace/solvespace/commit/f324477dd06c15022670a4486bf83be3abbf48b5) by @whitequark.  Perhaps and additional reviewer is warranted.

phkahler:
I will have a look this weekend on Linux. Probably tonight.

phkahler:
@robnee  This change doesn't seem to have any effect on Linux. I tried making a few sketches with different tool, and also sketches on extrusions to make more. If this is correct, it's not surprising is happened in that huge commit especially since it doesn't always cause a problem.  This is also an amazing example of finding the root cause vs just some weird fix.

@nabijaczleweli  or  @vespakoen  can you test this (one line delete) on macOS and verify that sketching and dragging still work as expected. This fix is in response to issue #720 which is only for some windows machines apparently.

robnee:
@phkahler Yes.  I'm glad to have received the push to continue to dig deeper into this.  None of the other new request dragging cases seem to require an Invalidate() so it makes sense that neither would segment/rectangle/ttf font.

kk6mrp:
This fixes the issue for me on Windows and it works fine with Linux as well!

nabijaczleweli:
Well *I* can't, never touched a macOS as long as I live.

phkahler:
> Well _I_ can't, never touched a macOS as long as I live.

Sorry about that, I think I'm getting confused. Will do better to verify stuff like that in the future.

ruevs:
I was experimenting with removing the exact same invalidate... but did not have enough time to convince myself it is not needed. It seems Whitequark might have added it by mistake in the big GUI refactoring. So this is good, let it be, and let's see if some regression pops op...

Issue #902, opened 2021-01-18
Add CLI option to start with custom configuration file (custom styles & workspace layout preset)


Symbian9:
### System information

- **SolveSpace version:** `3.0` (for release)
- **Operating system:** All supported

### Expected behavior

Add option to start SolveSpace with custom `*.json` (settings configuration) style.

- for GUI: `$ solvespace --conf {path_to_JSON_file}`
- for CLI: `$ solvespace-cli --conf {path_to_JSON_file}`

NOTE: If SolveSpace would start with that option, it should anyway read "Recent Files" list from default configuration file (e.g. `~/config/solvespace/settings.json`)

Implementing this feature SolveSpace users would get ability to create few presets and switch between them (without need to manually edit each style & resize windows) on star.

Each peset may include:

- custom color styles:
  - `white-black.json`, `bluepring.json`, `paper-wysivyg.json`, etc. 
- custom main window size/layout ("workspace"):
  - "fullscreen", "720x720", "half screen", etc. 
- custom tolerances & export preferences, etc.

Creating new configuration file would be simple:

1. Launch SolveSpace;
2. Set all styles & other properties for new preset (+ resize & change window position; optionally);
3. Save SLVS file & close SolveSpace;
4. Go to `~/config/solvespace/` & copy `settings.json` file;
5. Paste renamed `settings.json` as `my-new-preset.json` in same folder (or any other folder);
6. Start new SolveSpace session from terminal as `$ solvespace --conf ~/my-new-preset.json`.

### Actual behavior

Actually SolveSpace may use only `~/config/solvespace/settings.json` & overwrites it each time when on each "Save" action.

robnee:
I would find something like this useful.  On windows I haven't been able to find a documented way to set the default config.  I've manage to tweak the registry (so much fun converting doubles to QWORD) to fix constraint text height so I can read it but there should be a more straightforward way, hopefully one that doesn't involve the registry on Windows.

robnee:
@phkahler To start, could the json implementation of class Settings from gtk be moved to a common location for use by all three platforms as well as solvespace-cli?  Then the code from SolveSpaceUI::Init() that thaws all of the config options could be moved to an API outside SolveSpaceUI that platform/entrycli.cpp could use.  Then entry-cli.cpp could get things like ExportChordTolerance from a json Settings object rather than from command line options like --chord-tol.  An option to export the config as json would also be easy to implement across all three platforms.

If you think that's not completely misguided then I could take a look.,

[...]

PR #916, opened 2021-01-30
Add command line parsing to platform GUIs

robnee:
This refactoring adds command line functionality to the platform GUIs as discussed in #902.  Very little needed to be changed.  It mostly moves things around.  A few points of note:

- RunCommand (which was part of platform/entrycli.cpp) has a dependency on the Cairo renderer which must now be linked to the GUI targets.  Maybe not ideal.
- A change was made to the Mac platform to move the code that strips any "-psn_" command line argument to `initCli` instead of `initGui`
- `InitGui` functions still call `initCli` and return the args.  This is probably no longer needed. `InitGui` can be made void.
- Windows requires some code to enable the console output.  This is implemented in platform/entrygui.cpp.  It's kind of ugly and could use a once over.
- `solvespace-cli` is still built as a target


robnee:
Ugh, I need to review the Cairo dependency.  It builds locally for me on Mingw.

phkahler:
I have a bad feeling about that. Each platform defines some things differently and it may be that the cairo camera can not be built at the same time as the window/viewport camera. IDK. Either way, take your time. this is too big a change to put in 3.0.

Issue #918, opened 2021-01-31
Opening and Empty .slvs File Causes a Crash (Was: "Crash after autosave-restore dialog wth file from v2.3")


someSven:
### System information

- **SolveSpace version:3.0~3e3ccdca
- **Operating system:Raspian wth OpenGL active

### Expected behavior
Autosave file from v2.3 should open on v3 when I click to load it in the startup dialog.

### Actual behavior
Crash.

### Additional information
- I have OpenGL activated.
- The file was created in Solvespace 2.3, it works if I choose not to restore the autosave. 
- It's not important to me, just thought you should know.

Shell:
(solvespace:18054): Gtk-CRITICAL **: 10:57:23.745: gtk_file_chooser_widget_set_current_name: assertion 'priv->action == GTK_FILE_CHOOSER_ACTION_SAVE || priv->action == GTK_FILE_CHOOSER_ACTION_CREATE_FOLDER' failed
File ./src/graphicswin.cpp, line 394, function Init:
Assertion failed: !SK.groupOrder.IsEmpty().
Message: Group order can't be empty since we will activate the last group..
Aborted

[Chuwi-Guard.slvs.zip](https://github.com/solvespace/solvespace/files/5899573/Chuwi-Guard.slvs.zip)

phkahler:
Not sure what the issue is, but it looks like a problem in 2.3 or even current version with auto-save. There are 2 files in the zip archive. One of them is zero length so I opened the other one and it works fine. Is the problem actually the empty file? Does it actually work on 2.3?
![issue918](https://user-images.githubusercontent.com/14852918/106391855-4dca3b80-63bd-11eb-8957-30181e401b10.png)

someSven:
I tried again. The problem is the empty autosave file, which v2.3 created at some point. After putting it back into the same dir than the file I want to open, it crashes after trying to load the real file, if I click on "Load autosave". Same with both versions of Solvespace. If I click not to load it, then it works with both versions, the same if I delete the autosave-file.

ruevs:
There is a problem when opening empty (0 bytes long) .SLVS files.
I'll fix it.

![OpenEmptyFileProblem](https://user-images.githubusercontent.com/15338069/106463025-85111900-649f-11eb-8a8a-0c9b37ff27de.png)

The problem is that here:
https://github.com/solvespace/solvespace/blob/11a8a0abd5e21a0da6c818674f8b4b657a401a6c/src/file.cpp#L480
[...]

PR #919, opened 2021-02-01
Fix a crash when opening an empty file.

ruevs:
Fixes: https://github.com/solvespace/solvespace/issues/918

The problem was that here:
https://github.com/solvespace/solvespace/blob/11a8a0abd5e21a0da6c818674f8b4b657a401a6c/src/file.cpp#L480
we cleared all groups. Then we tried to read an empty file and therefore this `while` exited immediately:
https://github.com/solvespace/solvespace/blob/11a8a0abd5e21a0da6c818674f8b4b657a401a6c/src/file.cpp#L487
and we came to here:
https://github.com/solvespace/solvespace/blob/11a8a0abd5e21a0da6c818674f8b4b657a401a6c/src/file.cpp#L548
where there was no `fileLoadError` and thus we were left with no groups, and so this assert failed:
https://github.com/solvespace/solvespace/blob/11a8a0abd5e21a0da6c818674f8b4b657a401a6c/src/graphicswin.cpp#L394
since is is not allowed to have no groups :-)

### DIFF
--- a/res/locales/de_DE.po
+++ b/res/locales/de_DE.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: SolveSpace 3.0\n"
 "Report-Msgid-Bugs-To: whitequark@whitequark.org\n"
-"POT-Creation-Date: 2021-01-17 19:54+0200\n"
+"POT-Creation-Date: 2021-02-01 15:45+0200\n"
 "PO-Revision-Date: 2018-07-19 06:55+0000\n"
 "Last-Translator: Reini Urban <rurban@cpan.org>\n"
 "Language-Team: none\n"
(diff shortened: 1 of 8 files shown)


phkahler:
I'd prefer "The file is empty."  But it's OK either way.

phkahler (approved):

ruevs:
You mean remove "it may be corrupt"?
My logic was that a normal user should normally  not have a zero length slvs file, so for them " the file is empty" may be an unpleasant surprise. Not that telling them it may be corrupt will help :-)

rpavlik:
I'm not sure when this message would actually appear - just when e.g. the user has a complete 0 byte slvs file? As such I have no opinion on the strings wording.

That said, the code itself looks reasonable.

rpavlik (approved):

ruevs:
> I'm not sure when this message would actually appear - just when e.g. the user has a complete 0 byte slvs file?

Yes, only on a 0 byte file.
 Only then the first `fgets` would return 0 (a null pointer).

"If the end-of-file is encountered while attempting to read a character, the eof indicator is set (feof). If this happens before any characters could be read, the pointer returned is a null pointer (and the contents of str remain unchanged)."

Issue #797, opened 2020-11-15
Regression in property browser window position memorization


pwnage101:
I hesitate to call this a regression, but I do think the old behavior was better.

### System information

SolveSpace version: 3.0~ 6c2b967790b4554021b5f0fa8f16236456f95cbb

Operating system: Debian Stable (Buster)

### Old behavior

Here's the old behavior from Solvespace stable release (v2.3):

1. With the property browser open, drag it to to the corner of the screen.
2. Press tab to close it.
3. Press tab again to open it, and the window should appear where it was left (still in the corner of the screen)

### New behavior

3. After re-opening the property browser, it always appears back in the center of the screen.

### Additional information

This is quite a minor issue, but I did feel more productive with the old behavior (with the position of the propery browser memorized).  I have not attempted to bisect to find out which commit introduced the new behavior.

phkahler:
@pwnage101  When I read this my first thought was "Wayland" and I've confirmed (Fedora 32 with Gnome) that in an Xsession the behaviour is the way you expected - text window reappears where it was. Then I switched back to Wayland and for me the text window always appears in the upper left corner (unless I put something else there - gnome tries not to overlap windows). Let us know if your problem is due to running wayland vs X.

My thoughts on this: The Wayland security model makes sense to me - an application should not have access to the rest of the desktop environment and so should not even know where it is within that environment. That means an application can not position itself. I get it. But that means it becomes the compositors job to remember where windows last were and put them there when they appear again. To be fair, having that fall on every application was a stupid design decision in the first place but I can imagine the history...

Another issue, I'm not sure if SolveSpace destroys the text window or just hides it. If it's hidden then it seems really absurd that they move it when it's made visible again. But this is way outside my familiarity with the code so perhaps someone else can comment.

To recap: **Is this actually a Wayland issue?**

pwnage101:
I'm 99% sure I'm running X.  I start my desktop environment using `startx` from tty1 and when I do `ps aux | grep -i xorg` there's clearly Xorg processes running.  Maybe this is useful information: I run a tiling window manager ([i3](https://github.com/i3/i3)).

phkahler:
@pwnage101  Then it's probably not a Wayland issue. But it behaves correctly for me with X which means I'm not reproducing the problem here.

phkahler:
A few days after upgrading to Fedora 33, I applied a patch for PR #922 which seems totally unrelated. After testing that, my terminal window contained this new message:
```
[...]

PR #922, opened 2021-02-05
Permit toggling construction geometry while drawing lines

robnee:
This is for consideration for **post 3.0 release**

Currently attempting to toggle construction geometry while drawing line segments results in an error dialog.  This PR enables the construction geometry toggle command ('g' key) to work while drawing lines.  This does not affect other requests (but maybe it could)

I feel this is an improvement over the error dialog and is useful when sketching.  @phkahler I'm curious what you think of this.

### DIFF
--- a/src/graphicswin.cpp
+++ b/src/graphicswin.cpp
@@ -1304,6 +1304,14 @@ c:
             break;
 
         case Command::CONSTRUCTION: {
+            if(SS.GW.pending.operation == Pending::DRAGGING_NEW_LINE_POINT) {
+                Request* r = SK.GetRequest(SS.GW.pending.request);
+                r->construction = !r->construction;
+                SS.MarkGroupDirty(r->group);
+                SS.ScheduleShowTW();
+                SS.GW.Invalidate();
+                break;
+            }
             SS.GW.GroupSelection();
             if(SS.GW.gs.entities == 0) {
                 Error(_("No entities are selected. Select entities before "
--- a/src/platform/guiwin.cpp
+++ b/src/platform/guiwin.cpp
@@ -513,6 +513,7 @@ public:
     HWND hTooltip = NULL;
     HWND hEditor  = NULL;
     WNDPROC editorWndProc = NULL;
+    LONG lastMoveTime = 0;
 
 #if HAVE_OPENGL == 1
     HGLRC hGlRc = NULL;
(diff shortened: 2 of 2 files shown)


phkahler:
I like this idea a lot. It will reduce the need to re-select a line after its drawn to press G. It would be great to be able to do it while dragging new lines, arcs, rectangles.

@robnee when you updated your branch, this PR got the changes. It now has 5 commits. Fortunately we could still isolate the one you wanted for this PR.

phkahler:
This works even better than I thought it would. When drawing multiple segments once you toggle it stays that way for the rest of them unless you toggle back!

robnee:
Yes, sorry, I meant for that to only be a single commit.  I can resubmit if you need me to.  I can also try enabling it for other entities.

This started as simply suppressing the error dialog while drawing.  I frequently try to hit 'g' while drawing, never remembering it doesn't work.  It was very easy to take that and just make it work instead.

phkahler:
I need to leave an unreleated note here, pretty sure it was a fluke and not related to this patch. I recently upgraded to Fedora 33 so maybe a new message? After testing this, my terminal window contained this new message:

Gdk-Message: 17:45:14.723: Window 0x1370d90 is a temporary window without parent, application will not be able to position it on screen.

(solvespace:99349): Gdk-CRITICAL **: 17:45:14.728: gdk_wayland_window_handle_configure_popup: assertion 'impl->transient_for' failed

Since I've never seen that, I stashed the changes and didn't get the message next time I ran. I want to  investigate more as it seems to relate to issue #797   No idea why it chose now to happen.

robnee:
@phkahler Maybe this line causes the gtk issue you noticed?  I am not sure it's necessary for this to work.  I'll need to get my Linux build working anyway so I'll try this weekend.

SS.ScheduleShowTW();

robnee:
@phkahler I haven't been able to reproduce this on Linux.  I don't really know GTK well but [this documentation ](https://valadoc.org/gtk+-3.0/Gtk.Window.set_transient_for.html) states:

>Dialog windows should be set transient for the main application window they were spawned from.
>This allows window managers to e.g. keep the dialog on top of the main window, or center the dialog over the main window. Dialog.with_buttons and other convenience functions in GTK+ will sometimes call set_transient_for on your behalf.

I see `set_transient_for` call being made for the two top-level windows (in CreateWindow) but not for MessageDialogs.

robnee:
@phkahler I've also gotten this working for new rectangles, cubics, circles and arcs.

phkahler:
> @phkahler I've also gotten this working for new rectangles, cubics, circles and arcs.

@robnee  Nice!  Can you make a clean PR from it or unclog this one?

phkahler:
Replaced by #923. Closing.

PR #923, opened 2021-02-06
permit construction geometry toggle while drawing

robnee:
Currently attempting to toggle construction geometry while drawing results in an error dialog. This PR enables the construction geometry toggle command ('g' key) to work while drawing lines, rectangles, arcs, circles, and cubics.

I feel this is an improvement over the error dialog and is useful when sketching.

### DIFF
--- a/src/graphicswin.cpp
+++ b/src/graphicswin.cpp
@@ -1304,6 +1304,20 @@ c:
             break;
 
         case Command::CONSTRUCTION: {
+            // if we are drawing
+            if(SS.GW.pending.operation == Pending::DRAGGING_NEW_POINT ||
+               SS.GW.pending.operation == Pending::DRAGGING_NEW_LINE_POINT ||
+               SS.GW.pending.operation == Pending::DRAGGING_NEW_ARC_POINT ||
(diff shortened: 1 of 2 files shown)


phkahler:
Why is the first case for NEW_POINT in there? Can points be construction entities?

Also what is:  AddToPending(hr);
which seem unrelated. Just curious.

robnee:
Rectangle requests set `pending.operation = Pending::DRAGGING_NEW_POINT;` when processing the mouse click.  There's no `Pending::DRAGGING_NEW_RECTANGLE` which surprised me a little.

Circle requests (for some reason) don't add the handle for the new request to the `pending.requests` list like all of the other ones do.  Without this the code that flips the `construction` flag doesn't work for circles because `pending.requests` is empty.

ruevs:
This is absolutely beautiful! I love it! I'll suggest a simplification as a comment in the code.

phkahler:
> This is absolutely beautiful! I love it!
> In my opinion we should merge it as a feature for 3.0

@ruevs It is a huge time saver when drawing construction lines. Give it lots of testing. How about we merge it after the M1 stuff so we get an edge build with it to test.

PR #927, opened 2021-02-07
make scheduled/deferred task order deterministic

### LINKED ISSUE #920: "dof" in group list lags by "one click" after applying a constraint with the keyboard.
### System information

- **SolveSpace version:** 3.0~11a8a0ab (openmp edge build)

- **Operating system:** Win 10 2004

### Expected behavior
As soon as I hit `O` to constrain the point on the plane, it should have flipped back to the groups view and shown 1DOF.

### Actual behavior

It flipped back and showed 2 dof until I selected something then un-selected it.

### Additional information

GIF 2-3-2021 10-15-03 AM

You can do the same thing with just a rectangle, it turns out. If you click the "symmetric" toolbar button, you'll actually see the 4 turn into a 3 after it goes back to the groups screen. (Think my screen capture was too slow to see it though) If you do the keyboard, it won't go from 3 to 2 until you force something to happen. (You'll just have to imagine, based on the property browser, when I actually pressed `y` in this capture. when it shows down to 2, that's from  me just clicking) Really weird.

GIF 2-3-2021 10-23-15 AM

robnee:
Solvespace uses two timers (`generateAllTimer` and `showTWTimer`) to defer tasks until the event loop processing finishes.  This helps coalesce multiple calls into one.  You can call `scheduleGenerateAll` multiple times while processing UI messages but only trigger one `GenerateAll`. `scheduleGenerateAll` and `scheduleShowTW` do their scheuduling by setting timers with durations of zero.  These timers fire (at least on Linux and Windows) some time after all other events in the message queue have been processed.  This works fine when scheduling either one of these tasks.  However, there is no guarantee in what order the timers will fire (at least on Windows) regardless of which order the scheduling calls are made.  It's pretty easy to demonstrate (on some platforms) by adding logging to the scheduling calls and timer callbacks.

In many cases `TextWindow::Show` depends on `generateAll` happening first.  This causes UI glitches where displays don't update and their contents are stale.  Since this behavior is not deterministic it's easy to imagine how this problem could make certain bug reports difficult to reproduce and diagnose. #920 is a good example. 
 It also makes syncing up UI behavior across all platforms a challenge.

Solving this in the platform domain is tricky.  This is PR endeavors to make the ordering of deferred calls to `TextWindow::Show` and `generateAll` deterministic.  It does this by replacing `generateAllTimer` and `showTWTimer` with a single `refreshTimer`.  Calls to  `scheduleGenerateAll` and `scheduleShowTW` set flags to note the requested operations and schedule the refreshTimer.  A new callback function `SolveSpaceUI::Refresh` can then check the flags and ensure that `generateAll` happens first.  It fixes #920. Moreover, this PR makes it easy to observe and reproduce this problem reliably **and across all platforms** by simply reordering the calls in the `Refresh` callback.  

It's pretty clear that the ordering is important so some solution is needed, if for no other reason than the sanity of the devs.  I think this is a pretty good solution as it spells out the ordering. If nothing else this PR is helpful in further investigations.

@ruevs @phkahler I'd like to hear your thoughts.

### DIFF
--- a/src/solvespace.cpp
+++ b/src/solvespace.cpp
@@ -123,12 +123,8 @@ void SolveSpaceUI::Init() {
         SetLocale(locale);
     }
 
-    generateAllTimer = Platform::CreateTimer();
-    generateAllTimer->onTimeout = std::bind(&SolveSpaceUI::GenerateAll, &SS, Generate::DIRTY,
-                                            /*andFindFree=*/false, /*genForBBox=*/false);
-
-    showTWTimer = Platform::CreateTimer();
-    showTWTimer->onTimeout = std::bind(&TextWindow::Show, &TW);
+    refreshTimer = Platform::CreateTimer();
+    refreshTimer->onTimeout = std::bind(&SolveSpaceUI::Refresh, &SS);
 
     autosaveTimer = Platform::CreateTimer();
     autosaveTimer->onTimeout = std::bind(&SolveSpaceUI::Autosave, &SS);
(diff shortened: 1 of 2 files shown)


ruevs:
Will look, 2-3 days.

robnee:
@ruevs thanks.  To be clear, I am not proposing this for 3.0 (maybe slap the 4.0 label on it) so I think we can take our time looking at this.

ruevs:
@robnee it took me an year and three months instead of 2-3 days - and I apologize for this.

This is great and it totally makes sense it fixes https://github.com/solvespace/solvespace/issues/920 and https://github.com/solvespace/solvespace/issues/1143.

I rebased it on master here https://github.com/solvespace/solvespace/pull/1257 and will close this one.

PR #929, opened 2021-02-07
remember auto-removed constraints for undo

### LINKED ISSUE #928: Undo symmetric constraint does not restore auto-removed H or V constraints
### System information

- **SolveSpace version:** 3.0~02aa9ea8

### Expected behavior

Draw a rectangle about the origin and constrain one of the sides symmetric.  This will automatically remove a redundant H or V constraint.  Hit C-z to undo the symmetric constraint and restore the auto-removed H or V constraint.

### Actual behavior

While Undo removes the symmetric constraint it does not restore the auto-removed H or V.

### Additional information

@phkahler I've started working on a fix by adding an `UndoRemember()` call before auto-removing the H or V constraint.  This works but requires two undos to fully restore the sketch.  I suppose this is an improvement but is there a way to collapse this into a single undo stack entry so only one undo is required?

robnee:
This PR adds calls to `UndoRemember` before auto-removing constraints when adding midpoint or symmetric constraints.  The subsequent call to `AddConstraint` is changed to `rememberForUndo=false` in order to prevent requiring the user to undo twice to return to the starting state.

Fixes #928

### DIFF
--- a/src/constraint.cpp
+++ b/src/constraint.cpp
@@ -417,8 +417,11 @@ void Constraint::MenuConstrain(Command id) {
                 c.ptA = gs.point[0];
 
                 // If a point is at-midpoint, then no reason to also constrain
-                // it on-line; so auto-remove that.
+                // it on-line; so auto-remove that.  Handle as one undo group.
+                SS.UndoRemember();
                 DeleteAllConstraintsFor(Type::PT_ON_LINE, c.entityA, c.ptA);
+                AddConstraint(&c, /*rememberForUndo=*/false);
+                break;
             } else if(gs.lineSegments == 1 && gs.workplanes == 1 && gs.n == 2) {
                 c.type = Type::AT_MIDPOINT;
                 int i = SK.GetEntity(gs.entity[0])->IsWorkplane() ? 1 : 0;
@@ -493,6 +496,7 @@ void Constraint::MenuConstrain(Command id) {
                             "(symmetric about workplane)\n"));
                 return;
             }
+            // We may remove constraints so remember manually
             if(c.entityA == Entity::NO_ENTITY) {
                 // Horizontal / vertical symmetry, implicit symmetry plane
                 // normal to the workplane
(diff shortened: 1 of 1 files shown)


phkahler:
@robnee  Looks good except the midpoint one has an early exit in the else{} clause. Might want to put an undoremember in both the if and elseif.

robnee:
@phkahler Thanks!  I'll take a look at that.

also, there is a patch in there for the wayland error you noticed when using deprecated gtk popup.  Did you check that too?  I should have done that as a separate PR I suppose but I found these at the same time.  Do you want to close this and I can resubmit 2 separate PRs after I look at the elseif?

phkahler:
@robnee  No need to resubmit, just fix the one and I'll merge them both.  So yes, I tried the other commit and it seems to work. Getting rid of depricated GTK stuff is good ;-)

src/constraint.cpp
```diff
@@ -493,6 +496,8 @@ void Constraint::MenuConstrain(Command id) {
                             "(symmetric about workplane)\n"));
                 return;
             }
+            // We may remove constraints so remember manually
+            SS.UndoRemember();
             if(c.entityA == Entity::NO_ENTITY) {
                 // Horizontal / vertical symmetry, implicit symmetry plane
                 // normal to the workplane
```
ruevs, line 503: This needs to be taken into account as well.
Can be reproduced with "Sketch | Anywhere in 3D" then `Y`
Get the error dialog.
Then you need to undo two times, to undo the last operation before the failed symmetric constraint.

ruevs (changes requested):

robnee:
@ruevs. Thanks!  It started to get a little clumsy with the edge cases.  I rewrote it slightly to solve the edge cases and make the undo group handling clearer.  Thoughts?

phkahler:
Any chance you could squash the 3 undo related commits?

ruevs:
> @ruevs. Thanks! It started to get a little clumsy with the edge cases. I rewrote it slightly to solve the edge cases and make the undo group handling clearer. Thoughts?

@robnee Now it works as it should. But please squash (the non-GTK) commits into one.

robnee:
I squashed the commits and took out the gtk change which I will resubmit together with a few other deprecation fixes.  Are any other changes needed here?

phkahler:
I'll merge it. We like commits on the main branch to be clean, vs having the bugs worked out there over several commits. I think you've done the work on this one now ;-)

Issue #938, opened 2021-02-17
Property browser dissapeared


giannissc:
I would like to inform you of a rather weird bug I have found. Sometimes when you work on an external monitor with solvespace the next time you use solvespace without a monitor the property browser is "trapped" into an imaginary monitor. The only way to make it reappper is if you connect it again to a monitor drag both windows to the laptop monitor and disconnect from there on.

I am using Windows 10 (don't know if that helps)

I willing to help any way I can to fix this.

Thanks you for your good work :)

ruevs:
Or delete these registry keys:
```
HKEY_CURRENT_USER\Software\SolveSpace\
TextWindow_Bottom
TextWindow_Left
TextWindow_Right
TextWindow_Top
TextWindow_Maximized
```

Or if you are on an old version (2.3) `TextWnd_*`

ruevs:
At least on Windows there is code that tries to handle this situation:

https://github.com/solvespace/solvespace/blob/04e872a61fbf1492b5425c0cc674d3a10998496b/src/platform/guiwin.cpp#L1220

So it is interesting how it happened. Can you consistently reproduce it?

ruevs:
Fixed.

PR #940, opened 2021-02-17
Win32: If the main or text window is off-screen make it visible...

ruevs:
...by 1/40th of the scrreen width so that the user can notice it.

This can happen if the window was on a second monitor, which is
disconnected, or if the user knows about `Alt-Space | Move` and then
moving the window with the arrow keys.

Before this, if for example the left edge was off-screen it was moved
to 1920, which is just off-screen, so the window remained invisible.

Fixes: #938

### DIFF
--- a/src/platform/guiwin.cpp
+++ b/src/platform/guiwin.cpp
@@ -142,7 +142,18 @@ static std::string NegateMnemonics(const std::string &label) {
     return newLabel;
 }
 
-static int Clamp(int x, int a, int b) {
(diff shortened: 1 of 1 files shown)


phkahler:
I think your clamp function is too cryptic. Add some comments or stop bring clever?

ruevs:
OK, I'll "deobfuscate" it :-)

ruevs:
@phkahler done

phkahler:
Thanks @ruevs!

Issue #865, opened 2020-12-18
Rotating a "link / assembly" does not work with ctrl / shift down from the start


vespakoen:
### System information

SolveSpace version: edge

Operating system: macOS

### Expected behavior

When dragging a point, with ctrl or shift down before dragging the point, I would expect the model to rotate.

### Actual behavior

It doesn't, the only to get it to rotate is to start dragging without ctrl / shift down, then, while dragging add ctrl / shift to the combo and it starts working.

### Additional information

This is very confusing, It took me a while to figure out how to get it working consistently.

In `mouse.cpp`, there is a `if (!havePainted) { ... return; }`, when removing this, it works the way I expect it to work, but this is of course not the right solution.

I am not exactly sure what should be done to fix it.

ruevs:
Directly `Shift+LeftDrag` rotates the model around the selected point as expected for me (on Windows).

As you described `Ctrl+LeftDrag` only works if I am already `LeftDragging`. It works rather erratically, it can stop rotating the model "spoadically". And the rotation it does is the same as `Ctrl+WheelClickDrag" for the whole view. This is rather interesting - I did not even realize this function existed...

robnee:
I botched the first PR but but I've resubmitted a clean one.  There's a lot going on in the mouse movement handling code with many code paths.  Some code paths skip important steps like marking geometry dirty while dragging.  When that happens the havePainted check can "lock up" because SS gets into a state where it doesn't want to repaint (which would reset havePainted) but it also isn't marking anything dirty so no painting is deemed to be required.  This fix makes sure that the geometry gets consistently marked dirty (through all code paths) when dragging.

phkahler:
Does 945 (merged) close this?

vespakoen:
Yes it does 🎉, closing...

PR #945, opened 2021-02-27
Fix problems with ctrl/shift key drag modifiers

robnee:
This fixes handling of the shift and ctrl modifiers for dragging. There are many code paths in the mouse move handling code and some improperly skip necessary steps.

The were several code paths when handling mouse movement that skipped updating the dragged points which suppressed screen updates.
It was possible for the shift / ctrl drag rotation vector to be empty in some cases, like at the start of a drag, nulling out the rotation.
the code that prevented the rotation vector from being degenerate at the start of the drag also prevented starting a drag with ctrl down as noted in issue #865
This PR also separate the extraLine tracking code from the point dragging code for clarity.

### DIFF
--- a/src/mouse.cpp
+++ b/src/mouse.cpp
@@ -270,15 +270,16 @@ void GraphicsWindow::MouseMoved(double x, double y, bool leftDown,
         return;
     }
 
+    if(pending.operation == Pending::DRAGGING_POINTS && ctrlDown) {
+        SS.extraLine.ptA = UnProjectPoint(orig.mouseOnButtonDown);
+        SS.extraLine.ptB = UnProjectPoint(mp);
+        SS.extraLine.draw = true;
+    }
+
     // We're currently dragging something; so do that. But if we haven't
     // painted since the last time we solved, do nothing, because there's
     // no sense solving a frame and not displaying it.
     if(!havePainted) {
-        if(pending.operation == Pending::DRAGGING_POINTS && ctrlDown) {
-            SS.extraLine.ptA = UnProjectPoint(orig.mouseOnButtonDown);
-            SS.extraLine.ptB = UnProjectPoint(mp);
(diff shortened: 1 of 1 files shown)


vespakoen:
This is awesome!! Seems to work really well from my (somewhat) limited tests.
I think this is a very important feature to get merged into 3.0 because without it working with linked assemblies is quite annoying, especially for newcomers.

phkahler:
It seems confused. If I link a part, select a point A on the part, and then CTRL-SHIFT-drag another point B. It rotates the part around point B. If I don't first click A it rotates around either the screen center or the origin (I didn't pay close attention). So why do I need to select point A to rotate around point B?  ;-)

vespakoen:
Interesting, I also tested with a linked part and it's working correct for me.
It looks like you select points, and then try to rotate the assembly?

My "workflow" is, I hold shift of ctrl, then drag a point.

Another alternative is, mouse-down on a point -> add the modifier (ctrl or shift) -> move the mouse.

Before this PR, it only would rotate my assembly when I first dragged a point for some distance, and then add in the modifier key, this is now not necessary anymore for me.

robnee:
I meant shift drag _or_ ctrl drag above.  sorry for the confusion.  From looking at the code I'm pretty sure the ctrl and shift modifiers are independent.  Shift-drag on a point rotates the assembly freely around that point while ctrl-drag rotates around the point constrained to within the plane of the screen.  Drag with modifiers drags the assembly.  If you use both keys it looks like ctrl takes precedence.

I tried this on Centos8 and it behaves the way I think it should. But that's is based on reading the code and comments here, I wasn't originally even aware of this feature. :)

phkahler:
I didn't know about this feature. Shift-drag seemed to work for me prior to the patch, but ctrl-drag did not. now they both do. Nice.

Issue #936, opened 2021-02-14
Property browser: indicate distance also as vector components


pascal-niklaus:
### System information

Solvespace: V3.0

### Expected behavior

This is a feature request:

When a distance is displayed in the property browser (e.g. when two points are selected), it would be nice to not only have the 3d distance but also its projected components along the axis.

I often find myself calculating these from the coordinates displayed, but it seems easy enough to display these also in the property browser.

For example: distance = 5.0 (3.0, 4.0, 0.0)

robnee:
@phkahler How does one compute the three components given two point entities?  The UI work should be easy.

phkahler:
@robnee how about just ( B.x - A.x ,  B.y-A.y ,  B.z - A.z )

The real pain is dealing with the text window stuff, but this would be for display only so it's a good intro to that.

robnee:
@phkahler Thanks!  This works:
```
-        double d = (p1.Minus(p0)).Magnitude();
-        Printf(true, "  d = %Fi%s", SS.MmToString(d).c_str());
+        Vector dv = p1.Minus(p0);
+        Printf(true, "  d = %Fi%s " PT_AS_STR,
+            SS.MmToString(dv.Magnitude()).c_str(), COSTR(dv));
```
I can do a PR tomorrow.  At some point I suppose you are going to freeze feature changes for 3.0 though. :)

phkahler:
> I suppose you are going to freeze feature changes for 3.0 though. :)

@robnee  I thought we were feature frozen a  month or two ago. I'm just waiting for the mimalloc stuff and hopefully the Apple M1 builds. It's been way too long since the last "official" release and I want this one to be good for a long time - not that we should wait so long again.

The recent changes are bug fixes and reasonably small  enhancements that have low probability of causing regressions. I don't want to discourage anyone by not merging these things. We shall have our 3.0 final by pi-day, the 14th (I hope).

robnee:
@phkahler Ok, thanks.  That's great news about 3.0.  If there's anything I can do to help push it over the finish line please let me know.

phkahler:
> If there's anything I can do to help push it over the finish line please let me know.

[...]

PR #949, opened 2021-03-03
UI: Show XYZ distance components in the text window for two points

ruevs:
When two points are selected in addition to their distance show the X, Y and Z components of it in the text window.

Implements #936


robnee (commented):
Looks good to me.  :)

ruevs:
@robnee After one stupidity I do not merge my own pull requests any more ;-)

phkahler:
> @robnee After one stupidity I do not merge my own pull requests any more ;-)

Even my trivial ones I'll leave sit in a PR for a day or two in case I notice something dumb. Anything much bigger than that I like to have looked at  ;-)

Issue #682, opened 2020-08-22
Option to export construction entities (to 2d view / PDF)


gnbl:
SolveSpace version 3.0~6894b0c9 (https://ci.appveyor.com/api/buildjobs/rhtcbg18131udmto/artifacts/build%2Fbin%2FDebug%2Fsolvespace.exe) on Windows 10 64 bit.

I'd like to have construction entities (line segments) in the PDF export.
There does not seem to be an option to do this; points and construction entities are apparently not being exported.

The segment properties show

```
IN STYLE none
CONSTRUCTION
```
It would be nice to be able to click on `none` or another link to assign a style, which is possible via RMB context menu "Assign to Style". BTW, the menu item "Edit > Line Styles ..." does not work while an entity is selected, it is shown only when the entity is deselected.

Assigning a custom style to a construction line allows enabling the option  "export these objects", but **this does not take effect**.

The Line Styles list shows the default style

`s002-#def-construction`

that does not have an additional option for export as the custom styles do.

The 'configuration' menu does not have an option either to allow export of construction entities.

So currently I cannot use construction geometry for my purpose (annotation of secondary geometry like doors etc. in a floor plan) and have to ignore the "not closed contour, or not all same style" error (I might try and use another sketch).

Potential solutions

1. add global option for "export construction"
2. use "export these objects" from custom style to override default behaviour of not exporting construction entities or raise an error, since the option otherwise does not have the anticipated effect (or disable option for construction geometry)
3. show "export these objects" in default style for construction

Thanks!

robnee:
@phkahler When looking into recent SVG export fixes I notice that this issue might be possible to address.  I don't think we can make the construction style directly exportable.  "export these objects" isn't available to change for the default styles.  There is also no real way to "save" the construction style setting in the .slvs file since only custom styles are saved.

What we could do though is remove the explicit check that skips construction entities when exporting.  We would also force the exportable flag to false for the construction default style (currently all default styles set export to true).  This would preserve current behavior but If the user then restyled construction entities with a custom style that *did* have the export option enabled those entities would then export with the custom style.

it's not ideal and it is probably difficult for the user to discover but it should be simple to implement. It might be easier to demonstrate with a PR than to explain but I wanted your thoughts first.

phkahler:
I'm thinking the submitters idea 2 is the right thing to do since the option is already there but not working.

[...]

PR #952, opened 2021-03-07
Add exportable option to default styles

robnee:
Primarily, this enables the user to export of construction entities in 2D views such as SVG and PDF as noted in #682. Previously construction entities were always skipped. The "export these objects" is now available for all default styles. One may turn off export of constraints or the inactive groups for example.  Because export of construction entities is no longer blocked export option for custom styles also works for them.

This also adds the exportable flag to the factory defaults (so it can be reset) and support for saving the exportable option for default styles in the configuration. Construction entities with custom styles respect this option as well.  

This also addresses a problem with missing contour fill for exported SVG.

To test this try exporting a sketch with construction entities.  The default behavior is to skip them matching how SS behaves now.  Then enable the export option for the construction entity style and re-export.  You may also turn on/off other styles.  Contour fill will now be exported for each style when exporting SVG

**NOTE**:  Running this version will add new entries to the configuration (Registry, .config etc.) on exit when testing this code.

### DIFF
--- a/src/drawentity.cpp
+++ b/src/drawentity.cpp
@@ -462,7 +462,7 @@ void Entity::GenerateBezierCurves(SBezierList *sbl) const {
 
     // Record our style for all of the Beziers that we just created.
     for(; i < sbl->l.n; i++) {
-        sbl->l[i].auxA = style.v;
+        sbl->l[i].auxA = Style::ForEntity(h).v;
     }
 }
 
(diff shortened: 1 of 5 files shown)


phkahler:
Does this close #957 ?

robnee:
@phkahler No, this addresses a much older issue, #682.

Issue #951, opened 2021-03-07
Tangency between 2 Bezier splines


Tanzit555:
Hi guys,

  When I try to set a tangency between 2 Bezier splines connected together by their extremities, It only sets it on one connection, not the other... Normal behavior ???

Thanks for your replies

![tangent](https://user-images.githubusercontent.com/80163599/110225207-5d203580-7ee3-11eb-999a-b1baee42b2c8.png)

### System information

- **SolveSpace version:** 
- **Operating system:**

robnee:
I'm pretty hopeful that this can be fixed.  It's really two separate constraints, one for each pair of endpoints.  The problem is when attempting to apply the constraint the second time SolveSpace sees it as redundant and ignores the request.  The PR demonstrates that this can be fixed but still needs some work.

phkahler:
The work around is to make a construction line from the end of a curve to its adjacent control point. Then constrain the closest control point from the other curve to that construction line.

@robnee The fix for this may have to wait until after 3.0 because we don't want to break any existing sketches if it later needs to be tweaked. But we'll see ;-)

robnee:
@phkahler Yes, this is trickier that it first appeared.

ruevs:
Related to #937.

Evil-Spirit:
Don't forget about 2nd order tangent

ruevs:
With #937 is implemented in https://github.com/solvespace/solvespace/pull/1493 is is now possible to do this by first making the two ends tangent and then making them coincident. I should still look into why it does not work normally.

ruevs:
After the [above fix](https://github.com/solvespace/solvespace/issues/951#issuecomment-2460221949) it is possible to do this (constrain both ends tangent). To which point the constraint is applied depends in the order in which the beziers are selected. And I think this makes sense.

In my opinion we should close this @phkahler what do you think?

Issue #937, opened 2021-02-15
Allow tangents on curves/arcs/lines without coincident endpoints (Was: "Arc of circle tangential to line segment kind of works but is broken")


baryluk:
### System information

- solvespace from master @ 7795b0bc1911a01eaa4661e252488d8f07ec2135
- Debian testing, amd64
- gcc 10.2.1

[dual-tangent-arc-on-segment.slvs.zip](https://github.com/solvespace/solvespace/files/5980362/dual-tangent-arc-on-segment.slvs.zip)

So apparently, one is supposed to have a line segment that shares the end point with end of an arc, to make tangential constraint work.

But in fact solvespace allows right now to create a tangential constrain without that limtiation. One can for example do a tangential constrain of arc end with any line segment, or with a line segment that has end of an arc on line, but not on end of a segment. It still shows error message, but still creates a constrain and it works.

However, trying to create two such constrains (one for each end of an arc), create a wrong constraints.

![Screenshot at 2021-02-15 08-44-04](https://user-images.githubusercontent.com/110322/107923917-0f5e8000-6f6a-11eb-8c1d-1c66bbb7edb5.png)

https://user-images.githubusercontent.com/110322/107923761-d9210080-6f69-11eb-8ff3-a1294824c1f6.mp4

ruevs:
In some sense this is a regression - 2.3 does not do it.

On the other hand being able to apply the tangential in these "more relaxed" cases may be useful. What I think it does is apply the tangential constraint always to the "starting (first) point" of the ark - since if you think about it it is not clear which one should be used with the current UI.

It can potentially be changed to accept an ark, a line segment and one of the end points of the ark. I'm curious when and how this happened, but have no time to dig now... Maybe later.

And the error dialog shuld be rethought in any case.

ruevs:
By the way the situation is the same with cubic splines.

ruevs:
Aha! I know EXACTLY where this got introduced/broken: 96958f466395148ee7f0a252222a2a27e7345c22 the `return` statements got "swallowed" by the newly created functions. (e.g .https://github.com/solvespace/solvespace/commit/96958f466395148ee7f0a252222a2a27e7345c22#diff-49abc03ed071148c0ebae0c64aafb625fa6223135f77aeecdb47dab4cf8b940cL638 )

Here is the pull request where this was done https://github.com/solvespace/solvespace/pull/833.

[...]

PR #954, opened 2021-03-07
Allow constraining both sides of 2 curves or a line and a cubic

robnee:
There didn't really seem to be anything that prevented this from working from the data-model point of view.  The code that applies the constraints just needed to check if the other two points were already constrained.  If so it just tries to constrain the other two points.  You must constrain the two entities twice to constrain both ends.  That said, this feels a little ugly and I wonder if there is a more elegant way to do it.


robnee:
This doesn't properly handle attempting to constrain the entities tangent a second time if their remaining endpoints are not already coincident.  It redundantly constrains the first two points.

This also cheats when it detects the existing constraint.  It rewrites that first constraint to constrain the other points and then proceeds to constrain the first two points again.  It works but it might cause a problem somewhere else, changing the constraint after it's been written.

These can be fixed but the code started to get even uglier.  This is still a work in progress.

robnee:
@ruevs Didn't mean to conflict with your commit.  You might be in a better position to fix this since I think you are working on the tangent  #937.  It's tricky though.  There's a couple of frustrating edge cases.

ruevs:
Hmm I need time to think about this and debug through it...

ruevs:
See here https://github.com/solvespace/solvespace/issues/951#issuecomment-2888482170

ruevs:
I locally rabased this on master and it works.

It has an interesting side effect when a curve and line do not share end points - if the user selects the two entities and one end point of the curve and applies the tangent constraint it works as expected. If he repeats exactly the same a second time **the other end** of the curve gets constrained. Not surprising and perhaps not a big problem.

But see here https://github.com/solvespace/solvespace/issues/951#issuecomment-2888482170 in current master it is possible to achieve this without this PR by selecting the entities in a different order.

Hmmmm...

Issue #325, opened 2018-05-31
Change rotation behavior?


nazar-pc:
### System information

SolveSpace version: 2.3

Operating system: Ubuntu 18.10

### Additional information

I'm new to solvespace, but would say that the way rotation is implemented in solvespace is at very least strange. Every time I try to rotate model with mouse, it does almost exactly the opposite of behavior I'm used to in other programs like free online Tinkercad. To be clear, I'm not talking about mouse/keyboard keys, but rather about rotation process. I almost want to invert it, at least partially.

Is there any way to configure how rotation happens? I didn't find any relevant issues or settings in documentation or config file.

whitequark:
There is currently no way to configure the controls. This has been discussed before; pretty much every CAD program uses its own controls, so modifying ours would inevitably displease everyone except people who use one particular CAD program. It seems pointless.

nazar-pc:
I've tried FreeCAD briefly, it had a few hardcoded presets for this behavior that one can chose from. Do you think something like this can be added to SolveSpace?

whitequark:
Not easily, the code that handles controls is not currently written in a way that is amenable to this sort of customization. It'd require a major refactor.

nazar-pc:
I believe it would be useful for many people. So if there is a chance someone will tackle this in future, would be nice to label this as feature request.

notnot:
The rotation controls are also counter-intuitive to me. It is almost trial and error to get the view i am after...

whitequark:
Does it help if you tick the "use turntable mouse navigation" checkbox on master?

nazar-pc:
Unfortunately no, it is even more confusing with that option, can't imagine someone would actually want that.
The primary issue I have is that any circular-like motion turns the model in the opposite direction to what I expect. I've got used to most of other things.

hinell:
Thumbing up. Even with turntable is enabled it still feels quirky compared to blender.
I think rotation around the `X` axis is not correct. 

https://user-images.githubusercontent.com/8136158/151233874-cc101e33-cbdb-45aa-a91b-47adf80d08e6.mp4


https://user-images.githubusercontent.com/8136158/151233878-34df3b59-ebba-4831-9661-42ed34bdf304.mp4

Symbian9:
> solvespace-rotation.mp4

[...]

Issue #888, opened 2021-01-12
Feature Request: Trackball Rotation


giannissc:
I was wondering if it would be useful to have trackball rotation (as described [here](https://www.mattkeeter.com/projects/rotation/).

It seems that some people have trouble with the existing rotation system #325 so maybe that would solve it or at least offer another option for people who don't like the existing system.

Thanks for all your good work!

ruevs:
There is turntable navigation (find the option on the configuration screen).
There is also `CTRL+middle button drag` - try it it's useful.

phkahler:
So the turntable is kind of nice. That CTRL-middle-mouse is terrible. What is that even doing? It seems to affect the turntable somehow - messing it all up! Does anyone use that?

ruevs:
`CTRL+middle button drag` rotates around the center of the view port in the plane of the screen. It is very useful in the "normal" mode - I use it all the time.

When turntable navigation on it behaves stupidly - I do not know if anyone ever considered this - turntable navigation was merged in fairly recently (e67f967933e Author: Dynamo Dan 2016-12-29 16:34:30 Committer: whitequark 2019-05-23 18:45:15)

vespakoen:
It would be nice to have Turntable nav on all axes (I experimented with this once but failed to get the right transform for one of the axes)

Ah, just found it: https://github.com/solvespace/solvespace/compare/master...vespakoen:turntable-axes

I was also thinking that locking the rotation to a single axis while rotation an assembly could be helpful.

I am not sure what is the best way yet but this area could really use some improvement.

PR #956, opened 2021-03-07
Add config option for "camera" rotation navigation

robnee:
issue #325 requests an alternative to middle-button mouse navigation/rotation.  What I think the user is asking for is an option to manipulate the camera instead of model when rotating, i.e. reversing up/down and left/right directions.  This PR implements that feature (independent of turntable navigation) as a config option called "camera mouse navigation" (name could use some work).  It drives me a little bit crazy but I think some are used to this and would welcome this option.

### DIFF
--- a/src/mouse.cpp
+++ b/src/mouse.cpp
@@ -134,8 +134,9 @@ void GraphicsWindow::MouseMoved(double x, double y, bool leftDown,
         double dy = (y - orig.mouse.y) / scale;
 
         if(!(shiftDown || ctrlDown)) {
-            double s = 0.3*(PI/180)*scale; // degrees per pixel
-            if(SS.turntableNav) {          // lock the Z to vertical
+            double sign = SS.cameraNav ? -1.0 : 1.0;
+            double s = 0.3*(PI/180)*scale*sign; // degrees per pixel
+            if(SS.turntableNav) {               // lock the Z to vertical
                 projRight = orig.projRight.RotatedAbout(Vector::From(0, 0, 1), -s * dx);
                 projUp    = orig.projUp.RotatedAbout(
                     Vector::From(orig.projRight.x, orig.projRight.y, orig.projRight.y), s * dy);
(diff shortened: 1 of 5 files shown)


hinell:
Certainly thumb this up.

hinell:
image

An option to turn the turntable can be found in the config.

phkahler:
@hinell  That option switches to a different style of rotation. The "camera" rotation is very different and acts as though your mouse is moving the camera rather than the model/sketch. So dragging to the right causes you to view the model from the right instead of rotating the model to the right which would show you the left side of it. At least that my understanding. Moving the model seems very natural to me and the idea of moving the camera would be confusing, but I can see where camera movement seems natural to some people and moving the model like we do now would be very confusing.

We'll revisit this soon. When this PR came along none of this was clear to me ;-)

ruevs:
Note to self look at my pending reviews more often.
@phkahler this is pretty clean and looks nice, we should review/try it and consider it for 3.2.

src/mouse.cpp
```diff
@@ -134,8 +134,9 @@ void GraphicsWindow::MouseMoved(double x, double y, bool leftDown,
         double dy = (y - orig.mouse.y) / scale;
 
         if(!(shiftDown || ctrlDown)) {
-            double s = 0.3*(PI/180)*scale; // degrees per pixel
-            if(SS.turntableNav) {          // lock the Z to vertical
+            double sign = SS.cameraNav ? -1.0 : 1.0;
+            double s = 0.3*(PI/180)*scale*sign; // degrees per pixel
+            if(SS.turntableNav) {               // lock the Z to vertical
```
ruevs, line 139: And also I think this fixes the `CTRL-middle drag` "hack" of turntable navigation [with this line](src/mouse.cpp) .
https://github.com/solvespace/solvespace/issues/888#issuecomment-759855825
https://github.com/solvespace/solvespace/pull/1307#issuecomment-1304587916

phkahler:
As much as I hate it, I know people who want it find the existing behavior equally terrible ;-)  It works fine for me on Linux and it's a really small PR!

Issue #958, opened 2021-03-08
Cancel editing Dimension using ESCape key


gnbl:
This is SolveSpace version 3.0~36719c62 on Windows 10.

As editing dimensions uses the keyboard, it would be nice (and even expected) to be able to cancel editing by hitting the escape key, the same way return/enter "submits" the new value.

robnee:
I can reproduce this on Win 10 as well.  On Linux Esc cancels the edit as expected.  Seems like a bug on Windows.

PR #959, opened 2021-03-09
Hide editor window when user hits Esc on Windows

robnee:
#958 notes the hitting Esc does not cancel an edit as it does on Linux.  This PR hides the editor on Esc.

### DIFF
--- a/src/platform/guiwin.cpp
+++ b/src/platform/guiwin.cpp
@@ -1095,7 +1095,7 @@ public:
                         return 0;
                     }
                 } else if(wParam == VK_ESCAPE) {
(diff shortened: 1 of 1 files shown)


gnbl:
That was quick, thanks!
You are aware that the latest Workflow run (https://github.com/solvespace/solvespace/actions/runs/641166887) failed and did not upload any binaries to https://github.com/solvespace/solvespace/releases/tag/edge despite saying it built the artifacts and uploaded to releases?

gnbl:
Uh, I just now managed to download the artifact and the issue is indeed fixed. Thanks again.

PR #965, opened 2021-03-20
Update uk_UA.po

Symbian9:
- https://github.com/solvespace/solvespace/issues/890

### DIFF
--- a/res/locales/uk_UA.po
+++ b/res/locales/uk_UA.po
@@ -1,20 +1,21 @@
 # Ukrainian translations for SolveSpace package.
 # Copyright (C) 2017 the SolveSpace authors
 # This file is distributed under the same license as the SolveSpace package.
-# AppSoft4 <appsoft@ua.fm>, 2017.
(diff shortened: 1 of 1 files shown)


Symbian9:
This is NOT yet complete, BUT I'm working on it!

@phkahler, is there deadline date for `3.0` contributions?

phkahler:
> This is NOT yet complete, BUT I'm working on it!
> 
> @phkahler, is there deadline date for `3.0` contributions?

@Symbian9  There's no specific date, it's already longer than I wanted but I'm hoping mimalloc gets a new release first (they're late too). I think a quick update to the translation filling in the missing strings would be good, then if you want to go over it to tweak wording you could submit another one. That way we'd have a "good enough" update to go any time.

Symbian9:
> I think a quick update to the translation filling in the missing strings would be good, then if you want to go over it to tweak wording you could submit another one. That way we'd have a "good enough" update to go any time.

Okay, you may merge this pull now as is!

I will commit next updates as soon as possible in another pulls.

Issue #847, opened 2020-12-10
Space Navigator gone?


phkahler:
There are people on the forum noticing that Spacemouse is not working. i looked at the build instructions and under Windows it says it will no be supported.  I don't have one, but when was this removed? Is it still supported on Linux or macOS?

jwesthues:
I don't think it's actually removed? The code still seems to be there, but `#ifdef HAVE_SPACEWARE`. So I'm thinking the standard build just doesn't enable that, perhaps because the library in question is an opaque binary blob from the hardware vendor, rather impure. I'd have no objection to re-enabling, though.

phkahler:
Do we statically link the blob or is there a header file for it or what? Also, how does this vary by platform?

jwesthues:
Under Windows, statically linked, library and headers at

https://github.com/solvespace/solvespace/tree/master/extlib/si

This library is provided by the hardware vendor. It's a 32-bit library, so it's disabled for 64-bit builds, which perhaps is the reason for the statement in the build instructions you note above. That doesn't explain why it would be broken in the RC though. Perhaps check if the build process is actually finding the libraries and enabling `HAVE_SPACEWARE`?

I didn't write the support for other platforms, so I'm less familiar there. On brief inspection, looks like Mac uses the vendor-provided library, but dynamically links to a version that's already installed. Linux uses libspnav, an independent and free implementation.

phkahler:
@vespakoen and @ppd  I looked at a recent CI build from one of the PRs and it does print the line "Using prebuilt SpaceWare" for the Windows build.  From the CMakeLists.txt that would be here:

```
        if(MSVC AND ${CMAKE_SIZEOF_VOID_P} EQUAL 4)
            message(STATUS "Using prebuilt SpaceWare")
            set(SPACEWARE_FOUND TRUE)
            set(SPACEWARE_INCLUDE_DIR
                "${CMAKE_SOURCE_DIR}/extlib/si")
            set(SPACEWARE_LIBRARIES
                "${CMAKE_SOURCE_DIR}/extlib/si/siapp.lib")
        endif()
```
However I don't see any other references to SPACEWARE.  How to tell if it's actually getting linked?

macOS seems to always build with it, while for guigtk.cpp there is this nice condition:
```
#if defined(HAVE_SPACEWARE) && defined(GDK_WINDOWING_X11)
```
That GDK_WINDOWING_X11 has me scratching my head.

ruevs:
[...]

PR #968, opened 2021-03-24
Libspnavdev

rpavlik:
I hadn't realized spacemouse support was depending on a 32-bit-only binary library on Windows. I just got a SpacePilot. This is a WIP to use another very WIP library to access the space devices on Windows 64.


phkahler:
> Alternately, it would be slick if we could use a VRPN device as our input, since then we'd get all the spacemice on all platforms, plus DIY stuff, etc.

Was going to ask, but google.  I recently got a Quest 2 and was wondering how hard it would be to get SolveSpace running on that. Initially it could just be 2 floating windows with, as you say, some VR manipulation like spacemouse. It could later move to something more interesting perhaps moving around in model space and putting up a semi-transparent plane just for 2D sketches. Oh to have time...

ruevs:
:-D I bought a SpaceExplorer off of eBay a couple of months back (because of this https://github.com/solvespace/solvespace/issues/847) and just had to write a library. Then I found the  the one you found - it is too early WIP, so I wrote the USB device support :-) (https://github.com/FreeSpacenav/libspnavdev/pull/1)

Then I was thinking of doing (something similar to) what you just did here :-)

ruevs:
By the way the official 3Dconnexion SDK supports x64 in it's newer versions. I have **_not_** registered with them and downloaded it from the [official site](https://3dconnexion.com/us/software-developer-program/), but here is the newest copy I could find on GitHub:
https://github.com/n8vm/Foton/tree/master/external/spacemouse
look in `/Lib/x64`
The latest date in the headers is 2018-02-19, while [our current one](https://github.com/solvespace/solvespace/tree/master/extlib/si) is from 2005.

When I got the SpaceExplorer I updated the official library locally (by replacing our files with a minimal working set from the above link) and it compiles and works without any changes to the SolveSpace source code.

So this may be a simple and cleaner way to support 3Dconnexion on x64 on Windows. But I have not read the license carefully (says a lot of things about "do not copy", distribute etc.) and that is why I did not make a pull request.

What do you guys think? And @jwesthues ?

P.S. Yes is is more fun to write our own - and since these devices have been around for 30+ years the protocol(s) are well known - but it will be much more work to support all models (especially the buttons beyond "Fit") well.

ruevs:
@rpavlik Why are you using and x64 build of SolveSpace - do you notice some benefit? Or just for experimentation?

jwesthues:
Assuming the relevant license is

https://github.com/n8vm/Foton/blob/master/external/spacemouse/Doc/LicenseAgreementSDK.txt ,

the new official driver license is quite different from the old one. The advertising clause in (2) is obnoxious, but we could comply. The confidentiality clause in (5) is the bigger problem, and seems fundamentally inconsistent with use in any free software project. I doubt anyone at 3Dconnexion would sue us, but I'd rather be strictly compliant than rely on that.

So I'd be inclined either to keep the old official drivers, or to move to non-official drivers (which as you note may have worse compatibility across their line, but would also remove our only binary dependency). Thoughts?

ruevs:
Jonathan, after posting my comment I decided not to be lazy, read the license (the same one you linked) carefully and came pretty much to the same conclusion as you did.

The confidentiality clause makes it pretty much a "no go", but here is a full list of problems I noticed:

1. "non-transferable" "non-sublicensable"
2. We could add the "attribution" on the site - since that is our ""Help" file (or equivalent)" - so not a big problem.
3. OK
4. OK
5. Completely incompatible with GPL, we will be violating this clause as soon as we commit it to GitHub (we may not be violating GPL since one can always compile without the library).
6. Termination... what we want to do immediately activates this clause.
7. "Nice" :-D:
"3Dconnexion may, in its sole discretion, modify portions of this Agreement at any time."

So I'll continue to work on the library (libspnavdev) and perhaps when it becomes good enough to merge we can keep both. If the 3Dconnection software is installed and `si` linked - use it. Otherwise use our own.

By the way for just 6DOF motion events with USB devices it works just fine already. The button abstraction will need another (non-existing at his point) layer.

Yesterday I committed initial rough support corn(ancient) RS-232 devices. This is another advantage of writing ournown we can support all old devices on all platforms (unlike the official driver). They seem to be well made and last long (apart from the sticky sof touch coating) so they are easy to find and are cheap second hand, which is an advantage for our users. I bought my SpaceBall 3003 for 15€ :-)

rpavlik:
I default to 64 bit builds of stuff these days even on Windows. (And actually I first tried the ci build, which is also 64) I assume it's somewhat faster because of newer instruction sets.

VRPN is bsl-1.0 licensed (very permissive) and has both serial and hid space nav drivers in it. I only started this because there was already that one driver and I assumed it would be easier, but it's probably easiest to just generalize the 6dof input a bit more and make a VRPN module. Then you can even have your space mouse networked 😁

phkahler:
@rpavlik is there any chance VRPN could work with the emscrpten version? Solvespace can run in the web browser on the Quest and probably other VR setuos.

ruevs:
I asked a question about the license in the 3Dconnexion forum https://forum.3dconnexion.com/viewtopic.php?f=19&t=39870#p83415

krisrok:
@ruevs just a thought regarding licensing: Wouldn't it suffice to pull in the .dll from GAC and be compliant? The user would need the have the driver package installed anyway. That's windows-related only of course.

ruevs:
> @ruevs just a thought regarding licensing: Wouldn't it suffice to pull in the .dll from GAC and be compliant? The user would need the have the driver package installed anyway. That's windows-related only of course.

We do not need to "pull the DLL" from the [Global Assembly Cache](https://social.technet.microsoft.com/wiki/contents/articles/29637.understanding-the-global-assembly-cache-gac-and-managing-copying-dlls-from-global-assembly-cache.aspx) - it "Just Works". Our problem is that we use [their header files](https://github.com/solvespace/solvespace/tree/master/extlib/si) and we are linking against a small static library "[siapp.lib](https://github.com/solvespace/solvespace/blob/master/extlib/si/siapp.lib)" and the (new) [license](https://github.com/n8vm/Foton/blob/master/external/spacemouse/Doc/LicenseAgreementSDK.txt) states:

> 5. Confidentiality
The source code and instructions contained in the 3DCONNEXION SDK are trade secrets of 3Dconnexion and you agree to maintain the confidentiality of such source code

krisrok:
Ok thanks for clarifying.

ruevs:
It's been a while.

The original [libspnavdev](https://github.com/FreeSpacenav/libspnavdev) is "dead".

So I updated master on my fork (https://github.com/ruevs/libspnavdev) to my working implementation and added build instructions. It should work with all SmaceMouse devices and I have tested it with nine of them.

Since 64 bit SolveSpace on Windows seems to be more and more needed (faster SolveSpace, faster PCs, bigger models, 3/4Gb RAM runs out) I updated my  branch based on @rpavlik's here https://github.com/ruevs/solvespace/tree/rpavlik/libspnavdev. It is rebased on master and links to the latest version of  `libspnavdev`. To test it make sure to get submodules recursively (I include HIDAPI as a submodule) and build a 64 bit SolveSpace on Windows.

@phkahler @rpavlik @jwesthues we should consider if we want to merge it.

P.S. The 3DoF (USB) devices that I tested are marked here
https://github.com/ruevs/solvespace/blob/72a4d8026942717f1dae6d8c3fd41bcf23d73300/src/platform/spnavdevice.cpp#L107
if you have others - please test if they work.

ruevs:
Note to self on the subject of licensing - when Thomas Buck implemented 3Dconnexion support on macOS in fb62e0494b499532a97c59b660255a5e24ec446e he considered the license restrictions of the official SDK:

https://github.com/solvespace/solvespace/blob/a208201c32d52fac366da74e21dfb3ae4d8eff65/src/platform/guimac.mm#L1148

Issue #939, opened 2021-02-17
Poor Linking performance


giannissc:
I have noticed that the more complex the components become the slower the performance gets in solvespace. This is particularly apparent when linking said complex components. Just changing a size constraint can take up to 10 seconds to complete.

Do you know what might be causing this?
Is there anything I can do to get the most performance out of solvespace?

ruevs:
First:
- Use a release build (if you are building from source) - is is much faster than debug.
- Use the OpenMP (multi core) version - uses all the CPU cores for some operations and is somewhat faster for them.

That said SolveSpace does get in some cases. There are three main bottlenecks:
- Too many surfaces and thus triangles to display. You can alleviate this by increasing the `chord tolerance (in percents)` on the configuration screen.
- Too many constraints (complex sketches) - you can not do anything about it. (https://github.com/solvespace/solvespace/issues/841, https://github.com/solvespace/solvespace/issues/225)
- Too many entities (objects) - you can not do anything about it. (https://github.com/solvespace/solvespace/issues/932)

giannissc:
Would using a GPU help as opposed to just using the CPU?

phkahler:
> Would using a GPU help as opposed to just using the CPU?

Nope. Solvespace has several algorithms that have O(n^2) time complexity. Most of them could be improved to O(n log n) but it is a lot of work and won't really pay of well until we fix all of them.

A highly readable account of a similar situation is here: https://randomascii.wordpress.com/2021/02/16/arranging-invisible-icons-in-quadratic-time/

I've got a really efficient spatial index I'm going to try to dig up that may help some of the issues in solvespace. It was good enough to ray trace things like solvespace models in real time on a CPU so I'm kind of looking forward to locating it ;-)

ruevs:
The article is fantastic :-)

phkahler:
Linking gets slow I believe because it creates a group with a large number of constraints and only 6 degrees of freedom. Every point in the linked sketch gets a copy created with a rotate/translate transformation applied based on the same 7 parameters. I'm under the impression that the constraint solver/eigen improvements will bring a large performance win in this area.

pjanx:
Would money help?

pjanx:
I've tried the _eigen_over_master_ branch, and it doesn't help my [cursed model](https://github.com/pjanx/nyquist-case/tree/master/levinson_choc_case), seemingly at all, nor did OpenMP, seemingly at all. There don't seem to be any issues either, though.

[...]

Issue #932, opened 2021-02-11
Poor performance when extruding translate groups


baryluk:
### System information

- solvespace from master @ 7795b0bc1911a01eaa4661e252488d8f07ec2135
- Debian testing, amd64
- gcc 10.2.1

### Details

I was doing a small project for fun, and went a bit over board with some holes in a sheet metal, to the point where it takes about 4 minutes to render do anything useful.

I used an extrusion of translate of translate to generate a 2D pattern.

[case-vents.slvs.gz](https://github.com/solvespace/solvespace/files/5962704/case-vents.slvs.gz)  (~20MB unpacked, can take few minutes to open, be warned).

Looking at `perf top`, and a bit of code in `Generate`, it looks like the major culprit it CopyEntity, which triggers a copy of Entity classes. The base class (`EntityBase`) has two `std::string` fields `str` and `font`, which I believe should not be in the `EntityBase`, as most of the entities doesn't use these fields actually. Because they are there directly, each construction and destruction, will be called every time any entity is copied (I don't know if it will also trigger malloc/free, which could be even worse). In perf I see `~basic_string<...>`, using about 90% of the CPU. `Platform::Path file` from EntityBase, also shows quite a lot in `perf top`. Other option is to use reference counting for strings maybe. Not sure.

ruevs:
Yes it is slow on the path you described...
But for me on Windows (RelWithDebInfo) most of the time is spent in `SolveSpace::Entity::operator=` on this path.

```
Function Name	Total CPU [unit, %]	Self CPU [unit, %]	Module	Category
|||| + SolveSpace::Platform::RunGui	44611 (83.83 %)	0 (0.00 %)	solvespace.exe	UI | IO | Graphics | File System | Kernel
||||| + [External Code]	44611 (83.83 %)	55 (0.10 %)	Multiple modules	UI | IO | Graphics | File System | Kernel
|||||| - std::_Func_impl_no_alloc<<lambda_8b75e19112ec9ef5254c574e001b6b54>,void>::_Do_call	26118 (49.08 %)	0 (0.00 %)	solvespace.exe	IO | Kernel
[...]

PR #973, opened 2021-03-27
Fix pathologically slow translate groups on Linux

pjanx:
Major performance improvement on GCC with libstdc++, see my comments in #939. Relates to discussion in #932.

All tests passed, nothing interesting in memcheck showed up.

I think this belongs in 3.0 and is obvious enough.

I've tried to use `std::move_backwards` but it didn't look better.

### DIFF
--- a/src/dsc.h
+++ b/src/dsc.h
@@ -467,11 +467,17 @@ public:
         // Look to see if we already have something with the same handle value.
         ssassert(FindByIdNoOops(t->h) == nullptr, "Handle isn't unique");
 
-        // Copy-construct at the end of the list.
-        new(&elem[n]) T(*t);
+        // Find out where the added element should be.
+        int pos = LowerBoundIndex(*t);
+
+        // Shift everything from there to the end of the array.
+        new(&elem[n]) T();
+        for (int i = n; i > pos; i--)
+            elem[i] = std::move(elem[i - 1]);
+
+        // Copy-construct at the right place.
+        elem[pos] = T(*t);
         ++n;
-        // The item we just added is trivially sorted, so "merge"
-        std::inplace_merge(begin(), end() - 1, end(), Compare());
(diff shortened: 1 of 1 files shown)


ruevs:
Ohhhh I can't wait go get to a PC to try this!

ruevs:
OK I tried it - on this model https://github.com/solvespace/solvespace/issues/932#issuecomment-778506906 the time to set the last translate group to 6 copies is:
- 21 seconds before this patch
- 15 seconds after it.

ruevs:
By the way the hot path is still moving all the entities "one down" from the insertion point.
20210329_ProfilingAfter973

pjanx:
Yes, it is still doing a lot. The `inplace_merge` was just doing something barely comprehensible, as I saw in Debug callgrind output.

Just now I can make something as simple as a keyboard case without tearing my hair out. :)

Perhaps a different data structure entirely would be appropriate, but I don't have any obvious favourites.

The other thing that came to my mind was postponing the sort, which looked complicated.

rpavlik:
Wow, I think I touched that last before you. Blows my mind these are different performances, I would have expected nearly equivalent assembly after the optimizer had its way.

phkahler:
@rpavlik it seems the people on the C++ committee have a new definition of "in-place". It needs to allocate extra space, can take O(N logN) time, and uses custom comparator. No idea why it would take worse than linear time give the need to allocate more space. Seem like a good function to avoid.

rpavlik:
Could we just do an insert instead of manually moving things?

ruevs:
> Could we just do an insert instead of manually moving things?

I'll try it. Anyway I am deep into it https://github.com/solvespace/solvespace/issues/939#issuecomment-810641820 But `IdList` is not quite a container (in the stl sense).

PR #974, opened 2021-03-27
Save stipple settings for default styles to config

### LINKED ISSUE #957: stipple width
Hi guys,

In the settings, "stipple width" is the area where the targeted element is selected right ?
When I modify it in the settings, it doesn't keep the setting saved, when I restart SolveSpace, it returns back to the default value (15).
All the other settings I modify are kept.

Any ideas ??

Thanks a lot...

Win 10 home 64

3.0RC2

robnee:
This PR saves the stipple settings for default styles to the configuration and closes #957.

### DIFF
--- a/src/sketch.h
+++ b/src/sketch.h
@@ -884,11 +884,14 @@ public:
         double      width;
         int         zIndex;
         bool        exportable;
+        StipplePattern stippleType;
     } Default;
     static const Default Defaults[];
 
     static std::string CnfColor(const std::string &prefix);
     static std::string CnfWidth(const std::string &prefix);
+    static std::string CnfStippleType(const std::string &prefix);
+    static std::string CnfStippleScale(const std::string &prefix);
     static std::string CnfTextHeight(const std::string &prefix);
     static std::string CnfPrefixToName(const std::string &prefix);
     static std::string CnfExportable(const std::string &prefix);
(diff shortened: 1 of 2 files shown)


phkahler:
Does it fix #957?  You an link PRs to issues over on the right so merging will automatically close the issue if that's appropriate.

robnee:
@phkahler Yes it does.  There was some additional discussion about import/export of default styles but that should be it's own issue.  Also, thanks for the tip on linking to issues!

robnee:
@phkahler I linked to the issue using the "closes" keyword in the description.  I'm not sure if I can manually link it.  There is no gear icon on the RHS for me.  Is this something only available to the 4 SS members?

ruevs:
> @phkahler I linked to the issue using the "closes" keyword in the description. I'm not sure if I can manually link it. There is no gear icon on the RHS for me. Is this something only available to the 4 SS members?

It's possible. I'm not sure. By the way there are 6, but two are have set their membership to "private" and I presume are not visible for non-members.

phkahler:
> @phkahler I linked to the issue using the "closes" keyword in the description. I'm not sure if I can manually link it. There is no gear icon on the RHS for me. Is this something only available to the 4 SS members?

The sidebar is not visible to me on mobile, only desktop. And looking at repositories I have no rights to there is no gear. Can't find an options to enable it either :-(

robnee:
I was able to link it by editing by initial comment and replacing the word "addresses" with "closes"  Seems GH looks for certain keywords.

Symbian9:
@robnee, could you check, does latest commits in hidden style introduced incorrect default stipple type for `s00e-#def-hidden-edge` style:

- https://github.com/solvespace/solvespace/issues/1027#issuecomment-826293943

PR #976, opened 2021-03-27
Allow right click to end sketching of all entities

robnee:
Right click can be used to end sketching of new lines and cubics as a convenient alternative to ESC.  This PR extends this feature to work when sketching rectangles, circles, arcs, ttf text and images as well.  Currently right clicking while sketching, say a rectangle, just pops up a context menu with one option, "Select All".  While choosing Select All does work I can't figure out why that would be useful in the middle of sketching.

I've been sitting on this for a while because I wasn't sure if this was of general interest.  My fingers have gotten very used to the right click feature for lines and cubics so I thought I would submit to see others think.

### DIFF
--- a/src/mouse.cpp
+++ b/src/mouse.cpp
@@ -526,11 +526,16 @@ void GraphicsWindow::MouseRightUp(double x, double y) {
     }
 
     if(pending.operation == Pending::DRAGGING_NEW_LINE_POINT ||
-       pending.operation == Pending::DRAGGING_NEW_CUBIC_POINT)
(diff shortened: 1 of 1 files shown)


phkahler:
Maybe the "select all" option makes sense to toggle to construction after drawing something. If so, that's not needed now that we can toggle while drawing.

robnee:
@phkahler Yes, after drawing I can see that.  I can't think of why you might need to select all in the _middle_ of a drawing operation.  For instance, while setting the radius of a circle you can (prior to this PR) select all and then resume setting the radius.  I'm not sure why that would be useful so I'm hoping that this feature doesn't constitute a regression.

PR #978, opened 2021-03-28
GTK: make spacenavd work through its socket, too

pjanx:
This resolves the issue I discussed on the IRC channel where Blender worked but Solvespace didn't while both use libspnav.

I'm also strongly proposing this for 3.0.

```
E.g. for when you have more than one user logged in, in which case
spacenavd works only for the first of those X11 displays.

We try to do it the old way first, so there should be no regressions.

Also fixes storing non-booleans in bool variables.
```

### DIFF
--- a/src/platform/guigtk.cpp
+++ b/src/platform/guigtk.cpp
@@ -34,6 +34,11 @@
 #if defined(HAVE_SPACEWARE)
 #   include <spnav.h>
 #   include <gdk/gdkx.h>
+#   if GTK_CHECK_VERSION(3, 20, 0)
+#       include <gdkmm/seat.h>
+#   else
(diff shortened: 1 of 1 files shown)


pjanx:
It'd actually be possible to make it work on Wayland, too, with minor edits. Not that I care. :)

phkahler:
> It'd actually be possible to make it work on Wayland, too, with minor edits. Not that I care. :)

I run Wayland all the time and personally have no use for X. ;-) 
But I dont have one of those devices though, so can't test.

pjanx:
I don't want to change too much in the RC window. Still thinking about a version check `#if` for the part that definitely requires GTK+ 3.20, I'll probably amend that to keep some peace of mind.

PR #983, opened 2021-03-31
minor fix open/save dialogue on windows

dustinhartlyn:
On windows 10 the open/save dialogue box has an minor error, and I believe I fixed it. 

When "Open" is selected from the menu, the title of the dialogue box says "SolveSpace - Save File" and the entered file name is "united". My fix correctly titles the dialoged box, and leaves the address bar blank when a file is being opened because "united" is only needed as a default name when a file being saved. 

I found that class FileDialogImplWin32 from guiwin.cpp contains two if statements for "isSaveDialog". This is redundant. I removed the first where the title was originally set, but not working. I then set the title in the second if statement and moved the 'if isEmpty'' to this section.

### DIFF
--- a/src/platform/guiwin.cpp
+++ b/src/platform/guiwin.cpp
@@ -1621,16 +1621,16 @@ public:
     }
 
     bool RunModal() override {
-	if(isSaveDialog) {
-	    SetTitle(C_("title", "Save File"));
-	    if(GetFilename().IsEmpty()) {
-		SetFilename(Path::From(_("untitled")));
-	    }
-	    return GetSaveFileNameW(&ofn) == TRUE;
-	} else {
-	    SetTitle(C_("title", "Open File"));
-	    return GetOpenFileNameW(&ofn) == TRUE;
-	}
+        if(isSaveDialog) {
+            SetTitle(C_("title", "Save File"));
+            if(GetFilename().IsEmpty()) {
+                SetFilename(Path::From(_("untitled")));
+            }
+            return GetSaveFileNameW(&ofn) == TRUE;
+        } else {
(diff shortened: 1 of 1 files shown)


src/platform/guiwin.cpp
```diff
@@ -1626,15 +1621,16 @@ class FileDialogImplWin32 final : public FileDialog {
...
-        }
-
-        if(isSaveDialog) {
-            return GetSaveFileNameW(&ofn) == TRUE;
-        } else {
-            return GetOpenFileNameW(&ofn) == TRUE;
-        }
+	if(isSaveDialog) {
+	    SetTitle(C_("title", "Save File"));
+	    if(GetFilename().IsEmpty()) {
```
ruevs, line 1626: Use spaces, not tab for white space.

ruevs (changes requested):
This looks good to me and does the right thing. Please just fix the white space.
Welcome. And thank you for the contribution.

src/platform/guiwin.cpp
dustinhartlyn: I am still new to GitHub, do I create another pull request, or just update the code and let you know?

phkahler: Push the change to your repository and it should show up here.

Better yet, fix the whitespace and squash the 2 commits, then push to your repository and it will show up here as a single commit PR.

It does take time to get used to this stuff :-)

phkahler: After squashing it would require using --force to your repo.

ruevs: Change the file.
git add guiwin.cpp
git commit --amend
got push yourremote --force

PR #988, opened 2021-04-02
Clamp lighting value on export to fix flat colors

### LINKED ISSUE #986: Ambient 1.0 settings ignored on PDF and SVG export
### System information

- **SolveSpace version:** `3.0~` https://github.com/solvespace/solvespace/commit/10ca307d986ab0c52faaab4fcaedb26bd42d1fd6

### Expected behavior

Export of shaded triangles should depend on `ambient lighting` settings.

@robnee, See *"Flat colors render mode"* thread:

- https://github.com/solvespace/solvespace/issues/524

Special use-case: ***if*** `ambient lighting` set to `1.0` *shaded triangles should be exported as "flat color"*

pic.0

### Actual behavior

Actually if set `ambient lighting` to `1.0` — triangles with same color value (according solid group color) are NOT rendered as in pure "flat colors" (see PDF and SVG in attached ZIP):

pic.1

In same time if *"File > Export Image..."* — it shown exactly as expected (see PNG file in attached ZIP):

pic.2

### Additional information

- [slvs-ambient-export.zip](https://github.com/solvespace/solvespace/files/6246731/slvs-ambient-export.zip)

robnee:
This closes #986.  The computed lighting value when ambient lighting value == 1.0 needed to be clamped to 1.0 in order to achieve flat shading as expected.

Also, this fixes the Printf format for the ambient lighting entry on the configuration screen.  This was causing strange behavior when attempting to edit the ambient lighting value on Windows.  Oddly this didn't seem
to affect Linux.

### DIFF
--- a/src/confscreen.cpp
+++ b/src/confscreen.cpp
@@ -236,9 +236,7 @@ void TextWindow::ShowConfiguration() {
             CO(SS.lightDir[i]), i, &ScreenChangeLightDirection,
             SS.lightIntensity[i], i, &ScreenChangeLightIntensity);
     }
-    Printf(false, "%Bp         ambient lighting     "
-                  "%2 %Fl%D%f%Ll[c]%E",
-            (i & 1) ? 'd' : 'a', i,
+    Printf(false, "%Ba         ambient lighting     %2 %Fl%f%Ll[c]%E",
         SS.ambientIntensity, &ScreenChangeLightAmbient);
 
     Printf(false, "");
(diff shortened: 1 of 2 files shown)


robnee:
I was a little groggy last night (vaccine) and I don't think the confscreen.cpp  change is quite right.  The code is using `i` outside the for loop still which is bad form.  The function `ScreenChangeLightAmbient` doesn't use its arg so it can be hard coded to 0 in the Printf.  For that matter the Printf format doesn't need the `%D` in its format either since the argument to the callback will just default to zero anyway.  I'd like to revise and push again.

Issue #841, opened 2020-12-08
Couple of issues drawing a symmetric drawing with a lot of tangent arcs


vespakoen:
### System information

SolveSpace version: 3.0 / edge (Release build with OpenMP)

### The issue

Just wanted to drop this test file over here, I am trying to draw this 8040 aluminium profile, like this one:

![item-profile](https://user-images.githubusercontent.com/876117/101467574-c58e6380-3942-11eb-949b-f7a9ecd4dd7d.jpg)

And the experience is not optimal, I think it might have to do with having a lot of tangent arcs, in any case, it is becoming pretty slow to draw.

Another reason why I am dropping this file here is because wanting to draw this made me find a couple of problems / missing features in SolveSpace.

This attached sketch only contains a quarter of the final drawing, and mirroring / rotating it was not possible because of #833 and (the now merged) #830, but even with those changes merged, it complaints that there are `Too many items to paste`.
I think this should be a higher number by default, I doubled it and the actual pasting of the drawing wasn't slow at all (maybe it's memory intensive?).

Even with the possibility of rotating and mirroring the drawing, It also duplicates all the dimension constraints 4 times, and of course in this drawing it would make more sense to control all the dimensions in only a single place.

So I guess what makes more sense here is to make a new group by step rotating / translating, however, this does not allow us to mirror yet, so it's not a really good option.

I have experimented with drawing the complete "top half" of the drawing and step rotating it, which works, but then I run into the issue that I cannot extrude the resulting drawing when imported in another sketch.

So in conclusion:

1) The drawing is slow to solve, this might actually already be fixed with other PR's, but I guess this file can be handy as a test case.
2) When using `Paste Transformed`, I think there would be a lot of added value if there would be an option to change dimension constraints into equal length constraints.
3) When pasting with Paste Transformed, there shouldn't be a limit to the amount of items on the clipboard, there also seems to be a timeout anyways, and when the duration of that can be configured, all should be good.
4) There is probably value in having a "Symmetric" group tool that can copy and mirror the drawing over 1 or 2 axes.
5) There is an issue extruding a Step Rotated (haven't tested Step Translated drawings) drawing when imported into another sketch.

[8040.slvs.zip](https://github.com/solvespace/solvespace/files/5658434/8040.slvs.zip)

phkahler:
[...]

Issue #738, opened 2020-10-15
All NURBS and mesh bugs and their test models in one place


ruevs:
This is a list of all SolveSpace issues related to [NURBS](https://github.com/solvespace/solvespace/labels/NURBS) or [mesh](https://github.com/solvespace/solvespace/labels/mesh) bugs.
Links to the `.slvs` models reproducing the problems and the fixes may be useful for regression testing.
If the "Fixed on" column is empty the issue is probably still open (or I did not update this list :-))

Source | Model | Author | Fixed on | Fixed in
------ | ----- | ------ | -------- | --------
https://github.com/solvespace/solvespace/issues/32 <br/> https://github.com/solvespace/solvespace/issues/540<br/>https://github.com/solvespace/solvespace/issues/683<br/>https://github.com/solvespace/solvespace/issues/683#issuecomment-683274604<br/>https://github.com/solvespace/solvespace/issues/683#issuecomment-730602943<br/>https://github.com/solvespace/solvespace/issues/732 | [failed.lathe.slvs.zip](https://github.com/solvespace/solvespace/files/408873/failed.lathe.slvs.zip)<br/>[adapter-plate.tar.gz](https://github.com/solvespace/solvespace/files/4054379/adapter-plate.tar.gz)<br/>[wrong.zip](https://github.com/solvespace/solvespace/files/5112393/wrong.zip)<br/>[wrong2.zip](https://github.com/solvespace/solvespace/files/5145107/wrong2.zip)<br/>[wrong3.zip](https://github.com/solvespace/solvespace/files/5569373/wrong3.zip)<br/>[coincident4.zip](https://github.com/solvespace/solvespace/files/5357050/coincident4.zip) | @jockbert<br/>@baryluk<br/>@starsolaris<br/>@starsolaris<br/>@starsolaris<br/>@phkahler | Not really 2020-10-09 | https://github.com/solvespace/solvespace/pull/737 https://github.com/solvespace/solvespace/pull/737/commits/194945f173a4c4f32ba7005ba543091cf0545741 https://github.com/solvespace/solvespace/commit/ab10e38d44b26de3ee64b60db0d154278d1ee8ac https://github.com/solvespace/solvespace/pull/786 https://github.com/solvespace/solvespace/commit/8a3e5b4d56b644808b4378a1f068486151a69ab2
[...]

Issue #120, opened 2016-11-25
Regression with translate+extrude in 2.1, 2.2


ruevs:
The attached trivial sketch shows a regression in 2.1 and 2.2 releases.

The circle in g003 is translated twice in g005 and in g004.
The two translated groups are extruded in g007 and g008.
The extrude in g007 is visible only when it is selected as a working group. when any of the later groups are selected the holes disappear.

This is a regression in 2.1/2.2 if I open the file in 2.0 the holes from both extrusions are visible.

whitequark:
> Github does not allow attaching a zip file

GitHub does in fact allow attaching zip files, I do that all the time...

ruevs:
It does not allow me.
"We don’t support that file type. Try again with a PNG, GIF, JPG, DOCX, PPTX, XLSX, TXT, PDF, or ZIP. "
:-D

whitequark:
[bug.zip](https://github.com/solvespace/solvespace/files/613625/bug.zip)

whitequark:
Basically this boils down to: what happens if there are groups between source group (for some solid operation) and the solid operation group?

  * Clearly the entities should be generated from the source group.
  * Right now the source mesh/shell that undergoes the solid operation is also taken from the source group. But arguably it should be taken from the previous group in order.

I'm hesitant to just change where the mesh/shell are taken from because this might break some old files. I'll need to think about it.

ruevs:
That explains it.

What is the correct/logical behavior is hard to say. However the example came from a real detail I was modelling and I can not think of an (easy) way to do it in 2.2 with the current logic. I ended up deleting all the extrusions, opening the file in 2.0 and re-creating them - which placed them at the end.

But if you open bug.slvs in 2.0 (or 1.9 that's the oldest I have handy) the holes from both translates are present. That means that in 2.0 the logic is that the mesh undergoing the solid operation is taken from the previous group. Another interesting thing is that after complaining about unknown/newer format 2.0 rearranges the groups in order of creation. I discovered this only when creating the example file to report the bug - too late for the real model :-)

whitequark:
> That means that in 2.0 the logic is that the mesh undergoing the solid operation is taken from the previous group.

No. This means that 2.0 does not treat the files saved in 2.1 that have groups inserted in the middle like 2.1 does. This is in fact mentioned in CHANGELOG. The logic of mesh operations is exactly the same, only the group ordering has changed.

Evil-Spirit:
@whitequark,
[...]

PR #992, opened 2021-04-03
Eradicate the IdList bottleneck

### LINKED ISSUE #932: Poor performance when extruding translate groups

### LINKED ISSUE #939: Poor Linking performance

pjanx:
Should fix #939 and fix #932 almost entirely, shifting the bottleneck for large models. Passes tests and memcheck.

Properties of this changeset:
 - memory complexity of IdLists increases by `n * sizeof(int)` bytes, which is about as low as you can go here
 - `IdList::Add` now relies on a single `memmove` of PODs, which is faster than `std::move`-ing singular `elems`
 - `IdList::RemoveTagged` uses a bit of additional temporary memory to keep its complexity

Overall, I'm happy with this now, and I wouldn't be afraid of merging it. Commits should be reasonably atomic.

I think I inadvertently replicated half of [Evil-Spirit's change](https://github.com/Evil-Spirit/solvespace-master/commit/44eb1a4703a1be50d74fdfeacb97f89c06e08ab6). I haven't understood yet what he's doing there, though, even if it looks vaguely similar.


pjanx:
I would appreciate it if someone tried to resolve the Windows build failure, I don't speak MSVC++ error messages.

ruevs:
:-D everyone comes to more or less the same idea :-) I'll push mine tomorrow - just for comparison... But I haven't finished the iterator.

ruevs:
Here is mine https://github.com/ruevs/solvespace/tree/FixIdList_TheNextAttempt working and tests passing.

Very similar to yours with some differences - I used `vector`s as containers and I have a `freelist` to reduce reallocations to some cases.

I do not have the last huge "Remove `NextAfter`" (https://github.com/solvespace/solvespace/pull/992/commits/3b395bb5a7cc09023c9cb0c9d14ea63a5f5e07bf) commit in the branch, but it is critical for some workloads.

Without this commit it is almost as fast (or barely slower) on this https://github.com/solvespace/solvespace/files/5962704/case-vents.slvs.gz from here https://github.com/solvespace/solvespace/issues/932#issue-806016900 compared to current master.

But it is drastically faster on setting the last translate to 20 copies in this https://github.com/solvespace/solvespace/files/5974674/LeliaManya2.zip. It takes about 2s while setting it to 6 copies takes 15s on master (I did not wait for 20 to finish)!

Next I will try yours on Windows with Visual Studio 2019.

pjanx:
There should be a test for possible corner cases in `RemoveTagged`, I made a mistake there that passed but crashed on models.

I see how the freelist simplifies the algorithm in that method (at the cost of some more memory).

Like you, I also thought about putting some assertions in the iterator but... if they weren't there before... maybe it's not important.

I tried to change as little as possible but `NextAfter` just has to go.

Doing benchmarks was on my list, though the improvement was so immense that I just declared victory.

The next thing that irritates me on Solvespace is omnipresent NURBS failures and I can't really do math. How much do you with @phkahler charge? :) Triangle meshes also take a long time to compute.

phkahler:
> The next thing that irritates me on Solvespace is omnipresent NURBS failures and I can't really do math. How much do you with @phkahler charge? :) Triangle meshes also take a long time to compute.

@pjanx  There is a NURBS tag for issues. The failures have been getting better and most are tracked in #738.  It's an area of the code that is a bit outside most peoples experience, but @ruevs and I have been chipping away at it for a while now. A few other people have had a look too which has been very helpful even when we didn't take their changes directly.

Triangle meshes are IMHO not to be used except when NURBS fail. The code there has not had the optimizations or OMP love applied so it could be slower now than doing the same with NURBS. When STL linking lands I suppose we'll hear more reports of mesh bugs.

SolveSpace usage has been increasing as have the number of contributors, so there is hope ;-)

pjanx:
Yeah, I'm having some seemingly randomized mesh issues with my most complex model already (it NURBS-fails very early). Luckily Slicer mostly takes care of it on import, or the issues aren't too severe in 3D print.

phkahler:
@pjanx the evil spirit patches replaces the list of things with a list of pointers to things. That made all the moving faster and probably reduced cache misses while searching.

pjanx:
Replicating IDs in the index like he seems to be doing would eliminate some cache misses during search-by-ID, at the cost of even more memory. I'm not sure how much it is okay to waste on this--it'd be nice to have a memory profile. Binary search isn't particularly cache-friendly in any case, as it jumps around and prefetching fails to work optimally. In that sense, this general approach will always be slower during whole-list iteration. A lot of theorising here, besides the general observation of "large models are fairly snappy now".

ruevs:
> There should be a test for possible corner cases in `RemoveTagged`, I made a mistake there that passed but crashed on models.

The test suite is far, far from complete/exhaustive/ achieving good code coverage.

I broke RemoveTagged as well until the last commit (the 0xDEADBEEF caught it ;-) )
`elemidx.resize(n);  // Clear left over elements at the end.`

To make it work I wrote a quick and dirty "unit test" of sorts for IdList with a simple (int, char) struct. But there is no good place to commit this.

ruevs:
> How much do you with @phkahler charge? :)

For me it's just for fun ;-) Not to mention that I don't really feel qualified when hacking on the NURBS backend, I've forgotten all the linear algebra and calculus so long ago that I constantly have to consult references :-)

But having another mind to bounce ideas off of is extremely helpful. Thank you for that. For example you [starting to work on the IdList](https://github.com/solvespace/solvespace/pull/973) - it has been in the back of my mind since the [Whitequark commit and EvilSpirits patch](https://github.com/solvespace/solvespace/issues/939#issuecomment-810641820) (so at least four years), but you stirred things up, which lead to this pull request and my similar one.

Another one is Eigen - search the issues and pull requests ;-)

I'll make a tracking issue for performance - just as I did for NURBS in https://github.com/solvespace/solvespace/issues/738 .

pjanx:
> This is one reason it would have been extremely hard to merge your last commit here when she was the maintainer. One reason SolveSpace is so stable is that she was (quite rightly) very careful about merging such all encompassing changes that are hard to regression test and also "dirty up" git blame, thus making regressions harder to track down. That said, I totally agree that `NextAfrer` absolutely has to go - it simply makes it impossible to make IdList efficient for all cases.

As far as my experience goes, 2.3 was much less stable, and a performance _minefield_ that made me re-learn to save often. Converting models to master is a pain, but the software is usable. I'm not sure whom to blame for this newfound stability. :)

I don't really like the "git blame" argument, it often seems like a universal excuse to never change anything. Bisection should take care of tracking regressions.

What I've noticed is that the codebase is in the process of losing some sense of coherence, which this PR doesn't particularly help. And cleaning that up is certainly an offensive against "git blame".

> I broke RemoveTagged as well until my last commit (the 0xDEADBEEF caught it ;-) )
> `elemidx.resize(n); // Clear left over elements at the end.`

I just read the source code one more time and found it. I've spent quite a few hours trying to visualise that function and put it to code, mostly because I said I would do it. Sadly I don't have any good methods for making things _simpler_.

ruevs:
> Converting models to master is a pain, but the software is usable. I'm not sure whom to blame for this newfound stability. :)

[Whiteqark](https://github.com/solvespace/solvespace/graphs/contributors) ;-)

> I don't really like the "git blame" argument, it often seems like a universal excuse to never change anything. Bisection should take care of tracking regressions.

In general I agree, but contingent upon available maintainer time vs. desire for stability.

Anyway, I'm pretty sure either this/yours or mine, or a cleaned up amalgamation of the two (they are pretty much the same) will get merged after 3.0. 

( Both @phkahler and I have both been known to merge "stupid" stuff in the past ;-) )

src/dsc.h
```diff
@@ -417,46 +441,38 @@ class IdList {
...
-        auto idx = std::distance(begin(), it);
-        auto i = static_cast<int>(idx);
-        return i;
+        return std::distance(begin(), LowerBound(t));
     }
+
     void ReserveMore(int howMuch) {
         if(n + howMuch > elemsAllocated) {
             elemsAllocated = n + howMuch;
+            int *newIndexes = (int *)::operator new[]((size_t)elemsAllocated*sizeof(int));
```
ruevs, line 466: [Using `vector` for the storage](https://github.com/ruevs/solvespace/blob/6bc1aa91c6a0f1acfc986b34ff54ed36efcdbfc1/src/dsc.h#L404) gets rid of the manual memory management, and hopefully at least in release mode should be as fast.

src/dsc.h
```diff
@@ -467,16 +483,13 @@ class IdList {
         // Look to see if we already have something with the same handle value.
         ssassert(FindByIdNoOops(t->h) == nullptr, "Handle isn't unique");
 
-        // Find out where the added element should be.
+        // Shift indexes to the end of the array, if needed.
         int pos = LowerBoundIndex(*t);
+        memmove(indexes + pos + 1, indexes + pos, sizeof *indexes * (n - pos));
```
ruevs, line 488: I wonder if [insert in a `vector<int>`](https://github.com/ruevs/solvespace/blob/6bc1aa91c6a0f1acfc986b34ff54ed36efcdbfc1/src/dsc.h#L537) is as fast. I'll debug it tomorrow to see what it does.

src/dsc.h
```diff
@@ -486,55 +499,21 @@ class IdList {
...
-    T &Get(size_t i) { return elem[i]; }
-    T const &Get(size_t i) const { return elem[i]; }
+    T &Get(size_t i) { return elem[indexes[i]]; }
+    T const &Get(size_t i) const { return elem[indexes[i]]; }
     T &operator[](size_t i) { return Get(i); }
     T const &operator[](size_t i) const { return Get(i); }
 
-    T *begin() { return IsEmpty() ? nullptr : &elem[0]; }
-    T *end() { return IsEmpty() ? nullptr : &elem[0] + n; }
-    const T *begin() const { return IsEmpty() ? nullptr : &elem[0]; }
```
ruevs, line 534: As above - the const iterators may be useful for compiler optimisations, if we can add `const` in some places where IdLists are used.

src/dsc.h
```diff
@@ -548,23 +527,38 @@ class IdList {
...
-        for(src = 0; src < n; src++) {
-            if(elem[src].tag) {
-                // this item should be deleted
-                elem[src].Clear();
-            } else {
-                if(src != dest) {
-                    elem[dest] = elem[src];
-                }
-                dest++;
+        std::vector<int> elem2indexes(n);
```
ruevs, line 530: I need to trace through this with the debugger to really get it. Tomorrow.

ruevs (commented):
OK so overall we did roughly the same. I'll steal your "remove NextAfter" commit and try mine with it tomorrow.

I think we should try to "converge" on a single solution. If you have time I would love to hear your comments/critique on my version.

src/dsc.h
```diff
@@ -388,12 +388,36 @@ struct CompareId {
...
+        typedef T value_type;
+        typedef int difference_type;
+        typedef T *pointer;
+        typedef T &reference;
+
+        IdList<T, H> *list;
+        int offset;
+
+        Iterator(IdList<T, H> &list, int offset) : list(&list), offset(offset) {}
+        T &operator*() { return list->elem[list->indexes[offset]]; }
```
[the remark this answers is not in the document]
pjanx: That's one thing optimizers should be good at, unless some aliasing rules break it. I'd be somewhat surprised if it'd be measurable for the added complexity. Making `offset` point directly to the `indexes` array is more obvious. And both of these suffer if the array is resized during iteration, which STL just declares as undefined behaviour in such situations.

src/dsc.h
```diff
@@ -388,12 +388,36 @@ struct CompareId {
...
+        T *operator->() { return &list->elem[list->indexes[offset]]; }
+        const T &operator*() const { return list->elem[list->indexes[offset]]; }
```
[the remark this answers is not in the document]
pjanx: MSVC required me to add `const` versions here, for `std::lower_bound`. See the first comment of the PR.

src/dsc.h
```diff
@@ -388,12 +388,36 @@ struct CompareId {
...
+        const T *operator->() const { return &list->elem[list->indexes[offset]]; }
+        Iterator &operator--() { --offset; return *this; }
+        Iterator &operator++() { ++offset; return *this; }
+        Iterator &operator+=(int diff) { offset += diff; return *this; }
+        int operator-(const Iterator &i) const { return offset - i.offset; }
+        bool operator==(const Iterator &i) { return list == i.list && offset == i.offset; }
```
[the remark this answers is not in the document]
pjanx: It should rather be an assertion rather than a direct test like this.

src/dsc.h
```diff
@@ -417,46 +441,38 @@ class IdList {
...
     }
 
-    T * LowerBound(H const& h) {
-        if(IsEmpty()) {
-            return nullptr;
-        }
-        auto it = std::lower_bound(begin(), end(), h, Compare());
-        return it;
+    Iterator LowerBound(H const& h) {
+        return std::lower_bound(begin(), end(), h, Compare());
```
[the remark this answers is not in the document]
pjanx: I wouldn't expect a meaningful difference, it boils down to a few pointer operations. Didn't you have to add a forward declaration and complicate `CompareId` because of that?

ruevs:
> Converting models to master is a pain

By the way, why is this? Models from 2.0-2.3 should "just open™". If you have some that don't (and are not this https://github.com/solvespace/solvespace/issues/120) open an issue.

src/dsc.h
pjanx: That's one of the things that should be unified throughout the codebase, I'm just using what's already there.

I don't even know how the custom allocator library is hooked up.

@Evil-Spirit wanted to use `realloc`, which should be a trick that `std::vector` could be able to do as well, though I'm dubious whether it's useful for geometric steps.

src/dsc.h
pjanx: It should optimize that way.

pjanx:
Mostly getting rid of all those extra degrees of freedom, rotation takes a while to figure out. I don't remember much else.

src/dsc.h
pjanx: I'd be interested in seeing evidence of any difference in this case.

src/dsc.h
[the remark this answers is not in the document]
ruevs: It was an assertion before I removed it ;-)

src/dsc.h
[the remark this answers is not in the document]
ruevs: Yes. I called `CompareId" the predicate.

src/dsc.h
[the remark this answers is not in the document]
rpavlik: I'm inclined to think just implement the code that's needed on the YAGNI principle. Unused methods are untested and possible traps for the future.

src/dsc.h
rpavlik: The important bit is that idlist is reference semantics, not value semantics, so you can't embed a std::vector, but rather a std::shared_ptr<std::vector> - a lesson I learned painfully.

ruevs:
@pjanx 
Tried your version - it seems to work just as fast and as correctly as mine. RemoveTagged seems correct, but it's a bit complicated without a free element list.

I created a pull request with my version with a cleaned up history: https://github.com/solvespace/solvespace/pull/1002
It uses your "remove NextAfter" commit.

Next I will try to see if I can measure speed differences between our implementations but I highly doubt it that there will be any. 
I will also profile to see if there are some more bottlenecks related to IdList.

After that it boils down to style/aesthetics etc.

pjanx:
I think we need an arbiter. It'd be nice to have a memory profile as well, Valgrind can also do that, IIRC.

Evil-Spirit:
Hi there!

Things I remeber:

1.  memove and realloc is superfast
2.  if you are doing containers yourserf it is a bad idea to use another containers like std::vector. you can't control things under the hood of this. If you afraid of manual memory management - just remember, you are doing container for not doing manual memory management around whole program, but inside - you have to do it - instead, you will lost all the benefits of doing this job
3.  when you need map-lookup, you should use binary search. when you need interaing over elements - you may iterate only linear array of data. But NOT ALL THE TIME. Some operations should be performed on sorted arrays, like saving file. You can just break round-trip tests by not iterating in the hv alpahbetic order.

src/dsc.h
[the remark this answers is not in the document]
ruevs: Removed all unused methods from mine.

src/dsc.h
ruevs: @rpavlik Do you mean to say that once a pointer to an element is "taken out' of an IdList it may be kept around and dereferenced after the list has been modified? And thus the pointer may be invalid?

If this is the case can you point to an example?

But this does not make sense - the original implementation of IdList did not guarantee that the elemnts will not move - in fact it moved them around all the time!

In fact my implementation is "better" in this respect since it never moves elements once they are added because the `elemstore` is not rearranged when elements are removed:
https://github.com/ruevs/solvespace/blob/3c6e2be4b920d706d8231030467c5271a6d87d85/src/dsc.h#L570
the removed elements are simply added to the `freelist`.

baryluk:
A bit of my input. With concerns from #932 

I didn't check the memory usage, but did a bit of speed testing. For changing constraints and these PRs. Just dragging unconstrainted extrusion length by mouse, and looking when the UI is fully interactive again. I used gcc 10.2.1 on Linux, amd64. `-O3 -flto -march=native`. With no optimisations, these PRs are actually slower than current master, because of extra small functions and indirections in the iterators. But with optimizations on, they are a tiny bit faster. 

master, -O3 -flto: ~6.6s, 2.78s for Generate::DIRTY
ruevs, -O3 -flto: ~5.4s, 2.31s for Generate::DIRTY
pjanx, -O3 -flto: ~5.5s, 2.27s for Generate::DIRTY

That doesn't sounds like a lot to me, and I don't think it is worth merging any of them it at the current state. Don't get me wrong, these is good code, and a move in a right direction, but it might require more profiling, tuning and base data structure selection.

(No OpenMP used).

ruevs:
> A bit of my input. With concerns from #932
> 
> I didn't check the memory usage, but did a bit of speed testing. For changing constraints and these PRs. Just dragging unconstrainted extrusion length by mouse, and looking when the UI is fully interactive again. I used gcc 10.2.1 on Linux, amd64. `-O3 -flto -march=native`. With no optimisations, these PRs are actually slower than current master, because of extra small functions and indirections in the iterators. But with optimizations on, they are a tiny bit faster.
> 
> master, -O3 -flto: ~6.6s, 2.78s for Generate::DIRTY
> ruevs, -O3 -flto: ~5.4s, 2.31s for Generate::DIRTY
> pjanx, -O3 -flto: ~5.5s, 2.27s for Generate::DIRTY
> 
> That doesn't sounds like a lot to me, and I don't think it is worth merging any of them it at the current state. Don't get me wrong, these is good code, and a move in a right direction, but it might require more profiling, tuning and base data structure selection.
> 
> (No OpenMP used).
> 
> UPDATE:
> 
> LeilaManya2, changing last translate group from 3 to 20 repeats.
> 
> master: ~138s, 126119ms for Generate::DIRTY
> pjanx: ~1.2s, 427ms for Generate::DIRTY
> ruevs: ~1.1s, 403ms for Generate::DIRTY
> 
> Nice! Ok. I might be convinced.

I was just about to mention https://github.com/solvespace/solvespace/files/5974674/LeliaManya2.zip when I noticed you updated your comment ;-) Thanks for running it to 20 repeats on master - I did not have the patience :-)

Would you please do the same timing comparison when modifying some dimension in the first group of this model? https://github.com/solvespace/solvespace/issues/841#issuecomment-742153838

How do you do the measurements on Linux?

phkahler:
> By the way, why is this? Models from 2.0-2.3 should "just open™". If you have some that don't (and are not this #120) open an issue.

Off topic, but I know some Lathe groups can have issues. Lathe used to create 2 copies of the sketch entities like at the start and end of an extrude or revolve, even though both copies were exactly the same. I removed those redundant entities so any sketch that constrained against them could break in 3.0 depending which copy the constraint used.

phkahler:
Going with the @ruevs version unless something bad happens.

Issue #495, opened 2019-10-20
crash when trying to opening a file


bmorel:
### System information

SolveSpace version: 5d78f993cea3749b6644e0a2f9ddf89c9754e377
Operating system: Debian buster

### Expected behavior

File should be open.

### Actual behavior

Crash with that log:

```
File /home/foobar/contrib/solvespace/src/solvespace/src/platform/platform.cpp, line 400, function OpenFile:
Assertion failed: filename.raw.length() == strlen(filename.raw.c_str()).
Message: Unexpected null byte in middle of a path.
zsh: abort      solvespace
```

### Additional information

This crash seems to be semi-random, can't find a way to reproduce it. Seems to happen when using "Recently open" list. I wish I could do more.

whitequark:
You could try running it under gdb and see what `filename.raw` looks like

bmorel:
Sure, but for that I'd need to run it under gdb everytime since I can't find how to reproduce. Or maybe I could patch the assert to dump hexa code of both version of filename... seems less annoying to do that. I'll try to fetch more info.

mePy2:
I have the same problem.

macOS Mojave

Solvespace opens other files correctly. Not this one though.

whitequark:
Does it always reproduce with that file? If so, can you upload it?

mePy2:
Yeah... I mean, I created it and modified it with Solvespace. But now it will crash the app totally!
Since I have Wine, I downloaded the exe file and it does work and can open the file without issue.

Anyway, while typing this, I made a quick test with the Windows exec and I found out the problem was with the `force NURBS surfaces to triangle mesh` option.
I simply opened the file with WINE, deselected the option, saved and it now works with Solvespace for macOS too.

whitequark:
Does it crash with the "Unexpected null byte in middle of a path." assertion on macOS for you?

mePy2:
I got mad. And very unhappy, because I love Solvespace.
Also, while trying a way to open the file, I tried to download the `.dmg` file from here and I found out that it still does not work “out of the box”. (I wanted to see if the issue happened with a fresh Solvespace app package too.)

mePy2:
The app does not quit itself. The rolling circle appears and the app does not respond. Everything freeze.

mePy2:
Anyway now the problem happened again but just by _selecting_ the `force NURBS surfaces to triangle mesh` box.

The only way to stop it is to force close it.

whitequark:
[...]

Issue #995, opened 2021-04-03
Do not add relative paths to "Open Recent" list


Symbian9:
### System information

- **SolveSpace version:** `3.0~` https://github.com/solvespace/solvespace/commit/2df786e7415501c748f34741aa1d5c561e730814
- **Operating system:** Debian buster

### Expected behavior

If some files opened from folder via Terminal as `$ solvespace ./<filename>.slvs` its relative path should NOT be added to "Open Recent" list.

Alternatively add path to such files, BUT just write full path 
instead.

### Actual behavior

If some files opened from folder via Terminal as `$ solvespace ./<filename>.slvs` its relative path added to "Open Recent" list ­— as a result it would be an issue to open such files via GUI.

![pic.1](https://i.imgur.com/eIZUG6y.png)

ruevs:
https://github.com/solvespace/solvespace/commit/3ea8ebfaf57ab061f18f2838f6599d10a339d59b#diff-cc73b65c92fff23152e1a53f8f322c6e3e62f3c4efc6c3cd01ad66a7a93c1ab0

PR #997, opened 2021-04-04
Copy path to avoid it being invalidated by OkayToStartNewFile()

tomsci:
Fixes issue #495

I had a 100% reliable repro locally (as described [here](https://github.com/solvespace/solvespace/issues/495#issuecomment-813023035)) and this change fixes it for me, I am no longer able to reproduce the assertion. There might be a better way to fix it that could avoid similar problems in the future (change how the lambdas capture, rework how PopulateMenuWithPathnames and onTrigger work?) but I'm not familiar enough with the codebase to make that judgement, and this seems to do the trick!

### DIFF
--- a/src/graphicswin.cpp
+++ b/src/graphicswin.cpp
@@ -369,8 +369,11 @@ static void PopulateMenuWithPathnames(Platform::MenuRef menu,
 
 void GraphicsWindow::PopulateRecentFiles() {
     PopulateMenuWithPathnames(openRecentMenu, SS.recentFiles, [](const Platform::Path &path) {
+        // OkayToStartNewFile could mutate recentFiles, which will invalidate path (which is a
+        // refererence into the recentFiles vector), so take a copy of it here.
+        Platform::Path pathCopy(path);
         if(!SS.OkayToStartNewFile()) return;
-        SS.Load(path);
+        SS.Load(pathCopy);
     });
 
     PopulateMenuWithPathnames(linkRecentMenu, SS.recentFiles, [](const Platform::Path &path) {


phkahler:
@tomsci  Thanks for the fix ;-)

phkahler:
@tomsci  is this possibly why the following issue has come up on the forum:
https://solvespace.com/forum.pl?action=viewthread&parent=3601&tt=1617634061

I may have to revert this change.

tomsci:
@phkahler I don't immediately see any way how it could be? The only code I added was to copy a `Platform::Path`, and as far as I can there's just a default constructor on `Path` so there's no way it could be mangling the path, it's just doing a straight string copy. Unless I'm missing something really weird about how `Platform::Path` or `SolveSpaceUI::Load()` works...

Additionally that change is only when opening something from the recents menu, so if I'd messed something up I'd expect it to manifest on open, rather than on save. Sounds more like there's a relative path that's not being made absolute somewhere?

tomsci:
Maybe that forum issue is a side-effect of #995?

phkahler:
@tomsci  But that's an open issue - no change made for it.

The Platform Path stuff I think can behave differently on each OS. I noticed the original issue was on Debian and you reproduced and fixed is on MacOS. The person with the issue is on Linux. It may be they are tripping over an existing bug but they did claim it's new behavior and this was the only recent thing I could think of that might affect path related issues.

Either way it was probably a mistake on my part to merge a platform related fix right before pushing out a new release ;-)

tomsci:
Indeed, I believe the bug causing #995 (whenever that was introduced) is also what's causing the forum poster's problem, and that that issue is unrelated to this PR. I have just reproduced it (with the rc2 OSX binary, ie from before this PR was merged, 36719c62) by passing a relative path to the SolveSpace command line. I'll write something up on #995 if that would help?

tomsci:
I reckon it's probably https://github.com/solvespace/solvespace/commit/3ea8ebfaf57ab061f18f2838f6599d10a339d59b#diff-cc73b65c92fff23152e1a53f8f322c6e3e62f3c4efc6c3cd01ad66a7a93c1ab0 which removed the code which previously would have ensured paths were always absolute and thus introduced the forum issue and #995. I admit I'm still very new to this codebase, but I don't think I'm on the hook for this one :)

ruevs:
@tomsci @phkahler As I mentioned in the commit message 3ea8ebfaf57ab did indeed introduce #995.

However I can not reproduce the [forum problem](https://solvespace.com/forum.pl?action=viewthread&parent=3601&tt=1617634061) on Windows.
- I pass a file on the commend line as a relative path (e.g. ..\..\something.slvs)
- It shows up in the recent file list with the relative path and can be opened in a later session.
- I can `Save` it (CTRL-S) `Save As...`, `Export Triangle Mesh...` just fine.

`Path Path::Expand(bool fromCurrentDirectory)` is called at the proper time before saving and after hitting OK on the save dialog:

```
	solvespace.exe!SolveSpace::Platform::Path::Expand(bool fromCurrentDirectory) Line 268	C++
 	solvespace.exe!SolveSpace::Platform::OpenFile(const SolveSpace::Platform::Path & filename, const char * mode) Line 406	C++
 	solvespace.exe!SolveSpace::SolveSpaceUI::SaveToFile(const SolveSpace::Platform::Path & filename) Line 295	C++
 	solvespace.exe!SolveSpace::SolveSpaceUI::GetFilenameAndSave(bool saveAs) Line 524	C++
 	solvespace.exe!SolveSpace::SolveSpaceUI::MenuFile(SolveSpace::Command id) Line 630	C++
 	[External Code]	
 	[Inline Frame] solvespace.exe!std::_Func_class<void>::operator()() Line 986	C++
 	solvespace.exe!SolveSpace::Platform::WindowImplWin32::WndProc(HWND__ * h, unsigned int msg, unsigned int wParam, long lParam) Line 1071	C++
 	[External Code]	
 	user32.dll![Frames below may be incorrect and/or missing, no symbols loaded for user32.dll]	Unknown
 	solvespace.exe!SolveSpace::Platform::RunGui() Line 1711	C++
 	solvespace.exe!main(int) Line 29	C++
 	solvespace.exe!WinMain(HINSTANCE__ * hInstance, HINSTANCE__ * hPrevInstance, char * lpCmdLine, int nCmdShow) Line 41	C++
 	[External Code]	
```

Can you please describe in more detail how you cause the problem?

tomsci:
Suppose you run `solvespace ./foo.slvs` from a commandline shell in directory bar (the file's full path being `/bar/foo.slvs`) The open recent menu (and `SolveSpaceUI::saveFile`) now contains `./foo.slvs`.

Now let's suppose that at some point after the file is initially loaded, some bit of code (maybe in a library call) changes the process's current working directory to `/tmp`. I haven't gone through the entire codebase to see if/when that can happen, but it explains what JohnF was seeing. Now `saveFile` is no longer valid compared to what the cwd is now set to. And once _that_ is the case, I'd expect all kinds of things to go wrong including failing to save or export.

The other issue (that's much easier to reproduce) is if, having opened `./foo.slvs` once so it's saved in the open recent menu, suppose I now relaunch solvespace from a different directory. Now `./foo.slvs` will no longer resolve against the new current working directory, so attempting to open `./foo.slvs` from the recents menu will give an error.

Storing relative paths is a dangerous thing to do, basically, because cwd can change between the point when you know it was valid (when the path is initially passed in on the command line) and when it gets used, so it's safest to resolve them asap. The thing with Windows and max path limitations is a massive pain I agree, but I think it probably needs solving a different way :)

ruevs:
> Suppose you run `solvespace ./foo.slvs` from a commandline shell in directory bar (the file's full path being `/bar/foo.slvs`) The open recent menu (and `SolveSpaceUI::saveFile`) now contains `./foo.slvs`.
> 
> Now let's suppose that at some point after the file is initially loaded, some bit of code (maybe in a library call) changes the process's current working directory to `/tmp`. I haven't gone through the entire codebase to see if/when that can happen, but it explains what JohnF was seeing. Now `saveFile` is no longer valid compared to what the cwd is now set to. And once _that_ is the case, I'd expect all kinds of things to go wrong including failing to save or export.

This would explain the problem. But (at least on Windows) I do not know how to cause the `pwd` of the running SolveSpace to change. I do not think it is possible.

So how did you reproduce it on OSX?

> The other issue (that's much easier to reproduce) is if, having opened `./foo.slvs` once so it's saved in the open recent menu, suppose I now relaunch solvespace from a different directory. Now `./foo.slvs` will no longer resolve against the new current working directory, so attempting to open `./foo.slvs` from the recents menu will give an error.

This is completely normal for a relative paths. It's what the user asked for. One can argue that it is a feature in exactly the same way. You move a directory tree - with the executable and many slvs files scattered in different sub directories and remembered as relative paths in the recent file list - the files continue to open just fine from the recent files list because they are relative.

> Storing relative paths is a dangerous thing to do, basically, because cwd can change between the point when you know it was valid (when the path is initially passed in on the command line) and when it gets used, so it's safest to resolve them asap. The thing with Windows and max path limitations is a massive pain I agree, but I think it probably needs solving a different way :)

Nothing wrong with storing relative paths. Most development environments store almost everything as relative paths if you think about it ;-) The problems are:

- Users do not understand relative paths.
- The working directory (`pwd`) of the running process changing - we should concentrate on this. How can we cause this in SolveSpace? @phkahler ?

tomsci:
> So how did you reproduce it on OSX?

I couldn't reproduce the OP's problem, that of failure to save. I was able to reproduce putting a relative path into the recents list, relaunching from a different dir, then failing to open. I was speculating that the process cwd was changing for JohnF, `run/user/1000/doc/a0002d9/` looks a lot like a temp path that maybe is getting set by some library code. I don't know for sure that's what's happening. But even if the solvespace code itself is careful not to change cwd, I don't think it's necessarily safe to assume that none of the thirdparty stuff it links in won't... for example there's a `chdir` call in the zlib code although I don't know if that's the culprit or not. And of course there's no guarantee some other project wouldn't make a subsequent change. It's something you have to defend against, really.

phkahler:
> @tomsci @phkahler As I mentioned in the commit message [3ea8ebf](https://github.com/solvespace/solvespace/commit/3ea8ebfaf57ab061f18f2838f6599d10a339d59b) did indeed introduce #995.

@ruevs  That commit is from October. Seems we've had a couple RCs since then with no trouble.

Could the forum issue be due to #983 ?   The problem is happening on Linux since he's using a SNAP package.

ruevs:
> Could the forum issue be due to #983 ? The problem is happening on Linux since he's using a SNAP package.

Definitely not. It changes only guiwin.cpp and that is not even compiled on the other platforms.

PR #998, opened 2021-04-04
Correct libomp.dylib path in sign-macos.sh script

vespakoen:
Looks like I forgot to update this path, and that is causing the CI to fail at the moment.

@phkahler I also looked at the libpng patch, but that already seems to only be applied on macOS, so I think you can go into the submodule and run `git stash` or something to get rid of the changes that were made there

### DIFF
--- a/.github/scripts/sign-macos.sh
+++ b/.github/scripts/sign-macos.sh
@@ -16,7 +16,7 @@ lipo \
 
 cd build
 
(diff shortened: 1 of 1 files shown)


phkahler:
For the PNG thing I had to run git submodule update and include that one in the list (which is not in the instructions?).

PR #1000, opened 2021-04-04
Combine arm64 and x86_64 libomp libraries

vespakoen:
I cannot test the `sign-macos.sh` script locally unfortunately (I am out of code-signing certificates).
I just downloaded the last build and it crashes on my M1, because the libomp library is of the wrong architecture, this should combine the 2 architectures.

### DIFF
--- a/.github/scripts/sign-macos.sh
+++ b/.github/scripts/sign-macos.sh
@@ -2,6 +2,13 @@
 
 lipo \
     -create \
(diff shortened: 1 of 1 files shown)


vespakoen:
Screenshot 2021-04-04 at 21 14 15

Seems to work now, will test this some more tomorrow with SIP re-enabled, and I will also test it on a non-m1 iMac and old macbook air.

So far so good though...

PR #1001, opened 2021-04-05
revert PR 997 as it seem to be causing a regression for some.

phkahler:
There is a report of a new problem on the forum:
https://solvespace.com/forum.pl?action=viewthread&parent=3601&tt=1617634061

I suspect was introduced by that PR, so this reverts it.


tomsci:
PRs and comments are crossing :) As I wrote on https://github.com/solvespace/solvespace/pull/997#issuecomment-813470215 I'd bet the regression was caused by 3ea8ebfaf57ab061f18f2838f6599d10a339d59b (also leading to #995) and not by #997 so I don't think it will help to revert 997.

ruevs (commented):
@phkahler see my comment in https://github.com/solvespace/solvespace/pull/997#issuecomment-813501926

Issue #405, opened 2019-05-15
Regression: Hard to drag constraint labels


rpavlik:
### System information

SolveSpace version:  3.0~201e15e6

Operating system: Debian testing/buster

### Expected behavior
When I hover over a newly-added constraint label (like a distance or an angle) and it turns color, I should be able to click and drag to move. (And could, in 2.3)

### Actual behavior

I typically can't, at least not without zooming in or out. I think this is worse when the constraint happens to be near the origin, but not 100% sure. It's been this way on master for some time - since my earlier PR.

### Additional information
Here's the file I used in the screencast below - add an angle constraint to the tip of the slot sketch group. Sometimes it works, sometimes it acts up.

[sample.zip](https://github.com/solvespace/solvespace/files/3183960/sample.zip)

Screencast showing two attempts: the first one, I was able to zoom a bit and get it to drag eventually. The second clip, I couldn't get it to drag no matter what I did.

[slvs trouble dragging constraints.zip](https://github.com/solvespace/solvespace/files/3183971/slvs.trouble.dragging.constraints.zip)

whitequark:
Is this a duplicate of #161 ?

rpavlik:
Hmm. Possibly - wonder if the fact that I can't move the origin is why clicking and dragging on a near-origin constraint causes failure. I'll pay attention and see if I can figure it out better.

whitequark:
Actually, I think it's a duplicate of #174.

rpavlik:
ah, yeah, that matches better, Sorry for missing the dupe.

Issue #260, opened 2017-05-24
[Feature request] Ability to move entities with keyboard


Helios747:
So a lot of my workflows across my system for various reasons are very keyboard centric. What I would find nice to have is the ability to move selected entities via the keyboard arrows or through some shortcuts. If it's a single object, move that, if it's a group selection, move them all.

When I had the grid enabled, my intuition told me that I could use arrow keys and some sort of key modifier to move what I had selected and it would move by points on the grid.

whitequark:
@jwesthues So this is an accessibility issue. SolveSpace is already much more accessible than direct CADs (or so I'm told), so I'd quite like to make it even better. Any opinion on the modifier? Ctrl+arrows, then jump using the current grid step?

jwesthues:
Nudge with the arrow keys is pretty common, makes sense.

Do you need a modifier? The arrow keys alone don't do anything yet, and that's how Office etc. work.

And maybe nudge by a grid step if the grid is shown, otherwise by a constant distance in screen pixels? That would avoid surprising users who haven't set up a grid (which is probably a lot of them, since the workflow isn't generally too grid-centric).

Evil-Spirit:
@jwesthues 
I think by default arrows must control camera/view. This is needed when you are controlling without mouse (I don't like touch pad so much)

jwesthues:
Are there other applications where arrow keys alone pan/rotate the view, and modifier+arrow moves entities? The reverse would have seemed more intuitive to me, but I can't immediately think of examples of either.

whitequark:
> Are there other applications where arrow keys alone pan/rotate the view, and modifier+arrow moves entities?

Photoshop moves selection with plain arrows, Inkscape moves the canvas. I don't see either choice as more intuitive, personally, and in either case it seems trivial to adjust to it. We already have a highly nonstandard mouse controls for camera and it's not like many people complain.

@Helios747 any opinion here?

Helios747:
I think with no modifier, moving the camera is fine.

Symbian9:
> Are there other applications where arrow keys alone pan/rotate the view, and modifier+arrow moves entities? The reverse would have seemed more intuitive to me, but I can't immediately think of examples of either.

Maybe, better way is add "Keyboard settings" for SolveSpace, where user will map keys as want?

whitequark:
@marthinwurer A very substantial amount of work is required before that's feasible.

marthinwurer:
[...]

Issue #174, opened 2017-01-28
if a constraint is selected, prioritize it for starting a drag


wpwrak:
Lately, I'm having a lot of difficulties dragging distances. Selecting them is fine. I'm not entirely sure if it's because I unlearned some movement that happens to work better, or simply because my designs are getting more crowded.

When trying to drag a distance, I typically aim first for the text. If that doesn't work, I try to get the line(s) to the item being sized.

Perhaps the dragging function could, when deciding what to do with the drag movement, give preference to the item already selected.

whitequark:
@wpwrak do you have any sketch where this is reproducible? Please also post a screenshot with position/zoom settings that reproduce the issue.

wpwrak:
Here is one:
http://downloads.qi-hardware.com/people/werner/slvs/drag-diam-hard.png
The cursor (alas, xwd can't show it) is over the center of the circle-with-slash. Pressing the left button and dragging does nothing.

The .slvs is simple enough:
http://downloads.qi-hardware.com/people/werner/slvs/drag-diam-hard.slvs

whitequark:
confirmed

Evil-Spirit:
https://github.com/Evil-Spirit/solvespace-master/commit/3c8cc2e3cf22aad8546da4682938077df08e7361

baryluk:
Same here. I have a line-distance constrain selected, but I can't drag it, because it is overlapping with line behind it I guess. It happens even if the constrain is a reference value.

I am using git master 16c5fa6889202f2059b52c4a89599910541dd1a0

baryluk:
@Evil-Spirit I can confirm that the patch from https://github.com/Evil-Spirit/solvespace-master/commit/3c8cc2e3cf22aad8546da4682938077df08e7361 fixes the issue.

I do not know if it is the correct solution, but it does fixes this particular problem for me.

The patch is not really in mergable state tho. It should just remove the commented code, not comment it out. :D Or fix the logic there, otherwise it looks a bit more like a hack.

phkahler:
IMHO that patch is no good. It eliminates the ability to drag entities from prior groups. Normally if you extrude a rectangle you can push or pull the "new" end to change the length, while dragging points or lines of the original rectangle can change the width or height. The patch eliminates the option for the later - resizing the underlying sketch from the extrusion group.

phkahler:
[...]

PR #1003, opened 2021-04-05
Add an exploded view to sketches

tomsci:
This PR adds an option to workplane sketches to draw all the entities in the sketch "exploded" normal to the workplane. The reason for doing this is that it makes it much easier to select lines and points which entirely overlap with other lines/points. I've found that when drawing a sketch from scratch in SolveSpace, I can usually use the "requests in group" list to find what I just added, but when importing complex DXFs created by other tools, that have all kinds of weird unexpected lines and constraints, it can be very difficult to figure out what constraints violations the import has created.

For example, here's a really basic sketch with a construction line hidden under the top line of the rect:
Screenshot 2021-04-05 at 5 02 46 pm
Here's the same sketch rotated (obviously, this doesn't help identify the construction line):
Screenshot 2021-04-05 at 5 02 41 pm
But now if we enable the new "explode" option:
Screenshot 2021-04-05 at 5 03 44 pm

It's now really obvious where that line is:
Screenshot 2021-04-05 at 5 02 35 pm

Enabling the "explode" option does not affect the underlying data, the plane adjustment is only made when rendering, and only when the sketch is active, so it wouldn't show up in any exported view. ~~Hmm now I try it thought, I may need to tweak that because constraints _do_ show up in their exploded positions...~~ Edit: now fixed, exploded coords are now never exported.

I've tried to update all the constraints drawing code to put the constraints in the same plane as the entity they're constraining, although I gave up on some of the more complex constraints that aren't solely in the workplane to start with. I figured it might be more misleading to mess with projected points and lines and such, and also my geometry skills were starting to struggle to figure out how to actually code it. If anyone else wanted to fix that, be my guest.

The "explode sketch when active" checkbox and the `explodeDistance` is per-Group, but I haven't updated the file format to persist the choice. I wasn't sure whether to, or not. It'd be convenient to remember it, but updating the file format looks like it introduces a warning dialog if anyone opens a newer file with an older version of SolveSpace, so I wasn't sure if it'd be a good idea to introduce that for such a minor feature. Feedback welcome on how best to do that! Or whether to just not bother, of course.

### DIFF
--- a/src/confscreen.cpp
+++ b/src/confscreen.cpp
@@ -68,6 +68,11 @@ void TextWindow::ScreenChangeGridSpacing(int link, uint32_t v) {
     SS.TW.edit.meaning = Edit::GRID_SPACING;
 }
 
+void TextWindow::ScreenChangeExplodeDistance(int link, uint32_t v) {
+    SS.TW.ShowEditControl(3, SS.MmToString(SS.explodeDistance, true));
+    SS.TW.edit.meaning = Edit::EXPLODE_DISTANCE;
+}
+
 void TextWindow::ScreenChangeDigitsAfterDecimal(int link, uint32_t v) {
     SS.TW.ShowEditControl(14, ssprintf("%d", SS.UnitDigitsAfterDecimal()));
     SS.TW.edit.meaning = Edit::DIGITS_AFTER_DECIMAL;
@@ -269,6 +274,10 @@ void TextWindow::ShowConfiguration() {
     Printf(false, "%Ba   %s %Fl%Ll%f%D[change]%E",
         SS.MmToString(SS.gridSpacing).c_str(),
         &ScreenChangeGridSpacing, 0);
+    Printf(false, "%Ft explode distance%E");
+    Printf(false, "%Ba   %s %Fl%Ll%f%D[change]%E",
+        SS.MmToString(SS.explodeDistance).c_str(),
+        &ScreenChangeExplodeDistance, 0);
 
     Printf(false, "");
     Printf(false, "%Ft digits after decimal point to show%E");
@@ -457,6 +466,11 @@ bool TextWindow::EditControlDoneForConfiguration(const std::string &s) {
         case Edit::GRID_SPACING: {
             SS.gridSpacing = (float)min(1e4, max(1e-3, SS.StringToMm(s)));
             SS.GW.Invalidate();
+            break;
+        }
+        case Edit::EXPLODE_DISTANCE: {
+            SS.explodeDistance = min(1e4, max(-1e4, SS.StringToMm(s)));
+            SS.MarkGroupDirty(SS.GW.activeGroup, true);
             break;
         }
         case Edit::DIGITS_AFTER_DECIMAL: {
--- a/src/drawconstraint.cpp
+++ b/src/drawconstraint.cpp
@@ -267,7 +267,7 @@ void Constraint::DoEqualRadiusTicks(Canvas *canvas, Canvas::hStroke hcs,
     const Camera &camera = canvas->GetCamera();
     Entity *circ = SK.GetEntity(he);
 
-    Vector center = SK.GetEntity(circ->point[0])->PointGetNum();
+    Vector center = SK.GetEntity(circ->point[0])->PointGetDrawNum();
     double r = circ->CircleGetRadiusNum();
     Quaternion q = circ->Normal()->NormalGetNum();
     Vector u = q.RotationU(), v = q.RotationV();
@@ -291,7 +291,8 @@ void Constraint::DoEqualRadiusTicks(Canvas *canvas, Canvas::hStroke hcs,
 
 void Constraint::DoArcForAngle(Canvas *canvas, Canvas::hStroke hcs,
                                Vector a0, Vector da, Vector b0, Vector db,
-                               Vector offset, Vector *ref, bool trim)
+                               Vector offset, Vector *ref, bool trim,
+                               Vector explodeOffset)
 {
     const Camera &camera = canvas->GetCamera();
     double pixels = 1.0 / camera.scale;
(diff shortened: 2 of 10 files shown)


ruevs:
Interesting. Please be aware that you are attacking a problem (the selection of overlapping entities) that has been discussed many times in the past. Some I found easily:
https://github.com/solvespace/solvespace/issues/173 https://github.com/solvespace/solvespace/issues/161 https://github.com/solvespace/solvespace/issues/162 https://github.com/solvespace/solvespace/issues/174 https://github.com/solvespace/solvespace/issues/405 https://github.com/solvespace/solvespace/issues/260#issuecomment-575130517 https://github.com/solvespace/solvespace/issues/260#issuecomment-578089457
etc.

tomsci:
Thanks, I hadn't really dug through the issues and have been using this to familiarise myself with the codebase as much as anything. I like the sound of select-by-order as an option too, providing you do already know what you're trying to select, that'd probably be a quicker option. I haven't really decided myself if the exploded view is that useful so I thought I'd code it up and try it out for a bit. I think so far the most useful functionality it adds is the ability to better visualise a complex overlapping imported sketch. Selecting the right thing is probably a secondary issue to working out what the heck's going on in the first place, especially when there are conflicting constraints and/or open contours (which aren't immediately obvious) in the imported data.

Also, I didn't realise it'd kick off a test run by opening the PR as a draft, apologies for burning through CPU cycles on what won't be the final code. Should I close this and continue developing on a branch until it's ready for proper review?

ruevs:
> 
> 
> Also, I didn't realise it'd kick off a test run by opening the PR as a draft, apologies for burning through CPU cycles on what won't be the final code. Should I close this and continue developing on a branch until it's ready for proper review?

In my opinion it's fine - don't worry about it. At this point GitHub does not seem to complain about us using too much CPU for CI. If you want to "save cycles" you can of course push to your branch more rarely or even close the pull request. You can reopen it later. As you wish.

phkahler:
@tomsci  not sure if you're still interested in this, or if I like it ;-) It just occurred to me that you might want to keep non-overlapping entities that are the same style on single layers in the exploded view. That's even more code to sort them for a feature that seems to be uncertain anyway, but I wanted to share the thought.

tomsci:
I've been using it regularly for the last couple of weeks and have found it to genuinely pretty useful. There isn't any other good way that I've found to disambiguate the selection of, for example, one specific point belonging to line X when there are 4 or 5 other entities which also have a point at that location. I did read through all the issues referenced and none of them seemed to offer a visual way of doing that - having to either scroll through the text window, or repeatedly step through clicking keys/buttons whilst watching the text window would both quickly become unwieldy or impractical on large complex models, which is of course when you might need the feature the most!

My approach tends to be to start in the orientation of the workplane (ie by pressing 'W'), from that orientation the exploded view is of course indistinguishable from the normal view, and then rotate my view slightly to "peek" under the point/line I'm interested in to find the actual element I want to select. And then reorient myself - doing normal work in the exploded view at any other angle is pretty much impossible as you'd expect. But by repeatedly going back to the workplane aligned view I don't have to keep toggling exploded view on and off, so it works pretty well for me.

I decided against attempting any sort of overlap detection or grouping when calculating the z-offset, I wanted to keep the code as simple as possible, and once I added the `groupRequestIndex` var and the `explodeDistance` parameter, I haven't felt any need to further limit the amount of exploding. YMMV on that one of course but I've been using some models and sketches complex enough to slow solvespace to a crawl on my machine so I think I've been putting the feature through its paces to a reasonable degree :) (I'd love to see some of those solver performance improvements like #842 land on master by the way!)

The code changes are still more invasive than I'd like - suggestions welcome on any ways to improve on that! - but I've quickly come to rely on this feature myself and am keen to progress this PR if at all possible.

tomsci:
Updated branch to fix minor conflicts with latest master.

@ruevs Any more thoughts on this PR? I've been using this feature semi-regularly for the last few months on some models complex enough SolveSpace can't even calculate a valid mesh for (but that's a separate issue...) and haven't found any problems. At this point I'm relying on having this feature so I'm maintaining it on a branch for myself anyway (just the other day I had to build win32 binaries with it in so I could work on a beefier PC running windows), but it'd be nice to have it in master instead :)

tomsci:
> weather it should be triggered by a keypress rather than in the text window. I suppose the problem with that is setting the spacing between the exploded elements

It certainly could be a tickable menu item in the View menu which applied to whatever sketch was active, like "Use perspective projection". And I must admit I like having keyboard shortcuts for everything. In that case the explodeDistance could move to the configuration view. It would probably also simplify the code a little since each groups wouldn't have to track anything additional. I think I'm talking myself into some extra work, dammit :-)

tomsci:
I've made the discussed change to move the toggle option out of the per-group settings and in to the View menu ("Show Exploded View") and the explodeDistance to be global, in the configuration page. I picked the shortcut key `\` since there's almost nothing in the View menu without a shortcut. I think on balance I like it better like this, having a shortcut is pretty handy for jumping in and out of exploded view.

phkahler:
@tomsci  Can you squash some of these together and push it?

BTW nice choice of "\" for the key I think.

tomsci:
@phkahler have rebased and squashed.

ruevs:
@phkahler I've _not_ checked out, compiled and  teried this. I only looked at the code on GitHub a few months back. Give me a few days if you want an informed opinion.

ruevs:
OK I reviewed it and played with it.

It is nice. It can be quite helpful when many entities are on top of each other. It is especially convenient combined with `View - > Use Perspective Projection` (`) to avoid rotating away from the workplane view.

II like that the explode distance can be negative and it explodes "back" from the work plane.
And +-1000 may not be too much especially when used in combination with `View - > Use Perspective Projection`.

The code changes in src/drawconstraint.cpp and src/drawentity.cpp are "all over the place", but the way the drawing of constraints and entities is structured I don't see how it can be better with the current code. I have not tested them all.

ruevs:
@phkahler @tomsci  I merged it.

phkahler:
> The code changes in src/drawconstraint.cpp and src/drawentity.cpp are "all over the place", but the way the drawing of constraints and entities is structured I don't see how it can be better with the current code.

@ruevs  That was my thought as well. Most of the "all over the place" is just replacing one function call with another that returns position with the exploded modification, so it's actually kind of clean. Just wanted to see if you'd have an alternative idea on how to do that.  ;-)

PR #1010, opened 2021-04-11
Update property browser display live when dragging

robnee:
I believe there is a glitch in the way the Property Browser display is updated when dragging.  Currently, when one starts tp drag an entity such as a point it's details will briefly appear in the Property Browser.  This is due to the fact that initially the mouse-down event selects the entity being hovered.  Once the mouse begins to move the state transitions from selection to dragging.  This causes the Property Browser to be cleared (reverts to group display) during the duration of the drag which does not show anything useful.  Alternatively, when dragging an entity that was _already_ selected, the entity remains in the selected state and the Property Browser continues to show the dragged entity's details.  I feel the latter behavior is more informative.

This PR tweaks the selection to dragging state transition for previously-unselected entities so that the dragged entity remains selected until the mouse button is released and hence the property browser shows the details of the dragged entity while being dragged. 

Also, the "view" page of the property browser happens to update "live" already while the view is being manipulated.  I stole this idea and applied it while dragging entities as well.  The result is pretty cool.

https://user-images.githubusercontent.com/3731923/114316068-e5d85580-9acf-11eb-82fc-b546d0db9404.mp4

### DIFF
--- a/src/mouse.cpp
+++ b/src/mouse.cpp
@@ -10,6 +10,8 @@ void GraphicsWindow::UpdateDraggedPoint(hEntity hp, double mx, double my) {
     Vector pos = p->PointGetNum();
     UpdateDraggedNum(&pos, mx, my);
     p->PointForceTo(pos);
+
+    SS.ScheduleShowTW();
 }
 
 void GraphicsWindow::UpdateDraggedNum(Vector *pos, double mx, double my) {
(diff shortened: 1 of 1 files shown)


phkahler:
Haven't had a chance to try this yet.  I'm a little concerned about what happens when the dragged entity is not in the current group. Draw a simple sketch and extrude it. Then drag one of the points at the base (original end) of the extrusion and drag it. When clicking (selecting) one of those points we grab the one in the current group. When dragging we skip the one in the current group and let it drag the unconstrained one from the 2D sketch group. That behavior might be a little weird with this PR, but it might also be better than what it does now :-)

robnee:
I'd like to experiment with this and see if there is a regression but I'm not 100% sure I understand this test case.  Using the release 3.0 version I draw a rectangle with just the auto H and V constraints and extrude it.  I leave it in the ortho view and select one of the points but it always selects the one from the current group regardless of select or drag.  I'm aware of that there is code to make the choice of point to drag but I'm not sure how to set up this test case you describe.  Sorry, could you please elaborate a little?  Thanks.

phkahler:
So you draw a rectangle in group g002 and extrude it to form group g003. Your sketch is extruded toward you and you can drag the "new" end or "top" end to change the length of the extrusion. You see a bunch of white line entities and some green points. All these entities are in g003. The points and lines at the "original" or "bottom" of the extrusion are copies of the entities in the original 2d sketch in g002.  If you check the box "two-sided" in the text window of g003 you'll get an extrusion pulled symmetrically above and below the original sketch plane and you'll be able to see the entities from g002 - the lines are a dark color (brown?) and the points are still green. With the 2-sided extrusion it's easy to see the pointe in g002 and you can change the length by dragging points at either end of g003, but you can also drag the points in g002 to change the original sketch.

When you go back to a one-sided extrusion, some of the g003 points and lines are coincident with those from g002. If you select those points it will be the point in g003. However if you drag it, you'll be dragging the point in g002 and the one in g003 will just follow along since it's a direct copy of the one from g002. It feels like you're dragging the point in g003, but internally you're not and that will probably show in the text window.

I think this is just a more wordy version of my previous comment, so maybe it doesn't help :-)

The code that differs in Selection vs Dragging is in draw.cpp.
GraphicsWindow::Selection GraphicsWindow::ChooseFromHoverToDrag()
is a variation on the function above it which is
GraphicsWindow::Selection GraphicsWindow::ChooseFromHoverToSelect()

The primary difference is that dragging checks (!e->CanBeDragged())

This PR might actually help to illustrate the difference if it shows that a dragged point is in a prior group while it's happening.

robnee:
@phkahler Thank you!  Your explanation is excellent.  Now I understand that it's the coincident points from the two groups that you are talking about.  I tried this test case.  When you first select one of these coincident points it first selects the point from g003.  When you transition from select to drag, however, it switches to the point from g002.  

With this PR the property browser updates, live, to show that you are now dragging g002.  If the point from g003 is _pre-selected_ before you grab it to drag is will still show you are dragging your selected point from g003 but because this point is tracking the point from g002 it also updates live.

I feel that this is desirable behavior and helps make things clearer.  I'd love it if you could give it a try and let me know what you think.  I'm happy to continue tweaking it if necessary.  This is a piece of some other changes I am working on.  One is an improvement to the property browser display when multiple entities are selected to replace the fallback "SELECTED: n items" which isn't very helpful.  I've also built out some missing links in the browser in some places so you can fully "surf" the geometry.  I thought it would be easier to present them as separate PRs though.

PR #1015, opened 2021-04-18
CI: Never skip cancel_previous_runs

ppd:
The if condition was nonsensical and did not serve any
practical purpose; removing it allows the succeeding jobs
to run in any case without additional code.

Test for release event: https://github.com/ppd/solvespace/actions/runs/760219571

@phkahler We could re-create the release on this commit, or I could upload the assets manually and build the snaps locally. What do you prefer?

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -12,7 +12,6 @@ jobs:
   cancel_previous_runs:
     runs-on: ubuntu-latest
     name: Cancel Previous Runs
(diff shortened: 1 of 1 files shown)


phkahler:
@ppd either way is ok. Quicker is better, but not at the expense of correctness.

phkahler:
Actually, I wanted to make a branch after, just in case. So recreating it may be better.

Issue #344, opened 2018-07-28
Compile on Ubuntu 14.04


probonopd:
### System information

master

Operating system: Ubuntu 14.04

### Expected behavior

Can compile the GUI

### Actual behavior

https://github.com/solvespace/solvespace/blob/2b4c484dcb1031f9f196805def71887b92ab097f/.travis/build-debian.sh#L7

### Additional information

We need to find a ppa or other software source that provides a suitable version of Gtk+. Alternatively, we need to lower the requirements for Solvespace so that it can be compiled using the Gtk+ that comes with Ubuntu 14.04, the currently oldest still-supported LTS version.
I consider it good practice to target the oldest still-supported LTS versions of distributions when developing applications.

whitequark:
> Alternatively, we need to lower the requirements for Solvespace so that it can be compiled using the Gtk+ that comes with Ubuntu 14.04, the currently oldest still-supported LTS version.

This will not happen. Any GTKmm version under 3.16 (including GTK+ 2; there is a reason I don't support it anymore) has fundamentally unusable OpenGL support, and may not be used with SolveSpace. Actually, 3.16 is already supported using a number of workarounds; the original requirement was 3.22.

The GTK 3 release with stable ABI is 3.22, so I cannot and will not guarantee that SolveSpace works with any GTK 3 release that isn't 3.22 or later. If you think this is unacceptable, well, you can complain to GTK developers about their terrible release schedule.

> We need to find a ppa or other software source that provides a suitable version of Gtk+.

As discussed earlier, I have decided that packaging for extremely old Linux distro versions is out of scope of the SolveSpace project. Moreover, considering earlier discussions on shared libraries, theming, and user experience, and the fact that theming became stable only in GTK+ 3.22, I cannot endorse any AppImage except one that:
  * uses GTK+ 3.22 or later, and
  * uses theming plugins from the target system.

Such an AppImage could, in principle, carry its own copy of GTK+, but considering that using theming plugins from the target system requires that it would have a copy of GTK+ 3.22 *anyway*, it probably should just dynamically load GTK+ from the target system.

In other words, I don't think AppImages built against GTK+ earlier than 3.22 should be made. Of course, I cannot prevent you from doing so, but I do not think it is a good idea and I do not support this use case.

In terms of Ubuntu releases, this excludes both Trusty (which shipped with 3.10) and Xenial (which shipped with 3.18), leaving only Bionic (which finally shipped with 3.22.2).

[...]

Issue #336, opened 2018-07-13
Adding SolveSpace to package managers


Foadsf:
Not an issue, more of a feature request.

Most of the professional users, use package managers to keep their software updated. It would be great if SolveSpace team could make sure that the latest stable and nightly builds of the software is available in most of them. My suggestions:

1. I have listed available package managers for Windows [here](https://www.reddit.com/r/windows/comments/8nq62b/freeopensource_package_managers_for_windows/). The most famous ones are Chocolatey and Scoop. It would be great to make sure the latest SolveSpace is available there. 
2. [repology.org](https://repology.org/) is a great website which keeps track of different package managers. maybe you could add [their badge](https://repology.org/static/demobadge.v1.svg) to the github repo. You could go to [this page](https://repology.org/metapackage/solvespace/versions) and make sure all the package managers which have the SolveSpace are listed there, otherwise report it.

whitequark:
SolveSpace is distributed as a single executable file for Windows (and another one for the few people who want the CLI), and of course for macOS as well. Is there really any point in going for e.g. Chocolatey? As far as I'm aware package managers like Chocolatey and NuGet are typically used by developers; SolveSpace doesn't need and won't _install_ anything per se.

As for Linux, well, I tried building `.deb` packages, but this resulted in a generally awful experience for users (`.deb` packages aren't really portable even between Debian and Ubuntu), and the Debian upstream eventually specifically asked me to remove the Debian package from the repository. I have decided that until a technology like AppImage, but one that allows to link to libc and GUI toolkit dynamically without redistributing them, is available, the SolveSpace project will not provide prebuilt Linux binaries.

Foadsf:
It is great to have single file binaries like .exe, .dmg, .deb , snap and flatpack but there are a couple of issues. first of all if there is something wrong with the dependencies they can't handle it well. secondly keeping the software updated manually is a hassle. I personally use different operating systems per day and use different package managers to keep my software updated across these machines. I think it would be a great benefit for the SolveSpace community if they could have the software through package managers.

whitequark:
It's been a few years since the last release and I don't think SolveSpace will ever release more often than every few months, so it seems to me that it isn't a lot of hassle. Anyway, let's keep this open, if there is substantial interest I'll revisit this topic.

Foadsf:
[...]

PR #1022, opened 2021-04-21
Flatpak

rpavlik:
I was wondering why we didn't have a flatpak, just a snap, then I noticed we actually did have some flatpak-related stuff. It just needed some updating.

### DIFF
--- a/pkg/flatpak/com.solvespace.SolveSpace.json
+++ b/pkg/flatpak/com.solvespace.SolveSpace.json
@@ -1,19 +1,29 @@
 {
     "app-id": "com.solvespace.SolveSpace",
-    "runtime": "org.gnome.Platform",
-    "runtime-version": "3.30",
-    "sdk": "org.gnome.Sdk",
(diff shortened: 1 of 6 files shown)


ppd (commented):
Everything seems to be working; very nice. :+1: 

Just one thing that sticks out: 
```
GNOME 3.30 runtime is no longer supported as of 27th January 2020. Please ask your application developer to migrate to a supported platform
```
So we should also use a more recent runtime & sdk for the flatpak while we're (you) at it.

ruevs:
Just cross-linking things ;-)
https://github.com/solvespace/solvespace/issues/336
https://github.com/solvespace/solvespace/issues/347
https://github.com/solvespace/solvespace/issues/344#issuecomment-408624727

rpavlik:
Oh interesting I didn't see that eol notice. Yeah I'll see if I can update stuff a bit.

rpavlik:
There, updated as much as was reasonable and relatively easy.

Issue #880, opened 2021-01-09
Force distance constraints to plane while dragging


phkahler:
## Suggestion

When dragging a distance constraint, use SHIFT to force the constraint to the **nearest** workplane or face that contains the 2 points (or a point and line).  Another option might be to allow the user to select a face or workplane prior to dragging a constraint and have it stay within that plane.

### Use case
For technical drawing it looks good to have these items in-plane rather than at arbitrary angles to the sketch.

phkahler:
@Symbian9   I thought of this while looking at your usecase for "Dim Solid for Sketch Groups", so I'd like your thoughts in particular. I tried dimensioning a 3D sketch all in one group, and I kept needing to switch views to get the constraints to lie in the correct plane. Locking to the nearest axis-aligned plane might be enough? It is also interesting that radius/diameter constraints are already locked to the plane of an arc or circle.

Symbian9:
>  Another option might be to allow the user to
>
> I thought of this while looking at your usecase for "Dim Solid for Sketch Groups", so I'd like your thoughts in particular.

@phkahler, Yeah! I has planned to create such feature request few days ago.

JFTR, Actually in 3D mode sketching diameter constraint dimension is locked in circle/arc plane.

As for me, there should be few options:

1. Snap dimension constraint to plane in 3D (align distance constraint datum point to selected plane); 
2. Select a face or workplane prior to dragging a constraint and have it stay snapped within that plane;
3. `TBA` *(more suggestion soon!)*

### TBA

***NOTE:** I would add more details, as this feature is really important, but just in few days!* ;-)

PR #1034, opened 2021-05-09
Allow comment constraint associated with point entity (WIP)

### LINKED ISSUE #1032: The comments can't move with the shapes
I have made a 2D geometry sketch, and used the comments to mark the point name. When I moved the shapes by selecting and dragging them, the comments around the points didn't move at the same time. When I use the mouse to drag a box to select the whole shape, only the shape itself is selected, and the comment is not selected. And even I click the comments one by one and then selecting the whole shape, the shape and the comments still moved seperately.

I think if I could contrain the shape and the comments as one "group", or constrain the comments lock on the related points, this problem maybe be solved. I hope you can provide this feature.

Attached is the 2D geometric representation that I have drawn. I am trying to provide some basic parameter computing power in SolveSpace by means of plane geometry.
image

robnee:
Allows comment constraints to be associated with a reference point (in addition to the standard "free" mode) so that their position offset is relative to that point.  Moving or removing the reference point affects associated comments.

Things tested:
- does not impact "free" (created with no point selected) constraint behavior in 2D or 3D.
- if created with a selection checks if selection is a single point.
- saves and restores reference point `ptA` in .slvs file
- behavior degrades gracefully when .slvs files with associated comments are opened in version 3.0 and earlier, comment is just rendered offset relative to the workplane origin rather than the reference point.
- deleting the reference point deletes associated comment constraints as it would with any other constraint
- supports multiple comment constraints for a single point.
- rendering in 3D appears to be correct.
- property browser appears to work correctly

Note: this will require 1 new string to be localized.  The wording of the new string could probably be improved.  Perhaps the default comment when there is a reference point could be improved.

This closes #1032

### DIFF
--- a/res/locales/de_DE.po
+++ b/res/locales/de_DE.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: SolveSpace 3.0\n"
 "Report-Msgid-Bugs-To: whitequark@whitequark.org\n"
-"POT-Creation-Date: 2021-02-01 15:45+0200\n"
+"POT-Creation-Date: 2021-05-10 16:29+0300\n"
 "PO-Revision-Date: 2018-07-19 06:55+0000\n"
 "Last-Translator: Reini Urban <rurban@cpan.org>\n"
 "Language-Team: none\n"
@@ -66,15 +66,15 @@ msgstr ""
 msgid "No workplane active."
 msgstr "Es ist keine Arbeitsebene aktiv."
 
-#: confscreen.cpp:418
+#: confscreen.cpp:416
 msgid "Bad format: specify coordinates as x, y, z"
 msgstr "Ungültiges Format: geben Sie Koordinaten als x, y, z an"
 
-#: confscreen.cpp:428 style.cpp:659 textscreens.cpp:805
+#: confscreen.cpp:426 style.cpp:730 textscreens.cpp:805
 msgid "Bad format: specify color as r, g, b"
 msgstr "Ungültiges Format: geben Sie Farben als r, g, b an"
 
-#: confscreen.cpp:454
+#: confscreen.cpp:452
 msgid ""
 "The perspective factor will have no effect until you enable View -> Use "
 "Perspective Projection."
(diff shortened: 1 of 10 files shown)


ruevs (approved):
This looks good and works as expected. @phkahler what do you think?

I removed the extra empty lie at the end of drawconstraint.cpp and force pushed.

Also added a commit updating all translations.

ruevs:
@robnee @phkahler Thinking some more about this I came up with a slightly different behaviour:

- If a single point is selected associate a newly created comment with it.
- If nothing or more than a point is selected - create a floating comment as before.

This avoids the new error message and keeps the behaviour as it was in all cases but the "one point selected".

What do you think?

I created a pull request https://github.com/solvespace/solvespace/pull/1035.

robnee:
@ruevs I could go with either one.  I like that #1035 is simpler and I find the error dialogs annoying at times :)  I originally thought that new users would find the message helpful in learning about the feature.  I like the term "floating" comment too.

Separately, I'm working on an enhancement to make the property browser displays more informative in places where they are a little thin, like when 3 or more entities are selected etc.  That's almost ready and I was thinking to include an improvement to the comment display with that to mention the reference point and the offset.  Given that, I think I like your approach better.

phkahler:
@robnee  The only thing I don't like in your description is the new text string. Is there a new dialog box? If so, why?  If it's more text in an existing dialog then I'm fine with that kind of update. I also like @ruevs comment above and reference to #1035 which I also need to try.

BTW since you're both in this area of the code, any chance to address #880  IMHO that's a good one for placing dimensions in 3D.

robnee:
> > Separately, I'm working on an enhancement to make the property browser displays more informative in places where they are a little thin, like when 3 or more entities are selected etc.
> 
> While you're in there, see #165 :-) and probably some related issues.

Yes, I'm hoping to address this.

phkahler:
@robnee  I'm going to go with @ruevs version of this. I'm not a fan of adding dialogs in there. Keep the good ideas and commits coming though!

ruevs:
@robnee gets all the credit for this :-) I only got rid of the message :-)

PR #1035, opened 2021-05-11
Allow comments to be associated with point entities

### LINKED ISSUE #1032: The comments can't move with the shapes

ruevs:
If a single point is selected when a "Constrain | Comment" (`;`) is added
then the comment is associated with the point and its position becomes
relative to the point. In this way the comment will move with the point.

If nothing is selected or more than a single point is selected then
the behaviour is as before and the comment is "floating".

Closes #1032

Alternative to https://github.com/solvespace/solvespace/pull/1034

### DIFF
--- a/src/constraint.cpp
+++ b/src/constraint.cpp
@@ -767,10 +767,19 @@ void Constraint::MenuConstrain(Command id) {
             break;
 
         case Command::COMMENT:
-            SS.GW.pending.operation = GraphicsWindow::Pending::COMMAND;
-            SS.GW.pending.command = Command::COMMENT;
-            SS.GW.pending.description = _("click center of comment text");
-            SS.ScheduleShowTW();
+            if(gs.points == 1 && gs.n == 1) {
+                c.type = Type::COMMENT;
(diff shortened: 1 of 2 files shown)


ruevs:
Hmmm. Yes it does... Until you modify the comment from the default. After that it would add a new one. I have to search where is the logic that does this, but I kind of like it.

phkahler:
Oh that's actually really nice!

PR #1044, opened 2021-05-21
Misc icon fixes

pinotree:
- drop old XPM application icons, as needed for the old Debian menu
- fix naming of XDG mimetype icons for non-flatpak and non-snap installations


phkahler:
@pinotree  Does this affect packagers at all, or just remove unused files?

pinotree:
> @pinotree Does this affect packagers at all, or just remove unused files?

Yes, this removes & renames files currently installed. Checking the various packaging done in distros, it seems only the FreeBSD and openSUSE ones will require changes; however they are very small, and only as one-off.

Also, I'm puzzled by the CLA assistant bot, as I saw that cac3aae0c7d43413897490859b7d2518efd64c31 drops mentions of this requirement.

rpavlik:
yeah this is fine by me. I'm not sure what's up with the CLA bot either, I was pretty sure that was a thing of the past too.

res/CMakeLists.txt
```diff
@@ -199,20 +199,15 @@ else()
             RENAME      solvespace.svg)
         install(FILES freedesktop/solvespace-scalable.svg
             DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/scalable/mimetypes
-            RENAME      application.x-solvespace.svg)
+            RENAME      application-x-solvespace.svg)
```
rpavlik, line 202: Trying to figure out if this affects flatpak - have you tried building a flatpak with this patch in place? I can't find a mention of mime in the flatpak docs, but I was pretty sure that's why it was named that.

rpavlik (approved):
Just had one question about flatpak, but I think it's ok.

res/CMakeLists.txt
pinotree: Note this branch applies only to non-flatpak and non-snap builds. The flatpak branch has:
```cmake
        install(FILES freedesktop/solvespace-scalable.svg
            DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/scalable/apps
            RENAME      com.solvespace.SolveSpace.svg)
        install(FILES freedesktop/solvespace-scalable.svg
            DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/scalable/mimetypes
            RENAME      com.solvespace.SolveSpace.svg)
```
also with a custom XDG mime type definition that defines a different `<generic-icon>`. Not sure why that it is needed, instead of relying on the default name of XDG mimetypes (e.g. `application/x-foo` mime -> `application-x-foo` icon).

Issue #1048, opened 2021-05-31
Feature request: workplace creation and keyboard keys to rotate


simrim1:
This piece of software is unbelievably simple, efficient and robust. I haven't come across any other 2d/3d design software with such a small footprint and an output of such level. I have been using it for 3d design ever since I downloaded and tried it out and never looked around for any other software except for toolpath generation. Once it is equipped with toolpath generation ( I hope it will), I doubt there will be any need to use another software.

Working with solvespace I have faced a dire need of two capabilities though which will make working with it very easy and much more productive.

1. Creating a new workplane based on the selection of a line and its end points. The workplane created this way will be perpendicular to the line at the end point which is closer to the viewer.
2. As we can create workplane in 3d, there is no way ( or at least I don't know If there is any ) to precisely rotate in 3d to create a workplane at the desired precise place so keyboard shortcuts will be extremely helpful as mouse rotation is not as precise and efficient as keyboard shortcuts.

Thanks.

phkahler:
The usual thing to do is make a pair of construction lines starting at a point (the end of your line). You can constrain them perpendicular or at a specific angle relative to something else. Then select the 2 lines and point and press Shif->W.

Without at least on line in the workplace there is no clear way to orient the workplane - left/right or up/down.

simrim1:
Yes we can create a workplane by adding two lines ( most of the times construction lines) at a particular point and adding angle constraints to get the required surfuce on which to create a workplane. This suggestion is to increase productivity without adding lines and constraints on them for a surface which is perpendicular to the desired line. This will help to decraese a bunch of lines in a complex design. The less number of lines is less clutter in 3d view. 

Selecting a single line is enough to create a workplane perpendicular to it. The suggestion for selecting its endpoints as well is to ensure the user intention of creating workplane perpendicular to it.

Again, this is just a suggestuion. I think this can be tagged as UI?


As for the keyboard shortcuts for creating workplane in 3d Is actually, in my humble oinion, a necessity. because you keep moving mouse left/right, up/down and in the corners to attain the visual surface, in 3d, where you want to create a workplane. Even then you may not be able to get the precise position. With keyboard shortcuts for rotation in 3d will help pinpoint the desired plane. This will obviously be tagged an enhancement.

Thanks.

MaxiPaille:
Hello guys,

[...]

Issue #423, opened 2019-05-22
Start sketch on selected face


Harvie:
Please allow to start new sketch by simply selecting the face like this:

![image](https://user-images.githubusercontent.com/276504/58212658-d2f7e480-7cef-11e9-8015-a5c8240b32fd.png)

whitequark:
This has been implemented and then reverted because of file format compatibility issues. Don't expect this to happen any time soon.

Harvie:
That sounds very sad...

whitequark:
AutoCAD has hundreds of people maintaining it, SolveSpace has one. Make your conclusions.

Evil-Spirit:
Actually, this is simplier than it was when I was doing this. So, this is not so hard to implement, and... In NoteCAD I still have no such feature :)

whitequark:
> Lacking such basic feature, but holding onto file format which does not support it? I understand there can be some issues between versions,

To be more clear, the problem is not that the current file format inherently cannot support a feature like "sketch on face". It can. But the maintenance overhead is very high, and testing is so complex, that I find it hard to justify shipping that feature with the existing format.

baryluk:
I remember using this feature some time ago, and it was amazing and speed up things a lot. Would be nice to have it back one day.

A library of tests (including UI interactions) would be something worth improving.

Also, do you have a way to donate to the project?

hinell:
I think this doesn't require any extra functionality to be embedded.
This is might be a mere automatic "_create a point on a surface -> create a workplane at this point -> constrain angle of every point on the workplane relative to surface ..._" macro. All necessary components supported by `.slvs` are already present.

phkahler:
Here is the challenge as I understand it:

A workplane in solvespace is represented by a point and a "normal". I put that in quotes because it's not a vector perpendicular to the plane as one might expect. A "normal" in this case is a quaternion used to represent the u,v basis vectors for the sketch plane. It is also possible to extract a conventional N vector from it. The reason for this is that 2d points in a sketch are stored with only 2 coordinates (u,v), they are not 3d points constrained to a plane. This all works fine until we try to define a workplane from a "face".

Faces are represented by... a point and a normal vector (hacked by storing it in 3 elements of a quaternion). Aside from causing much confusion for me, a face does not have enough information to construct U,V basis vectors for the workplane.

IIRC there are ways to define the u,v determinidtically but it is still arbitrary and can probably do strange thing at times.

[...]

PR #1054, opened 2021-06-06
Improve "Sketch in New Workplane" by adding two way of construction

MaxiPaille:
I improved "Sketch in New Workplane" by adding two way of construction:
- point and normal
image

- point and face
image

I also added description in error popup:
image

I tried to be as less invasive as possible, and also tried to respect the coding style.
I have tested this code on Windows only.

### DIFF
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -138,12 +138,26 @@ void Group::MenuGroup(Command id, Platform::Path linkFile) {
                         g.predef.q = wrkplg->predef.q;
                     } else ssassert(false, "Unexpected workplane subtype");
                 }
+            } else if(gs.anyNormals == 1 && gs.points == 1 && gs.n == 2) {
+                g.subtype = Subtype::WORKPLANE_BY_POINT_ORTHO;
+                Vector direction = SK.GetEntity(gs.anyNormal[0])->VectorGetNum();
+                g.predef.q       = Quaternion::From(direction, 0);
+                g.predef.origin  = gs.point[0];
+            } else if(gs.faces == 1 && gs.points == 1 && gs.n == 2) {
+                g.subtype = Subtype::WORKPLANE_BY_POINT_ORTHO;
+                Vector direction = SK.GetEntity(gs.face[0])->FaceGetNormalNum();
+                g.predef.q       = Quaternion::From(direction, 0);
+                g.predef.origin  = gs.point[0];
             } else {
                 Error(_("Bad selection for new sketch in workplane. This "
                         "group can be created with:\n\n"
                         "    * a point (through the point, orthogonal to coordinate axes)\n"
                         "    * a point and two line segments (through the point, "
-                               "parallel to the lines)\n"
+                        "parallel to the lines)\n"
+                        "    * a point and a normal (through the point, "
+                        "orthogonal to the normal)\n"
+                        "    * a point and a face (through the point, "
+                        "parallel to the face)\n"
                         "    * a workplane (copy of the workplane)\n"));
                 return;
             }
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -139,15 +139,13 @@ void Group::MenuGroup(Command id, Platform::Path linkFile) {
                     } else ssassert(false, "Unexpected workplane subtype");
                 }
             } else if(gs.anyNormals == 1 && gs.points == 1 && gs.n == 2) {
-                g.subtype = Subtype::WORKPLANE_BY_POINT_ORTHO;
-                Vector direction = SK.GetEntity(gs.anyNormal[0])->VectorGetNum();
-                g.predef.q       = Quaternion::From(direction, 0);
-                g.predef.origin  = gs.point[0];
-            } else if(gs.faces == 1 && gs.points == 1 && gs.n == 2) {
-                g.subtype = Subtype::WORKPLANE_BY_POINT_ORTHO;
-                Vector direction = SK.GetEntity(gs.face[0])->FaceGetNormalNum();
-                g.predef.q       = Quaternion::From(direction, 0);
-                g.predef.origin  = gs.point[0];
+                g.subtype       = Subtype::WORKPLANE_BY_POINT_NORMAL;
+                g.predef.q      = SK.GetEntity(gs.anyNormal[0])->NormalGetNum();
+                g.predef.origin = gs.point[0];
+            //} else if(gs.faces == 1 && gs.points == 1 && gs.n == 2) {
+            //    g.subtype = Subtype::WORKPLANE_BY_POINT_FACE;
+            //    g.predef.q      = SK.GetEntity(gs.face[0])->NormalGetNum();
+            //    g.predef.origin = gs.point[0];
             } else {
                 Error(_("Bad selection for new sketch in workplane. This "
                         "group can be created with:\n\n"
@@ -156,8 +154,8 @@ void Group::MenuGroup(Command id, Platform::Path linkFile) {
                         "parallel to the lines)\n"
                         "    * a point and a normal (through the point, "
                         "orthogonal to the normal)\n"
-                        "    * a point and a face (through the point, "
-                        "parallel to the face)\n"
+                        /*"    * a point and a face (through the point, "
+                        "parallel to the face)\n"*/
                         "    * a workplane (copy of the workplane)\n"));
                 return;
             }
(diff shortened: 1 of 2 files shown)


phkahler:
@MaxiPaille  This is a nice change. Not sure if you realize what's going on here vs what the original request was in #1048.  They wanted to select a line and a point to create a workplane. What you've implemented is to select a normal and a point.  A normal in Solvespace is represented by a quaternion, which is a full orientation and not just a direction vector. A workplane has to have a U and V axis, whcih can be pulled from the quaternion, or derived from 2 line segments and some cross products to ensure they're orthognal.

What you've implemented here is definitely a nice improvement, but I wanted to point out how its different from the original issue/request.

Another thing might be to allow using just a face without having to also select a point. I think each face is defined by a point and a quaternion. What you've implemented should also allow creating a workplane parallel to a face but at some distance depending which point was selected. It seems like another line or two of code should allow it to work by selecting only the face.

These are my comments after looking at the commit without even trying it yet ;-) It looks good!

phkahler:
@MaxiPaille  There is something wrong. If I use extrude to create a cylinder they way you've shown it works.

When I draw a rectangle and use lathe to create a ring, select a point and the top face and press shift->W I get a workplane through the cross-section of the ring rather than on the top face.

I'm not sure why this happens. This may need a little more understanding on how a face is represented. @jwesthues can you offer any insight here?

phkahler:
The lathe faces are created in Group::MakeLatheSurfacesSelectable()

I tried changing from:
```
en.numNormal` = Quaternion::From(0, n.x, n.y, n.z);
```
to:
```
en.numNormal = Quaternion::From(u,v);
```
Which seem to be a very different function (it's in util.cpp) but there was no change. The workplane is still wrong when selecting a face created by lathe.

MaxiPaille:
Hey @phkahler,
Thanks for your quick review.

I'm happy to hear that this is a nice improvement :).

I understand that my pull request do not answer what is discussed in the feature request #1048 originally. I was mainly focus on the fact that the simpliest way to describe a plane is a combination of a point and a normal and the software did not support it out of the box. Sorry for the confusion.
However, the original request should be easy to implement. I will investigate this.

I wanted to implement the "face only" way you suggested but i'm not sure how to get the _hEntity_ of the origin point from a face _Entity_. The only method I found is FaceGetPointNum() that return the Numerical value but the _predef_ structure takes a _hEntity_ as origin. If you can help me on this, it would be very cool.

I'm also think about implementing a "3 points" way. I imagine we could take the orientation of the camera to infer the normal direction, as we do with the "1 point" method. What do you think?

MaxiPaille:
Regarding the lathe bug, I've just break into the "point + face" method. The normal of the face seems ok. I bet there is an issue with the _Quaternion::From(direction, 0)_.

jwesthues:
I think the confusion is that SolveSpace normals in general are "oriented", representing 3 DOF equivalent to a unit quaternion (or 3x3 rotation matrix), as @phkahler notes above. We need this when we create the workplane, since that determines the rotation of the u and v vectors about n (which determines e.g. the rotation within the plane of our monitor when we View -> Align View to Workplane, or what the horizontal and vertical constraints do).

But for "workplane from face", the normal to a face is un-oriented, representing 2 DOF equivalent to a unit 3-vector. So we need some algorithm to choose that last degree of freedom, the twist about n. We could allow the user to select a line and use the projection of that line into the plane for u, or if all we have is the face we could just use `u = n.Normal(0)` and `v = n.Normal(1)`, which chooses them with an arbitrary map. That map is ugly (e.g. contains discontinuities), but the hairy ball theorem means it's the best we can do.

For "workplane from [oriented] normal", there's no fundamental problem; the code just seems wrong. We just need to copy all four elements of the original normal's unit quaternion, and not reduce it to a 3-vector like now. Probably also good to name new workplane subtypes, since these aren't `Subtype::WORKPLANE_BY_POINT_ORTHO`.

Final observation, note that even though the "oriented normal" has 3 DOF, we represent it as four scalars in a unit quaternion, and even though the "unoriented normal" has 2 DOF, we represent it as three scalars in a unit 3-vector. Neither geometric concept can be represented in any well-behaved mathematical form without that extra scalar and extra (unit length) constraint, due again to something like the hairy ball theorem.

MaxiPaille:
Hello @jwesthues,

Thanks for you detailed answer.
Obviously, i was missing some fondamentals! I appologize for that.

What I understand is that you have an extra scalar for each type of data to be able to keep orientation when needed (creation of workplanes for example).

Knowing that, my "improvements" are indeed totally wrong. Now I understand why it's not simple to implement.
The "workplane with normal" should be easy to fix, I will use _NormalGetNum()_ instead of _VectorGetNum()_.
For the other ways, I will need to increase my knowledge.

Regarding the Subtype, I wanted to be as less invasive as possible. But if you think it's worth it, I will create a subtype for each new way I'll add.

Thanks for your time.

jwesthues:
@MaxiPaille No problem at all, and the feature makes sense to me too. SolveSpace's use of the word "normal" to mean "oriented normal" is nonstandard, and thus a bit of a trap for new contributors.

The solution for "workplane with normal" is as you say, straightforward. The solution for the others requires a bit of math; but if you plan to do more work in SolveSpace (or any other software development involving solid geometry), then I believe you'll find the effort to learn that worthwhile. Let us know if you have questions...

phkahler:
> But for "workplane from face", the normal to a face is un-oriented, representing 2 DOF equivalent to a unit 3-vector.

@jwesthues  For Lathe and Revolve groups we could define the U vector as being in the direction of the line that sweeps out the face. V would be the rotation axis.Cross(U). Would it break compatibility if we redefined it as an "oriented normal"? We might want to revisit all of the face creation if possible - the idea of selecting a face and pressing Shift->W to draw on it is really nice, but probably not at the cost of incompatibility.

jwesthues:
@phkahler Yeah, I was thinking that too. The source geometry always provides a u vector, so it was a bad decision to discard that information. I think you could add that back without breaking backwards compatibility, since you're strictly adding information, though you'd have to support both the old and new representations to avoid breaking old linked files.

MaxiPaille:
If I understand correctly, you aim to refacto the face creation to always keep an "oriented normal" in the face entity?
I would like to help you but it seems too easy for me to break compatibility ^^.

And there is too much type of face I don't understand yet...
FACE_XPROD
FACE_N_ROT_TRANS
FACE_N_TRANS
FACE_N_ROT_AA
FACE_ROT_NORMAL_PT
FACE_N_ROT_AXIS_TRANS

But if I could help, let me know. I will continue to dig the source code on my side.

phkahler:
@MaxiPaille   That was the idea above. I still need to understand more and I think there will be a compatibility issue for sure. Face entites are created in the following functions:

```
Group::MakeExtrusionLines
Group::MakeLatheSurfacesSelectable
Group::MakeRevolveEndFaces
Group::MakeExtrusionTopBottomFaces
```
There may be some other places like when a group is copied.

Trouble is, if we change the definition for a face normal to include the U,V coordinate system definition that is probably going to break some things like point-on-face constraints. I'm not sure if those equations are regenerated which might define them in new terms if so.

Some stuff I've written about the entity types can be found in the [solve transforms document](https://github.com/solvespace/solvespace/tree/master/developer_docs)  More documentation on the internals really needs to be written ;-)  Redefining face normals would require updating some of the functions in entity.cpp to return numeric values and expression differently.

It's a fairly big job, and a really big one to ensure comaptibility. If you want to attempt it I'd suggest changing one of the functions above to produce a normal that's compatible with your pull request here, along with the normal functions in entity.cpp. If you can then define a work-plane using the modified surface then you're on the right track. Changing the other 3 functions should go a long way. Then you (or someone) needs to look at the constraint creation involving faces. I'm not sure there is anything else to change for new sketches - there almost always is - but the ability to use old sketches will be a requirement to merge it into master and I have no idea how to do that. Well some ideas but none that I think are good yet.

Solvespace does need more developers, so if you're interested keep hacking away at it. There are people familiar with most if not all of the code so just ask.

phkahler:
Also, see issue #423

MaxiPaille:
Thanks for your detailed answer. Glad to see that other people want this feature too.

I will read your documentation and try to understand everything. Does the project have a discord or somewhere to discuss ? Not sure if i should flood this thread in case of basic questions about code and stuff.

phkahler:
@MaxiPaille  I'm currently working on a PR for mirror groups which involves making a new copy-type in entity.cpp.  It's partly working, but faces are normals are maybe not right yet ;-) BTW I am using the existing convention for normals which I understand better now after all the discussion.

phkahler:
Merging this. While using a normal this way may not be the most common thing to do, there's no reason it shouldn't work if someone tries it. Also a very small code change.

hinell:
Which version this feature is available in? I have 3.0, but there is none...

kk6mrp:
It is part of this commit: https://github.com/solvespace/solvespace/commit/4308dc136b03d6727da96ec75cc7ae27d169040a

What commit are you on?

PR #1057, opened 2021-06-10
Add mirror groups. WIP.

phkahler:
This is my 4th try at mirror groups. Face entites are not working, normals may not either. Other entities work. The shell works, but faces can not be selected yet. Plans are to include options to mirror the group across selected points and to have it work in 2D as well.

Shift->M will create the new group.

### DIFF
--- a/src/describescreen.cpp
+++ b/src/describescreen.cpp
@@ -171,6 +171,8 @@ void TextWindow::DescribeSelection() {
             case Entity::Type::FACE_N_ROT_TRANS:
             case Entity::Type::FACE_N_ROT_AA:
             case Entity::Type::FACE_N_TRANS:
+            case Entity::Type::FACE_MIRROR:
+            case Entity::Type::FACE_N_COPY:
                 Printf(false, "%FtPLANE FACE%E");
                 p = e->FaceGetNormalNum();
                 Printf(true,  "   normal = " PT_AS_NUM, CO(p));
--- a/src/drawentity.cpp
+++ b/src/drawentity.cpp
@@ -86,6 +86,7 @@ void Entity::GetReferencePoints(std::vector<Vector> *refs) {
         case Type::POINT_N_ROT_TRANS:
         case Type::POINT_N_ROT_AA:
         case Type::POINT_N_ROT_AXIS_TRANS:
+        case Type::POINT_MIRROR:
         case Type::POINT_IN_3D:
         case Type::POINT_IN_2D:
             refs->push_back(PointGetNum());
@@ -96,6 +97,7 @@ void Entity::GetReferencePoints(std::vector<Vector> *refs) {
         case Type::NORMAL_N_ROT_AA:
         case Type::NORMAL_IN_3D:
         case Type::NORMAL_IN_2D:
+        case Type::NORMAL_MIRROR:
         case Type::WORKPLANE:
         case Type::CIRCLE:
         case Type::ARC_OF_CIRCLE:
@@ -121,6 +123,8 @@ void Entity::GetReferencePoints(std::vector<Vector> *refs) {
         case Type::FACE_N_TRANS:
         case Type::FACE_N_ROT_AA:
         case Type::FACE_ROT_NORMAL_PT:
+        case Type::FACE_MIRROR:
+        case Type::FACE_N_COPY:
         case Type::FACE_N_ROT_AXIS_TRANS:
             break;
     }
@@ -529,7 +533,8 @@ void Entity::Draw(DrawAs how, Canvas *canvas) {
         case Type::POINT_N_ROT_AA:
         case Type::POINT_N_ROT_AXIS_TRANS:
         case Type::POINT_IN_3D:
-        case Type::POINT_IN_2D: {
+        case Type::POINT_IN_2D:
+        case Type::POINT_MIRROR: {
             if(how == DrawAs::HIDDEN) return;
 
             // If we're analyzing the sketch to show the degrees of freedom,
@@ -573,6 +578,7 @@ void Entity::Draw(DrawAs how, Canvas *canvas) {
         case Type::NORMAL_N_COPY:
         case Type::NORMAL_N_ROT:
         case Type::NORMAL_N_ROT_AA:
+        case Type::NORMAL_MIRROR:
         case Type::NORMAL_IN_3D:
         case Type::NORMAL_IN_2D: {
             const Camera &camera = canvas->GetCamera();
@@ -783,7 +789,9 @@ void Entity::Draw(DrawAs how, Canvas *canvas) {
         case Type::FACE_N_TRANS:
         case Type::FACE_N_ROT_AA:
         case Type::FACE_ROT_NORMAL_PT:
+        case Type::FACE_MIRROR:
         case Type::FACE_N_ROT_AXIS_TRANS:
+        case Type::FACE_N_COPY:
             // Do nothing; these are drawn with the triangle mesh
             return;
     }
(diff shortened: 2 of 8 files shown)


kk6mrp:
This is pretty cool! Even as it is seems to work pretty well!

phkahler:
@kk6mrp Thanks for trying it out. If you mirror arcs they will fail, at least in 3D.

kk6mrp:
The 3D rotation seemed a little different than other objects rotate, not necessarily bad but just different. Is that because it isn't mirroring across anything?

I'll have to test it out more tonight, last night I just tried out mirroring a rectangle and cube...

phkahler:
> The 3D rotation seemed a little different than other objects rotate, not necessarily bad but just different. Is that because it isn't mirroring across anything?

It's mirroring across an invisible mirror plane. If you drag a point, the mirror plane is adjusted to that point and it's reflection are on opposite sides of the mirror. This means there are 3 degrees of freedom in the reflected part, not a full 6. When you do it in a 2D sketch, it will constrain the mirror to be perpendicular to the sketch plane so the reflected sketch elements are also in that plane.

I plan to add an option so any 1,2, or 3 points selected (or ends of a selected line segment) will be constrained on the mirror, or rather the mirror will be constrained to pass through those points. Also hoping to constrain the mirror to a face if one is selected, which would just reflect a solid across that face and join the 2 together. Just have to get the basics working first!

phkahler:
This is turning out to be hard. In EntityBase::NormalGetNum()  for the mirrored normals...
You can't use the Mirror() function of a normal/quaternion. That function builds a new "normal" by flipping the U and V basis vectors defined by the quaternion. the problem is negating those has no effect on the N basis vector which is needed for ARC entities. There is no way to represent a left handed coordinate system using the quaternions here.  So I'm creating a rotation quaternion to rotate the "normal" through the mirror planes normal. the x,y,z for that rotation quaternion should be the original normal.RotationN() vector cross product with the mirror normal vector (params 0,1,2 of the mirror group). The 4th (W) parameter would just need to be sqrt(1.0 - (x*x + y*y + z*z)). This seems to work half the time. If you extrude something with an arc and then mirror the solid, the mirrored arcs will appear correctly about half the time. If you drag them around the originals it will spin around correctly, and then again incorrectly. I'm wondering if there are times when the square root needs to be negative or something.

I'm also wondering if Quaternion.Mirror() is ever used and if we can change the definition so the resulting quaternion will have its RotationN() vector flipped, as that would be useful here.

kk6mrp:
Is this what you were referring to that faces cannot be selected yet? I get an assert when selecting any face of the mirrored object or when selecting two faces of the mirror I get an assert.
```
Missing (absent) translation for group-name'mirror'
File ./src/dsc.h, line 526, function FindById:
Assertion failed: t != nullptr.
Message: Cannot find handle.
Aborted (core dumped)
```

phkahler:
@kk6mrp  Yes, that looks like an example. The latest version of this doesn't crash for me, but the faces don't highlight when hovering over them, and they don't do anything when clicked. I think it has to do with having consistent parameters passed to Remap and RemapFaces. I think something is missing like remapping the surfaces?

phkahler:
@jwesthues  Could you look at this? I'm not sure why the surfaces of a mirrored extrusion can not be selected. I've had trouble with that before. I might not be defining the new face entities correctly, or might not be doing something right with remap.

Other entities are working now.

Adding options (constraints) when creating the group will get added after the groups mirror correctly ;-)

hydromisc:
Very cool. It looks like you're aiming to create two copies of the original geometry, one unchanged with `REMAP_BOTTOM` and the other mirrored with `REMAP_TOP`? I don't immediately see anything wrong with your remapping, though perhaps it would make sense to make just one mirrored copy. If I understand correctly, you couldn't use your current implementation e.g. to create an extrusion of just the mirrored version of a plane sketch, since you'd have both copies in the resulting group.

I do notice that you're setting `timesApplied` to 0 and 1 for the two copies (made with `CopyEntity()`) but never using it. That seems unnecessary but not harmful.

The problem you note with the quaternion's `n` is fundamental--if we had an entity that generated points/curves lying outside the uv plane, then there would be nothing we could do with that quaternion to mirror that entity. But I think all entities defined in terms of quaternions lie entirely in the uv plane, so that's not actually a problem? If it were then we should hit that when mirroring by importing scaled by a negative number too, and I don't recall any trouble there.

Have you tried graphically debugging the faces, like drawing the point and normal vector for each face in the group? I suspect that will be much easier than trying to get all the math right blind.

jwesthues:
Very cool. It looks like you're aiming to create two copies of the original geometry, one unchanged with REMAP_BOTTOM and the other mirrored with REMAP_TOP? I don't immediately see anything wrong with your remapping, though perhaps it would make sense to make just one mirrored copy. If I understand correctly, you couldn't use your current implementation e.g. to create an extrusion of just the mirrored version of a plane sketch, since you'd have both copies in the resulting group.

I do notice that you're setting timesApplied to 0 and 1 for the two copies (made with CopyEntity()) but never using it. That seems unnecessary but not harmful.

The problem you note with the quaternion's n is fundamental--if we had an entity that generated points/curves lying outside the uv plane, then there would be nothing we could do with that quaternion to mirror that entity. But I think all entities defined in terms of quaternions lie entirely in the uv plane, so that's not actually a problem? If it were then we should hit that when mirroring by importing scaled by a negative number too, and I don't recall any trouble there.

Have you tried graphically debugging the faces, like drawing the point and normal vector for each face in the group? I suspect that will be much easier than trying to get all the math right blind.

phkahler:
I think making 2 copies in the new group is consistent with step-rotating and step-translating. That way after mirroring half of something to create a whole, the entire thing can be step-copied or rotated again.

By setting timesApplied to 0 or 1 we have the option of making both copies copy-as mirror (the mirror math would only apply to one), but then I decided the first copy could just be a numeric copy, so that's redundant.

Arcs seem to be defined by 3 points in 3D and a normal. In order to mirror an arc, I think the normals N vector has to flip. Negating U and V just rotated 180 degrees in the plane. That might need some more investigation. Text entities don't scale by -1 properly but they do mirror correctly.

jwesthues:
> I think making 2 copies in the new group is consistent with step-rotating and step-translating. That way after mirroring half of something to create a whole, the entire thing can be step-copied or rotated again.

That's a good point. Perhaps then use the same flag that distinguishes start "with original" vs. "with copy <span>#</span>1" for translate/rotate, and make it user-selectable?

> Text entities don't scale by -1 properly but they do mirror correctly.

What fails in the text entities? I did a quick test, and they seem to work linking scaled by -1. The behavior of normals there is rather subtle and I can't immediately articulate why it works, but as far as I knew it did.

image

phkahler:
The problem with selecting faces was sovled by adding the MIRROR type to the condition in Group::Activate which sets ShowFaces = true;  Now they highlight when hovered and the "new" object faces can be clicked/selected. The original crashes when clicked. I've seen this before but don't recall what causes it ;-) more investigating....

phkahler:
Numeric copies of faces are crashing when selected because that type of face tries to access its point via handle when doing things like putting info in the text window. It seems to be an old bug that was never hit because nothing has made numeric copies of faces until now.

This is one of those times I wish I'd just fired up the debugger instead of trying to reason it out. I searched for the FACE enums and found one missing in the text window. Adding that in made it crash even more often, so I narrowed it down to 4 lines of code in there. And here I thought "obviously" I was remapping faces incorrectly and it was failing on face handle...

Symbian9:
> I've seen this before but don't recall what causes it ;-) more investigating...

Any news? "Mirror" feature is really needed to design various parts geometry features.

phkahler:
Had to manually rebase changes on master. Did that in a new branch so closing this PR

ruevs:
https://github.com/solvespace/solvespace/pull/1109 is the new pull request - just for historical cross linking.

PR #1074, opened 2021-07-25
Add optional helix pitch constraint.  WIP

phkahler:
The idea here is to use the group valB as the pitch parameter. I believe the math and constraint equation are correct.

There is a problem in the text window - it crashes when setting the pitch parameter.

Also not sure about using the parameter itself as a flag for enabling the constraint, but that's how it is now.

### DIFF
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -813,6 +813,12 @@ void Group::GenerateEquations(IdList<Equation,hEquation> *l) {
         AddEq(l, (EC(axis.z))->Minus(EP(6)), 5);
 #undef EC
 #undef EP
+        if(type == Type::HELIX) {
+            if(valB != 0.0) {
+                AddEq(l, Expr::From(h.param(7))->Times(Expr::From(PI))->
+                Minus(Expr::From(h.param(3))->Times(Expr::From(valB))), 6);
+            }
+        }
     } else if(type == Type::EXTRUDE) {
         if(predef.entityB != Entity::FREE_IN_3D) {
             // The extrusion path is locked along a line, normal to the
--- a/src/textscreens.cpp
+++ b/src/textscreens.cpp
@@ -299,6 +299,23 @@ void TextWindow::ScreenChangeGroupScale(int link, uint32_t v) {
     SS.TW.edit.meaning = Edit::GROUP_SCALE;
     SS.TW.edit.group.v = v;
 }
+void TextWindow::ScreenChangeHelixPitch(int link, uint32_t v) {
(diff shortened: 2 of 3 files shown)


phkahler:
OK, it works (crash fixed) but I'm still wondering if the setup is OK. Should there be a checkbox to "fix pitch" or is this enough - just click "set" and type a value. The value will change if you stretch or twist the helix, but setting it will lock it.

It's not obvious but setting it back to 0.0 should unlock it, but then you should have used a Revolve group ;-)

phkahler:
Added the check box to lock the pitch. This is more consistent and obvious.

Symbian9:
> Added the check box to lock the pitch.

Awesome! What about `number of turns`?

phkahler:
> > Added the check box to lock the pitch.
> 
> Awesome! What about `number of turns`?

Maybe? The length and angle are free to drag on the sketch. The "pitch" is adding a constraint so the angle and length are not independent.

I'm cheating by using the pitch parameter itself as a flag to indicate that the constraint should be applied. I suppose the same could be done with an additional "turns" parameter?

Do people really want to specify a number of turns? For small numbers it seems (to me) just twist manually and add a constraint (point on plane, lines perpendicular). For screws it seems like setting the length would be better.

BTW I'm going to try drawing some standard bolts to test this better.

Symbian9:
> Do people really want to specify a number of turns?

Guess it may be useful for Imperial units users, where "number of thread per inch" used instead of metrical "pitch" (single thread length in mm) — guess it could realized with some *radio buttons* to switch helix definition type:

- `(*) pitch: {pitch_length_value}mm`
- `( ) number of thread per inch: {actual_unit}`

- http://www.afi.cc/?page=customer&file=customer/asfain/customerpages/Blog/ScrewThreadTerminology.htm

> BTW I'm going to try drawing some standard bolts to test this better.

Try next:

- `#4-40 x 0.5`
- `1/4-20 x 5/8`
- `M3-0.50 x 10`

Symbian9:
Here is *Metric Bolt* `M6x1-35` with *twisted manually* threads:

- SLVS - [bolt_m6x1-35.zip](https://github.com/solvespace/solvespace/files/6882348/bolt_m6x1-35.zip)

slvs-m6x1-35

phkahler:
For now there isn't going to be an option for turns-per-unit length. If you're in inches and want 20 turns per inch, just type 1/20 for the pitch value. The input box can do math. I do believe I've got the units working correctly which was an issue, so thanks for reminding me to handle it correctly.

PR #1077, opened 2021-07-25
Add STL Linking. Experimental.

phkahler:
Some edge vertexes fail to add a point entity. Not sure why. Otherwise this is working OK.


phkahler:
Updated to use the point.Equals() function. This reduces the number of verticies and results in more edge vertexes. Still doesn't get all of them. I thought we had something for equality checking that allows a bit more error?

PR #1080, opened 2021-08-01
Don't reset showFaces every time a group is activated

tomsci:
I prefer not to have showFaces set and it really bugs me that it is reset to true every time a non-drawing group is activated.

Assuming that most people want the current behaviour (show faces for non-drawings, hide for drawings) I've added 2 booleans, `showFacesDrawing` and `showFacesNonDrawing`, and clicking the show faces button updates whichever of those is appropriate for the current group type, so that showFaces is preserved independently between drawing and non-drawing groups.

The current behaviour will be preserved if you never touch the show faces button (because of the `showFacesNonDrawing = true` in `Init()`), but if you eg untick it when a non-drawing group is active, showFaces will stay disabled whenever a non-drawing group is activated (and vice versa).

### DIFF
--- a/src/graphicswin.cpp
+++ b/src/graphicswin.cpp
@@ -407,6 +407,8 @@ void GraphicsWindow::Init() {
     showEdges = true;
     showMesh = false;
     showOutlines = false;
+    showFacesDrawing = false;
+    showFacesNonDrawing = true;
     drawOccludedAs = DrawOccludedAs::INVISIBLE;
 
     showTextWindow = true;
@@ -1364,6 +1366,14 @@ void GraphicsWindow::ToggleBool(bool *v) {
     Group *g = SK.GetGroup(SS.GW.activeGroup);
     if(*v && (g->displayOutlines.l.IsEmpty() && (v == &showEdges || v == &showOutlines))) {
         SS.GenerateAll(SolveSpaceUI::Generate::UNTIL_ACTIVE);
+    }
+
+    if(v == &showFaces) {
+        if(g->type == Group::Type::DRAWING_WORKPLANE || g->type == Group::Type::DRAWING_3D) {
+            showFacesDrawing = showFaces;
+        } else {
+            showFacesNonDrawing = showFaces;
+        }
     }
 
     Invalidate(/*clearPersistent=*/true);
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -416,11 +416,10 @@ std::string Group::DescriptionString() {
 }
 
 void Group::Activate() {
-    if(type == Type::EXTRUDE || type == Type::LINKED || type == Type::LATHE ||
-       type == Type::REVOLVE || type == Type::HELIX || type == Type::TRANSLATE || type == Type::ROTATE) {
-        SS.GW.showFaces = true;
+    if(type == Type::DRAWING_WORKPLANE || type == Type::DRAWING_3D) {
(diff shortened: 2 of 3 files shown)


tomsci:
The macOS test failure seems to be an infrastructure issue unrelated to the changes - it builds fine for me in Xcode :)

Symbian9:
>  it really bugs me that it is reset to true every time a non-drawing group is activated

Same with me, but opposite: it really bugs me that *"Make faces selectable with mouse"* deactivated on activating each *sketch group*.

Setting *"Make faces selectable with mouse"* ON / OFF option choice to work *globally* is what I really would be happy to get.

> The current behaviour will be preserved if you never touch the show faces button (because of the showFacesNonDrawing = true in Init()), but if you eg untick it when a non-drawing group is active, showFaces will stay disabled whenever a non-drawing group is activated (and vice versa).

It would be better if an extra option `[v] deactivate faces selecting in sketches` (enabled by default) would be added into *"Property Browser > configuration"*.

tomsci:
I think what I've implemented covers your usecase too:

# You open a new file. Face selection is on in non-sketches, off in sketches (the hard-coded defaults)
# You select a sketch. Face selection switches off (because you're now in sketch mode).
# If you now re-enable sketch mode, it will stay on in all subsequent switches to sketch mode.

So it requires just one toggle of the show faces button per launch, to get the behaviour you prefer, which is the same as what's needed to get the behaviour I prefer too, so I don't think it biases either usecase much.

I do agree it would be nice to be able to set the default behaviour in configuration, eg:

```
[x] Make faces selectable by default in sketches 
[x] Make faces selectable by default elsewhere
```

So then you could leave both those ticked, and faces will stay selectable in all views, and I could untick both, and makes faces not selectable in all views by default, and neither of us would need to click the showFaces button unless for when we really want to toggle the desired behaviour.

Although that begs the question of should all the other show/hide toggles have defaults in configuration too? Maybe it needs a "Features shown by default" section? I think that might be nice but I don't want to get too carried away?

Symbian9:
> So it requires just one toggle of the show faces button per launch

Good suggestion on "per-launch"

> I do agree it would be nice to be able to set the default behaviour in configuration, eg:
>
> ```
> [x] Make faces selectable by default in sketches 
> [x] Make faces selectable by default elsewhere
> ```

Shorten it a little:

```
[x] faces not selectable in sketches by default
[x] faces selectable in non-sketches by default
```

phkahler:
@tomsci   I was going to ask if this should be a saved configuration. But it really seems more like a toolbar item. I'm inclined to go with it the way you have it.

This really comes down to use-cases. I do a lot of point-on-face constraints in 3D so I need to have faces selectable. I also hope to add some additional constraints in the future like faces-perpendicular. But I can also see how they can be annoying in a number of situations so it comes down to what kind of design you've got and what kinds of construction you prefer.

tomsci:
Agreed, I wouldn't want to see it removed from the toolbar. The question is whether it'd be worthwhile to have the _default_ value for `showFacesDrawing` and `showFacesNonDrawing` be read from configuration, rather than being hardcoded in `GraphicsWindow::Init()`? If the answer is no, then I think the PR as-is is good to go? If yes, I'm happy to add that to the PR.

phkahler:
@tomsci  I'm merging it as-is. I think that's best or close to best option.

Issue #813, opened 2020-11-22
Suggestion: Use feet for imperial units with large values


phkahler:
The largest imperial unit of measure in SolveSpace is currently the Inch. For long lengths it would be good to switch the displayed units to feet.  For area square feet, and volume cubic feet.. These would only be for display, all inputs would still be in inches.

At what length, area, and volume should the displayed values change to feet?

Indoor dimensions are often in inches (I don't want to do mixed feet/in but anything is possible) so the change should be somewhere between 10 and 20 feet?  For area I'd guess more than a few hundred sq inches should be in sq feet? For volume perhaps change to cubic feet after 1000 in^3 which is about 0.58 cubic feet.

ruevs:
Mmm let's not forget [cubic shaftments](https://en.wikipedia.org/wiki/English_units#Length) ;-)

phkahler:
@ruevs  You may laugh, but in the US most home construction is done in feet and inches (and fractions of an inch). Putting windows in your house?  Width and height in inches. New flooring? square feet.  Sidewalk work? Length in feet.  A refrigerator may well indicate volume in cubic feet. Getting a load of dirt, gravel, or crushed stone?  Order that in cubic yards.

ruevs:
I know, but never got used to it. Sorry for the inappropriate joke.

phkahler:
> I know, but never got used to it. Sorry for the inappropriate joke.

Oh no. Joke was fine! I think our situation with units is kind of absurd ;-) I need to use more emoji.  It's really absurd since there was a major push to switch to SI as long ago as 1980 and we are still not there ;-(  it was really sad when NASA crashed a probe into Mars due to failure to convert units in the 90s. But I digress...  :-)

Symbian9:
As SolveSpace has *"Dimensions in Meters"* and  *"Dimensions in Meters"* for `SI` system of measurements, it would be good if there would ***"Dimensions in Feets"*** additionally to *"Dimensions in Inches"* for `Imperial` system of measurement too.

Also it would be cool if there would be a way to show Imperial dimensions with `ft`/`in` and alternatively with `'`/`"`:

- in Feet:
  - 2.67ft
  - 2' 8"
- in Inches:
  - 32in
  - 32"

phkahler:
Not sure why this wasn't closed when #1088 was merged.

PR #1088, opened 2021-08-12
Add option for displaying dimensions in feet and inches

tomsci:
So that eg 30 inches is shown as `2' 6"`.

The rationale for this additional display mode is that I found myself transcribing a blueprint in feet and inches and while entering eg `2*12+6` for "2 feet 6 inches" was easy enough, not being able to see the dimensions in the constraints in the same format as the original made it difficult to go back and verify that everything had been entered correctly.

When editing, the text field still shows the value in inches, eg `30.00` in this example. In other words `Unit::FEET_INCHES` behaves mostly the same as `Unit::INCHES`. I did debate having the editable value match the display value, but that would require modifying `LexNumber()` and `StringToMm()` to understand the `xx' yy"` syntax and it felt like that might be a bit too invasive a change for something that maybe only I would ever use. Especially given that specifying specific units while editing is not something that's allowed for anything else.

That being said, if there's a demand for the ability to do that I don't mind looking at adding an awareness of units to the expression parser. For eg I have occasionally found myself needing to enter a value in inches into a predominantly mm-based model and having to remember to switch dimension units before setting the constraint, and then switching back again (compared to simply writing "0.3in" and having that translated for me into mm) is a little tedious. But maybe that should be a separate PR?

### DIFF
--- a/src/confscreen.cpp
+++ b/src/confscreen.cpp
@@ -64,7 +64,7 @@ void TextWindow::ScreenChangeCameraTangent(int link, uint32_t v) {
 }
 
 void TextWindow::ScreenChangeGridSpacing(int link, uint32_t v) {
-    SS.TW.ShowEditControl(3, SS.MmToString(SS.gridSpacing));
+    SS.TW.ShowEditControl(3, SS.MmToString(SS.gridSpacing, true));
     SS.TW.edit.meaning = Edit::GRID_SPACING;
 }
 
@@ -89,7 +89,7 @@ void TextWindow::ScreenChangeExportScale(int link, uint32_t v) {
 }
 
 void TextWindow::ScreenChangeExportOffset(int link, uint32_t v) {
-    SS.TW.ShowEditControl(3, SS.MmToString(SS.exportOffset));
+    SS.TW.ShowEditControl(3, SS.MmToString(SS.exportOffset, true));
     SS.TW.edit.meaning = Edit::EXPORT_OFFSET;
 }
 
@@ -171,7 +171,7 @@ void TextWindow::ScreenChangeCanvasSize(int link, uint32_t v) {
     }
     int col = 13;
     if(v < 10) col = 11;
-    SS.TW.ShowEditControl(col, SS.MmToString(d));
+    SS.TW.ShowEditControl(col, SS.MmToString(d, true));
     SS.TW.edit.meaning = Edit::CANVAS_SIZE;
     SS.TW.edit.i = v;
 }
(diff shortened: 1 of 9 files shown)


phkahler:
This is something I was considering along with #813 but is definitely not the same thing. I like that the input reverts to numbers and doesn't require parsing mixed units.

phkahler:
Did you want to go so far as fractions of an inch?  2 3/8"  I'm thinking you can get fractions by multiplying the fractional inches by 64 (with rounding) and then shifting the zeros right. So 24/64  is 11000 / 100000 in binary - right shift until there are no trailing zeros in the numerator gives 11/ 100 or 3/8.  Just thinking this goes along with mixed feet and inches.

tomsci:
> Did you want to go so far as fractions of an inch

That hadn't occurred to me but sure, happy to give it a go :) I don't think unicode fractions go as far as 64ths (let alone the font being used) so it might look a little ugly. I'll see how it looks.

It also occurred to me that having a "toggle imperial/metric" menu item might be useful too, complete with keyboard shortcut, which did meters<->feet and mm<->inches?

phkahler:
Not sure how far unicode fractions go. I wasn't even thinking of special characters, but now that you mention it yeah!

Keyboard shortcuts for this - nope. Available keys are is short supply and this is a per-model kind of feature.

tomsci:
I've pushed a change to add fractions - here's how it looks for variously-sized constraints:
image

Symbian9:
> I've pushed a change to add fractions - here's how it looks for variously-sized constraints:

Nice! 

@tomsci, Please, add also `-` character between inches & feets (without spaces, but keep space between integers & fractions) to follow *"Architectural"* format:

- `3'-0"`
- `⌀8'-2 1/8"
- `2'-1 1/4"`
- `11 3/16"`

FTR, Here is *"QCAD 3.x > Drawing Preferences"* screenshot (in QCAD maximum fraction precision could be set to `0'-0 1/256"`):

- https://qcad.org/rsforum/viewtopic.php?t=5464
  pic

tomsci:
> Please, add also - character between inches & feets

👍

> `3'-0"` - is it trailing zeros?

The reference linked on #813 https://www.construction53.com/2011/08/understanding-dimensions/ says "Even if the dimension has no inches, the zero remains as part of the designation."

> `11 3/16"` - should it be `0'-11 3/16"`?

Same reference says "When dimensions are small (less than 1 foot), only inches are used."

This seems to line up with some random internet searches for blueprints such as https://prestigelivingspaces.com/wp-content/uploads/2018/04/Blueprint-of-a-house-19157126.jpg which has `6'-0"` and `7 3/4"`.

Although one of the diagrams on that page appears to use both `12"` and `1'-0"` to represent 12 inches, so I'm not sure what the correct thing to do there is. I've gone for `1'-0"`.

Updated example:
image

phkahler:
Going to merge this. But if anyone is feeling ambitious we might use the superscript and subscript unicode characters for the fraction, assuming they are in the vector font.

PR #1090, opened 2021-08-14
Fix/silence mac build warnings

tomsci:
As per Xcode 12.4 you can at least do a warning-free incremental build
with these changes. There are still plenty of warnings in a full build
(mostly from thirdparty components) but with these changes you can at
least develop on mac and see if/when you've added any new warnings when
doing incremental builds.

This PR ended up mostly being to add `-DGL_SILENCE_DEPRECATION`. The fact that OpenGL is deprecated in the first place on macOS is concerning, but other projects seem to be OK with [ignoring that](https://www.sublimetext.com/blog/articles/hardware-accelerated-rendering)...

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -117,6 +117,7 @@ if(WIN32)
         ${SPACEWARE_LIBRARIES})
 elseif(APPLE)
     add_compile_options(
+        -DGL_SILENCE_DEPRECATION
         -fobjc-arc)
 
     list(APPEND platform_SOURCES
--- a/src/drawconstraint.cpp
+++ b/src/drawconstraint.cpp
@@ -1006,7 +1006,6 @@ void Constraint::DoLayout(DrawAs how, Canvas *canvas,
             Vector center = SK.GetEntity(circle->point[0])->PointGetNum();
             Quaternion q = SK.GetEntity(circle->normal)->NormalGetNum();
             Vector n = q.RotationN().WithMagnitude(1);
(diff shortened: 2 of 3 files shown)


ruevs:
> The fact that OpenGL is deprecated in the first place on macOS is concerning

The fact that they still have not removed it is a testament to how important/fundamental it is. Apple just love to deprecate stuff and break backward compatibility. I think it's due to them controlling the "whole stack" - hardware and software.

phkahler:
@tomsci  so you're doing development on a mac? We need more of that. BTW you can build native for M1 as well.

tomsci:
> The fact that they still have not removed it is a testament to how important/fundamental it is. Apple just love to deprecate stuff

Agreed :/ I don't really see how they could ever remove OpenGL, but you never know with them sometimes...

> so you're doing development on a mac? We need more of that.

Mostly, yes. I was thinking about some more mac-y improvements like supporting multiple windows/documents since macs don't really give you a good way of running multiple instances of the process at once (at least, not through the UI, you can if you drop to the command line) - but the amount of global data makes that a tricky proposition. So instead I've been working on a clipboard serialisation scheme that allows cross-process copy-and-paste using the system clipboard, to make it easier to work with multiple SolveSpace instances, but it's currently mac-only until I can get round to adding Windows and GTK clipboard glue code. Plus there's a few bugs I haven't ironed out yet.

I don't have an M1 yet, waiting for something with more ports. It may be a while :-)

PR #1092, opened 2021-08-15
Set OSX minimum supported version

### LINKED ISSUE #1086: Feature Request: Backwards compatibility for macOS 10.14.5
### System information

- **SolveSpace version:** 3.0~0e0b0252 (release)
- **Operating system:** macOS 10.14.5

### Expected behavior

Application opens under macOS 10.14.5

### Actual behavior

Popup saying Solvespace is only compatible with macOS 10.15.x

### Additional information

For organizational reasons I cannot upgrade yet would like to still use this software... Is there any way to maintain a single major OS version of backwards compatibility?

tomsci:
As per [cmake docs](https://cmake.org/cmake/help/v3.21/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html#variable:CMAKE_OSX_DEPLOYMENT_TARGET) this is how you configure `-mmacosx-version-min` in cmake land.

To my surprise this is all that was needed to get a binary compiled on 11.5.1 to launch on 10.12.6, which is as far back as I can test. 10.12.6 is supported on macs at least as far back as mid 2012.

Fixes #1086.

It did cause some build warnings from libomp:

    ld: warning: dylib (/usr/local/lib/libomp.dylib) was built for newer macOS version (11.0) than being linked (10.12)

But again to my surprise this didn't seem to actually cause any problems (although different/future libomp versions may behave differently). In my very limited testing this change doesn't appear to have any noticeable impact when running on recent macOS versions.

cc @ruevs

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,11 @@ if(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
 endif()
 
+if (APPLE)
+    # Docs say this must be set before the first project() call
+    set(CMAKE_OSX_DEPLOYMENT_TARGET "10.12" CACHE STRING "macOS minimum supported version")
+endif()
+
 # project
 
 # NOTE TO PACKAGERS: The embedded git commit hash is critical for rapid bug triage when the builds


tomsci:
I've tested this change also builds ok for arm64 (on an intel machine) although I don't have anything I can actually run the arm64 version on.

ruevs:
Well this looks beautiful! @vespakoen what do you think?

@phkahler should we rebuild the 3.0 macOS binaries? Or maybe make a 3.1? A nontrivial amount of improvements for all platforms has accumulated since 3.0.

phkahler:
@ruevs  I was thinking about a 3.1 maybe by the end of the year, but fixing the mac issues may warrant doing it sooner. The other big thing IMHO would be getting the releases to include all source in the zip - including dependencies?

vespakoen:
Looks good to me, I guess this will also work on arm64, i'll test it out when I get the chance.

ruevs:
@tomsci you chose 10.12, because that was the oldest you could test on I presume? Because it looks to me that we should run fine on 10.8 (the oldest x64 only version), maybe even 64bit 10.7? https://en.wikipedia.org/wiki/MacOS#Release_history

Theoretically I could test on 10.8 if I install it on my 2008 MacBook Pro... but that would be quite an undertaking :-)

tomsci:
Yep 10.12 is the only thing I had lying around. My recollection was 10.13 dropped a bunch of hardware support so 10.12 was a good reference point that more machines would be capable of upgrading to, but wikipedia doesn't seem to agree. The 10.12 machine I have here isn't being offered any further updates but maybe there are other reasons for that... We might start seeing more critical missing functionality, and have to spend time finding workarounds, if we go back too far.

In any case 10.12 support is a step in the right direction and exceeds what the reporter was originally asking for :)

phkahler:
@ruevs rebuilding the 3.0 mac binaries might be a good idea. Not sure how to go about that :-)

Symbian9:
> Theoretically I could test on 10.8 if I install it on my 2008 MacBook Pro... but that would be quite an undertaking - the machine has not been used in 10+ years and is on 32bit 10.6 :-)

@tomsci, It would cool to support as oldest OS/hardware/platforms, as possible. Please, at least try do that ;-)

> I was thinking about a 3.1 maybe by the end of the year... rebuilding the 3.0 mac binaries might be a good idea. Not sure how to go about that :-)

@phkahler, I think actually there are a lot of new features & bugfixes for release `3.1` this month. And for the end of this year `3.2` which, I suppose, also would include few more new features & bugfixes.

Also It would be cool to add `solvespace-xp.exe` (OpenGL1.x/32bit) builds for `3.1` release (https://github.com/ruevs/solvespace/releases/tag/WindowsXPPost3.0) and include its build scripts & patch to GitHub Actions workflow for `edge` releases.

tomsci:
> It would cool to support as oldest OS/hardware/platforms, as possible

Agreed, but you know what they say - if you don't test something, it doesn't exist. Given the vast amount of churn that goes on in macOS dev environments I'd rather not promise support that I haven't verified actually works :) I've not sure how far back the current Xcode 12.5 toolchain even _can_ build for - it's not guaranteed that the answer is infinitely far back, Apple happily drop support for stuff like that according to their own schedules. 10.12 is already 5 years old, not a patch on Windows compat but still pretty good for macOS.

But if someone with more old macs (and time) than me wants to test it, hopefully this PR provides a place to start.

Symbian9:
>  I'd rather not promise support that I haven't verified actually works... But if someone with more old macs (and time) than me wants to test it, hopefully this PR provides a place to start.

@tomsci, I will try find testers ;)

PR #1095, opened 2021-08-20
Show suppressed groups in gray in the text window

tomsci:
Throwing this one out there to see what people think.

I often lose track of which bits of a model I've temporarily hidden via the "suppress this group's solid model" tickbox. I could really do with a way to see what's suppressed and what's not in the main text window. So I made a simple patch to show suppressed groups in gray rather than the usual link blue:

image

### DIFF
--- a/src/textscreens.cpp
+++ b/src/textscreens.cpp
@@ -129,7 +129,7 @@ void TextWindow::ShowListOfGroups() {
                "%Ft%s%Fb%D%f%Ll%s%E "
                "%Fb%s%D%f%Ll%s%E  "
                "%Fp%D%f%s%Ll%s%E "
-               "%Fl%Ll%D%f%s",
+               "%Fp%Ll%D%f%s",
                // Alternate between light and dark backgrounds, for readability
                backgroundParity ? 'd' : 'a',
                // Link that activates the group
@@ -146,6 +146,7 @@ void TextWindow::ShowListOfGroups() {
                ok ? ((warn && SS.checkClosedContour) ? "err" : sdof) : "",
                ok ? "" : "ERR",
                // Link to a screen that gives more details on the group
+               g->suppress ? 'g' : 'l',
                g->h.v, (&TextWindow::ScreenSelectGroup), s.c_str());
 
         if(active) afterActive = true;
--- a/src/textwin.cpp
+++ b/src/textwin.cpp
@@ -203,7 +203,7 @@ const TextWindow::Color TextWindow::fgColors[] = {
     { 'r', RGBi(  0,   0,   0) },  // Reverse   : black
     { 'x', RGBi(255,  20,  20) },  // Error     : red
     { 'i', RGBi(  0, 255, 255) },  // Info      : cyan
-    { 'g', RGBi(160, 160, 160) },
+    { 'g', RGBi(128, 128, 128) },  // Disabled  : gray
     { 'b', RGBi(200, 200, 200) },
     { 0,   RGBi(  0,   0,   0) }
 };


Symbian9:
> Throwing this one out there to see what people think.

I already thought about this too, but suppose it should not be in color (it is difficult to recognize in the list where each odd line is gray), BUT instead show *few character prefixes* before `group name` to represent some important group properties:

- `+g{id}-{name}` - Union group;
- `-g{id}-{name}` - Difference group;
- `%g{id}-{name}` - Assembly group;
- `#g{id}-{name}` - Intersection group;

For "Linked/Assembled group" add also `L` prefix:

- `+Lg{id}-{name}` - Linked/Assembled Union group;
- `-Lg{id}-{name}` - Linked/Assembled Difference group;
- `%Lg{id}-{name}` - Linked/Assembled Assembly group;
- `#Lg{id}-{name}` - Linked/Assembled Intersection group;

Also there should be prefix `Δ` for groups with *"force NURBS surfaces to triangle mesh"*:

- `Δ+g{id}-{name}` - Union group with forced NURBS to triangle mesh;
- `Δ+Lg{id}-{name}` - Linked/Assembled Union group with forced NURBS to triangle mesh;

And finally for *"Suppressed"* groups show only `$` and `Δ` before it (if it has forced NURBS to triangle mesh) - until suppressed it does not matter which type it is (but even it suppressed forced NURBS to triangle mesh still is in action):

- `$g{id}-{name}` - Suppressed group;
- `$Lg{id}-{name}` - Suppressed Linked/Assembled group;
- `Δ$Lg{id}-{name}` - Suppressed Linked/Assembled group with forced NURBS to triangle mesh.

ruevs:
No one will remember the meaning of all these prefixes. Even the person who implemented them three days later :-)

tomsci:
Your mileage may vary on the legibility of gray-on-black of course, but it's worth noting that pasting the screenshot into github (and the associated conversion into a jpeg) did rather mangle the colour profile. The blue should be little brighter and the contrast between it and the gray text is actually pretty good. Even the gray text on the alternate dark gray background is perfectly legible to my eyes since the brightness of the text is about the same as the normal link colour. I can't speak for how it looks on an old LCD with bad colour reproduction of course.

The colourblindness accessibility aspect is a good point that I hadn't considered, but if we're going down that route those green construction lines versus white normal lines must surely be a bigger problem :-)

I wouldn't object to having more info encoded in the text window but I don't personally find a need on a regular basis to know what's a revolve versus an extrude at-a-glance, whereas finding the things I've temporarily toggled to hidden, I do.

phkahler:
@tomsci  This looks good. We've had people want the ability to (un)suppress the groups right from the list among other things. This at least helps to find those groups without taking any additional space. I've used characters in my group names to indicate what they are -  sketch, extrusion, etc. It is quite useful, but I think the user can add their own preferred indicators in the name.

As for the color choice, you might want to darken it just a little bit more for better contrast? Oh, I see you're using predefined colors. Is that one used somewhere else?

tomsci:
I'm using the pre-existing `'g'` colorcode which has no description in the source, and I can't see that it's being used anywhere else (either via `%Fg` or `%Fp` followed by `'g'`). The other predefined options are `'h'` (dark gray) and `'b'` (light gray). `'h'` is a bit too low contrast with the alternate-row background, and `'b'` is a bit too bright for a grayed-out option, to my eyes at least. Neither of these are used anywhere either that I could see.

If we do want to keep using `'g'` I guess I should add a comment in the `fgColors` to say something like 

```c++
    { 'g', RGBi(160, 160, 160) },  // Disabled  : gray
```

phkahler:
That sounds good. Keep it 'g' and maybe darken it a little more.

PR #1096, opened 2021-08-22
Add a "∆" suffix to groups which have "force to triangle mesh" ticked.

tomsci:
Following on from comments by @Symbian9 https://github.com/solvespace/solvespace/pull/1095#issuecomment-903062126 I thought I'd try adding a `∆` indicator on groups that have 'force to triangle mesh' ticked, since that's also something that sometimes you need to scan down the entire group list to find (in my case, usually when trying different things to fix a NURBS issue).

I didn't go the whole way to adding identifiers for different group types, as @ruevs mentioned there's a point where you're making things less clear rather than more, if your group names are all prefixed with a string of punctuation. I did like the idea of using a `∆` character to indicate `forceToMesh` is set though, and I added it as a suffix in parentheses outside of the link colouration instead, to make it clearer it's an annotation rather than actually being part of the group name.

I think going for the modest change of just showing `forceToMesh` is a better plan, rather than attempting something more sweeping that not everyone will agree on, although I don't particularly mind if you don't want to take this one into master. It may well be something that not many other people really need to see. I just thought I should see how it looks given the interesting suggestions @Symbian9 proposed, and I think it actually looks pretty useful :-)

image

In case anyone is wondering, the triangle character is U+2206 INCREMENT from the "mathematical operators" category, and not a repurposed greek capital delta. It looked better than some of the most obscure triangle options. I've checked it will render right on macOS and Windows, haven't checked linux because my ubuntu VM is busy updating itself (very slowly).

### DIFF
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -404,7 +404,11 @@ bool Group::IsForcedToMeshBySource() const {
 }
 
 bool Group::IsForcedToMesh() const {
-    return forceToMesh || IsForcedToMeshBySource();
+    return forceToMesh || IsTriangleMeshAssembly() || IsForcedToMeshBySource();
+}
+
+bool Group::IsTriangleMeshAssembly() const {
+    return type == Type::LINKED && linkFile.Extension() == "stl";
 }
 
 std::string Group::DescriptionString() {
--- a/src/importmesh.cpp
+++ b/src/importmesh.cpp
@@ -171,7 +171,6 @@ bool LinkStl(const Platform::Path &filename, EntityList *el, SMesh *m, SShell *s
         addUnique(verts, tr.b, normal);
         addUnique(verts, tr.c, normal);
     }
-    SK.GetGroup(SS.GW.activeGroup)->forceToMesh = true;
     dbp("%d verticies", verts.size());
 
     BBox box = {};
--- a/src/sketch.h
+++ b/src/sketch.h
@@ -269,6 +269,7 @@ public:
     void Generate(EntityList *entity, ParamList *param);
     bool IsSolvedOkay();
     void TransformImportedBy(Vector t, Quaternion q);
+    bool IsTriangleMeshAssembly() const;
     bool IsForcedToMeshBySource() const;
     bool IsForcedToMesh() const;
     // When a request generates entities from entities, and the source
--- a/src/textscreens.cpp
+++ b/src/textscreens.cpp
@@ -123,13 +123,18 @@ void TextWindow::ShowListOfGroups() {
               sprintf(sdof, "%-3d", dof);
             }
         }
+        std::string suffix;
+        if(g->forceToMesh) {
+            suffix = " (∆)";
+        }
+
         bool ref = (g->h == Group::HGROUP_REFERENCES);
         Printf(false,
                "%Bp%Fd "
                "%Ft%s%Fb%D%f%Ll%s%E "
                "%Fb%s%D%f%Ll%s%E  "
                "%Fp%D%f%s%Ll%s%E "
-               "%Fp%Ll%D%f%s",
+               "%Fp%Ll%D%f%s%E%s",
                // Alternate between light and dark backgrounds, for readability
                backgroundParity ? 'd' : 'a',
                // Link that activates the group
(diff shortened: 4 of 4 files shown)


Symbian9:
Great! I may just recommend to show `∆` in first place (e.g. `∆ g{id}-{name}`), instead of show it in parenthesis after whole group name.

phkahler:
This one I like. People submit models with issues that they worked around by forcing to mesh and it can take a while to figure out which group/s have that checked.

@tomsci  FYI the text window font is identical on all platforms, its GNU unifont - bitmap font that contains every glyph in the multilingual plane. Its compiled in, as is the vector font used on sketches.

tomsci:
> the text window font is identical on all platforms, its GNU unifont

Ah, nice. That simplifies matters :)

ruevs:
This is not bad. And I like it as a suffix. In "serious" models I give the groups meaningful names and want to see them, rather than (a bunch of) flags.

It should be noted (unless my memory is totally confused), that once a group is "force triangle mesh" everything after it is forced as well. So this marker is essentially "this and everything after this is not NURBS".

phkahler:
> It should be noted (unless my memory is totally confused), that once a group is "force triangle mesh" everything after it is forced as well. So this marker is essentially "this and everything after this is not NURBS".

@ruevs Correct. Every group is actually NURBS, but for combining with the previous group one of 2 things happens:

1) combine NURBS with a boolean and then convert entire model (up to that point) to mesh for display.
2) Convert just 1 group to mesh and then combine with the previous model using mesh boolean. No NURBS shell will exist for the entire model after this.

Linking an STL will force to mesh, while linking an IDF will create a NURBS shell for the group.

phkahler:
Does this show if you link an STL file? It should.

tomsci:
>> Linking an STL will force to mesh

I wasn't aware of this (never used linked STLs in SS), it doesn't seem to behave that way in the UI? If I link an STL then do a subsequent sketch-and-extrude, the extrude group does not show "model already forced to triangle mesh".

In fact, unless I manually tick "force NURBS surface to triangle mesh" on the STL assemble group, I don't get any solid model for the STL whatsoever...

> Does this show if you link an STL file? It should.

It only shows on groups that explicitly have `forceToMesh` set, not ones that inherited forced-ness from an ancestor group. And STL link groups do not (currently) seem to set `forceToMesh` (nor do they otherwise force the force-to-mesh setting, see above!).

phkahler:
I tried linking as STL and it works and shows up fine. I don't have the "force to triangle mesh" option on that group. I also created a rectangle and extruded-as-difference to cut part of the STL object off.

BTW github has STL file 3d previews now. Click an STL [here:](https://github.com/phkahler/mellori_ITX/tree/master/STL)

tomsci:
Hmm ok I must've fat-fingered the test and linked a .slvs file instead of the .stl I intended, sorry for the confusion. (The .slvs in question wouldn't display right unless forced to triangle, because it was one of my fine NURBS problem projects!) However the link of an STL file is still a bit weird. Here's what happens when I do it more carefully:

* Start with an empty project, with the default groups g001-#references and g002-sketch-in-plane
* Choose "New Group->Link / Assemble", select an STL file.
* g003-stl-filename is created and is showing "model already forced to triangle mesh"

The weirdness is that the act of linking has set `forceToMesh` to true on `g002`, ie the group _before_ the STL import. This seems very odd to me because surely it should be `g003` that has that set, a link/assemble operation shouldn't modify existing groups should it?

If I manually untick force from g002, the solid model from g003 disappears (kindasorta fair enough) and reappears if I set forceToMesh on g003 (again, makes sense). But assuming that the force should be on g003 and not g002, shouldn't there be no way in the UI to untick it, given that (if I've understood this right) an STL must be forced to triangle mesh in order to display at all?

Assuming I've not got this all wrong, would you like me to fix those two issues as part of this PR, ie change the force to be set on the STL group itself not the previous one, and prevent it from being deselected in the UI?

phkahler:
@tomsci  you are right, the force to mesh should be on g003 not g002. Ideally it would not be optional. I guess I thought it wasn't optional since g003 doesnt have the checkbox, but apparently that's because it was auto-checked in the wrong group.

If you're in there and see how to fix go ahead :-)

tomsci:
@phkahler have pushed 0495a54 which avoids selecting the wrong group by instead changing `IsForcedToMesh()` to always return true for STL assembly groups via a new helper function `Group::IsTriangleMeshAssembly()`. The code which previously set forceToMesh=true didn't have direct access to the group info so this seemed like a simpler change than reworking which function knows what about linking. I did check for anything else which implicitly requires the group to be meshed, like the STL import, but didn't find anything, at least not that did it by automatically setting `forceToMesh=true`.

I've also made STL groups always show the `∆` which feels like it makes sense as that group is what's forcing the model to now be meshed, even though the group doesn't actually have `forceToMesh=true` set by the user. Hopefully it makes more sense when you look at it than my explanation :-)

phkahler:
@tomsci  Both your explanation and the code seem clear to me.

PR #1097, opened 2021-08-22
Use std::map to store STL link indices to speed up initial load time

DevyHeavy:
Please compare logic to what was there before. Seems like it is OK.

No recorded definite metrics/benchmarks, but 124941 vertex count mesh initially loads much faster, I may even say it cuts load time by about 50%+ for huge meshes.

Still room for improvement in parts of code not touched by this PR.

### DIFF
--- a/src/importmesh.cpp
+++ b/src/importmesh.cpp
@@ -10,6 +10,8 @@
 
 #define MIN_POINT_DISTANCE 0.001
 
+typedef std::map<long int, std::map<long int, std::map<long int, long int>>> indexMap;
+
 // we will check for duplicate verticies and keep all their normals
 class vertex {
 public:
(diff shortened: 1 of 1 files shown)


phkahler:
@DevyHeavy  This might be OK, but I have some comments.

You're effectively rounding the coordinates to the nearest point on a 3d grid. That means points within the same cell will be considered the same which they would before. However, points at the "high" end of one cell and the "low" end of the adjacent one will be considered different regardless of how close they are. That's probably not an issue considering what this is meant to do.

What it does: It's meant to collect the normals of triangles that share a vertex in order to 1) eliminate redundant verticies 2) determine which verticies are on an edge (meaning sharp edges like the perimeter of a hole, edge of a cube etc). So we can create Solvespace Point entities in the sketch that the user can build on.

Having too many points will kill performance. Having redundant points is somewhat irrelevant. So given all that, this will produce different sets of points in some cases, but that's probably OK. The feature is still subject to breakage in my mind since it's not part of a numbered release. Changing the number of points will change their indexes, which will cause constraints in linked sketches to apply to the wrong point. This could probably be minimized by numbering them with a unique hash of the integer coordinates of the point. The index (ID) doesn't have to be sequential, but I forget how big it is.

ruevs:
> However, points at the "high" end of one cell and the "low" end of the adjacent one will be considered different regardless of how close they are.

[KD tree](https://en.wikipedia.org/wiki/K-d_tree) searches have the same problem. So at the end they check adjacent cells. With the map(map(map)) implementation here this will not be possible.

DevyHeavy:
Sorry it has been a while.

Good points. Haven't looked into KD trees yet, I apologize if what I am about to say is covered in that.

I guess I could check the distances between the points in the 24 neighboring cubes as well as the central 25th.

I also realize that `MIN_POINT_DISTANCE` should probably be based on the size of the bounding box of the points in some way. Alternatively, maybe there could be a way to quickly find the average density of the points? Like the average distance between points in a sample (but how do you take a sample that reflects the space of ALL points, probably can't). The problem is that an STL with a large scale means points are more spread apart, and a small `MIN_POINT_DISTANCE` will basically end up comparing every point to each other again (the problem trying to be solved here).

phkahler:
@DevyHeavy  No worries. I'm not sure KD-trees are the answer either. They tend to want all the points ahead of time to construct the tree. We kind of want to check each point as it's added to see if it's "the same" as a previous point. Same points keep track of all the related surface normals for edge identification.

The MIN_POINT_DISTANCE is really meant to see if points are identical except for things like rounding error. In my mind this should be independent of the bounding box. However, if someone tries to model a planet a tiny number might be the wrong thing.

I suspect checking the 26 neighbors would still be a lot faster than what I wrote even if it's a little ugly.

Issue #39, opened 2016-09-15
Single window UI


Symbian9:
Make UI look like 'single window' by merge "Property Browser" inside main window UI.

mmiscool:
I vote for this one also. Would make it easier to manage and use. I frequently maximise windows and this would make life easier. 

Also it might provide and opertunity to make the interface a bit easier to modify.

whitequark:
> Also it might provide and opertunity to make the interface a bit easier to modify.

It won't. For that matter the UI code will become more complex if/when this lands.

ruevs:
No! Haven't you heard of dual/multiple monitors?!

whitequark:
@ruevs No one is going to remove the two-window UI completely, if it's implemented it will just be an option. Like in GIMP.

ruevs:
These are the kind of pointless options that just cause boat.

whitequark:
If you don't see a need for it, that doesn't mean it's pointless. For example, SolveSpace is currently rather inconvenient to use with my window manager, i3, because, although it is tiling, there's no good way to tell it to lay out the graphics and text windows as they should be.

virtualritz:
+1

andrew-ayers:
Throwing my 2 cents in - worthless, I guess - I just installed and tried out SolveSpace last night, and I really like it. I went thru the first tutorial, and it was fairly intuitive and easy to use. I didn't find having two windows an issue, as I am a fan of that style (gimp, etc); in fact, I would love if the tool window inside the view was also separate. 

But my main gripe - unless there's a way I haven't found (I was more interested in how to use the software than other things - so I only spent a very brief time looking for a solution) - was the fact that the tool icons and other text was so darn small on my 4k monitor. The menus and such were all fine, of course, because they are controlled by the system. But the icons, tooltips, etc - it's really tiny. Not impossible to use, but it would be nice if it could be changed.

I know that isn't a part of this particular issue - but I think it is something that should be considered for the future, if it hasn't already (I haven't explored the issue space for future changes or whatnot - maybe it already exists).

whitequark:
> I just installed and tried out SolveSpace last night, and I really like it.

Which version? 2.x does not have HiDPI support, the master branch does, and should scale the icons appropriately for the rest.

andrew-ayers:
[...]

PR #1101, opened 2021-08-28
mac: Remove spurious view menu items

tomsci:
Which are either not applicable for SolveSpace (the tabs ones) or are
already handled in the platform-independent code (the fullscreen item).

@vespakoen would you mind looking over this? I assume support for tabbed windows is unlikely to be needed any time soon :-)

(given `SS` and `SK` are globals thus preventing you ever having multiple sketches loaded in a single process, I actually started thinking about how to support opening multiple windows in the same process, and quickly gave up...)

### DIFF
--- a/src/platform/guimac.mm
+++ b/src/platform/guimac.mm
@@ -1621,6 +1621,14 @@ std::vector<std::string> InitGui(int argc, char **argv) {
     ssDelegate = [[SSApplicationDelegate alloc] init];
     NSApplication.sharedApplication.delegate = ssDelegate;
 
+    // Setting this prevents "Show Tab Bar" and "Show All Tabs" items from being
+    // automagically added to the View menu
+    NSWindow.allowsAutomaticWindowTabbing = NO;
+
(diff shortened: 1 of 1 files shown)


ruevs:
Just for reference https://github.com/solvespace/solvespace/issues/39 this will likely not go anywhere, since no one seems to want it.
As for multiple tabs/MDI interface - what is the point ? One can open multiple instances.

tomsci:
Multiple instances don't really work well on mac. You get multiple dock icons, you have to do it from the command line, document opens don't necessarily go where you expect when you do, Open Recent items aren't updated in sync, the system doesn't really intend for apps to work like that. Also you can't copy and paste between different different projects (although that's a whole separate problem).

I wasn't suggesting MDI or tabs as a serious option :-) This PR is just about deleting some menu items that don't do anything useful, but macOS adds automatically. Although I think multiple window support would have merit for the reasons above, if not for the fact that the global data thing would make it really difficult to implement.

vespakoen:
LGTM

PR #1102, opened 2021-08-29
move perspective, lighting, and explode distance from configuration s…

### LINKED ISSUE #1081: Move light direction and perspective factor to view tab
### System information

- **SolveSpace version:** `3.0~` https://github.com/solvespace/solvespace/commit/1b8e1dec652eef7916a94b6c9db404457dacb0db
- **Operating system:** Debian 10.x buster

### Expected behavior

`light direction` and `perspective factor` options should be on `view` tab in "Property Browser", as they directly related to view properties in main window (aka *scene, work area*).

Additionally rename light sources `#0` and `#1` to `#1` and `#2` respectfully - no need to confuse users with numbering started with "Zero" (same for `user colors` numbering, BUT for group numbering `g001-#references` should be renamed to  `g000-#references`)

### Actual behavior

Actually `light direction` and `perspective factor` options are placed on `configuration` tab in "Property Browser".

Also light sources (and user colors) numbering started from "Zero" (`#0` for 1st, and `#1` for 2nd) confuses users.

phkahler:
…creen to view screen.

Closes #1081

### DIFF
--- a/src/confscreen.cpp
+++ b/src/confscreen.cpp
@@ -9,24 +9,6 @@
 #include <omp.h>
 #endif
 
-void TextWindow::ScreenChangeLightDirection(int link, uint32_t v) {
-    SS.TW.ShowEditControl(8, ssprintf("%.2f, %.2f, %.2f", CO(SS.lightDir[v])));
-    SS.TW.edit.meaning = Edit::LIGHT_DIRECTION;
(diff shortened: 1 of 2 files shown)


ruevs:
Meeeeh OK... Same difference :-)

PR #1107, opened 2021-09-05
mac: Don't interpret single-touch scroll events as pan gestures

### LINKED ISSUE #1106: Mac Magic Mouse scroll gestures are misinterpreted
The mac Magic Mouse has a touch-sensitive surface instead of scrollwheel, and its 'scroll' gesture is misinterpreted as a trackpad pan by the logic I added in #1093.

_Originally posted by @nanoant in https://github.com/solvespace/solvespace/issues/1093#issuecomment-913132466_

tomsci:
To handle the Magic Mouse, whose single-finger touch gestures should be interpreted as scrollwheel events rather than pan gestures.

fixes #1106

@vespakoen does this look ok to you? Having to track the current number of active touches as an independent bit of state is slightly messy, but it does seem to do the trick without breaking actual trackpad support. Providing someone doesn't try zooming with a Magic Mouse single-finger gesture at the same time as panning with two fingers on a trackpad, I suppose...

### DIFF
--- a/src/platform/guimac.mm
+++ b/src/platform/guimac.mm
@@ -372,6 +372,7 @@ MenuBarRef GetOrCreateMainMenu(bool *unique) {
     double             rotationGestureCurrent;
     Point2d            trackpadPositionShift;
     bool               inTrackpadScrollGesture;
+    int                numTouches;
     Platform::Window::Kind kind;
 }
 
@@ -397,6 +398,8 @@ MenuBarRef GetOrCreateMainMenu(bool *unique) {
         editor.action = @selector(didEdit:);
 
         inTrackpadScrollGesture = false;
+        numTouches = 0;
+        self.acceptsTouchEvents = YES;
         kind = aKind;
         if(kind == Platform::Window::Kind::TOPLEVEL) {
             NSGestureRecognizer *mag = [[NSMagnificationGestureRecognizer alloc] initWithTarget:self
@@ -573,7 +576,9 @@ MenuBarRef GetOrCreateMainMenu(bool *unique) {
     using Platform::MouseEvent;
 
     MouseEvent event = [self convertMouseEvent:nsEvent];
-    if(nsEvent.subtype == NSEventSubtypeTabletPoint && kind == Platform::Window::Kind::TOPLEVEL) {
+    // Check for number of touches to exclude single-finger scrolling on Magic Mouse
+    bool isTrackpadEvent = numTouches >= 2 && nsEvent.subtype == NSEventSubtypeTabletPoint;
+    if(isTrackpadEvent && kind == Platform::Window::Kind::TOPLEVEL) {
         // This is how Cocoa represents 2 finger trackpad drag gestures, rather than going via
         // NSPanGestureRecognizer which is how you might expect this to work... We complicate this
         // further by also handling shift-two-finger-drag to mean rotate. Fortunately we're using
@@ -624,6 +629,23 @@ MenuBarRef GetOrCreateMainMenu(bool *unique) {
     event.scrollDelta = [nsEvent scrollingDeltaY] / (isPrecise ? 50 : 5);
 
     receiver->onMouseEvent(event);
+}
+
+- (void)touchesBeganWithEvent:(NSEvent *)event {
+    numTouches = [event touchesMatchingPhase:NSTouchPhaseTouching inView:self].count;
+    [super touchesBeganWithEvent:event];
+}
+- (void)touchesMovedWithEvent:(NSEvent *)event {
+    numTouches = [event touchesMatchingPhase:NSTouchPhaseTouching inView:self].count;
+    [super touchesMovedWithEvent:event];
+}
+- (void)touchesEndedWithEvent:(NSEvent *)event {
+    numTouches = [event touchesMatchingPhase:NSTouchPhaseTouching inView:self].count;
+    [super touchesEndedWithEvent:event];
+}
+- (void)touchesCancelledWithEvent:(NSEvent *)event {
+    numTouches = 0;
+    [super touchesCancelledWithEvent:event];
 }
 
 - (void)mouseExited:(NSEvent *)nsEvent {


phkahler:
Going to merge this. Let's get some mac testing on it.

vespakoen:
@tomsci This PR "broke" trackpad scrolling a bit.
When scrolling fast, there are a bunch of scroll events where numTouches are `0`, because the scrolling continues after lifting the fingers from the trackpad.
So I get a pan + zoom effect happening.

I tried all kinds of things to differentiate between the Magic Mouse and "normal" Trackpads, to no avail.

I think we might have the most luck by using the `NSPanGestureRecognizer`, I know this will complicate the `shift-two-finger-drag` rotating, but it is probably the only way we can make Magic Mouse AND Trackpad users happy?

Without this PR, Trackpad works perfect, Magic Mouse does not.
And with this PR, Magic Mouse works perfect but the Trackpad doesn't.

Have you tried implementing the `NSPanGestureRecognizer` and run into troubles?

I just attempted it over here and got it kind of working accept for the the offset being wrong, hoping you can help out =)

tomsci:
@vespakoen I'm not sure I understand what you are seeing? I've just rebuilt from master, and if I do a two-finger drag on the trackpad I get the expected inertial scrolling that carries on for a bit after I stop touching the trackpad. I haven't seen anything that incorrectly triggers a zoom. What do I need to do to reproduce the problem you're seeing?

vespakoen:
Ah interesting...
I get the weird behaviour on the built in trackpad on the mbp m1 pro.
I didn't try it out on the external trackpad yet, maybe they behave differently?

It happens when panning fast and lifting the fingers off while panning, more like "flicking"?

I'll send a screen recording of the behaviour later.

tomsci:
I'm using an 2015 intel MBP built-in trackpad on 11.6 Big Sur, and flicking seems to work ok for me, but the hardware is probably completely different to what's in a modern M1 mac. Or Monterey has changed how the APIs work, that's always a possibility too. I did notice you get weird interactions if you have a mouse connected and move that while the inertial scrolling is slowing down, the mouse movement affects the inertial scrolling when it shouldn't, but I'm not sure how one would fix that. I don't seem to be able to reproduce any weirdnesses using just the trackpad, even when doing lots of quick flicks in quick succession, cancelling the flicks by touching the trackpad again, etc.

Regarding using a pan gesture recognizer, I did consider it and decided against, although it's long enough ago I can't remember the exact reason. Possibly I couldn't get shift-two-finger-drag to work with it either. The gesture recognizers are great when they do exactly what you want, but if they don't they're even harder to customise that just doing it yourself...

vespakoen:
Hope this makes sense:

https://user-images.githubusercontent.com/876117/151660827-d3c591b8-b212-47b9-87ed-81ecb9175e1e.mov

vespakoen:
I added some logging to the `touched*WithEvent` functions, and here is what I get:

```
2022-01-29 13:57:31.096 SolveSpace[74037:14383128] Began
2022-01-29 13:57:31.184 SolveSpace[74037:14383128] Moved
2022-01-29 13:57:31.192 SolveSpace[74037:14383128] Moved
2022-01-29 13:57:31.201 SolveSpace[74037:14383128] 2
2022-01-29 13:57:31.201 SolveSpace[74037:14383128] Moved
2022-01-29 13:57:31.210 SolveSpace[74037:14383128] 2
2022-01-29 13:57:31.210 SolveSpace[74037:14383128] Moved
2022-01-29 13:57:31.215 SolveSpace[74037:14383128] 2
2022-01-29 13:57:31.215 SolveSpace[74037:14383128] Moved
2022-01-29 13:57:31.218 SolveSpace[74037:14383128] 2
2022-01-29 13:57:31.220 SolveSpace[74037:14383128] Moved
2022-01-29 13:57:31.224 SolveSpace[74037:14383128] 2
2022-01-29 13:57:31.226 SolveSpace[74037:14383128] Moved
2022-01-29 13:57:31.226 SolveSpace[74037:14383128] 2
2022-01-29 13:57:31.227 SolveSpace[74037:14383128] Moved
2022-01-29 13:57:31.234 SolveSpace[74037:14383128] 2
2022-01-29 13:57:31.234 SolveSpace[74037:14383128] Moved
2022-01-29 13:57:31.235 SolveSpace[74037:14383128] 2
2022-01-29 13:57:31.235 SolveSpace[74037:14383128] Moved
2022-01-29 13:57:31.243 SolveSpace[74037:14383128] 2
2022-01-29 13:57:31.243 SolveSpace[74037:14383128] Moved
2022-01-29 13:57:31.244 SolveSpace[74037:14383128] 2
2022-01-29 13:57:31.244 SolveSpace[74037:14383128] Moved
2022-01-29 13:57:31.250 SolveSpace[74037:14383128] 2
2022-01-29 13:57:31.250 SolveSpace[74037:14383128] Moved
2022-01-29 13:57:31.252 SolveSpace[74037:14383128] 2
2022-01-29 13:57:31.252 SolveSpace[74037:14383128] Moved
2022-01-29 13:57:31.259 SolveSpace[74037:14383128] 2
2022-01-29 13:57:31.259 SolveSpace[74037:14383128] Moved
2022-01-29 13:57:31.260 SolveSpace[74037:14383128] 2
2022-01-29 13:57:31.260 SolveSpace[74037:14383128] Ended
2022-01-29 13:57:31.274 SolveSpace[74037:14383128] 0
2022-01-29 13:57:31.291 SolveSpace[74037:14383128] 0
2022-01-29 13:57:31.308 SolveSpace[74037:14383128] 0
2022-01-29 13:57:31.324 SolveSpace[74037:14383128] 0
2022-01-29 13:57:31.341 SolveSpace[74037:14383128] 0
2022-01-29 13:57:31.357 SolveSpace[74037:14383128] 0
2022-01-29 13:57:31.374 SolveSpace[74037:14383128] 0
2022-01-29 13:57:31.390 SolveSpace[74037:14383128] 0
2022-01-29 13:57:31.407 SolveSpace[74037:14383128] 0
2022-01-29 13:57:31.424 SolveSpace[74037:14383128] 0
2022-01-29 13:57:31.441 SolveSpace[74037:14383128] 0
2022-01-29 13:57:31.457 SolveSpace[74037:14383128] 0
2022-01-29 13:57:31.474 SolveSpace[74037:14383128] 0
2022-01-29 13:57:31.491 SolveSpace[74037:14383128] 0
2022-01-29 13:57:31.507 SolveSpace[74037:14383128] 0
2022-01-29 13:57:31.524 SolveSpace[74037:14383128] 0
```

So there are scroll events happening after `touchesEndedWithEvent` fired, at least on my mac ;)

vespakoen:
I think I just figured out that the `NSPanGestureRecognizer` only works with a `buttonMask`, with other words, it doesn't work without any buttons pressed, which is what we want to achieve.

This is probably the reason for not using the `NSPanGestureRecognizer`.

I guess the only way we can fix this is by adding a checkbox to the menu to change the behaviour, so it can work correctly with the Magic Mouse and Trackpads.

ruevs:
@vespakoen a platform specific check box would be a first and sounds to me like a pretty bad idea. You should think carefully what it does, what would it be called an would it disappear when not on macOS? Also - his does other software on the Mac handle this? Do they have "a checkbox"

tomsci:
Agreed I don't think we should have to resort to UI toggles for this. I think most other apps don't try and do such specific and advanced things with the mouse/trackpad, and aren't interacting with a cross-platform codebase that assume the mouse event model works in a particular way. Got to say, Apple's APIs for this are less than helpful when they're trying to be "helpful"...

I'll take another look at this and @vespakoen's vid when I get a sec, hopefully there's something we can do that satisfies both usecases. If it really is super broken for the trackpad maybe we should just consider reverting this PR, and go back to having fully-working trackpad and somewhat-broken magic mouse? I'd speculate more people have either trackpads (thanks to being a mac laptop) or are using a standard PC-compatible mouse, compared to the number of people with Magic Mice. (I only have a PC mouse at home thanks to it being half the price, and actually works with PCs...)

nanoant:
> If it really is super broken for the trackpad maybe we should just consider reverting this PR, and go back to having fully-working trackpad and somewhat-broken magic mouse?

So it looks like this happened in #1218 unfortunately without mentioning this discussion and we now somewhat-broken magic mouse as a trade-off for having nice trackpad padding and no system specific settings.

Afaik we should be able to get touch device type from:
~~~objc
- (void)touchesBeganWithEvent:(NSEvent *)event {
    NSTouch *touch = [event touchesMatchingPhase:NSTouchPhaseTouching inView:self].anyObject;
    NSTouchDevice *touchDevice = (NSTouchDevice *)touch.device;
    NSLog(@"touchesBeganWithEvent %p", touchDevice);
    if (touchDevice != nullptr) {
        touchDeviceType = touchDevice.deviceType;
        NSLog(@"touchDeviceType %ld", touchDeviceType);
    }
    [super touchesBeganWithEvent:event];
}
~~~

Interestingly this method does not fire at all on my Magic Mouse. I'll look into that on my MacBook later this week. Maybe we could actually distinguish if the touches come from Magic Mouse or any mouse or a trackpad in the end.

vespakoen:
From the top of my head:
I also had difficulty = was not able to differentiate between trackpad and magic mouse.
It works nicely now with the trackpad, and there is room for improvement for the Magic Mouse, but I am not sure yet how we can implement it.
I will take a look at this when I am returned from my holiday in a couple of weeks.

phkahler:
IIRC my old MBA trackpad allows 2-finger dragging to rotate and 1 finger to pan on Solvespace version 2.3 but the 2-finger doesn't work on 3.0+  It really was nice to not have to use modifier keys.

nanoant:
I was able to unbreak Magic Mouse with #1274. Tested on Magic Mouse and Mac mini running Monterey, and MBP 2013 trackpad running Big Sur.

@phkahler I am bit confused how can you pan with 1 finger? Isn't 1 finger used to move the mouse cursor over the view?

Issue #955, opened 2021-03-07
Hole Tool - feature


phkahler:
## Proposed operation (3D groups):
A new sketch tool would be added. Once selected and the mouse hovers over a plane face, that face is highlighted as usual. When left mouse click, a hole entity is created in the 3D sketch group AND the solid model is modified. Multiple holes could be created in a single group.

The sketch entity might be a circle with cross-hairs and a center point. The point would be constrained to the face clicked on and the normal for that face would be used to orient the entity and hole. The point could then be dragged or constrained as usual.

## implementation
One key issue is how to define a hole. Using 2 diameters, a depth (of the counterbore), and a checkbox (counterbore, countersink, normal) would cover the basics. Some people prefer to specify a countersink by the angle as on the [wikipedia page](https://en.wikipedia.org/wiki/Countersink). I assume the definition will be in the text window. It's not clear to me how to define the final depth - should it be parametric (per hole) like a normal extruded hole, or defined in the text window? What about non-through holes?

If the depth is draggable/constrainable, an additional line or circle may be needed but will these will all need to be part of the single hole entity. Also, how would that work with the 2D option below?

Once a hole has been created, its definition should be used for subsequent holes until one is edited. It might also be possible to copy/paste a hole entity, which would bring along its parameters but take its normal from the new location.

## 2D option
People will probably want to use the new sketch entity for 2D drawings. and of course they will want those holes to be formed if they extrude the sketch. The 3D option is more universal but not as nice for simple designs mostly covered by a 2D drawing, so this seems important. Biggest issue is that the 3D NURBS behavior would be quite different between the two. Hole would be created from previous 2D sketch entities AND current 3D group entites, both after the solid model is created. Maybe that's not so bad.

### Hole styles?
Would it be feasible to allow the user to create named hole definitions, similar to line styles? Would the hole style be shown on the sketch?

### Key questions:
1) what is the best way to define a hole profile?
2) how to handle the final depth of the hole?
3) are there any other important types of holes to handle?
4) is this too much violation of the SolveSpace group structure?

ruevs:
This is a "weird" idea in context of the "ideology" of SolveSpace so far. Making "holes" with their "machining" features is a common and useful task, but so far all tools have been pure geometry. If we start down this path it leads to treads, bolts, nuts, axles, gears...

jwesthues:
[...]

PR #1132, opened 2021-10-13
+ Safe height parameter for the GCODE files export functionality

MX-Master:
Allow user to setup own safe height (retract height) value. Instead of using hardcoded (+/- 5mm) one.

### DIFF
--- a/src/confscreen.cpp
+++ b/src/confscreen.cpp
@@ -497,6 +497,11 @@ bool TextWindow::EditControlDoneForConfiguration(const std::string &s) {
             if(e) SS.gCode.depth = (float)SS.ExprToMm(e);
             break;
         }
+        case Edit::G_CODE_SAFE_HEIGHT: {
+            Expr *e = Expr::From(s, /*popUpError=*/true);
+            if(e) SS.gCode.safeHeight = (float)SS.ExprToMm(e);
+            break;
+        }
         case Edit::G_CODE_PASSES: {
             Expr *e = Expr::From(s, /*popUpError=*/true);
             if(e) SS.gCode.passes = (int)(e->Eval());


MX-Master:
> @MX-Master I don't think so? We would need a multi-line text input box, or a dialog with one, and it needs to work on all 3 OSes. I dont think we've got that. 

Got it. So a single line solution is the easiest way to do this now. I'll try to do some tests soon.

phkahler:
> > 1. In ui.h  G_CODE_SAFE_HEIGHT  is out of order in the enum.
> > 2. Can you squash into a single commit?
> 
> Got it. Will be done soon :)
> 
> One more question. Is it possible to add a multiline text field to the gcode exporting settings? Or single line text field with correct /r/n interpreting? I want to add two text fields - startupGcode and endupGcode.

@MX-Master  What about adding a G-code template file that has starting and ending G-code with some kind of marker to indicate where solvespace should put the sketch related g-code. The text window/settings would then store the path/filename last used. This would also make it easy to change the setup for whatever reason (different CNC etc). It also doesn't require a multi-line text input box ;-)

ruevs:
Isn't this superfluous? Anyone capable of preparing such a template and configuring it should be more than capable of copy-pasting the start/end G-code in the exported files? But maybe as a type of "automation" - when always exporting the files in the same way for a particular machine?

phkahler:
@ruevs  Yes, it's equivalent to copy/paste. But so is storing pre- and post- G-code into a config dialog.  It doesn't really belong in the solvespace config because it's machine (or material) specific. You know slicers keep separate configs for machine and material. Why not solvespace?  I don't use if for generating g-code and sadly I don't have a CNC, so I have to throw ideas around to people who do ;-)

MX-Master:
Template file is a good idea 😉

CNC machines - it's really fun. You can easily build a small laser/printer machine by yourself. And when you will taste it, your life will never be the same 😁

At this moment, I'm using SolveSpace for the quick laser engraving.

phkahler:
@MX-Master  It would be great if a template could run on your CNC without modification. That means the "place G-code from solvespace here" should be indicated by some G-code or M-code.  After looking around, semantically M98 would be appropriate, but that may cause problems if you run it prior to substituting output from solvespace. G04 with a magic value would work almost universally, but I'm not big fan of magic values.

Maybe you've got an idea? Maybe ask on a CNC forum? We could also as @jwesthues since he wrote this thing for CNC  :-)

MX-Master:
Popular CNC software supports gcode comments. So

`"place G-code from solvespace here"`

can be a

`;(gcode)`

LEGEND:
"; comment" - uses by 3D printers
"(comment)" OR "; comment" - uses by the other popular CNCs (LinuxCNC, Mach3, etc.)

jwesthues:
The safe height change is definitely good, annoying for that to be hard-coded. For additional changes, SolveSpace will presumably never be a full CAM tool; but if there's stuff we can do to make more basic CAM tasks possible from SolveSpace alone, then I think there's value there, especially given the limited state of free CAM software. 

As to the template thing specifically, I see many places where that would be useful. Ideally we'd find a way to add value beyond what's already relatively easy with `cat template-before from-solvespace template-after > final`, though I can't immediately think what myself.

phkahler:
@MX-Master  can you open an issue for the g-code template file?

I agree with what @jwesthues said, but I think there is a lot of room for incremental additions to solvespace CAM capabilities. I'd say supporting G02 and G03 arcs, and then if we add feature #955 (hole entities) you might want to add tool changing and fancy hole drilling. If someone is ambitious, maybe even more. I don't really see it going beyond what a 3-axis can do though.

MX-Master:
> @MX-Master can you open an issue for the g-code template file?

I'll try to make this functionality by myself, no need to add a new issue ;)

PR #1133, opened 2021-10-13
Add safe height gcode parameter (as single commit)

MX-Master:
Allow user to setup own safe height (retract height) value. Instead of using hardcoded (+/- 5mm) one.

1132 as single commit.


MX-Master:
Is there any problems to commit this simple update to the main branch? :)

PR #1135, opened 2021-10-22
modify the spaceware code to also work on wayland

xzcvczx:
use the spacenav socket only on wayland, as the "official 3dconnexion" linux sdk is no longer maintained it is unlikely to ever be useable under wayland anyway. 

use the recommended compile-time and run-time checks for x11 and wayland based on
https://github.com/GNOME/gtk/blob/b539c92312d449f41710e6930aaf086454d667d2/docs/reference/gdk/wayland.md

### DIFF
--- a/src/platform/guigtk.cpp
+++ b/src/platform/guigtk.cpp
@@ -33,6 +33,7 @@
 
 #if defined(HAVE_SPACEWARE)
 #   include <spnav.h>
+#   include <gdk/gdk.h>
 #   if defined(GDK_WINDOWING_X11)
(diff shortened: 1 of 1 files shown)


phkahler:
@xzcvczx  Do we actually define both of these in our builds?
GDK_WINDOWING_X11
GDK_WINDOWING_WAYLAND

Thanks for looking into this!

xzcvczx:
they are defined in gdk/gdkconfig.h (included by gdk/gdk.h) based on what gtk was built with

PR #1140, opened 2021-11-04
GUI: Flexible vertical space above the toolbar

### LINKED ISSUE #1130: Do not reserve space above toolbar in minumum window size
### System information

- **Operating system:** Debian 11 bullseye (Fluxbox)

### Expected behavior

There should be no extra `32px` space in height reserved in minimum window size.

To allow record full window (with title & toolbar ON) at least in 720p video standard (1280x720).

### Actual behavior

Actually impossible to record full window at least in 720p even with minimalistic window decoration, due to minimum height of full window (with tittle &, toolbar ON) out of 720 px in height:

ruevs:
If the main window is not high enough allow the default 32 pixel padding
between the menu bar and the toolbar to shrink down to zero.

This allows the main window height to be a minimum of 688 pixels (on
Windows 10) so it is possible to capture 720P video tutorials.

Fixes #1130

image

### DIFF
--- a/src/graphicswin.cpp
+++ b/src/graphicswin.cpp
@@ -430,7 +430,7 @@ void GraphicsWindow::Init() {
             using namespace std::placeholders;
             // Do this first, so that if it causes an onRender event we don't try to paint without
             // a canvas.
-            window->SetMinContentSize(720, 670);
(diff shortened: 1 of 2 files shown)


phkahler:
Thank you @ruevs!

Symbian9:
Thank you all!

PR #1141, opened 2021-11-05
Add Reddit community, Replace Freenode, Fix HTTPS

Symbian9:
NOTE: As Freenode IRC already deprecated buy most of the FLOSS projects, no need to confuse users with it.

Instead, replace links to active IRC channel on LiberaChat.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,9 @@ The official SolveSpace [website][sswebsite] has [tutorials][sstutorial],
 [reference manual][ssref] and a [forum][ssforum]; there is also an official
 IRC channel [#solvespace at irc.libera.chat][ssirc].
 
-There is also [r/SolveSpace][ssreddit] — a fan driven users community on
-Reddit, which is a good place to share your showcases, tutorials and ask
-for help with a general use of SolveSpace app for creating anything.
+Also check [r/SolveSpace][ssreddit] — a fan driven users community on
+Reddit, which is used for sharing showcases, tutorials and ask for help
+with a general use of SolveSpace app.
 
 [sswebsite]: https://solvespace.com/
 [ssref]: https://solvespace.com/ref.pl
--- a/README.md
+++ b/README.md
@@ -9,20 +9,23 @@ SolveSpace
 This repository contains the source code of [SolveSpace][], a parametric
 2d/3d CAD.
 
-[solvespace]: http://solvespace.com
+[solvespace]: https://solvespace.com
 
 Community
 ---------
 
 The official SolveSpace [website][sswebsite] has [tutorials][sstutorial],
-[reference manual][ssref] and a [forum][ssforum]; there is also an official
-IRC channel [#solvespace at irc.freenode.net][ssirc].
+[reference manual][ssref] and a [forum][ssforum].
 
-[sswebsite]: http://solvespace.com/
-[ssref]: http://solvespace.com/ref.pl
-[sstutorial]: http://solvespace.com/tutorial.pl
-[ssforum]: http://solvespace.com/forum.pl
-[ssirc]: https://webchat.freenode.net/?channels=solvespace
+There is also [r/SolveSpace][ssreddit] — a fan driven users community on
+Reddit, which is a good place to share your showcases, tutorials and ask
+for help with a general use of SolveSpace app for creating anything.
+
+[sswebsite]: https://solvespace.com/
+[ssref]: https://solvespace.com/ref.pl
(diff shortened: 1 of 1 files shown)


ruevs:
Google exists. The Reddit link is redundant.
I don't use IRC so someone else will have to comment on the change of the IRC server.

Symbian9:
> I don't use IRC so someone else will have to comment on the change of the IRC server.

To be clear, `#solvespace` on LiberaChat created at 5/19/2021 and there is @rpavlik already

- https://web.libera.chat/#solvespace

ppd:
This kind of "fluff" content could/should also be placed in a wiki. GitHub offers such an option.

Symbian9:
> also be placed in a wiki. GitHub offers such an option.

I may agree, but there are no GitHub Wiki enabled for SolveSpace repo yet.

@phkahler, may you enable GitHub Wiki feature for actual repo?

Symbian9:
@ruevs, please, fix at least HTTPS links and replace Freenode IRC with LiberaChat.

ruevs:
Complete silence rules in both IRC channels. I'll leave them open for a while to see if anything happens.

The https I'll can when there is something more to change in the readme. It's not worth adding a commit for one letter when the http URL gets redirected anyway.

bmorel:
Since un-identified person on libera.net/#solvespace said something along the lines of "is this active? Is it worth pointing at #issue"... let me say that, even if, yes, activity is sporadic, it is still active.
There's a small group of users idling around, sure (around 7), but when someone comes by to ask and wait around long enough, they get an answer.
Yes, waiting its required, because everyone can't be present 24/24, and sure, 7 persons are not much. Still, that's very different from a "complete silence".

There was actitivity today, ofc, but before that, a discussion started on 1st december, which ended the 3rd  (yes, discussion can span on several days). Even before, 30 nov to 30 nov. Before, on 26 nov. Before again, on 25 nov.
Is that a complete silence?

ghane83:
I agree with @bmorel, there are sporadic discussions about Solvespace in libera.chat/#solvespace even if it doesn't happen every day. It is not too uncommon that someone with a problem joins and asks for help or want to discuss bugs.

Btw, I am the user that registered the channel on libera.chat just to make the topic (copied over from Freenode and updated according to latest release) stay in the initial chaos of netsplits when "everyone" was switching IRC network. If the project desires to register the namespace and use it as a official project channel and/or have ownersip transfered to actual project developers or such, then let's do that. :)

ruevs:
Cool. The unidentified person was me. So I will change the IRC URL to Libera Chat.

By the way what happened at Freenode? I haven't used IRC since the nineties so I'm completely out of the loop. The #solvespace channel is still there with no one inside.

Is there an archive of the Libera channel?

ghane83:
@ruevs you had already left when we noticed you. ;)

Long story short: New management led to most (all?) of the old staff resigning from Freenode and starting up Libera Chat. Subsequent event lead to most projects (that I know of) moving to the Libera or  OFTC networks.

I do not remember if the old #solvespace@freenode even was registered (do not think so), the topic stated that the latest release was 2.1 or something.

No archiving of the channel set up as far as I know except for possibly some local logs... ;)

ruevs:
@whitequark kept a log of the Freenode channel here https://freenode.irclog.whitequark.org/solvespace/ but I'm not sure if she will be willing to redirect it to Libera?

Symbian9:
> but I'm not sure if she will be willing to redirect it to Libera?

`@whitequark` asked to not cast anymore:

- https://twitter.com/whitequark/status/1343664895506702337
> I am no longer involved with the SolveSpace project

rpavlik:
She logs some other libera channels now (e.g. for glasgow)

Symbian9:
> She logs some other libera channels now (e.g. for glasgow)

@rpavlik, you missed my previous comment, but TL;DR: she probably would not be interested to log `#solvespace` channels anymore.

ruevs:
IRC link and https changes in https://github.com/solvespace/solvespace/commit/2450010bbfa20a1ffbf5fd90b85c4b941f56cb16

ruevs:
@ghane83 @bmorel
So Whitequark has kindly set up IRC logging at https://libera.irclog.whitequark.org/solvespace.
She notes that "Libera guidelines require channel topic to state that
it is logged."
Would you please change the topic and add a link to the logs.

I posted a similar message in the channel and it immediately showed up in the log.

ghane83:
Nice! I've added this to the entrymsg and topic of the channel.

PR #1145, opened 2021-11-20
fix STL linking issue. Model was disappearing after the link group.

phkahler:
This issue was mentioned on the forum. Linked STL groups don't have a NURBS shell, so they need to be forced to triangle mesh. The original STL code forced the previous group by mistake, and the fix for that didn't quite work. This seems to fix it.

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -909,6 +909,8 @@ try_again:
                     return false;
                 }
             }
+            if(g.IsTriangleMeshAssembly())
+                g.forceToMesh = true;
         } else if(linkMap.count(g.linkFile) == 0) {
             dbp("Missing file for group: %s", g.name.c_str());
             // The file was moved; prompt the user for its new location.


phkahler:
@tomsci   the issue was reported here:

https://solvespace.com/forum.pl?action=viewthread&parent=3973&tt=1636468199

I'm wondering if you think any other changes or deletions should go along with this.

tomsci:
I think a neater way of fixing it might be to change this line:

```c++
return g->forceToMesh || g->IsForcedToMeshBySource();
```

in `Group::IsForcedToMeshBySource()` to:

```c++
return g->IsForcedToMesh();
```

That way IsForcedToMeshBySource will pick up the fact that the source group is forced to mesh for whatever reason, without imports having to set `forceToMesh`, because everywhere else that's used to mean "the user has explicitly elected to force this specific group to mesh" which is why we have `Group::IsForcedToMesh()` which is "this group is forced to mesh for whatever reason".

Symbian9:
> I think a neater way of fixing it might be to change this line:

Nice, @tomsci!

PR #1151, opened 2021-12-16
Fix hang when trying to display characters missing from the embedded font

### LINKED ISSUE #1150: SolveSpace freeze on enter U+EA00 character for GNU Unifont Text
### System information

- **SolveSpace version:** 3.0~https://github.com/solvespace/solvespace/commit/85f6ec414492d07b512fdfc6f97559ec11ebb36c

### Expected behavior

SolveSpace should not be freezed by input any Unicode symbol for TrueType Text.

### Actual behavior

SolveSpace fully freeze after input `` (U+EA00) character into text input field and apply (click <kbd>Enter</kbd>)

1. Launch SolveSpace;
2. Add TrueType Text object (press <kbd>T</kbd>);
3. Select it & input `` into text input field (no matter what font is selected);
4. Press <kbd>Enter</kbd> to apply.

Occasionally I also double clicked on "Property Browser" window title (after SolveSpace frozen) — as a result I got "two" windows of "Property Browser" (all are totally frozen including main window).

Screenshot_2021-12-15_02-34-30

Screenshot_2021-12-15_02-35-15

Screenshot_2021-12-15_02-39-22

### Additional information

FTR, There is Twitter font with the only character symbol is `` (U+EA00).

- https://github.com/ctrlcctrlv/TwitterEA00
  - https://twitter.com/fr_brennan/status/1435780076537274372

BUT, I tested with other fonts (including in-built Bitstream Vera) — SolveSpace always freezes on `` (U+EA00) input, NO matter which font is selected.

cc @ctrlcctrlv

ruevs:
When SolveSpace tries to display an Unicode code point (e.g. U+EA00 ) that does not exist in the embedded vector font (unifont.hex.gz) it hangs in an endless loop in `const BitmapFont::Glyph &BitmapFont::GetGlyph(char32_t codepoint)`.

The reason is that the binary search through the text file unifont-8.0.01.hex that does the "lazy loading" of glyphs does not end.

Here is a short excerpt from the file:

```
D7FE:00007FFE63866DF66DEE6DDE63DE7FFE7FFE61866FBE638E6FBE6F867FFE0000
D7FF:00007FFE63866DF66DEE6DDE63DE7FFE7FFE61866FBE638E6FBE6FBE7FFE0000
F900:0080108810881FF8100800007FFE00000FF008100FF00810042002447FFE0000
F901:00047FFE008010841FFE10841FFC10841FFC10840880050003000CC0703E0000
```

When searching for `0xEA00` after some iterations of the while loop https://github.com/solvespace/solvespace/blob/2450010bbfa20a1ffbf5fd90b85c4b941f56cb16/src/resource.cpp#L567
both `first` and `last` end up pointing to F900. After that on each consecutive iteration of the loop `last` ends up pointing to the LF (0x0A) just before F900 https://github.com/solvespace/solvespace/blob/2450010bbfa20a1ffbf5fd90b85c4b941f56cb16/src/resource.cpp#L585
and then `mid` ends up back on F900 here https://github.com/solvespace/solvespace/blob/2450010bbfa20a1ffbf5fd90b85c4b941f56cb16/src/resource.cpp#L570 and this will repeat enlessly.

The solution is to do
```
            last = mid - 2;
```
here https://github.com/solvespace/solvespace/blob/2450010bbfa20a1ffbf5fd90b85c4b941f56cb16/src/resource.cpp#L585, which will make `first>last` and allow the while loop to exit.

The lazy parsing of the vector front was introduced here.
https://github.com/solvespace/solvespace/commit/645c2d90ace3e48c639c49512fe1d7a43beb1755

Fixes #1150

### DIFF
--- a/src/resource.cpp
+++ b/src/resource.cpp
@@ -564,7 +564,7 @@ const BitmapFont::Glyph &BitmapFont::GetGlyph(char32_t codepoint) {
     // Find the hex representation in the (sorted) Unifont file.
     auto first = unifontData.cbegin(),
          last  = unifontData.cend();
-    while(first <= last) {
+    while(first < last) {
         auto mid = first + (last - first) / 2;
         while(mid != unifontData.cbegin()) {
             if(*mid == '\n') {
@@ -588,7 +588,10 @@ const BitmapFont::Glyph &BitmapFont::GetGlyph(char32_t codepoint) {
         if(foundCodepoint < codepoint) {
             first = mid + 1;
             while(first != unifontData.cend()) {
-                if(*first == '\n') break;
+                if(*first == '\n') {
+                    first++;
+                    break;
+                }
                 first++;
             }
             continue; // and last stays the same


rpavlik:
Is there a good reason why we're not just using a standard library algorithm here? Too hard to bodge in right now? (On my phone ATM)

rpavlik:
Oh interesting. I have now read the full code there. We're using iterators but into the string, so not easy to binary search the normal way.

Do we need a bounds check before subtracting 2 to avoid running off the front? (I don't remember if that font gets compiled in or if somebody could give us a bad data file)

Other than that it looks great, thanks!

ruevs:
(On the phone myself now). The font (https://github.com/solvespace/solvespace/blob/master/res/fonts/unicode.lff.gz) is linked into our binary and we are unzipping it in memory at runtime. Then glyphs are parsed and loaded "lazily" (on demand) by the function I modified. The binary search looks the way it does since the parsing is of the plain text of the font. This approach was first introduced by Whitequark in two consecutive commits here https://github.com/solvespace/solvespace/commit/645c2d90ace3e48c639c49512fe1d7a43beb1755#diff-596d9576b1809607ab9159b8e03a5049bc880fb49029317d0425da41cf07dde9 https://github.com/solvespace/solvespace/commit/fc796427887e66e8e6667477c581530b9a251f6e#diff-e99cc430914390ed01b41b9eeefd419ec4c099e6b15d4fd5b29612e21cb8cfa0 . Before that there were lff2c and unifont2c tools that parsed the fonts into C arrays of structures and the binary search was less "crazy" but still not using a standard algorithm :-)

ruevs:
> Do we need a bounds check before subtracting 2 to avoid running off the front?

I think not, but I will debug it when I'm back on a PC to be absolutely sure.

phkahler:
Is this code actually doing a binary search over the characters of the decompressed .hex file? That's quite different than doing it over the *lines* of the file, so it's hard to follow the logic.

Since the font never changes (unless we update it) it should suffice to test it for failure on 0000 and FFFF.

unifont is weird in abusing .hex format, and by mixing 8-wide and 16-wide character. It seems like that also needs to be considered in the binary search code.

ruevs:
It is doing a binary search over the lines (the 16 bit code point values for each glyph, ornn other words the first 4 characters on each line) of the decompressed file. However the file is stored in a `std::string` and `fitst`, `last` and `mid` are iterators on it. That is why the code is a bit complicated.

phkahler:
My concern is that because it's doing binary search over a string, lines like this:
```
mid = first + (last - first) / 2;
```
don't cause mid to point at the beginning of a line. Some lines are longer than others depending on the width of the character, so in terms of line count, mid could be within the same line as first, or many other lines between there and last.

ruevs:
Of course. That is why thisese are there:
https://github.com/solvespace/solvespace/blob/2450010bbfa20a1ffbf5fd90b85c4b941f56cb16/src/resource.cpp#L569

https://github.com/solvespace/solvespace/blob/2450010bbfa20a1ffbf5fd90b85c4b941f56cb16/src/resource.cpp#L590

ruevs:
OK I debugged and tested with:
- 0xEA00 - non existent, not found in 16 iterations.
- 0xF900 - exists but takes exactly 16 iterations of the binary search to finish
- 0x0000 - found in 16 iterations
- 0xFFFD - the replacement Unicode code point for non-existing glyphs. Also the end of the font.
- 0xFFFFF - a codepoint beyond the end, not found and does not cause an exception

And as a result did the fix in a different way, And this is more "correct". The previous one worked but it was a bit "hacky".

phkahler:
When I first looked at the code I thought (first <= last) should not include the equal condition but I couldn't say why. It was just a feeling based on what the code does and the existence of first, mid, last.

PR #1157, opened 2021-12-22
Normalize namespaces and fix uninitialized

rpavlik:
The namespace thing should make editors and IDEs like the code more, reducing maintenance burden. The other commit fixes a bug identified by a warning on my machine.

### DIFF
--- a/src/importidf.cpp
+++ b/src/importidf.cpp
@@ -332,8 +332,9 @@ bool LinkIDF(const Platform::Path &filename, EntityList *el, SMesh *m, SShell *s
     
     double board_thickness = 10.0;
     double scale = 1.0; //mm
-    bool topEntities, bottomEntities;
-    
+    bool topEntities       = false;
+    bool bottomEntities    = false;
+
     Quaternion normal = Quaternion::From(Vector::From(1,0,0), Vector::From(0,1,0));
     hEntity hnorm = newNormal(el, &entityCount, normal);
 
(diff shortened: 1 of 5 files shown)


ruevs:
Can you elaborate more on how the namespaces hep IDEs?

The initialized members are of course fine.

rpavlik:
So, no includes were added to the main header, they were just moved. The added include lines in the other headers are ones that "should have been there" all along. (see for example https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-contained https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-include-order )  Previously, headers weren't safe to be loaded standalone: they were assumed to be included within `namespace SolveSpace { }` in `solvespace.h`. However, IDEs, tooling, etc don't necessarily realize that, and that's not a common pattern either. So, e.g. you'd get lots of noise like all these "... is ambiguous" in your editor: 

image

image

That's just a sample from VS Code, other editors show it other ways, but basically including something inside a namespace in another header is so unconventional it doesn't work very well. (Actually, when I'm making a library, I will often add tests to make sure that each header will compile in its own translation unit, meaning it is complete with its dependencies, etc.)

Additionally, this should let us cut build time by reducing coupling and reducing extra includes where possible. We'll be able to move more toward an "include-what-you-use" model, so not every header change will necessarily affect solvespace.h triggering a full rebuild of everything, etc.  Basically, making the codebase more "conventional" and contributor-friendly.

phkahler:
I like your intent, and I trust your understanding of how things *should* be arranged in this case better than my own, and all 3 platforms passed so merging.

PR #1158, opened 2021-12-23
Solver changes without eigen

rpavlik:
This is just the solver changes from #721, skipping the final commits that add Eigen, and fixed so that every commit successfully builds. I do not fully understand the changes, I just can rebase them sensibly. @Evil-Spirit is the real author here.

I'll put the Eigen stuff in a separate PR.

This does run.

With sanitizers enabled, looks like no new  leaks thru here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,6 @@ endif()
 list(APPEND CMAKE_MODULE_PATH
     "${CMAKE_SOURCE_DIR}/cmake/")
 
-cmake_policy(SET CMP0048 OLD)
 set(CMAKE_CXX_STANDARD 11)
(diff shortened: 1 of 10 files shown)


phkahler (commented):
I can't speak to the more complex changes, but nothing "looks wrong".

rpavlik:
With this plus the "no sparse expr" pr, I can actually say I understand the Jacobian code moderately ok now, at least the part about computing and evaluating it. Haven't looked at the other methods in the file enough to learn those. For a fun time, look at branches I pushed to my fork recently named things like "another way to cause memory corruption" 😅. I should probably script running over the test case repo with sanitizers on to make sure it works on more than just the normal test suite and the 2 models I tested on .

rpavlik:
OK this has been rebased.

PR #1155, opened 2021-12-22
Build simplify

rpavlik:
Builds on #1154 

Some cleanups of the cmake files so they're easier to understand and edit. Also fixes a bunch of CMake warnings.

PR #1159, opened 2021-12-23
Evilspirit eigen over ryan buildsys

rpavlik:
This is the Eigen-related stuff from #721, but put over my Eigen work excluding my work in the system.cpp file (so most but not all of #432 ).  Builds on #1158 and #1155 

I also split out the "too many unknowns" removal and put it at the end, in case we want to keep a now-artificial limit.

No new leaks.

### DIFF
--- a/.github/scripts/install-macos.sh
+++ b/.github/scripts/install-macos.sh
@@ -13,4 +13,4 @@ else
     brew install libomp
 fi
 
-git submodule update --init extlib/cairo extlib/freetype extlib/libdxfrw extlib/libpng extlib/mimalloc extlib/pixman extlib/zlib
+git submodule update --init extlib/cairo extlib/freetype extlib/libdxfrw extlib/libpng extlib/mimalloc extlib/pixman extlib/zlib extlib/eigen
--- a/.github/scripts/install-ubuntu.sh
+++ b/.github/scripts/install-ubuntu.sh
@@ -5,6 +5,6 @@ sudo apt-get update -qq
 sudo apt-get install -q -y \
   zlib1g-dev libpng-dev libcairo2-dev libfreetype6-dev libjson-c-dev \
   libfontconfig1-dev libgtkmm-3.0-dev libpangomm-1.4-dev libgl-dev \
-  libgl-dev libglu-dev libspnav-dev 
+  libgl-dev libglu-dev libspnav-dev
 
-git submodule update --init extlib/libdxfrw extlib/mimalloc
+git submodule update --init extlib/libdxfrw extlib/mimalloc extlib/eigen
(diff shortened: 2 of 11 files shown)


rpavlik:
Hmm, something in there is making things slower, not sure what. No time to test today, already spent too much time on this :) I did manage to get a speed bump with #1160 (test suite from 5.0s to 4.9s) but still getting slower on my test model, not sure why.

phkahler:
You've included the commits from #1158 in this. Can you remove those?

And thank you for doing all this!

rpavlik:
Sure, guess I can split out a few of mine from here. The idea was that the build system in the original pr wasn't great, and all but one of my eigen changes were little "peephole" optimizations based on hotspots i saw when profiling 4 years ago 😅 we'll see when I get a chance, I did way more Solvespace work the past few days than I was supposed to

rpavlik:
Oh, ok, so I got the PR's mixed up in my head. The changes in here are necessarily dependent on at least a few commits from #1158.  I did drop most of my (minimal) changes from #432 because I thought that's what you had asked for :facepalm: . I also dropped the "remove too-many-constraints" commit for now. In any case, this is now rebased, builds on #1158 and incorporates #1155 . Rebasing now to drop as much of #1158 as possible.

rpavlik:
OK, the version of this with as little of #1158 as possible, is here: https://github.com/solvespace/solvespace/compare/master...rpavlik:evilspirit-eigen-over-ryan-buildsys-fewer-solver-changes  I could drop the first three commits of #1158 but the rest were built upon by this code.

phkahler:
@rpavlik  Looks like the first 9 commits here conflict since they were merged in the previous PR. What's the best way to handle that?

rpavlik:
Weird, must be squash merging or something. I'll just rebase this if you don't beat me to it.

rpavlik:
OK, rebased!

phkahler:
@rpavlik   Lets see how this goes ;-)

phkahler:
@rpavlik  question, in testing I ran into "Too many unknowns". The define is still set at 1024, but I thought the code was modified to not have an actual limit.  Should we remove some checks, or increase the value?

BTW I made a long chain (snake) of connected rectangles much like evil_spirit showed in a video. This runs dramatically smoother when dragging it around.  ;-)

phkahler:
Closes #225

vespakoen:
This is an awesome way to start the year!! Happy new year everybody! Thanks @Evil-Spirit and @rpavlik for making this happen!!

rpavlik:
Awesome, thanks! The limit is now artificial, but left in to avoid getting to a point of being unresponsive. We can increase the define for Max unknowns as we see fit. Probably double it if we're a lot smoother now.

ruevs:
After the fact I code-reviewed this fairly carefully and it is fine! Thank you very much for cleaning in up, making the integration clean  and rebasing it @rpavlik and thank you very much @Evil-Spirit for writing it in the first place back in 2017 :-)

PR #1174, opened 2022-01-02
CMake Fixes + Snap port to core20

### LINKED ISSUE #1173: Snapcraft build issue
### System information

Ubuntu 20.04.3 (in Github Actions)

### Expected behavior

Snap should build successfully

### Actual behavior

Fails with ld linking errors

### Additional information

Working fine here: https://github.com/solvespace/solvespace/actions/runs/1485714200
Broke in the next push to master: https://github.com/solvespace/solvespace/actions/runs/1564051582

The commits in between are: https://github.com/solvespace/solvespace/compare/85f6ec4..91db627 and are unrelated / don't touch any snap stuff.

So I guess the issue is in the environment.
Both runs seem to use Snapcraft 5.0, the Ubuntu version is also the same, although the Github Actions image changed a little.

I tried fixing the missing cairo library with this commit: https://github.com/solvespace/solvespace/commit/b71c7282622b74f3f74a33d7801357d95f8a4d2e
 (here is the action run for it: https://github.com/solvespace/solvespace/actions/runs/1646157739)

Which made the missing -lcairo go away, but presents other missing libraries.

So my guess is that some library path is not provided anymore or is incorrect.

I don't have a linux computer setup at the moment, so I can only test things by going through CI, which is a bit annoying.

Hoping @ppd can help here ;)

ppd:
In 006539b, solvespace_MAJOR_VERSION etc. were removed.
However, these variables were still referenced in some places.
Solution: Use PROJECT_VERSION instead.

006539b removed the call to link_directories for gtkmm, jsonc & fontconfig.
This leads to linking errors if those libraries are in "non-standard"
paths.
Fix this by introducing a target specific target_link_directories call.

Fixes #1173

Moving to core20 was long overdue anyway, and
the recent CMake changes necessitated some fixes.
Also switch to LZO compression for (way) better cold start
performance.

I left the CMake commits separate for easier review. I/we can squash them before merging.

### DIFF
--- a/cmake/MacOSXBundleInfo.plist.in
+++ b/cmake/MacOSXBundleInfo.plist.in
@@ -15,9 +15,9 @@
         <key>CFBundlePackageType</key>
         <string>APPL</string>
         <key>CFBundleVersion</key>
-        <string>${solvespace_VERSION_MAJOR}.${solvespace_VERSION_MINOR}~${solvespace_GIT_HASH}</string>
+        <string>${PROJECT_VERSION}~${solvespace_GIT_HASH}</string>
         <key>CFBundleShortVersionString</key>
-        <string>${solvespace_VERSION_MAJOR}.${solvespace_VERSION_MINOR}</string>
+        <string>${PROJECT_VERSION}</string>
         <key>NSHumanReadableCopyright</key>
         <string>© 2008-2016 Jonathan Westhues and other authors</string>
         <key>NSPrincipalClass</key>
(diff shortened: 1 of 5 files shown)


vespakoen:
Looks good to me, thanks a lot for doing this!

ppd:
Oh no, I deleted g++ from the build packages :man_facepalming: 
See https://github.com/solvespace/solvespace/pull/1175

PR #1178, opened 2022-01-04
Fix various typos

luzpaz:
Found via `codespell -q 3 -S ./res/locales,./extlib -L asign,ba,hsi,mata,pinter,tothe,wser`

### DIFF
--- a/src/ttf.cpp
+++ b/src/ttf.cpp
@@ -11,7 +11,7 @@
 #include FT_ADVANCES_H
 
 /* Yecch. Irritatingly, you need to do this nonsense to get the error string table,
-   since nobody thought to put this exact function into FreeType isself. */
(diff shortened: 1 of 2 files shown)


src/ttf.cpp
```diff
@@ -11,7 +11,7 @@
 #include FT_ADVANCES_H
 
 /* Yecch. Irritatingly, you need to do this nonsense to get the error string table,
-   since nobody thought to put this exact function into FreeType itsself. */
+   since nobody thought to put this exact function into FreeType isself. */
```
ppd, line 14:
```suggestion
   since nobody thought to put this exact function into FreeType itself. */
```

luzpaz: :facepalm:

Issue #1020, opened 2021-04-21
Create an automated performance benchmark


phkahler:
We should have a performance benchmark that can be run in an automated way, probably using solvespace-cli. Maybe this is just loading an existing sketch that stresses a lot of code paths and produces an image?  Frame rates are not relevant because sketch regeneration when geometry changes takes way longer than rendering.

If there is an existing benchmark it's probably out of date with all the recent improvements. We should have something that takes on the order of seconds (10's of seconds?) where variations of a few percent to 5x can be observed.

This is based on an inquiry from Michael over at Phoronix, so keep in mind the kind of testing he's known for. If we're lucky he'll start doing benchmarking for us, but it should be easy/trivial for him ;-)

ruevs:
This would be extremely useful. However some performance tests depend on changing something. E.g. changing constraints in the fist group of this https://github.com/solvespace/solvespace/issues/841#issuecomment-742153838

solvespace-cli can not do that.

vespakoen:
I have a vague memory of seeing a benchmark function in someones fork at some point, I think it added something to the "test" functions that allowed one to see how long solving took.

Was it in your fork maybe @Evil-Spirit ?

Profiling on macOS is also pretty good, showing how long some lines of code took to run, I should probably open a whole bunch of big files and see where the bottlenecks are & share back.

phkahler:
@ruevs agreed. It's probably a good idea to stress the solver as well as everything else. Maybe we need a benchmark build (since we dont have scripting) that programmatically creates something and stresses everything?

Do groups get solved when opening a file? That should be easy though because it should be saved in a solved state.

PR #1184, opened 2022-01-08
Add some gitignore stuff

rpavlik:
Visual Studio 2022 creates these files that we want to ignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,8 @@
 /obj-*/
 /*.slvs
 .vscode/
+
+# Visual Studio
+out/
+.vs/
+CMakeSettings.json


ruevs:
?!? Did you create a `build` subdirectory? VS 2019 does not do that.

phkahler:
@ruevs  no.

ruevs:
From: https://github.com/solvespace/solvespace#building-on-windows

> Create a directory build in the source tree and point cmake-gui to the source tree and that directory.

or later

> `mkdir build`

Try it and if it works we can get rid of this.

rpavlik:
What are we getting rid of? I use vs and vs code and just the command line, and a directory named build is common in the instructions, and standard CMake usage, and default vs code behavior, and... Pretty frustrating to have a simple quality of life commit like this be immediately opposed after merge. I think I'll take the day away from the project to cool down.

ruevs:
> If you mean the lines added here instead: The ones I added here are the ones you get if you do "open folder" in vs with the built in CMake support in modern visual studio, which works well (and faster than the old generators since it uses ninja to build) and it's the recommended way by the vs team.

@rpavlik that's what I meant and reading my comments above I see that I completely failed to make it clear - sorry:-)
Now I understand - I had never tried to "open folder" in VS and did not even realize the build in CMake existed - so I was surprised where the new files you added to the ignore list were coming from. And guessing you missed to make a `build` directory was ridiculous - sorry.

> Pretty frustrating to have a simple quality of life commit like this be immediately opposed after merge. I think I'll take the day away from the project to cool down, which is a bummer, since I enjoy contributing and I've enjoyed the community working on this project. 🙁 😥

I apologize for annoying you. It was not my intention at all - sorry again.

rpavlik:
Thanks for following up and apologizing. A day away did help, I'm OK now. I think the force-push undo just probably got me a bit bothered, then opposition to a gitignore change was just too much for my early morning brain.

Yeah, the VS cmake support is pretty slick  and has been improving since they added it in the 2019 version. It's way faster to build this way than with the vs generators because ninja is much faster than msbuild, too. The most annoying bit is probably the files that it drops...

I think my Solvespace time this week will probably be hammering out the flatpak stuff since I can use a little work time for that. After that I really want to figure out a way to do some performance regression testing. Will probably involve a separate repo of publicly-sharable models we can use for testing, to avoid bloating this repo. I have seen some substantial "warm up" costs even with all the iterations the benchmark tool runs, so I ignore the first run. I might adjust it so that the first few runs don't get counted in the average. We should also figure out a general "philosophy" for what we'll merge: my tendency is to merge things that improve maintainability even if they don't improve performance or if they introduce a very small slowdown, and similarly probably avoid performance improvements that make maintainability worse unless they're really substantial improvements. But, that doesn't mean that the project's philosophy has to be that, we should just document what it is.

phkahler:
@rpavlik  Glad you're back so quick. I get frustrated here too sometimes, and in almost every case it's not because anyone meant to be annoying ;-)   As for gitignore, I figure if someone adds something it's because the way they do things is different and we should help them out so long as it doesn't impact the rest of us ;-)

From a performance benchmarking PoV, see issue #1020 - making something Michael can use might get the project noticed a lot more.

ruevs:
> We should also figure out a general "philosophy" for what we'll merge: my tendency is to merge things that improve maintainability even if they don't improve performance or if they introduce a very small slowdown, and similarly probably avoid performance improvements that make maintainability worse unless they're really substantial improvements.

My attitude is essentially the same, with a very small tolerance for slowdowns in maintainability improvements. On the other hand "truly ugly" things should really give substantial performance improvements to be worth it.

Everything else being equal, I have an extra, rather intangible, criteria that I can best describe as churn vs. increase in abstraction. I hate the former and love the latter, so my "feelings" for such changes are very context dependent and subjective.

SolveSpace is rather small - team wise - so we may be OK without written rules by simply discussing less obvious pull requests before merging them. And not force pushing master as I did ;-)

Issue #1172, opened 2022-01-02
Feature request: Add build date to the Help->About dialog


anymouse404:
Sometimes, it would be nice to easily find out how old my installed build is.
It would be nice if in addition to the revision number, the Help>About dialog contained a date of either the last commit or the build.

ruevs:
SolveSpace has always had a rather "simple" versioning approach. For example during the whole five years between v.2.3 and v.3.0 all builds from master were simply 3.0 and the commit hash. If we add the date of the last commit it is redundant - one can simply look up the commit on GitHub by going to (example for 3.0) `https://github.com/solvespace/solvespace/commit/0e0b025` where the last part of the URL is (a part of) the commit hash.

If on the other hand we add the build date there could be multiple binaries built by different people at different times from the same commit showing different dates in the About dialog - this does not carry information - since anyone can build whatever they want - but adds confusion.

Because of the above I think we should not add the date.

Finally @phkahler, @ppd, @jwesthues now that 3.0 is released should we change the minor version in master to 3.1 to match the approach used so far?

phkahler:
@ruevs  I'd like to get a 3.1RC out soon, so maybe wait until then? I think leaving it 3.0 for all the edge builds since 3.0 might have been a mistake. It shouldn't be 3.0 for pre and post 3.0 official ;-)

BTW @ppd  I saw your email on my phone, just need to get to it. Not feeling well this week.

anymouse404:
That makes sense from dev's perspective, but it is very cumbersome and confusing from user's perspective. I get that the build date is not reliable, but there really should be *some* information about the age so that a user doesn't have to cross-reference the commit hash, if they know how to make sense of it at all. Redundancy is good in this case IMO.
I don't know about what the build system and the program allows. Would it perhaps be possible to show date of the commit the same way that it currently shows its hash?

ppd:
Versioning is really a matter of philosophy, I guess. I'm rather fond of bumping the version just before/when tagging a new release. Dev/edge versions should then always be named according to the git describe scheme, e.g. `v3.0-101-g985e4fb`, which clearly shows how far we've come since the last release. 

Advancing the in-source version before tagging the release leads to confusion when you then decide to do major release instead. This is probably not really a concern for solvespace though.

ruevs:
[...]

Issue #1098, opened 2021-08-23
CLA Assistant Agreement check needs to be removed


Symbian9:
### System information

- **SolveSpace version:** `3.0~master`
- **Operating system:** All supported

### Expected behavior

CLA Assistant should be fully un-linked/removed from pull request checks for SolveSpace repo due to violating SolveSpace licensing, because *"SolveSpace is licensed under the GPLv3 or later and any contributions must be made available under the terms of that license"* 

### Actual behavior

New SolveSpace contributors (e.g., https://github.com/solvespace/solvespace/pulls/DevyHeavy) may not aware that *CLA Assistant Agreement* (Version: 2018-07-13) violates actual SolveSpace licensing (see https://github.com/solvespace/solvespace/issues/943):

- https://cla-assistant.io/solvespace/solvespace
> ...
>
> 2.3 Outbound License
>
> Based on the grant of rights in Sections 2.1 and 2.2, if We include Your Contribution in a Material, **We may license the Contribution under any license, including copyleft, permissive, commercial, or proprietary licenses**. As a condition on the exercise of this right, We agree to also license the Contribution under the terms of the license or licenses which We are using for the Material on the Submission Date.
>
> ... 

Actually (as has already been finally clarified by @jwesthues in https://github.com/solvespace/solvespace/commit/2fcc933aa9c7d8f30a35d94fdb760b343e4562bf) all contributions to SolveSpace *"must be made under the GPLv3 or later"*, to prevent including of SolveSpace source and contributions to any proprietary products.

- https://github.com/solvespace/solvespace/blob/master/CONTRIBUTING.md#licensing
> SolveSpace is licensed under the GPLv3 or later and any contributions
must be made available under the terms of that license.

### Additional information

- https://github.com/cla-assistant/cla-assistant/issues/691

phkahler:
Yeah, does anyone know how to disable that? We're just GPL3 now since things have been merged without the CLA already.

Symbian9:
@phkahler, please, look here:

- https://github.com/cla-assistant/cla-assistant/issues/561

Symbian9:
@phkahler, please "pin" this issue to make more visible for new contributors.

ppd:
Wouldn't it be sufficient to remove the webhook to get rid of it? There  should be one in the GH settings.

Shegox:
So there are quite a few ways to disable it: 
[...]

PR #1185, opened 2022-01-09
CMake: use git rev-parse to get GIT_COMMIT_HASH

herrgahr:
The old approach of reading .git/HEAD does not work when using git worktrees, where the folder layout looks roughly like:

```
solvespace.git/                      - bare clone (.git dir)
solvespace.git/work                  - example worktree containing master
solvespage.git/worktrees/work/       - .git dir of worktree
solvespage.git/worktrees/work/HEAD   - actual HEAD ref for master
```

First attempt was to just get GIT_ROOT from `git rev-parse --git-dir` but that wasn't enough, since:

1. GIT_ROOT points to solvespage.git/worktrees/work/
2. GIT_ROOT/HEAD points to refs/heads/master
3. GIT_ROOT/refs/heads/master does not exist but the old implementation would want to use this to get the sha

so we need two invocations of git rev-parse

1. `git rev-parse --git-dir` to get GIT_DIR needed for setting GIT_DEPENDS
2. `git rev-parse HEAD` to get the sha of the worktree's HEAD


herrgahr:
Note: I don't use CMake much so this might not be the most sensible approach to this. I'm also not certain at which cmake version this was added so I'm not sure if `cmake_minimum_required(VERSION 3.9...3.19)` covers `execute_process`

cmake/GetGitCommitHash.cmake
```diff
@@ -1,31 +1,11 @@
...
-            else()
-                set(HEAD_REF "")
-            endif()
-        else()
-            set(HEAD_REF "")
-        endif()
-    endif()
     set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS ${GIT_DEPENDS})
 
-    string(STRIP ${HEAD_REF} HEAD_REF)
```
herrgahr, line 28: this is redundant since `execute_process`'s `OUTPUT_STRIP_TRAILING_WHITESPACE` option takes care of trailing whitespace and the output of  `git rev-parse` doesn't containg leading whitespace. But: if `git rev-parse` fails, `${HEAD_REF}` is undefined and cmake gives a confusing error message regarding the number of arguments give to `strip`

ruevs:
@herrgahr welcome to the project :-)

It is rather interesting that you submit this pull request now since there is an active discussion relating to this for the last few days here  https://github.com/solvespace/solvespace/issues/1172 :-) Take a look.

herrgahr:
Oohhh, I wasn't aware that this problem was being discussed there, the title didn't give it away ;)

This was the typical drive-by fix, I just happened to check out the sources today after a looong time

I already suspected rummaging through .git folder and HEAD files manually was for edge cases where git is not installed or somesuch. So I guess it makes sense to close this in favor of integrating @rpavlik 's CMake magic.

Either way, I'd be a happy camper if the build worked with worktrees :)

Should I close the PR and open an issue for this?

ruevs:
> Either way, I'd be a happy camper if the build worked with worktrees :)
> 
> Should I close the PR and open an issue for this?

I'd say leave the PR open for now, but it's up to @phkahler ultimately. And I'm not at all proficient with CMake to say whether Rpavlik's solution works with worktrees :-)

Symbian9:
> @herrgahr welcome to the project :-)

I am second with such greetings ;)

JFTR, @herrgahr, there no need to sign CLA anymore, as it violates GPL:

- https://github.com/solvespace/solvespace/issues/1098

rpavlik:
I'm pretty sure my module works with work trees, but not 100% sure. Welcome!

phkahler:
> JFTR, @herrgahr, there no need to sign CLA anymore, as it violates GPL:

No @Symbian9  it does not violate the GPL. However we do not require it anymore, the code shall stay GPLv3+

Symbian9:
> No, it does not violate the GPL.

It is *violate* in case that any changed file licensed under GPL (SolveSpace source actually GPL, not double licensed) could NOT be double licensed as stated in CLA Assistant:

- https://cla-assistant.io/solvespace/solvespace

I already highlighted it here: https://github.com/solvespace/solvespace/issues/1098#issue-977035998

phkahler:
> I'd say leave the PR open for now, but it's up to @phkahler ultimately. And I'm not at all proficient with CMake to say whether Rpavlik's solution works with worktrees :-)

Oh man, don't leave stuff like that to me! If the @rpavlik solution does the job, lets go with that. Otherwise @ruevs you can merge this PR ;-)

phkahler:
On second thought, this is a fix and simplification that is available now. If it gets supplanted later that's fine too.

ruevs:
@phkahler @herrgahr unfortunately this breaks the git commit in the version string when building in the IDE in VisualStudio since (I guess) `git` is not in the "path" of the "build environment". I do have `git` and it is in the global `path` of the system.

So the version ends up "3.0~" :-(

phkahler:
@ruevs  so I guess I'll revert it tonight.

PR #1189, opened 2022-01-13
Add a link to the GitHub commit at the bottom of the config screen

ruevs:
The "Go to GitHub commit for 3.0~a88f5d8b" link at the bottom of the configuration screen opens the URL [`https://github.com/solvespace/solvespace/commits/a88f5d8b`](https://github.com/solvespace/solvespace/commits/a88f5d8b) and allows the user to see the date of the commit that SolveSpace was built from.

image

Also revert f1e47e655 https://github.com/solvespace/solvespace/pull/1185

Closes #1172

PR #1191, opened 2022-01-14
Revert "CMake: use git rev-parse to get GIT_COMMIT_HASH"

phkahler:
This reverts commit f1e47e6554b469ebe2cbac849e6264a3012e4c31.

That method did not work properly on Windows.


ruevs:
@phkahler as you noticed if you decide to merge either #1190 or #1189 (or both?) this is already in those.

PR #1193, opened 2022-01-17
Add normals to linked STLs

phkahler:


src/importmesh.cpp
```diff
@@ -59,20 +59,43 @@ static hEntity newPoint(EntityList *el, int id, Vector p) {
...
+    Entity en = {};
+    en.type = Entity::Type::NORMAL_N_COPY;
+    en.extraPoints = 0;
+    en.timesApplied = 0;
+    en.group.v = 472;
+    en.actNormal = normal;
+    en.construction = false;
+    en.style.v = Style::NORMALS;
+    // to be visible we need to add a point.
+//    en.point[0] = newPoint(el, id, Vector::From(0,0,0));
```
ruevs, line 87: This is left over early code, before you added `hEntity p` as a parameter I presume.
It should be removed.

src/importmesh.cpp
```diff
@@ -85,7 +108,8 @@ static hEntity newLine(EntityList *el, int id, hEntity p0, hEntity p1) {
     en.actVisible = true;
     en.forceHidden = false;
 
-    en.h.v = id + en.group.v*65536;
+    en.h.v = *id + en.group.v*65536;
+    *id = *id + 1;
```
ruevs, line 112: This replaces the `id++` in the calls below. OK.

src/importmesh.cpp
```diff
@@ -183,35 +212,34 @@ bool LinkStl(const Platform::Path &filename, EntityList *el, SMesh *m, SShell *s
     }
 
     hEntity p[8];
-    int id = el->n+2;
```
ruevs, line 186: What was this for? I obviously have not gone deep enough to understand.

src/importmesh.cpp
ruevs: But why did you decide to do it? Just aesthetics or am I missing something?

ruevs (commented):
Overall it works just fine :-)

src/importmesh.cpp
phkahler: This was borrowed from the way it's done in idfimport. Over there I did it because that's how I saw them combined somewhere else? Using a pointer to the id was also borrowed from the idf code because I'd like to eventually use the same set of entity creation functions in both. The ID stuff is still strange to me overall, so I'm not real sure about it.

src/importmesh.cpp
phkahler: Yes, I need to remove that.

src/importmesh.cpp
phkahler: I honestly don't recall. I may have wanted to leave a gap in there to add a normal? It turns out the 3 normals require 4 entities if they share the same point.

PR #1194, opened 2022-01-19
Fix resolve of relative path of assembly group.

hzeller:
If the path of the slvs file that refers to assemblies, the
Parent() of the file is empty, thus the Join() operation returns
an empty string, and loading a file with assemblies fails.
Fix: Don't Join() in that case.

Signed-off-by: Henner Zeller <h.zeller@acm.org>

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -415,7 +415,8 @@ void SolveSpaceUI::LoadUsingTable(const Platform::Path &filename, char *key, cha
                 case 'P': {
                     Platform::Path path = Platform::Path::FromPortable(val);
                     if(!path.IsEmpty()) {
-                        p->P() = filename.Parent().Join(path).Expand();
+                        const Platform::Path &parent = filename.Parent();
+                        p->P() = parent.IsEmpty() ? path : parent.Join(path).Expand();
                     }
                     break;
                 }


hzeller:
I think the fact that [`Path::Join()`](https://github.com/solvespace/solvespace/blob/5cb299b2bef43634bb6c6e0cb22508b1e16aaa0e/src/platform/platform.cpp#L242) does have the behavior if any of the joined elements is empty, it returns an empty path, is somewhat surprising and error-prone (Also, it is only documented in the cpp file, [not the header](https://github.com/solvespace/solvespace/blob/5cb299b2bef43634bb6c6e0cb22508b1e16aaa0e/src/platform/platform.h#L49-L50).

hzeller:
Example of use-case
```
  solvespace my-assembly-file.slvs
```

This failed to load and only worked if the filename had a path attached to it, even if it is just `./`
 
```
  solvespace ./my-assembly-file.slvs
# or
  solvespace `pwd`/my-assembly-file.slvs
```

ruevs:
You are right - I debugged it. It is a problem. I guess no one has ever ran SolveSpace with a single file name as parameter (from the command line) and that file being an assembly.

The code here:
https://github.com/solvespace/solvespace/blob/b399d9a1ec50086cdfc827a43d661ef79052e175/src/platform/platform.cpp#L252
specifically the `IsEmpty()` is indeed what causes the problem and I am wondering if anything relies on this behaviour.

Obviously the code:
https://github.com/solvespace/solvespace/blob/b399d9a1ec50086cdfc827a43d661ef79052e175/src/platform/platform.cpp#L258
does - otherwise it would append a `\\` to an empty string, and will point to the root directory, which is certainly not correct.

But there are overall 29 places that call `Join` I'll have to review them all.

In the meantime your fix does work but I wonder if it is the "right thing to do.

hzeller:
I think indeed the 'right thing to do' would be to have `Path::Join()` _not_ return an empty path in this case, as it is suprising and behaves unlike typical implementations of the same functionality. My patch is just a band-aid for a particular use-case, but I wonder how many other bugs are lurking due to the Join()-behavior.

(I wonder if c++17 is considered for solvespace, then one could use [`std::filesystem`]( https://en.cppreference.com/w/cpp/filesystem) for platform-independent handling of these things)

ruevs:
All of the above was on Windows, since there are platform specific parts in the file handling (even though this is not one of them) this should at the very least be tested on Linux and MacOS before merging.

hzeller:
I developed and tested this patch on Linux.

Symbian9:
> I guess no one has ever ran SolveSpace with a single file name as parameter (from the command line) and that file being an assembly.

I did it last summer:) 

- https://twitter.com/app4soft/status/1406372114630852609

And here it's SLVS: 

- `WARNING!` Cursion Assembly: [Burgerecursion_slvs.zip](https://github.com/solvespace/solvespace/files/7909604/Burgerecursion_slvs.zip)

Screenshot_2021-06-20_14-10-40

ruevs:
> I wonder if c++17 is considered for solvespace

C++ 11 for now:
https://github.com/solvespace/solvespace/blob/master/CONTRIBUTING.md#contributing-code

And specific "praise" for I/O in C++ ;-)
https://github.com/solvespace/solvespace/blob/master/CONTRIBUTING.md#libraries

I have not looked at `std::filesystem` but my guess is it's not such an amazing thing to require C++17 just for using it.

hzeller:
Yes, IO in C++ is awful.
Most of the time we switched to C++ 17 in other C++ projects, that I am involved in, was in part due to `std::filesystem` making it easier to write portable code that also has to work on Windows.  There are other features of course as well; my favorite thing in 17 is `std::string_view`, but there are other features such as `constexpr`, structured bindings and variable declaration right in an `if` that make it worthwhile considering. In 2019 I was still wary of doing such switch because of old compilers, but these days one can assume c++17 availability in all relevant distributions.

Anyway, this is an entirely different discussion orthogonal to this PR :)

hzeller:
How do we proceed with this PR, which fixes a real problem ?
Use as-is as stop-gap, or should I investigate all the places that use path-joining and determine if a call-site is expecting the quirky behavior of if it is safe to change it to behave as expected.

ruevs:
@hzeller I would like to review the places that call `Join` and if possible modify it. As you pointed out it's behaviour is a bit counter intuitive. But I have not found time. If you want to do it - by all means go ahead - that would be great.

phkahler:
> How do we proceed with this PR, which fixes a real problem ?
> Use as-is as stop-gap, or should I investigate all the places that use path-joining and determine if a call-site is expecting the quirky behavior of if it is safe to change it to behave as expected

It's always preferable to get to the root problem rather than fix one instance. If nobody is really up for it, I'll take it as-is. This is an area where I agree with previous maintainer (who would probably not accept this) in principle, but am willing to bend in practice ;-)

hzeller:
I am happy to do it the 'right way', check and document all the Join() places.
This will have to wait for the weekend when I have more time.

phkahler:
@hzeller  Any update on this?

hzeller:
Yes, I did the changes and got distracted, I should make it a PR.

phkahler:
> Yes, I did the changes and got distracted, I should make it a PR.

@hzeller  ???  :-)

hzeller:
Oh, uhm, yes!
Should swap in my state back into brain this evening...

phkahler:
@hzeller I've been busy too. I'd like this for the 3.1~RC1 which I'd like to push out very soon.

rpavlik:
We would love to have this fix, it seems. Anyone have bandwidth to make the needed changes?

ruevs:
@hzeller please take a look at https://github.com/solvespace/solvespace/pull/1421 is shoudl be the "clean" alternative to this pull request.

phkahler:
@hzeller and @rpavlik  can we close this? Did the other merged PR fix the issue well enough?

hzeller:
Yes, I think this can be closed now.

Issue #1084, opened 2021-08-07
SolveSpace runs out of RAM and swap on linking STL files


Symbian9:
### System information

- **SolveSpace version:** `3.0~` https://github.com/solvespace/solvespace/commit/1b8e1dec652eef7916a94b6c9db404457dacb0db
- **Operating system:** Debian 10.x buster

### Expected behavior

SolveSpace should detect any potential issue on linking 'dangerous' STL files to prevent overload RAM.

### Actual behavior

SolveSpace fully blows my RAM with linking next STL files:

- `break_openscad.stl` - https://forum.freecadweb.org/viewtopic.php?t=60912&p=522689#p522689
- `Suzanne.stl` - https://en.wikipedia.org/wiki/File:Suzanne.stl
- `Utah_teapot_(solid).stl` - https://commons.wikimedia.org/wiki/File:Utah_teapot_(solid).stl

I just kill SolveSpace when it stopped to be responsible and its process loaded RAM at 100% and started to overload swap.

### Additional information

In same time here are STL files of Blender's default "Cube" (`cube.stl`) & "Suzanne" (`monkey.stl`) mesh objects, which I exported from Blender 2.49b myself (mesh quads converted to triangles before export) and *which are linking successfully* into SolveSpace without RAM issues:

- [blender_cube_monkey_stl.zip](https://github.com/solvespace/solvespace/files/6950016/blender_cube_monkey_stl.zip)

![pic](https://i.imgur.com/A6ofvac.png)

ruevs:
This is true - there are no provisions in the code to restrict RAM usage. Probably due to the fact that it would become unusably slow, when trying to construct complex objects in the "normal" way, long before RAM usage became excessive.

Linking STL is the first time we have been able to "trick" SolveSpace into importing something "too complex".

Symbian9:
>  Probably due to the fact that it would become unusably slow, when trying to construct complex objects in the "normal" way, long before RAM usage became excessive.

Issue here is not about its complexity, but about 

Just compare next files:

- `break_openscad.stl` (blows out RAM) VS `cube.stl` (not issued RAM)
- `Suzanne.stl` (blows out RAM) VS `monkey.stl` (not issued RAM)

Maybe there are some digit precision or big distance values in those STL files which triggers SolveSpace to blows RAM? Or maybe there some difference in format structure, characters, etc (in comparison to `cube.stl` and `monkey.stl`).

phkahler:
[...]

PR #1197, opened 2022-01-23
Check for text STL when linking

phkahler:
This is more of a WIP.  It no longer locks up if you link a text formated STL file, but the dialog we get is not helpful.

### DIFF
--- a/src/importmesh.cpp
+++ b/src/importmesh.cpp
@@ -130,7 +130,7 @@ bool LinkStl(const Platform::Path &filename, EntityList *el, SMesh *m, SShell *s
     char str[80] = {};
     f.read(str, 80);
     
-    if(str[0]=='s' && str[1]=='o' && str[2]=='l' && str[3]=='i' && str[4]=='d') {
+    if(0==memcmp("solid", str, 5)) {
     // we could display a message that text stl files are not supported
     // just returning false will trigger the warning that linked file is not present
     // best solution is to add an importer for text STL.
--- a/src/importmesh.cpp
+++ b/src/importmesh.cpp
@@ -131,9 +131,9 @@ bool LinkStl(const Platform::Path &filename, EntityList *el, SMesh *m, SShell *s
     f.read(str, 80);
     
     if(0==memcmp("solid", str, 5)) {
-    // we could display a message that text stl files are not supported
     // just returning false will trigger the warning that linked file is not present
     // best solution is to add an importer for text STL.
+        Message(_("Text-formated STL files are not currently supported"));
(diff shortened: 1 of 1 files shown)


src/importmesh.cpp
```diff
@@ -130,6 +130,13 @@ bool LinkStl(const Platform::Path &filename, EntityList *el, SMesh *m, SShell *s
     char str[80] = {};
     f.read(str, 80);
     
+    if(str[0]=='s' && str[1]=='o' && str[2]=='l' && str[3]=='i' && str[4]=='d') {
```
ruevs, line 133: I love this! :-D
C++ be damned, STL be damned even C and `strstr` be damned; we're just gonna check the bytes one by one! ;-)

In fact I am not even entirely joking, enen though I did write `strstr` EXACTLY on this line in my local copy [when we discussed the bug originally](https://github.com/solvespace/solvespace/issues/1084#issuecomment-896969284). Then I decided I would rather write text format STL parsing... and never got to it (so far) :-(

In fact I would almost merge this :-)

phkahler: @ruevs We really can't use standard string functions here because str may not be null terminated. I spent a few minutes looking for a standard comparison that would work and just decided that this one liner is self explanatory and simple enough ;-)

I'm more concerned about the useless error message it produces. Making a dialog for that doesn't seem great either since the "correct" thing to do is handle the text STL file, but that will still need this check.

ruevs: I did do
```C++
char str[81] = {}
```
 to guarantee the null termination for `strstr` ;-).

 But anyway I totally agree with everything you wrote.

phkahler: You should have submitted a PR...

phkahler: @ruevs  look at the first function in file.cpp   I think it fits the bill.

ruevs: It does not, because of `strlen` but `if(0==memcmp("solid", str, 5))` should. I should have thought of that :-) (writing on the phone, so not tested)

ruevs: OK added a commit with `memcmp`.

It does not matter at all but it optimizes beautifully :-)
image

phkahler: StrStartsWith only uses strlen on the second string, which would be "solid" and would be terminated in this case.

phkahler:
Added a useful message box, but it only show *after* the useless one?  Also, if we want to use StrStartsWith it should probably be moved to util, and must be added to a header somewhere.

PR #1201, opened 2022-01-30
Fix normal issue, when a workplane is created from a point and normal…

### LINKED ISSUE #1120: New sketch from point and normal does not follow its parent transforms
### System information

- **SolveSpace version:** `3.0~` https://github.com/solvespace/solvespace/commit/6bc63e92b0d908aa751ae06063b2ec00fec4cc57

### Expected behavior

If new sketch created from point & normal, this new sketch should follow transforms of parent point and normal.

### Actual behavior

Actually if create revolve and use normal from resulted revolve ends for new sketch - new sketch created in same plane as revolve end face, but any changes in parent normal does not applied to new sketch orientation & new sketch origin does not follow parent point changes:

https://user-images.githubusercontent.com/13188580/134754585-09178328-a152-420b-bf23-5ab306659694.mp4

### Additional information

Related pull

- https://github.com/solvespace/solvespace/pull/1054

phkahler:
… we were only storing the numeric normal so the plane (and subsequent extrusion) wouldn't follow if the sketch was on the face of a revolve. Now we store the handle to the defining normal it predef.entityB so it can update properly on regeneration.

Fixes #1120

### DIFF
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -136,6 +136,9 @@ void Group::MenuGroup(Command id, Platform::Path linkFile) {
                         g.predef.negateV = wrkplg->predef.negateV;
                     } else if(wrkplg->subtype == Subtype::WORKPLANE_BY_POINT_ORTHO) {
                         g.predef.q = wrkplg->predef.q;
+                    } else if(wrkplg->subtype == Subtype::WORKPLANE_BY_POINT_NORMAL) {
+                        g.predef.q = wrkplg->predef.q;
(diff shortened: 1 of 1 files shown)


ruevs:
@phkahler I added a [commit](https://github.com/solvespace/solvespace/pull/1201/commits/2e553b310fecb0ec071d5719682112f34d5f3b82) fixing the bug I described here https://github.com/solvespace/solvespace/issues/1120#issuecomment-1025791804

ruevs (approved):

Issue #1211, opened 2022-02-07
Display a circle length in the browser


hinell:
![image](https://user-images.githubusercontent.com/8136158/152760009-9d2988df-af8f-497c-97f9-5c36ccaacedb.png)

I request to enable displaying a circle length in the browser window when selected.

### System information
- **SolveSpace version:** 3.0.rc2+repack1-3
- **Operating system:** Linux/Microsoft

phkahler:
Should we also add arc length?

hinell:
@phkahler May be useful, I'm not sure. I was making  a gear and found myself calculating its length it manually.
@EkaterinaEvdokimova Екатерина, ты из России?

EkaterinaEvdokimova:
@hinell Да, я из России

phkahler:
@hinell   I just checked, and you can measure lengths with Analyze->Perimeter in the menu. You select any number of lines, arc, circles (but no points, not sure why those aren't auto-excluded) and it will give the total length.

It still seems reasonable to display the length of a single selected entity in the text window though.

phkahler:
> @hinell Да, я из России

I've looked at google search data, and some other things and it seems Solvespace is quite popular in Russia and parts of eastern Europe. Does anyone know why that might be? Just curious.

hinell:
@phkahler It's handy to have this information listed right away in the browser data in unit-less format. I used this information to calculate the pitch for the gear (in non standard way of course). Would be best if we can have _cog tool_ for that at some point in the future though.

Regarding geographical distribution of users, well I'm not sure why. Probably because it's free. Personally I started to use it only recently. It's a good (not perfect though) alternative for Linux.

ruevs:
There are better ways.

You can type formulas in any dimension field. If you want a circle with a circumference of 500 just type `500/(2*pi)` for the radius.

You can use `sqrt`, `square`, `sin`, `cos`, `asin`, `acos`, `pi`, `+`, `-`, `*`, `/`, `()`

![CircleCircumference1](https://user-images.githubusercontent.com/15338069/152974170-11b3c842-5244-4161-8ded-605b6fae51ff.png)
![CircleCircumference2](https://user-images.githubusercontent.com/15338069/152974181-33cc9962-2ad0-445b-87c9-d61b42984c81.png)

And don't ask for the formulas to be remembered it is the oldest and best feature request  https://github.com/solvespace/solvespace/issues/77 ;-)

ruevs:
[...]

PR #1215, opened 2022-02-09
Add display of a circle length

EkaterinaEvdokimova:
Added display of a circle length referring to issue #1211 . Also added "CMakeLists.txt.user" to .gitignore as this unnecessary user-specific file is being generated by Qt/


ruevs:
The circle length is called circumference https://github.com/solvespace/solvespace/pull/1216.

PR #1218, opened 2022-02-12
Revert "mac: Don't interpret single-touch scroll events as pan gestures"

vespakoen:
This reverts commit 91db627a81a1438ca4e02f0b27c6f17ceac1051c.

### DIFF
--- a/src/platform/guimac.mm
+++ b/src/platform/guimac.mm
@@ -372,7 +372,6 @@ MenuBarRef GetOrCreateMainMenu(bool *unique) {
     double             rotationGestureCurrent;
     Point2d            trackpadPositionShift;
     bool               inTrackpadScrollGesture;
-    int                numTouches;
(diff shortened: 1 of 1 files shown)


nanoant:
Sorry to bother you @vespakoen and @phkahler but why did you revert this? Unfortunately there's no description given in this PR.
Now Magic Mouse again acts like a touch pad (again), and there's no way to zoom anymore with the mouse.
I though we had this settled in #1106 but now I'd need to open #1106 again? Or am I missing something 😮

ruevs:
The discussion that lead to this rollback is towards the end of https://github.com/solvespace/solvespace/pull/1107 (I think, I don't follow Mac too closely since I don't have a modern one). @tomsci should be able to explain it.

nanoant:
Thanks @ruevs! I'll continue there.

PR #1220, opened 2022-02-13
Move shell functions out of surface.cpp and into shell.cpp

phkahler:
This was the first thing I wanted to do when starting to write code for #955 Hole Tool. Then decided it might be frowned upon, so proposing it separately.


rpavlik (approved):
This is just moving the code to a separate file, right? I don't have to review every line? ;) Sounds good to me, long source files often make me nervous

phkahler:
> This is just moving the code to a separate file, right? I don't have to review every line? ;) Sounds good to me, long source files often make me nervous

@rpavlik Yes, that's it - just moving functions. I don't like huge source files either and this split naturally almost in half.

ruevs:
A perfect split :-)

phkahler:
I remember whitequark not wanting to refactor just for the sake of refactoring - something about using git blame - so I wanted to get opinions on it. Looks like we're on the same page. If the change makes sense or makes things easier then make it.

ruevs:
In principle I'm with Whitequark on this. But this particular change makes sense.

On the other hand when someone submitted a pull request that split solvespace.h in "50" separate headers  - totally stupid (albeit perhaps not totally pointless).

Or - for example - globally finding and replacing `unsigned int` with `uint32_t` I would never approve of.

rpavlik:
For things that don't move files (like spacing) there's actually a mechanism for telling git blame to ignore certain commits now. I don't think GitHub handles it yet, but there is hope on the horizon, I've used it in other projects already. Just an fyi, I don't intend to submit a reformatting commit or anything.

PR #1221, opened 2022-02-21
Fix for zoom in/out error

dustinhartlyn:
When creating any multi point shape (such as rectangle), if zoom in/out is used from the keyboard or menu after setting the first point of the object, the mouse will be misaligned with the current point. 
- Select rectangle from tool window
- Click on sketch to set first point-point
- Press – or + on keyboard to zoom in or out to see alignment issue that occurs.

When I dug into the code I found that there are two different function responsible for zooming in/out. It seems the first was originally created in graphicswin.cpp for the menu option while the second was made in mouse.cpp for the scroll-wheel

Unfortunately having two different functions created to accomplish the same behavior left the opportunity for anomalies and I believe I have fixed the problem by making a single zoom function in graphicswin.cpp. This function takes a single argument which is the zoom multiplier and adjusts the scale centered on the mouse position (same behavior as the zoom function on mouse.cpp).  The scroll-wheel event goes to the same function in mouse.cpp, but instead of performing actions directly it points to the shared ZoomToMouse function placed bellow the ZoomToFit function in graphicswin.cpp.  The keyboard also points to this function instead of modifying the 'scale' variable directly. Triggering the event from the menu appears to work naturally even though the mouse is positioned over the menu. 

In addition to fixing the bug and eliminating a redundant function, I think the ZoomToMouse function offers the opportunity for future development where a call can be easily made to zoom in/out. It could also be modified to be a ZoomToPoint function if needed. The ZoomToMouse function would be changed to forward the mouse position as the point of focus, but I did not see a need to go that far just yet.


phkahler:
Also want to bring in @vespakoen  since he's been in the mouse code and has pending PR #1218 which might conflict?

dustinhartlyn:
Just following up on rather this code conflicts with work from @vespakoen
I looked over the edits of PR #1218  and don't see any conflicts myself. Especially since I left the original function used for zooming in/out within mouse.cpp, I think the edits I made are fairly sensitive to any work being done by others.

phkahler:
@dustinhartlyn thanks for that update.

ruevs:
@phkahler @dustinhartlyn I am not sure it this is a bug or a feature.

Before this pull request was merged scroll wheel zooming was centered on the mouse pointer and `+` `-` zooming was centered on the view-port (the center of the drawing area). To me this was a feature.

phkahler:
> Before this pull request was merged scroll wheel zooming was centered on the mouse pointer and `+` `-` zooming was centered on the view-port (the center of the drawing area). To me this was a feature.

@ruevs I still see that behavior on Windows. Have not specifically checked in Linux yet.

@dustinhartlyn  If I draw a rectangle and zoom while dragging, the zoom with mouse wheel is centered on the point/mouse pointer. If I zoom with +/- it zooms relative to the screen center, but the point being dragged is no longer under the mouse pointer. This seems like the "bug" you set out to fix, or is it different. Either way I don't think this is done yet.

On a somewhat related note, I've always wanted the ability to select a point and have rotation be around that point. As it is, rotation when zoomed in usually results if seeing a completely different part of the model.

ruevs:
> @ruevs I still see that behavior on Windows. Have not specifically checked in Linux yet.

@phkahler strange... Ara you sure you have master at fc16cdb370f60 ?

With that commit I see that zooming in both ways (scroll wheeel and +-) is always "centered" on the mouse pointer - irrespective of whether one is still drawing a rectangle or not and whether LMB is pressed or not.

phkahler:
> @phkahler strange... Ara you sure you have master at [fc16cdb](https://github.com/solvespace/solvespace/commit/fc16cdb370f6004e5a2135ec55cd7a4c0c8b5f9c) ?

I'll check on my linux system later. I used the latest edge build on windows, but that may not have updated with recent commits.

ruevs:
Hmmm the latest edge build on windows is from this commit https://github.com/solvespace/solvespace/commit/c5ea9a44
Something is wrong...

PR #1224, opened 2022-03-06
Repeated tool use and spacebar to alternate on/off

dustinhartlyn:
Totally fine if this pull request is rejected. It is something I did for myself, but think it is an improvement to the UI. 

First
I found it annoying to click on the same tool each time I wanted to draw a shape. My alteration reselects the tool after it is used. For example this allows the user to draw rectangles continuously after only selecting the tool once. 
Second
I find it really useful to switch between the mouse and the most recently selected tool with a hotkey. 

Esc still clears the selected tool, and hotkeys switch between tools as expected. 
ClearPending() is used to process through tools - such as drawing the first point of a rectangle, and then placing the second point. I altered the function to store the original tool command and reactivate it when the ClearPending has reached the end of its command chain.  Since all the tools funnel through this function, the one change works for all without any additional change. However I realized that it might not always be desired to reactivate a tool, so I altered the Toolbar[] array which stores all the tools. I added the 'keep_active' bool which allows authors to specify which tools should be reactivated (1) and which should be unselected after a single use (0). 

To accomplish the flip-flop between the last selected tool and the mouse I added the AlternateTool() which is called by the Command::ALTERNATE_TOOL case.


dustinhartlyn:
I see that the test failed for mac and Ubuntu. I will look into that.

ruevs:
This changes a very fundamental UI "presumption" that has always been true so far - the mouse can always select/modify/drag entities unless one activates a drawing tool. When using keyboard shortcuts  this allows for a very fast and efficient workflow (there are shortcuts for everything on the toolbar and I personally keep it hidden). Lets say the goal is to draw a lot of rectangles.:
1. `R`
2. Click - Drag - Click (draw a rectangle)
3. Resize if needed, maybe constrain
4. Go to 1

This new approach causes a situation where one never "knows" what the mouse will do unless one looks at the text window or had the tool bar enabled to see that a button is highlighted. In addition the workflow above will actually be less efficient:

0: `R`
1. Click - Drag - Click (draw a rectangle)
2. `Space`
3. Resize if needed, maybe constrain
4. `Space`
5. If constraining in step 3 go to 0 otherwise go to 1

In my opinion a definite no.

Unrelated - when creating a pull requests please rebase your branch on current master and clean up the commits. For example this PR should contain a single commit.

phkahler:
@dustinhartlyn I'm in agreement with @ruevs on this. Learning the shortcut keys for the tools is fundamental to using Solvespace efficiently and make this redundant, or even more complicated.

I do appreciate people making changes and proposing them for inclusion, so thank you for the effort. If it works better for you, by all means maintain it on your own branch.

Another thing to think about (and I'm not sure I'd like it either) is using right-click to distinguish what behavior you want. Line segments for example are chained together with LMB but terminate with RMB.  You might try keeping the behavior standard for RMB termination of things like rectangle, while repeating the tool use with LMB? But again, don't take this as an approval it's just a random thought I'm having. You may or may not find it better or worse than what you've got now ;-)

dustinhartlyn:
I love GitHub because it allows for disagreements. I come from a drafting background and think it simplifies the process greatly.
I find it annoying to press the hotkey each time I want to draw a shape.  I would much rather select the tool a once and have it be active until I select another tool or cancel out of it. I know you don't want to be other drafting programs, but I think there is a good reason that virtually everyone of them keeps a tool active until unselected. 

I have farther developed the code to undo (and forget) the draw action if escape/right-click is pressed while in process (I.E. first point of rectangle is placed, but not the second).  If you want me to create a pull request for that code I will happily do so. I think it avoids the concern that was raised for the wrong tool being used when drawing. I also found it solved my annoyance with the last line segment being placed when I pressed escape before the curve is closed. 

Really impressed by this repository though and plan to continue pulling from the master while working my fork as the desire strikes me.

phkahler:
> I would much rather select the tool a once and have it be active until I select another tool or cancel out of it. I know you don't want to be other drafting programs, but I think there is a good reason that virtually everyone of them keeps a tool active until unselected.

@ruevs would it makes sense to put that behavior under a configuration checkbox?

>  "I also found it solved my annoyance with the last line segment being placed when I pressed escape before the curve is closed."

That is a normal behavior that I think is a bug. I agree that ESC should not place the last point. When I got tired of deleting that last point, I though it might be good to have right mouse button place the last point and not start a new one - tried it and OMG that was already a feature. But I agree that ESC should not place an extra point. You could make a PR for that separately and I'd accept it ;-)  Does it cancel other operations too? I see that hitting ESC while drawing a rectangle also places the point rather than escaping from anything.

ruevs:
> I agree that ESC should not place the last point.

That is anti-intuitive indeed. Perhaps no one will cry if we change it.

As for making this PR an option - [I hate options](https://www.youtube.com/watch?v=Xzat1_2jxbA) ;-)

dustinhartlyn:
My two cents is to try using it... or have a armature user test it out (they don't have a particular pattern engrained). I can push the most recent iteration that has the bug fix which caused it to fail GitHub's test and also has the cancel-action feature. I'm still new to GitHub though so I just want to make sure that sending a new pull request is the preferred next step. I can sent a pull request to one of the forks to be evaluated if that is better.

rafo:
Another way could be that the user actively decides that a tool remains permanently switched on. For example, by clicking on an icon with ALT+click. The same would also work for the key combination, so that ALT+R can be used to permanently create rectangles.

Alternatively, a right mouse click on the icon in the toolbar could also work.

All suggestions have in common that somewhere it has to be indicated that you are in “permanent mode”. Work plane (colour, line type), coordinate cross, mouse icon or attribute browser would be suitable for this.

In the end, this is a pretty default behaviour for most CAD systems (and many other software solutions) and doesn't need additional options.

PR #1226, opened 2022-03-15
Export selection and slight update to right-click operation

dustinhartlyn:
I have continued to update my fork. As noted in pull request #1224, tools stay active until canceled. A cancel operation (escape, spacebar or right-click) Will first cancel any current drawing action, if none it will unselect the selected tool. This update allows right click to unselect a tool before brining up right-click menu. 

The second addition is an option to export only the currently selected lines  (File->Export 2d Selection). I find this valuable because I often want to export only a portion of my design to a dxf. I know its possible to move those lines to a specific layer(s), but I think it tedious to move lines from group to group depending on what I am exporting. I find the flow much better when I can simply select what I want and export it regardless of groups. 

As before, I offer these alterations only because I appreciate what others have done for this repository.  Not offended if they are rejected because I intend to keep developing my fork in the ways I think best.


phkahler:
I'd appreciate a few things with PRs. Please provide one feature per PR so we can easily review and merge/reject them. From looking at the list I see:
1) fix a dialog on Windows
2) a zoom to mouse function - which probably fixes the recent breakage
3) escape cancels the last drawing operation rather than completing it
4) right click behavior that I don't think we want
5) the main behavioral change in tool use

Assuming 1-3 work as expected and are PRs with 1 commit each, I'd accept them. It would then be easier to evaluate 4 and 5 independently and accept, reject, improve, or offer feedback.

Looking at "Update gui win" I see "Export selection and slight update to right-click operation" but the export selection feature doesn't even appear to be part of that commit. We try to keep a fairly clean git history, so comments should match commits. It's not clear what the right-click change does, or why its Windows specific. It's things like this that drive the need to do one PR per feature (even if the feature is more than one commit).

dustinhartlyn:
I believe you have already accepted the pull request for the window fix and the fix for the zoom. I am still new to Git so I am sorry if my pull request was messy.  The referenced pull request contains the cancel-operation behavior and to switch to tools being used until canceled. I can look into separating the cancel behavior out but right now I am getting into the task of adding line offsets / contours so it might not be immediate.

phkahler:
@dustinhartlyn  Oh yes, i accepted a previous PR. Since you're new to git I have some suggested things to learn. Of course the first is to learn vim, since git commands are sometimes going to dump you into vi to edit some text. I learned vim by running "vim tutor" every day for a week and learning the first few sections. (or maybe you're not new to vi).

Assuming you forked solvespace on github and then cloned it to your local repository, you should have the ability to reference "origin" as your github repo. You'll also want to setup "source" to point at the solvespace/solvespace repositiory. I don't recall how to do that, but once done you can checkout your master branch and type "git pull upstream master" to pull all the changes from our master then "git push origin master" to push them up to your github master.

If your changes are on a branch and you've updated your local master to match ours, you can switch to your branch and do "git rebase master" which should bring your branch history up to date and not show it differing from master by so many commits. Then you can push it to the branch on your origin and the PR will update as well.

Squashing commits is done with "git rebase -i HEAD~X" where X is the number of commits you want to rearrange. You'll need to read some instructions on rebasing and squashing.

I hope some of this helps. We can use more contributors to Solvespace ;-)

PR #1235, opened 2022-04-30
German translation completed

carelinus:
Remaining strings translated and some minor fixes.


phkahler:
@carelinus  Thank you!   If there's one thing I will never be able to do myself, it's translations ;-)

carelinus:
> @carelinus Thank you! If there's one thing I will never be able to do myself, it's translations ;-)

You're welcome! :-)

PR #1242, opened 2022-05-13
Update CMakeLists.txt

ktrace:
Remove duplicated

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,9 +185,6 @@ endif()
 message(STATUS "Using in-tree libdxfrw")
 add_subdirectory(extlib/libdxfrw)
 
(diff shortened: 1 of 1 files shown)


rpavlik:
I hit some issues regarding this in packaging, fixed it a little differently in #1243

rpavlik:
Looks like we do actually need SparseCore in solvespace.h because the insides of System require Eigen::SparseMatrix (and Eigen::VectorXd). However, that include is in fact the minimal include to get those types.

phkahler:
So close this in favor of #1243 then?

ktrace:
> So close this in favor of #1243 then?

Yes, of cause. #1243 solve problem better.

PR #1243, opened 2022-05-13
Eigen includes are needed in more places, etc.

rpavlik:
Also remove erroneous redundant extlib/eigen mentions, and fix lintian warnings about the desktop files, etc.

Found when packaging rc1 for debian.


res/freedesktop/solvespace-snap.desktop
```diff
@@ -3,7 +3,7 @@ Version=1.0
 Name=SolveSpace
 Comment=A parametric 2d/3d CAD
 Exec=solvespace
-MimeType=application/x-solvespace
+MimeType=application/x-solvespace %f
```
ktrace, line 6: I don't think that mimetype need parameter, look like it need in line 5

rpavlik: Gah good catch

rpavlik: Fixed!

rpavlik:
OK, fixed my goof from earlier

phkahler:
@rpavlik  you can merge this when you're happy with it.

PR #1248, opened 2022-05-24
C# sample project

TrevorDArcyEvans:
* conversion of VB.NET sample to C#
* factored out SolveSpace functionality to a separate file & namespace
* C# project


phkahler:
Does this actually work? I never noticed the VB.NET one but C# seems like an even better idea.

TrevorDArcyEvans:
@phkahler Does it work?  LOL!  It definitely works ie outputs something to console and doesn't crash.

exposed/CsDemo/.gitignore
```diff
@@ -0,0 +1,369 @@
+## Ignore Visual Studio temporary files, build results, and
```
ruevs, line 1: Adding such a huge .gitignore seems like bad form to me. One that ignores only the specific files and directories created by VS when building the example is (in my opinion) better. Currently I do not have the C# build tools in my VS installation but I will add them (when I get to my development system) and try the example.

TrevorDArcyEvans: It is a 'standard' .gitignore from [here](https://github.com/github/gitignore/blob/main/VisualStudio.gitignore)

ruevs: That is clear from the comment at the top.

vespakoen:
Please take a look at the new C library, especially `ExampleStateful` in CDemo.c, it probably makes `Solver.cs` a lot simpler and you just have to wrap the methods and the Slvs_Entity / Slvs_SolveResult.

Also you might want to use `LibraryImport` over `DllImport` in C#.

Issue #1210, opened 2022-02-05
Change function names (quaternions)


phkahler:
This would be strictly a change of naming convention in the code.

Quaternions currently have 3 functions called RotationU, RotationV, and RotationN.

What would you expect they do based on the names? What if I then told you the return a vector? Now what you think they do? There are not comments in the code documenting these (adding some might also be a solution)

A quaternion can be used to apply a rotation to a vector or another quaternion. They can also be used as a compact representation of a coordinate systems orientation. As such, these 3 functions can be used to extract 3 basis vectors from a quaternion. RotationU returns the U axis (I think x) as is the vector (1,0,0) were rotated by the quaternion. Similar for V and N (y and z). With this understanding, the naming makes more sense. The problem is that the naming is not explanitory.

I'm wanting to use BasisU in place of RotationU, but I'm not sure what others think about that. Other suggestions are welcome too. I just dont like the current situation.

Something similar is done with "normals" which are stored as quaternions, and have similar functions with different names that just call the quaternion functions.

Thoughts on this?

Symbian9:
> A quaternion can be used to apply a rotation to a vector or another quaternion.

Some related reference info:

1. Quaternions — https://liorsinai.github.io/mathematics/2021/11/05/quaternion-1-intro.html
1.1 https://news.ycombinator.com/item?id=29510237
2. Rotations with Quaternions - https://imadr.github.io/rotations-with-quaternions/
2.1. https://news.ycombinator.com/item?id=27353263
3. Quaternions: The Strange Numbers That Birthed Modern Algebra - https://www.quantamagazine.org/the-strange-numbers-that-birthed-modern-algebra-20180906/
3.1. https://news.ycombinator.com/item?id=23966755
4. Understanding Quaternions - https://www.3dgep.com/understanding-quaternions/
4.1. https://news.ycombinator.com/item?id=18265355
5. How Quaternions encode rotations – derivation and sample code - http://loopspace.mathforge.org/HowDidIDoThat/Codea/Quaternions/
5.1. https://news.ycombinator.com/item?id=12441656
6. Let's remove Quaternions from every 3D Engine - https://marctenbosch.com/quaternions/
6.1 https://news.ycombinator.com/item?id=29512302

hinell:
Refactoring.

phkahler:
I'm leaning toward VectorU()  VectorV() and VectorN() 

[...]

PR #1250, opened 2022-06-01
Use nullptr instead of NULL 

pap1rana:
I saw that you are developing with std=c++11, and using [nullptr](https://en.cppreference.com/w/cpp/language/nullptr) is recommended.
I excuted cmake
```sh
cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_OPENMP=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=ON .. 
```
with extra flag for  `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON`  clang-tidy purpose.
From build directory executed:
```sh
run-clang-tidy -checks='modernize-use-nullptr'
```
modernize-use-nullptr
After that i ran same command with -f to auto apply fixes:
diffs
After that i built everything again, ran app with no problems:
ApplyfixTestingProg
I even have executed `solvespace-testsuite`:
TestPassed

This was done for [issue clang-tidy](https://gitlab.com/pap1rana/azrs-tracking/-/issues/8) for my faculty project.

### DIFF
--- a/src/bsp.cpp
+++ b/src/bsp.cpp
@@ -23,8 +23,8 @@ SBsp3 *SBsp3::FromMesh(const SMesh *m) {
         swap(mc.l[k], mc.l[n]);
     }
 
-    SBsp3 *bsp3 = NULL;
-    for(auto &elt : mc.l) { bsp3 = InsertOrCreate(bsp3, &elt, NULL); }
+    SBsp3 *bsp3 = nullptr;
+    for(auto &elt : mc.l) { bsp3 = InsertOrCreate(bsp3, &elt, nullptr); }
     mc.Clear();
     return bsp3;
 }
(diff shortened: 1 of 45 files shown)


ruevs:
@pap1rana thank you for the contribution. While `nullptr` is the correct thing to use since C++11 SolveSpace is from 2008 and thus uses `NULL` in most places.

Globally replacing `NULL` with `nullptr` at this point simply introduces "noise" in the GIT log by changing all file in a single commit. Therefore I can not accept your pull request. Just in case - @phkahler if do not agree with me just re-open it.

phkahler:
@ruevs  I agree. OTOH we should probably make a list of big changes like that, and do them all in a row? For example, I opened #1210  which is just renaming.  Another issue is that such changes may also cause existing WIP patches and branches to not automatically merge any more.  OTOH there may never be a "good" time to do this. So no for now.

pap1rana:
I agree with you, this should be done as part of big changes list.

kk6mrp:
You can ignore commits with Git Blame like here: https://akrabat.com/ignoring-revisions-with-git-blame/

It is only applied on a user by user basis though.

rpavlik:
Github now has support for the git blame ignore file.

kk6mrp:
You can ignore changes like this with Git blame as described here: https://akrabat.com/ignoring-revisions-with-git-blame/

The bad part is that you need to apply on a user by user basis, but that could be documented.

PR #1252, opened 2022-06-06
WIP First try at faces perpendicular constraint

phkahler:
This is WIP. It works for making 90 degree revolves, and also in assembly groups!  Looking for feedback.
Some of my comments:

I don't like all the checking IsFace() that needed to be added here. OTOH by doing that there was almost nothing needed to add the constraint. This should also be applicable to a new face-face-angle constraint.

The little purple perpendicular constraint on the sketch is drawn OK I suppose? If we made a new constraint type we'd have to add more code to draw it. not sure how to display face-face-angle anyway.

Message box text is not updated to indicate face-face as an option for perpendicular constraint.

The additional cases in describescreen.cpp are needed independently of this feature - they were missing. I'll make a different PR for those and maybe some other missing cases in there...

### DIFF
--- a/src/constraint.cpp
+++ b/src/constraint.cpp
@@ -721,7 +721,11 @@ void Constraint::MenuConstrain(Command id) {
         }
 
         case Command::PARALLEL:
-            if(gs.vectors == 2 && gs.n == 2) {
+            if(gs.faces == 2 && gs.n == 2) {
+                c.type = Type::PARALLEL;
+                c.entityA = gs.face[0];
+                c.entityB = gs.face[1];            
+            } else if(gs.vectors == 2 && gs.n == 2) {
                 c.type = Type::PARALLEL;
                 c.entityA = gs.vector[0];
                 c.entityB = gs.vector[1];
@@ -765,6 +769,7 @@ void Constraint::MenuConstrain(Command id) {
             } else {
                 Error(_("Bad selection for parallel / tangent constraint. This "
                         "constraint can apply to:\n\n"
+                        "    * two faces\n"
                         "    * two line segments (parallel)\n"
                         "    * a line segment and a normal (parallel)\n"
                         "    * two normals (parallel)\n"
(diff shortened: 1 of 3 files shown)


phkahler:
This is accidentally allowing a line segment and a face to be constrained perpendicular, which makes the line segment parallel to the face (perpendicular to the face normal). This is probably because a face is returning true for HasVector which is allowed by a different condition than 2 faces.

jwesthues:
That makes sense to me. The face-line behavior seems confusing but geometrically fundamental, worth documenting and maybe warning about but not worth trying to fix.

phkahler:
> That makes sense to me. The face-line behavior seems confusing but geometrically fundamental, worth documenting and maybe warning about but not worth trying to fix.

Maybe checking if 1 "vector" and one face are selected we could actually create a parallel constraint instead. That would behave the way a person thinks about lines perpendicular to a face, but it would be weird to see a parallel constraint in the text window when the user created a perp [ constraint. Or I could disallow it for cases of only 1 face entity among the 2 selections.

jwesthues:
I'd have no objection to making face-nonface behavior less surprising (swap parallel and perpendicular as you say, and also make the angle dimension show the complement I guess). I'd probably be too lazy to do it myself, though.

phkahler:
@ruevs  are you saying this allows parallel constraints with faces? That was not even intended. That's a problem with making HasVector() return true for faces. They will be usable everywhere lines are, and that's not at all what I wanted. I suppose I didn't even need to add 2 faces as an option to the perpendicular code. I started with the [ code and ended up having to add the other changes as a result of that.  I'll look into making this work without advertising faces as having a vector. I want it to behave the way a normal person thinks, not the way a mathematician thinks ;-)

It was really nice to get something working with so little code though.

phkahler:
It works fine if we don't advertise faces as having vectors in HasVector(). I explicitly added the parallel faces constraint as well and updates the message boxes.

phkahler:
Not happy about the post-squash commit message though ;-)

phkahler:
closing in favor of#1254 and #1255.

Issue #29, opened 2016-08-08
Generalized constraint creation


mmiscool:
Would be useful to display a list of valid constraint types for the selected objects.

Desired workflow for creating a constraint.

Click "Constraint" command --> Select items to be constrained --> Select from a list in the "property browser window" of possible valid constraints. 

As more items get selected the list would get smaller. For example I could select 2 lines. It would show all the constraints I could do with 2 lines. As soon as I click on a point. It shows things I could do with 2 lines and a point.

mmiscool:
Would also be good to show a list of invalid constraints and what is missing to make it valid. 

For example.

phkahler:
No. Selecting a general constraint command, then selecting entities, then a specific constraint? That first command is almost completely redundant.

As it is, you just select entities and apply a specific constraint.

@ruevs I think we should close this.

ruevs:
@phkahler the part where, once one selects some entities a list of the possible constraints that can be applied to them is shown, could be useful for a complete novice. We could do it in this "beautiful" function - it will just get slightly longer ;-) https://github.com/solvespace/solvespace/blob/b5333608e9d4826fcb6dc497a9e94e162267a84c/src/describescreen.cpp#L64

Anyway, just like you I find this superfluous.

PR #1258, opened 2022-06-19
Add ability to select 3 faces

phkahler:
This allows the user to select 3 faces vs the old limit of 2.  I would like to allow constraining a point to more than one face at a time, and 3 would be the maximum. This is a prerequisite to that.

The remaining problem is that the 3rd selected face is not shown as selected. The selection code indicates that the limit of 2 was for rendering performance reasons, which may or may not be a problem today. I'm not sure where the relevant rendering code is to show the selected faces. Help would be appreciated but this is a low priority. Might be a good way for someone to dig into more obscure parts of the code.

### DIFF
--- a/src/describescreen.cpp
+++ b/src/describescreen.cpp
@@ -428,6 +428,26 @@ void TextWindow::DescribeSelection() {
             double d = (p1.Minus(p0)).Dot(n0);
             Printf(true,  "      distance = %Fi%s", SS.MmToString(d).c_str());
         }
+    } else if(gs.n == 3 && gs.faces == 3) {
+        Printf(false, "%FtTHREE PLANE FACES");
+
+        Vector n0 = SK.GetEntity(gs.face[0])->FaceGetNormalNum();
+        Printf(true,  " planeA normal = " PT_AS_NUM, CO(n0));
+        Vector p0 = SK.GetEntity(gs.face[0])->FaceGetPointNum();
+        Printf(false, "   planeA thru = " PT_AS_STR, COSTR(SK.GetEntity(gs.face[0]), p0));
+
+        Vector n1 = SK.GetEntity(gs.face[1])->FaceGetNormalNum();
+        Printf(true,  " planeB normal = " PT_AS_NUM, CO(n1));
+        Vector p1 = SK.GetEntity(gs.face[1])->FaceGetPointNum();
+        Printf(false, "   planeB thru = " PT_AS_STR, COSTR(SK.GetEntity(gs.face[1]), p1));
+
+        Vector n2 = SK.GetEntity(gs.face[2])->FaceGetNormalNum();
+        Printf(true,  " planeC normal = " PT_AS_NUM, CO(n2));
+        Vector p2 = SK.GetEntity(gs.face[2])->FaceGetPointNum();
+        Printf(false, "   planeB thru = " PT_AS_STR, COSTR(SK.GetEntity(gs.face[2]), p2));
+
+        // We should probably compute and show the intersection point if there is onw.
+
     } else if(gs.n == 0 && gs.constraints == 1) {
         Constraint *c = SK.GetConstraint(gs.constraint[0]);
         const std::string &desc = c->DescriptionString().c_str();
--- a/src/describescreen.cpp
+++ b/src/describescreen.cpp
@@ -64,11 +64,6 @@ void TextWindow::ScreenConstraintShowAsRadius(int link, uint32_t v) {
 void TextWindow::DescribeSelection() {
     Printf(false, "");
 
-    auto const &gs = SS.GW.gs;
-    if(gs.n == 1 && (gs.points == 1 || gs.entities == 1)) {
-        Entity *e = SK.GetEntity(gs.points == 1 ? gs.point[0] : gs.entity[0]);
-        Vector p;
-
 #define COSTR_NO_LINK(p) \
     SS.MmToString((p).x).c_str(), \
     SS.MmToString((p).y).c_str(), \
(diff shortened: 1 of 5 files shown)


phkahler:
@Evil-Spirit do you know where/how selected faces get rendered vs not-selected?

ruevs:
@phkahler there you go ;-)

How would the "point on three faces" constraint be stored? A completely new type?

We should not merge this before it can do something useful - it would be rather confusing.

phkahler:
> @phkahler there you go ;-)
> 
> How would the "point on three faces" constraint be stored? A completely new type?

No. We shall create 3 point-on-face constraints. I've been pondering how to do the variadic constraints request and looking at the neglected PR on the topic.

I'm not sure why whitequark didnt accept it. It isnt really conforming to the style, nor is the existing style really great. I think creating multiple constraints inline would be fine in most cases vs creating a list of them.

ruevs:
OK understood - select one to three faces, hit `O` - get one to thee constraints automatically generated. Makes sense - of course then there would be three to delete but that could legitimately be called a feature.

In the meantime the face selection code was too ugly for me - I cleaned it up a bit.

phkahler:
I'd like to just merge it now.

>> There are two PR-s on the topic https://github.com/solvespace/solvespace/pull/530 and https://github.com/solvespace/solvespace/pull/836 which one are you pondering? ;-)

Well, #530 is a little odd in making a list and then creating them all at once. #836 was designed for a specific use-case and assumes pairs of things to constrain (not what we want) but I think the style of the code is more in line with how I'd like it. So I was planning to just start rewrite the code for each type of constraint over again. If anyone wants to give that a go, I'm happy to merge PRs for these.

ruevs:
> I'd like to just merge it now.

Please don't.

If for no other reason, than for the awful "loop unrolling" in describesreen. And even if we were to merge it, we have to set `MAX_SELECTABLE_FACES` to 2 until selecting three does something (but then the commit messages make no sense and need to be changed).

In general a lot of code in constraint and describescreen fells like "copy-paste, but not quite" - it annoys me every time I look at it. I've dealt with this pattern in the past and have some ideas. But there is A LOT of code and logic, and thus huge opportunity for regressions in both files. And the test suite covers none of it :-( So it will be a lot of work to plan, refactor and test.

And in my opinion neither #530 nor #836 nor your planned  "a point and one to three faces coincident" goes in before such a clean up.

src/describescreen.cpp
```diff
@@ -428,6 +428,26 @@ void TextWindow::DescribeSelection() {
             double d = (p1.Minus(p0)).Dot(n0);
             Printf(true,  "      distance = %Fi%s", SS.MmToString(d).c_str());
         }
+    } else if(gs.n == 3 && gs.faces == 3) {
+        Printf(false, "%FtTHREE PLANE FACES");
+
+        Vector n0 = SK.GetEntity(gs.face[0])->FaceGetNormalNum();
```
ruevs, line 434: What is wrong with a for loop?

src/describescreen.cpp
```diff
@@ -428,6 +428,26 @@ void TextWindow::DescribeSelection() {
...
+
+        Vector n1 = SK.GetEntity(gs.face[1])->FaceGetNormalNum();
+        Printf(true,  " planeB normal = " PT_AS_NUM, CO(n1));
+        Vector p1 = SK.GetEntity(gs.face[1])->FaceGetPointNum();
+        Printf(false, "   planeB thru = " PT_AS_STR, COSTR(SK.GetEntity(gs.face[1]), p1));
+
+        Vector n2 = SK.GetEntity(gs.face[2])->FaceGetNormalNum();
+        Printf(true,  " planeC normal = " PT_AS_NUM, CO(n2));
+        Vector p2 = SK.GetEntity(gs.face[2])->FaceGetPointNum();
+        Printf(false, "   planeB thru = " PT_AS_STR, COSTR(SK.GetEntity(gs.face[2]), p2));
```
ruevs, line 447: A [C++11 style range-based for loop.](https://en.cppreference.com/w/cpp/language/range-for)

ruevs, line 447: See, copy-pasta already bit you. (`planeB`) ;-)

phkahler:
> In general a lot of code in constraint and describescreen fells like "copy-paste, but not quite" - [it annoys me](https://github.com/solvespace/solvespace/issues/29#issuecomment-1175295702) every time I look at it.

Agreed, but it is fairly readable. Many areas of solvespace look like this often because the failure to use polymorphism, but I can understand and sometimes appreciate the simplicity of it.

One thought I had on the constraints was to make a table indicating what constraint type could be applied to which combinations of entities, but I haven't really dug into that idea enough to like it or not. That might even simplify the code for dialogs that pop up if you misapply a constraint. The complexity will never go away, so it's a question of how to embody it in code so people can read and understand it.

phkahler:
Allowing multiple point-on-plane constraints at once should be simple by modifying this:
```
            } else if(gs.points == 1 && gs.workplanes == 1 && gs.n == 2) {
                c.type = Type::PT_IN_PLANE;
                c.ptA = gs.point[0];
                c.entityA = gs.entity[0];
            } else if(gs.points == 1 && gs.lineSegments == 1 && gs.n == 2) {
```
The condition changes to:
```
            } else if(gs.points == 1 && gs.workplanes > 0 && gs.n == (gs.workplanes+1) ) {
```
Then just do the body of the condition in a loop over all the workplanes including the "AddConstraint(&c);" line from further down.

Then we either have to add a "break" when done, or move the "AddConstraint(&c);" to all the cases within ON_ENTITY. Neither of those seems desirable, but the other option seems to be to make a list of constraints like PR #530 and loop over them where "AddConstraint()" is called now.

@ruevs I don't see a cleaner way to do this. Please fill me in on what you think it should look like. Maybe @rpavlik can recommend something too, he seems to appreciate clean C++ .

phkahler:
@ruevs  I want to push this in. I made a 3.2 milestone and I intend to do variadic constrain equal length and diameter, as well as probably 1-3 point-on-face constraints. However, my local and personal repos on github are quite broken so I want to re-fork and I don't want to lose these changes since they work and are not that big.

ruevs:
@phkahler I rebased on master, fixed the copy-pasta in describescreen.cpp and also pushed the branch to my fork here: https://github.com/ruevs/solvespace/tree/wip_phkahler_three-faces
so it is safe.

If you want I can also push the branch to the main repository, but in my opinion it is unwise to merge it into master- after all one can select three faces but can not do anything with them for now.

P.S.  I also pushed to your branch.

src/describescreen.cpp
ruevs: I "fixed" this in https://github.com/solvespace/solvespace/pull/1258/commits/c59e93c3e37a0b3e07d9346f4087fa2a7264798d

src/describescreen.cpp
ruevs: I fixed his in https://github.com/solvespace/solvespace/pull/1258/commits/c59e93c3e37a0b3e07d9346f4087fa2a7264798d

phkahler:
@ruevs  I rebased this on master after merging variadic constraints. Then I added a commit here to allow constraining a point to multiple faces at once. This is very handy for getting a point on a vertex that is a result of boolean operations where no point gets created.

ruevs:
> @ruevs I rebased this on master after merging variadic constraints. Then I added a commit here to allow constraining a point to multiple faces at once.

Happy New Year @phkahler. This looks reasonable, but I have not tried it. Tomorrow I'll try to take a look at (and test) this and all the "multi-constraint" changes from the last few days.

phkahler:
> > @ruevs I rebased this on master after merging variadic constraints. Then I added a commit here to allow constraining a point to multiple faces at once.
> 
> Happy New Year @phkahler. This looks reasonable, but I have not tried it. Tomorrow I'll try to take a look at (and test) this and all the "multi-constraint" changes from the last few days.

In that case I'll just merge it. I tested it on 2 and 3 faces. Merging multi-constraints broke a bunch of stuff and needed fixing on master. It's slightly messy but I'd rather just do testing with master since it's still changing.

PR #1260, opened 2022-06-24
czech translations

strzinek:


strzinek:
All translated texts edited by human.
I noticed that Property Browser window contents are not prepared for localization, which is a pitty. Is there a plan to make it localisable?

phkahler:
@strzinek  The property browser needs work. The other big issue is that the text can not be resized. Unfortunately there is no plan to address either of these issues, and both are likely to require a lot of code changes and probably a new dependency or use of a common toolkit across OSes.

ruevs (commented):
Looks good - to the extent I superficially looked at a few random places and the little I decipher from Czech.

Issue #1261, opened 2022-06-24
32bit SolveSpace crashes when it runs out of memory (was SolveSpace 3.1 is crashing so frequently without any warning/error on assembling parts.)


simrim1:
System information

SolveSpace version: 3.1~70bde63c
Operating system: Windows 10

Expected behavior

I am designing a CNC machine in SolveSpace and trying to combine (assemble) different parts (SolveSpace sub assembly files) in a separate file which has its own drawing as well. 
It should work as normal and should not crash.

Actual behavior

It keeps crashing and has become almost unusable.

Additional information

The files sub assemblies which further consists of sub assemblies are larger than usual files. The final assembly has become of size 45MB. It looks like memory leak issue. The application crashes silently without showing any error.

ruevs:
@simrim1 would you be willing to share your model? If not publicly et least privately?
Does it crash with 3.0?
Are you by any chance the author of this video https://www.youtube.com/watch?v=iuBfzpcCO5c ?

phkahler:
I'm wondering if this is a new issue or if it's the size of the sketch/assembly running into a limitation.

We have seen cases where the performance increases allowed people to hit limitations before.

So yeah, is this unique to version 3.1 or does it happen with 3.0 as well? Assuming the files are compatible (no 3.1 specific features).

simrim1:
@ruevs Spot on man. Your searching capabilities are better than Google. I am really curious how you related my post to my video on Youtube.
Coming back to the issue, I have not tested it in 3.0 and I don't think I can because I have made all my designs in 3.1 taking benefit of better Helix feature in 3.1.

I can send you files privately. I am going to public them all in my next 2-3 videos. I am keeping them private for the time being. Please let me know how to send you files.

@phkahler I guess this is memory leak issue when it comes to assembly but I am not sure. In the final assembly when we have done 2-4 operations doesn't matter it is saving the file, moving something with mouse. applying any constraint or something else which needs a geometry calculation, it crashes.

ruevs:
> Please let me know how to send you files.

Email "dpr" at the domain you can see in my profile.

ruevs:
@simrim1 First thank you for sending me your CNC lathe/router model! Very impressive job! It is by far the most complicated SolveSpace project I have seen.

[...]

PR #1262, opened 2022-06-26
Win32: Allow 32 bit SolveSpace to access up to 4GB of RAM.

ruevs:
Link 32 bit SolveSpace for Windows with /LARGEADDRESSAWARE which allows
it to access up to 3GB of RAM on a properly configured 32 bit Windows and
up to 4GB on 64 bit.

See: https://msdn.microsoft.com/en-us/library/aa366778
https://docs.microsoft.com/en-us/cpp/build/reference/largeaddressaware-handle-large-addresses
https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#characteristics
https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2003/cc786709
https://docs.microsoft.com/en-us/windows/win32/memory/4-gigabyte-tuning

Fixes: #1261

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,10 @@ endif()
 if(MINGW)
     set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} -static-libgcc")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -static-libgcc -static-libstdc++")
+    # Link 32 bit SolveSpace with --large-address-aware which allows it to access
+    # up to 3GB on a properly configured 32 bit Windows and up to 4GB on 64 bit.
+    # See https://msdn.microsoft.com/en-us/library/aa366778
+    set(CMAKE_EXE_LINKER_FLAGS "-Wl,--large-address-aware")
 endif()
 
 # Ensure that all platforms use 64-bit IEEE floating point operations for consistency;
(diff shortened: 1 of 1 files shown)


ruevs:
@rpavlik @ppd would you please take a look at this since it affects CMake and I am not exactly "proficient" in it :-)

rpavlik:
Looks ok (though it would be better to use target_compile_options on the targets directly)

ppd:
Looks good and is consistent with the rest of the cMake hodgepodge :)
The only issue I have is that the comment is the same for both toolchains: 
> Link 32 bit SolveSpace with /LARGEADDRESSAWARE

For MinGW, it links with the equivalent to /LARGEADRESSAWARE, but not exactly with it. That'd be my only nitpick.

ruevs:
> it would be better to use target_compile_options on the targets directly

@rpavlik If I only knew what "target_compile_options" were in the context of cmake :-) Unfortunately I have no time to learn it.


> The only issue I have is that the comment is the same for both toolchains:

@ppd Good catch - I fixed it and force pushed.

ruevs:
> I'm more concerned about what Solvespace models you're using that need that much memory 🤣

@rpavlik not me :-) But it is pretty impressive what people are doing with the "new fast SolveSpace" :-)
See here https://github.com/solvespace/solvespace/issues/1261

phkahler:
BTW this doesn't actually "fix" the issue, it just pushes it out a bit further ;-)  "Use 64bit Solvespace" is a nice workaround.

PR #1267, opened 2022-07-08
Win32, MSVC: Enable Multi-processor Compilation (/MP) with Visual Studio

ruevs:
Makes compiling from the Visual Studio IDE much faster when using the solution and projects generated by cmake.

For the externals I hijack the `disable_warnings` function.


rpavlik (approved):
Looks good, this one is easy to forget to do. I should probably check a few other projects I work on...

ruevs:
I've been doing it "by hand" from time to time in the generated VS projects but recently switched too many build types and branches and got annoyed :-)

PR #1270, opened 2022-07-11
snap: Port to core22

ppd:
- Migration according to https://forum.snapcraft.io/t/micro-howto-migrate-from-core20-to-core22/30188
- Also declare all snaps stable

What's left?
- [x] Wait for the gnome extension to appear in a stable snapcraft release (very soon)
- [x] Testing


ppd:
Everything seems to be working; now's a good time to test this in `edge`, I think.

PR #1273, opened 2022-07-27
ci: re-enable arm64 snap builds via remote-build

ppd:
Use snapcraft's remote-build feature to outsource building
on arm64 to Launchpad's build servers.

This needs a new secret: LAUNCHPAD_CREDENTIALS
It can be sourced from a local snapcraft installation that has
performed a remote build, or a Launchpad login, before:
~/.local/share/snapcraft/launchpad/credentials

Also fix a problem related to file permissions in override-pull
and declare all snaps stable, regardless of whether they are built
from tagged commits.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -142,6 +142,46 @@ jobs:
         snap: ${{ steps.build.outputs.snap }}
         release: edge,beta
 
+  deploy_snap_arm64:
+    needs: [test_ubuntu, test_windows, test_macos]
+    concurrency: launchpad-remote-build
+    name: Deploy ARM64 Snap
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install snapcraft
+      run: sudo snap install --classic snapcraft
+    - name: Set up Launchpad credentials & git
+      env:
+        LAUNCHPAD_CREDENTIALS: ${{ secrets.LAUNCHPAD_CREDENTIALS }}
+      run: |
+        mkdir -p  ~/.local/share/snapcraft/provider/launchpad
+        echo "$LAUNCHPAD_CREDENTIALS" > ~/.local/share/snapcraft/provider/launchpad/credentials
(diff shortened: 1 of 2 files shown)


ppd:
We should probably register a separate Launchpad account if we decide to go that route.

ppd:
@phkahler This strategy appears to be working well for the official KDE snaps: https://invent.kde.org/packaging/snapcraft-kde-applications/-/jobs/423670

So, (a) Do you want this at all? and (b) Do you want me to register a solvespace Launchpad account for this? Best would be to register with an email that you control. We could do your gmail with a '+launchpad-solvespace' suffix (https://support.google.com/a/users/answer/9308648?hl=en).

ppd:
This is the old SolveSpace launchpad account, registered by whitequark: https://launchpad.net/~solvespace
Maybe we can re-use this one.

ruevs:
@ppd @phkahler @vespakoen  Since this is the currently open CI issue I'll write here to avoid extra "noise". Sorry for the unrelated topic and spamming you all.

The macOS build is failing (again) https://github.com/solvespace/solvespace/runs/7938462623?check_suite_focus=true :

> 2022-08-21 11:39:11.482 altool[22875:73353] *** Error: code -1013 (Failed to get notarization info. You must specify a uuid)'
Error: Process completed with exit code 1.

phkahler:
> @phkahler This strategy appears to be working well for the official KDE snaps: https://invent.kde.org/packaging/snapcraft-kde-applications/-/jobs/423670
> 
> So, (a) Do you want this at all? and (b) Do you want me to register a solvespace Launchpad account for this? Best would be to register with an email that you control. We could do your gmail with a '+launchpad-solvespace' suffix (https://support.google.com/a/users/answer/9308648?hl=en).

@ppd if you continue to handle snaps, just use my email as you suggest and let me know when to do what. If not, I'll probably end up disabling snap builds or something just to make the noise go away. IMHO these things (snaps, flatpaks, etc...) are optional and a way to push package management upstream, which I want no part of.

PR #1274, opened 2022-07-27
mac: Distinguish trackpad from Magic Mouse scroll

nanoant:
This is the second attempt to distinguish trackpad scrolling that when used
should yield panning from Magic Mouse scrolling that should control zoom.

Since trackpad will begin with a touch prior to the scroll event, but Magic
Mouse not, we can use it to make sure we trigger panning only on a trackpad.

Previous "mac: Don't interpret single-touch scroll events as pan gestures"
that was flawed was reverted which ultimately lead to being unable to zoom using
Magic Mouse.

This fixes again #1107 and is also related to #1218

### DIFF
--- a/src/platform/guimac.mm
+++ b/src/platform/guimac.mm
@@ -372,6 +372,8 @@ MenuBarRef GetOrCreateMainMenu(bool *unique) {
     double             rotationGestureCurrent;
     Point2d            trackpadPositionShift;
     bool               inTrackpadScrollGesture;
+    int                activeTrackpadTouches;
+    bool               scrollFromTrackpadTouch;
     Platform::Window::Kind kind;
 }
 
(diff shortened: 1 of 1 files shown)


ruevs:
Just for cross-reference - the original issue is here https://github.com/solvespace/solvespace/issues/1106

nanoant:
@vespakoen Thanks for the report. I managed to reproduce and then fix the problem. Can you pull the latest update.

TL;DR Actually we can have multiple touches when we accidentally touch trackpad with another finger when scrolling fast. Using counter instead of simple flag seems to resolve the problem to me and I am no longer able to trigger zoom with scrolling widely on my MBP with 2 fingers, while Magic Mouse on my Mini works great as well.

vespakoen:
This is working for me now as well.

Thanks for working on this and figuring out a way to distinguish between mice ;)

Merging...

PR #1275, opened 2022-08-04
Emscripten: Fix to build with latest toolset, and merge recent commits

verylowfreq:
This PR
 - fixes for able to built with latest emscripten toolset, and
 - merges recent commits in the master branch.

Related PR: https://github.com/solvespace/solvespace/pull/419
This PR resolves the issue https://github.com/solvespace/solvespace/issues/1037

My build is hosted on GitHub Pages ( https://verylowfreq.github.io/experimental-solvespace-on-browser/ ).

## Details:

The changes of Emscripten toolset broke the code introduced in 4d7145a763e5c4198b7a348dbbd2f93ab701b8bf . This PR fixes it. 

My build environment info:
 - OS: Ubuntu Linux 22.04 x86_64 (on WSL2)
 - Emscripten SDK: 26a0dea0d3bbf616fa7f0a908e5b08aab406f7c4 (2022-08-01)
 - emcc: 3.1.18
 - clang: 14.0.0

src/platform/guihtml.cpp is based on, and src/CMakeList.txt is partialy picked up from https://github.com/zopieux/solvespace/tree/emscripten  (Thanks, @zopieux !)

This PR merges the commits from the master branch. The latest commit that merged is bc4244e0990af31ff326a72e2152fed662fca3e9 (2022-07-08).

No functional improvement or fix is included.

PR #1280, opened 2022-08-19
Rebased @verylowfreq Emscripten Fixes

vespakoen:
I rebased @verylowfreq's changes from #1275 to the latest master branch over here, will review this later today.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,6 +186,10 @@ if(APPLE)
     set(CMAKE_FIND_FRAMEWORK LAST)
 endif()
 
+if(EMSCRIPTEN)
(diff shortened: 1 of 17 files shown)


verylowfreq:
@vespakoen Thanks :smile:
I checked this and this is same that I thought to do. This can be built successfully and runs. And thank you for reviewing.

ruevs:
@verylowfreq I think it is better to keep most of the history of your changes when merging in master - it has some interesting details.

verylowfreq:
@ruevs I agree. That would be better.

ruevs:
Equivalent changes merged in master here f7415048a5769c78fc4bb02afa426d101059374d .

PR #1285, opened 2022-08-27
README: Move the check out instructions to one place.

ruevs:


ppd (approved):
Lgtm. The only very minor downside is that a few unnecessary submodules are pulled on Linux, where we default to linking to system libs.

ruevs:
> The only very minor downside is that a few unnecessary submodules are pulled on Linux, where we default to linking to system libs.

Hmmm, @ppd you are right, I did not consider that. In this case I'm ambivalent to this change. Let's see what the others tink.

phkahler:
I'm kind of in favor of it.  It seems OK to pull in and update the submodules even if they're not all built. But we also don't want to build them when not needed right? So long as we still used system installed dependencies when appropriate I think it's fine. It also clarifies (not that anyone cares?) what the dependencies are.

phkahler (approved):

Issue #1288, opened 2022-09-03
Duplicate operations - lathe and revolve


nick87720z:
Just one revolve operation is enough.
There could be switch for angle mode with example values "Lathe" and "Partial". Of course, current way of getting lathe-like result by overrotating revolution works, but it leaves redundant control points, which contribute to clutter, so in my opinion - mode switch should have better usability.

ruevs:
Lathe existed back to version 1.9 at least. Revolve was added later. the reason both exist is backward compatibility.

nick87720z:
Btw, I just realized - no need to change file format. Change to UI would be enough, so no breakages are possible even theoretically.

Yet must admit - I made this proposal only because localized action names for these two are same for me (though icons are fine - hard to mix up).

ruevs:
> localized action names for these two are same for me (though icons are fine - hard to mix up).

Which language?

nick87720z:
Russian

PR #1290, opened 2022-09-04
Enhance russian translation

nick87720z:
Lathe and Revolve words were maped to Rotate during translation.
However, russian has words to not lose meaning.


ruevs:
Токарный станок = lathe - и поетому - "Вы&точить". Понятно.
But I'd like to hear what @Evil-Spirit thinks. I have no clue what the proper terminology is in Russian in the context of mechanical engineering.

ruevs:
Relates to #1288

nick87720z:
Yeah, me too.
Somewhat hard case as I don't remember CADs, using such difference
(only tried autocad, kompas and freecad). "Тело точения" or "Тело
поворота" - sound rather unusual despite valid.

ruevs:
@Evil-Spirit do you have an opinion on this?

Evil-Spirit:
"Тело вращения" [solid of revolution](https://www.multitran.com/m.exe?s=solid+of+revolution&l1=1&l2=2) is more common for both lathe operators and students. This is geometrical definition. "Выточить" - this will cause brain damage to every who are not lathe operator. So, need to revert all this since I don't like this at all (pointless to remame to less understandable in order to support initial "lathe" instead of "revolve" )
https://github.com/solvespace/solvespace/pull/1290/commits/7bf427e1f6cece5bb937c2eaece9155fb9bd71cd

Evil-Spirit:
Hey, I just realized that now there is 3 different operations like:

*   revolve by 360
*   revolve by less than 360
*   revolve by more than 360 and shifing by axis

This all should be just single operation! I believe it is possible to merge them into single group without breaking backward compatibility. but if you want to make translation:

Lathe = Тело вращения

Angular lathe / Lathe by angle = Угловое тело вращения  / Тело вращения на угол

res/locales/ru_RU.po
```diff
@@ -932,11 +932,11 @@ msgstr "Тело Винтовое"
 
 #: graphicswin.cpp:118
 msgid "&Lathe"
-msgstr "Тело В&ращения"
```
Evil-Spirit, line 935: leave this or make it "Тело вращения на 360"

res/locales/ru_RU.po
```diff
@@ -932,11 +932,11 @@ msgstr "Тело Винтовое"
...
+msgstr "Вы&точить"
 
 #: graphicswin.cpp:119
 msgid "Re&volve"
-msgstr "Тело В&ращения"
```
Evil-Spirit, line 939: Тело вращения на угол / Угловое тело вращения

Evil-Spirit (changes requested):

ruevs:
@nick87720z @Evil-Spirit merged here 31ac8083ae27733b18a6c2f6454104b9efd9fc29 with the discussed changes. I'm pretty sure I did not screw up anything but please take a look.

nick87720z:
@Evil-Spirit Just for interest - how about "Lathe" -> "Токарный станок", that was my original idea. Also most verbatim translation, carrying original author idea totally preserved.

ruevs:
> how about "Lathe" -> "Токарный станок",

But in the English version " lathe" is used as a verb - the literal translation is indeed [выточить]()

PR #1301, opened 2022-10-10
Fix inconsistent French translation of "Rotate Imported"

progval:
"Rotation Importation 90°" was the only noun group in this part of the menu.

This replaces it with an infinitive group like the rest of the menu, which is also easier to understand.


phkahler:
@progval  Thank you for these updates!

PR #1302, opened 2022-10-15
fr_FR: Add missing translations,review existing ones, and fix typographic style

progval:
You can review commit per commit.

"fr_FR: Add missing translations, and review existing ones." contains the content change, the rest fixes only inconsistencies and typography.

FWIW, I am struggling with the translation of "Lathe" because the French word ("Tour") has many meanings ("lathe", "turn", "tower", "rook", ...), and this one is rarely the one that comes to mind. To disambiguate, I translated this way:

* "Lathe" (in menu) -> "Tour complet" (literally: "Complete turn", but sounds more natural in French than in English)
* "Revolve" (in menu) -> "Révolution" (I'm considering "Partial revolution" too, but the shorter version seems good enough)
* "New group rotating active sketch" (in left toolbar) -> "Nouveau groupe de révolution du dessin actif (tour complet)" (literally: "New revolution group of active sketch (complete turn)")
* "New group revolve active sketch" (in left toolbar) -> "Nouveau groupe de révolution partielle du dessin actif" (literally: "New partial revolution group of active sketch", with "partial" unambiguously qualifying "revolution" rather than "group")


ruevs (commented):
Are these « » really the "proper" quotes to use in French?

progval:
Yes. https://en.wikipedia.org/wiki/Quotation_mark#French describes their use

phkahler:
In english, Lathe is name of the machine tool that rotates the part for cutting. As opposed to a drill or mill that has a rotating cutter. As a result of spinning the work piece, all cuts go the full 360 degrees.

I borrowed "revolve" from commercial CAD tools that use that word for extruding through an angle less than 360.

Comparing to the French words used in commercial CAD would make the choice consistent if not ideal.

progval:
What other tools make the distinction between 360° and <360° revolutions? The ones I am familiar with have a single operation for both

TristeFigure:
Please consider this one-line PR too which fixes the _conStraintes_ typo. #1306 

As for the translation of "lathe", the ambiguity is not just in the translation but in the French language and the activity/semiotic context we're in (namely CAD). Imagine if a "lathe" in english was called a "turn" and a "turn" meant "rotation" too. Knowing that you're in the context of a CAD program what would you think "tour" could mean ? It could mean "lathe" and it could mean "rotation".

What I suggest: "_**Tour (usinage)**_"
Literally: "_Lathe (machining)_"

PR #1307, opened 2022-11-04
[Graphics][Bugfix] Make "Nearest Isometric" follow turntable navigation

Lunderberg:
Prior to this commit, the "Nearest Isometric" GUI command searched all 24 possible isometric views.  When using turntable navigation, this could result in the z-axis no longer being oriented vertically.  This commit restricts the views being searched while in turntable navigation mode to those that follow this restriction.

### DIFF
--- a/src/graphicswin.cpp
+++ b/src/graphicswin.cpp
@@ -6,6 +6,8 @@
 //-----------------------------------------------------------------------------
 #include "solvespace.h"
 
+#include <iostream>
+
 typedef void MenuHandler(Command id);
 using MenuKind = Platform::MenuItem::Indicator;
 struct MenuEntry {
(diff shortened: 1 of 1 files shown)


src/graphicswin.cpp
```diff
@@ -6,6 +6,8 @@
 //-----------------------------------------------------------------------------
 #include "solvespace.h"
 
+#include <iostream>
```
ruevs, line 9: What?!?! No!

ruevs:
@Lunderberg makes perfect sense! Thank you for your first SolveSpace contribution.

FWIW it is still possible to "screw up" turntable navigation with `CTRL + middle drag`.

ruevs:
An extra observation experimenting with your commit - something I had never noticed before - there are 8 views (all with the Z axis projection on the screen vertical) that are not reachable with `F3` - even with normal navigation. I need to think about it.

src/graphicswin.cpp
Lunderberg: Whoops, thank you for the catch!  Looks like I left that in after some initial debugging.

Lunderberg:
> FWIW it is still possible to "screw up" turntable navigation with CTRL + middle drag.

Hmm.  I'm not sure what the most compatible way to handle that within turntable mode would be, since it would always rotate the z-axis.  It looks like the normal middle-click rotation assumes that z is already up, and starts having weirder movement when that assumption is broken, so maybe there's a way to improve it there.

> An extra observation experimenting with your commit - something I had never noticed before - there are 8 views (all with the Z axis projection on the screen vertical) that are not reachable with F3 - even with normal navigation. I need to think about it.

Good point.  Testing here, there's a similar effect with turntable notification, as it preferentially places the viewer looking down from above, even if the view was previously looking down from above.  Even when already aligned along an isometric view, every computed values of `d` is relatively far from zero.

PR #1308, opened 2022-11-05
Web: Add missing features

verylowfreq:
This PR introduces missing features into Web porting.

Live demo: https://verylowfreq.github.io/experimental-solvespace-on-browser/

Sorry for the big and complex patch, but I decided make a pull-request as is. I couldn't break down to individual small patches.

Added features missing in previous emscripten port:
 - Opening/saving file.
 - Touch device support.
 - Layout corruption fixed.

### Known issue

 - When exporing OBJ triangle mesh file, .obj and .mtl file are generated. But you can download only the obj file (download dialog for the mtl will not be shown.)
 - UI is implemented both of C++ and Javascript.
   - UI that implemented in Javascript doesn't refer translated texts provided by gettext.
 - UI for uploading/downloading file is not good and not stable.
   - Downloaded filename includes fullpath in psuedo filesystem that not useful for endusers.
   - File can download only when save dialog closed.
 - Zoom in/out with mouse wheel is not optimized. Too big step on some device.
 - Exporint Triangle Mesh is often not working.
   - Too much memory requests (> 2GB) and crash or the tab will freeze.

### Note: Filesystem

Emscripten provides the wrapper for C standard file I/O functions. Psuedo filesystem is created in RAM and able to read/write in it.
If the tab in Web-browser is closed, all files are lost.

### Note: MIME type

In `src/platform/html/filemanagerui.js`, MIME type for .slvs file is set to `application/x-solvespace`, as already declared in xdg desktop entry file.
REF: https://solvespace.com/forum.pl?action=viewthread&parent=4052&tt=1638119260

### DIFF
--- a/src/solvespace.cpp
+++ b/src/solvespace.cpp
@@ -560,12 +560,18 @@ bool SolveSpaceUI::GetFilenameAndSave(bool saveAs) {
 
     if(saveAs || saveFile.IsEmpty()) {
         Platform::FileDialogRef dialog = Platform::CreateSaveFileDialog(GW.window);
+        // FIXME(emscripten):
+        dbp("Calling AddFilter()...");
         dialog->AddFilter(C_("file-type", "SolveSpace models"), { SKETCH_EXT });
+        dbp("Calling ThawChoices()...");
         dialog->ThawChoices(settings, "Sketch");
         if(!newSaveFile.IsEmpty()) {
+            dbp("Calling SetFilename()...");
             dialog->SetFilename(newSaveFile);
         }
+        dbp("Calling RunModal()...");
         if(dialog->RunModal()) {
+            dbp("Calling FreezeChoices()...");
             dialog->FreezeChoices(settings, "Sketch");
             newSaveFile = dialog->GetFilename();
         } else {
@@ -578,6 +584,9 @@ bool SolveSpaceUI::GetFilenameAndSave(bool saveAs) {
         RemoveAutosave();
         saveFile = newSaveFile;
         unsaved = false;
+        if (this->OnSaveFinished) {
+            this->OnSaveFinished(newSaveFile, saveAs, false);
+        }
         return true;
     } else {
         return false;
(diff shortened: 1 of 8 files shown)


src/solvespace.cpp
```diff
@@ -560,12 +560,18 @@ bool SolveSpaceUI::GetFilenameAndSave(bool saveAs) {
 
     if(saveAs || saveFile.IsEmpty()) {
         Platform::FileDialogRef dialog = Platform::CreateSaveFileDialog(GW.window);
+        // FIXME(emscripten):
+        dbp("Calling AddFilter()...");
```
ruevs, line 564: Should these `dbp` calls stay?

phkahler:
Is the file load/save actually useful? It says it's a RAM FS and everything is lot on exit. That's not really useful is it?

ruevs (commented):
I've been looking forward to this. I see that you made significant improvements and changes compared to https://github.com/verylowfreq/solvespace/tree/feature-em-fileio.

In fact I do like that you dropped the ZIP file support - one less dependency.

ruevs:
> * Zoom in/out with mouse wheel is not optimized. Too big step on some device.

Indeed on Firefox 106.0.2 for me the scroll wheel zoom step is too big. Luckily we added `shift+scroll` fine zoom some time ago. So this is a minor problem.

ruevs:
> Is the file load/save actually useful? It says it's a RAM FS and everything is lot on exit. That's not really useful is it?

@phkahler have you tried it? It actually allows you to download the files in the pseudo file system and to upload local files into it on `File | Open`

ruevs:
Rebased onto master and merged. @verylowfreq my question about the `dbp`'s stands https://github.com/solvespace/solvespace/pull/1308#discussion_r1014655305

src/solvespace.cpp
verylowfreq: @ruevs Oops, I forgot removing many debugging messages I've added; including it.

PR #1310, opened 2022-11-05
Platform: Fix GUI scaling on devices with non-integer pixel ratio

ruevs:
The fuction GetDevicePixelRatio now returns a `double` instead of an `int`. This should allow the scaling of the GUI on devices where the pixel ratio is non integer to work properly. For example a monitor on Windows where the DPI is not a multiple of 96. It may help with the Web Emscripten port on tablets and phones as well.

In addition on Windows the mouse wheel delta calculation is fixed.

### DIFF
--- a/src/platform/gui.h
+++ b/src/platform/gui.h
@@ -230,7 +230,7 @@ public:
     virtual double GetPixelDensity() = 0;
     // Returns raster graphics and coordinate scale (already applied on the platform side),
     // i.e. size of logical pixel in physical pixels, or device pixel ratio.
-    virtual int GetDevicePixelRatio() = 0;
+    virtual double GetDevicePixelRatio() = 0;
     // Returns (fractional) font scale, to be applied on top of (integral) device pixel ratio.
     virtual double GetDeviceFontScale() {
         return GetPixelDensity() / GetDevicePixelRatio() / 96.0;
--- a/src/platform/guigtk.cpp
+++ b/src/platform/guigtk.cpp
@@ -889,7 +889,7 @@ public:
         return gtkWindow.get_screen()->get_resolution();
     }
 
-    int GetDevicePixelRatio() override {
+    double GetDevicePixelRatio() override {
         return gtkWindow.get_scale_factor();
     }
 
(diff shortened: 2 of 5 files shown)


ruevs:
@verylowfreq can you please check if this helps with the GUI scaling and mouse pointer/touch location on Android/phones/tablets. In particular I think your `this->useWorkaround_devicePixelRatio` logic may not be required after this.
https://github.com/solvespace/solvespace/blob/31ac8083ae27733b18a6c2f6454104b9efd9fc29/src/platform/guihtml.cpp#L621

At least for me this pull request fixes the Web version on an Android tablet (without removing your workaround). On the same tablet with https://verylowfreq.github.io/experimental-solvespace-on-browser/bin-2022110501/solvespace.html I have a big offset between the place where I touch and the location where the touch is "interpreted" (e.g. the selection marquee).

ruevs:
@verylowfreq OK some more experiments show that removing `this->useWorkaround_devicePixelRatio` (an the related code) it keeps working fine on my tablet. The UI looks a bit rough since the font and icons are re-scaled  by a non integer ratio which also changes based on the orientation of the tablet - portrait or landscape. But this happens with the workaround as well - bot scaled to different size (?!).

An alternative would be to change GetDevicePixelRatio like this:

```C++
    double GetDevicePixelRatio() override {
        if(this->useWorkaround_devicePixelRatio) {
            return 1;
        }
        
        return emscripten_get_device_pixel_ratio();
    }
```
It also works

verylowfreq:
@ruevs Thanks! This PR (6fc84ae) works good also in my test.

Test mobile devices:
 - Android 11 Tablet (Nexus 7 2013; custom firmware)
 - Android 9 Smartphone (HTC HTV33; Japan-localized HTC U11)
 - iPad Pro 2020 (iPad OS 15.4)
   - Works good as previous.

And as you noted, also works good on PC browsers (Firefox and Google Chrome) when zoom in/out.

ruevs:
@verylowfreq I am merging it then. I also added and adjustment for the scroll wheel zoom sensitivity that (at least on my system) makes it mach the desktop version.

PR #1323, opened 2023-01-07
Fix snap release

vespakoen:


vespakoen:
Looks like the credential has to be re-newed, see: https://github.com/snapcore/action-publish/issues/28#issuecomment-1157061194

vespakoen:
@ppd do you know who created the credentials for this / can re-new them?
For the time being we should also be able to lock the snapcraft version to something below v7.

ruevs:
This commit https://github.com/solvespace/solvespace/commit/10cb310f18a0cd44954c01f41cd6b7f1c0bc8067 sis not help. The snap store upload is still failing. https://github.com/solvespace/solvespace/actions/runs/3891048562/jobs/6641004302

PR #1326, opened 2023-01-10
Chinese translation updated.

lomatus:


ruevs:
@youzhou2020 since you are the original author of the Chinese translation can you please review this update?

phkahler:
@ruevs I think it might be wise to merge this before having them update the constraint messages.

ruevs:
I agree. But let's see if @youzhou2020 or maybe @KmolYuan is around and wants to take a look.

Issue #1070, opened 2021-07-19
Do not hide comments with a custom style when hide constraints enabled


Symbian9:
### System information

- **SolveSpace version:** `3.0~`
- **Operating system:** Debian 10 buster

### Expected behavior

*Comments* with a custom style (with *"show this object on screen"* turned ON) should be *visible* if enable *"Hide constraints and dimensions"*

### Actual behavior

If enable *"Hide constraints and dimensions"* all *comments* also hides even some of them has a custom style (with *"show this object on screen"* turned ON)

Issue #1046, opened 2021-05-30
Ability to independently show/hide: H/V constraints, dimensions, REF and comments


baryluk:
### System information

- **SolveSpace version:** 3.0-rc2
- **Operating system:** Debian testing

### Expected behavior

I really love SolveSpace, and it is my tool of choice for 2D and 3D designs, metal, 3D printing, woodworking, sketching, and even planning space use for many years, but there is one thing that bother me often, that is visual noise of all the constraints on the sketch.

There is a button to hide / show "All constraints and dimensions", and option to change the line style (including text style) of constraints and dimensions. But it is all or nothing. I wish there was a bit more flexibility here.

For example:

I would like to keep the dimensions, just hide H and V constraints. Often I have very rectilinear design, with many rectangles, that I already know have H and V on their side segments. 99% of the I don't need to see them. Even the line segments that I do manually, I often will see that it has H or V constraints, and once I know it has one (either automatic or manual), I really don't want to see it again, ever, until maybe the very last verification at some point later.

Similarly, I sometimes might want to see H and V constraints, but not dimensions or references. This would be exactly for the remaining 1% of cases, where I want to ensure that H and V constraints are where they should be, but other things should be hidden so I can easily verify it.

I want to hide dimensions and all constraints, but keep the "Comment" texts, still visible. This is again, for clarity, and sketching.

I also want to be able to hide equal length constraint markers (small perpendicular magenta line on line segments and arcs) . They are nice, and I love them, but there are times, where I want to hide them and only them.

And be able to hide point on line constraint markers (square around the point). Again, I love them, but often when I am done with them, I don't care, and wish to not see them 95% of the time, to focus on other things.

And I think it would be also nice to be able to separately show/hide "reference dimensions" independent from "normal dimensions".

On a related note, I wish there was easy way to set color and text size of Comments, independently of other constraints and dimensions style. Often I want Comment to be slightly bigger and bolder, but dimensions be smaller. And sometimes H/V constrain marker, even smaller (if shown). Just some ideas.

It would be best to be able to set this per SolveSpace model file.

Symbian9:
This feature request is what I'm itself wanted for SolveSpace!

phkahler:
@baryluk  There is a lot covered in this request. I think it's good that you put it all in one place because it's not really a series of specific features, but wants a rather comprehensive change in displaying constraints. The "On a related note" part about comments is the only part I'd consider separate ;-)

[...]

PR #1329, opened 2023-01-13
dimonly fix upd

77maxikov:


phkahler:
Not a big fan of "none" for the hover text, is that what it was for hiding constraints?

Unrelated to this PR, I tried tweaking the color of shown constraints in style.cpp by changing the RGB value of their style and it has no effect after recompiling.

ruevs:
@77maxikov I really like this.

I pushed some improvements. @phkahler @77maxikov please take a look.

phkahler:
What about #1046 and #1070 ? Does this address either, and if not, should it?

src/textwin.cpp
```diff
@@ -219,15 +153,13 @@ static ShowHideButton pointsButton =
...
-      {"constraints and dimensions","dimensions","none"},
-      {"constraint","constraint-dimo","constraint-wo"}
-    };
-
+static TriStateButton constraintsButton  = {
+    (unsigned *)(&(SS.GW.showConstraints)),
+    {(unsigned)GraphicsWindow::ShowConstraintMode::SCM_SHOW_ALL,
+     (unsigned)GraphicsWindow::ShowConstraintMode::SCM_SHOW_DIM,
+     (unsigned)GraphicsWindow::ShowConstraintMode::SCM_NOSHOW},
+    {"Show only dimensions", "Hide constraints and dimensions", "Show constraints and dimensions"},
```
77maxikov, line 161: {"Show constraints and dimensions", "Show only dimensions", "Hide constraints and dimensions"}

ruevs: @77maxikov The idea is that the tool tip should show what will happen when the user clicks the button instead of the current state. Take a look at how the hidden line one worked even before.

ruevs:
> What about #1046 and #1070 ? Does this address either, and if not, should it?

It doesn't address either of those. #1046 is a very general one asking for fu fine grained come nstrol over what is shown. #1070 is a bit more reachable but...

src/textwin.cpp
77maxikov: Ok, got it.

PR #1330, opened 2023-01-16
Don't try to drag points with pt-coincident constraint to a previous …

### LINKED ISSUE #1012: Bug: Can not drag the datum point sometimes
SolveSpace version 3.0~f4ad8205
Operating system: Win7 Ultimate Edition sp1

During I learing the tutorials at the https://solvespace.com/tutorial.pl , there is 6 video tutorials, in the No.4 video, just as shown below,
image

at 3:24, the teacher dragged the datum point, and then the circle and the sketch-in-plane was moved with the drag, just as shown below.
image

But I can not drag the point to move in SolveSpace version 3.0~f4ad8205! The svls doc is here.
(https://github.com/solvespace/solvespace/files/6310333/can.not.drag.the.circle.by.the.datum.point.zip)

I have tried the same operations in SolveSpace version 2.3-7c1ca460, the result is very OK! I can drag the datum point and circle pretty well in SolveSpace version 2.3-7c1ca460. **So it's must a bug in the SolveSpace version 3.0~f4ad8205.**

phkahler:
…group. Fixes #1012 and makes dragging as done in one of the tutorials possible again.

### DIFF
--- a/src/drawentity.cpp
+++ b/src/drawentity.cpp
@@ -180,19 +180,43 @@ bool Entity::IsVisible() const {
     return true;
 }
 
+static bool PtCanDrag(hEntity pt) {
+    Entity* p = SK.GetEntity(pt);
+    // a numeric copy can not move
+    if(p->type == Entity::Type::POINT_N_COPY) return false;
+    // these transforms applied zero times can not be moved
+    if(((p->type == Entity::Type::POINT_N_TRANS) ||
+       (p->type == Entity::Type::POINT_N_ROT_AA) ||
+       (p->type == Entity::Type::POINT_N_ROT_AXIS_TRANS))
+        && (p->timesApplied == 0)) return false;
+    return true;
+}
+
 // entities that were created via some copy types will not be
 // draggable with the mouse. We identify the undraggables here
 bool Entity::CanBeDragged() const {
-    // a numeric copy can not move
-    if(type == Entity::Type::POINT_N_COPY) return false;
-    // these transforms applied zero times can not be moved
-    if(((type == Entity::Type::POINT_N_TRANS) ||
-       (type == Entity::Type::POINT_N_ROT_AA) ||
-       (type == Entity::Type::POINT_N_ROT_AXIS_TRANS))
-        && (timesApplied == 0)) return false;
+    if(IsPoint()) {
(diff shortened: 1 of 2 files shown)


phkahler:
I have not verified this with the actual tutorial scenario yet, but it allows dragging in similar cases.

phkahler:
Added some more logic and confirmed it fixes the tutorial example in #1012

src/drawentity.cpp
```diff
@@ -192,7 +212,11 @@ bool Entity::CanBeDragged() const {
         && (timesApplied == 0)) return false;
     // for these types of entities the first point will indicate draggability
     if(HasEndpoints() || type == Entity::Type::CIRCLE) {
-        return SK.GetEntity(point[0])->CanBeDragged();
```
ruevs, line 195: Hmm if removed from here `CanBeDragged` remains called in only one place in the code.... I've never looked in these parts much... I need to think....

phkahler:
@ruevs  I made it non-recursive in the latest version. This strictly adds more conditions where points can not be dragged. The original reason to create CanBeDragged() was to fix the dragging of entities from prior groups. Obviously the user wants to drag the same entity that is highlighted by hovering, but there are times when that item is fully constrained and we know it. In those cases, by not dragging the entity the drag code is free to grab the next highest priority entity to drag, and that can be in a prior group. That function is the key to dragging stuff in prior groups that are covered by things in the present group. It fixed a few drag-selection issues while making the code cleaner. Selecting and Dragging are now almost identical except for calling CanBeDragged.

phkahler:
In the tutorial, the workplane origin is a numeric copy of the datum point in the prior group. The old version of CanBeDragged knew about that and returned false, preventing it from dragging the copy and allowing it to drag the datum point.  When a circle was drawn in the workplane on that same point, the circle center is a new point with a coincident constraint to the workplane origin. So the only way to prevent trying to drag the circle center is to check if it's constrained to something that can not be dragged. For now that's only if its constrained to a numeric copy point. We could go further, but this gets the tutorial working ;-)

phkahler:
And I made some regressions - now fixed. CanBeDragged IIRC was what allowed non-point entites to drag the length of an extrusion (drag by a line or circle for example). I had broken that, but fixed now.

Issue #1239, opened 2022-05-10
Multiple "Message" dialogs when deleting elements with dependant constraints (minor, UI)


gnbl:
Deleting elements that other elements are constrained to, this message is shown:

```
---------------------------
SolveSpace - Message
---------------------------
Additional sketch elements were deleted, because they depend on the element that was just deleted explicitly.

These include: 
     0 requests
     2 constraints
     0 groups

Choose Edit -> Undo to undelete all elements.
---------------------------
OK   
---------------------------
```

Unfortunately I can't provide a minimal example of this, but sometimes this automatic deletion appears to cause another automatic deletion and a corresponding message dialog. I've only noticed because the second message takes the focus, so hitting Return to accept with `OK` does not work and I have to use the mouse...

This could be resolved by restoring focus to the other dialogs, or by catching all automatic deletions/dependencies before showing the dialog.

phkahler:
@gnbl what OS is this on? I thought those message boxes were modal so there couldn't be 2 at the same time. An example could be helpful.

gnbl:
windows-openmp 3.0~b334996c on Windows 10 64 bit.

Sorry, I tried to come up with an example, but unsuccessful so far.
Next time it happens, I'll take a closer look and will attempt to reproduce in minimal example file.

gnbl:
Here's what it looks like, even three dialogs:
![autodelete](https://user-images.githubusercontent.com/1181708/167914570-5b9197dd-c50d-4900-b2d5-dca29b097449.png)
Cannot reproduce in example sketch.

iscgar:
So, the cause for this is quite simple once it is reproduced:

![Image](https://github.com/user-attachments/assets/53025080-f6ef-4d6b-9cbe-bf8a75f011e8)

Calling `Message()` causes events to be handled, and because of that, the timer may fire with a scheduled regeneration after entity splitting, causing a call to `GenerateAll()` while we're showing the message box, but before it's dismissed. Because we still didn't zero the `deleted` member, the code path happily shows the message box again, which can lead to another scheduled regeneration to be fired, etc.

https://github.com/solvespace/solvespace/blob/2c5bb5d475e03a73da685649d2009539bfa96a7e/src/generate.cpp#L331-L352

I'll open a PR shortly to fix this issue.

ruevs:
Wow! This was easy (in retrospect) :-)

PR #1333, opened 2023-01-21
Fix MessageBox Avalanches due to Message blocking

### LINKED ISSUE #1320: message box apocalypse on split lines when endpoint is on line
### Preamble

Yes, I know, I'm using the "Split lines / ..." function not in the correct way; I'm really missing the trim-line-to-next-point function most CAD programs have *; but this is far from the failure mode I expected .... it didn't do that previously and just created a zero-length edge instead of crashing after a message box invasion.

\* some things are much easier and faster to draw that way

### System information

- **SolveSpace version:** 3.1~3833dd02 (built a few minutes ago)
- **Operating system:** W10 22H2

### Expected behavior

From optimal to worst:
1. just split the one line at the endpoint of the second line
2. create zero-length line segment like previous versions
3. error out ("can't do that")

### Actual behavior

linecrash

### Additional information

🙏〰✂

MrBlockers:
This fixes #1320, but not #1239. The root cause for the avalanche of messages is due to how the refresh timer system calls GenerateAll. When GenerateAll is called by Refresh, if a Message occurs, that Message will block GenerateAll. It _doesn't_ block subsequent calls from the timer to Refresh (presumably from a separate thread). Because the "scheduledGenerateAll" flag is not cleared until after the generation is unblocked, each following refresh triggers another call to GenerateAll. By reversing the flag clear and call, it breaks the cycle. I don't think this matters for scheduledShowTW, but I updated it as well. 

I am going to look into what might be causing #1239 next, but I think it is due to something unrelated.

P.S. I am still getting my bearings on the codebase, so please nitpick away!

### DIFF
--- a/src/solvespace.cpp
+++ b/src/solvespace.cpp
@@ -301,12 +301,14 @@ void SolveSpaceUI::Exit() {
 void SolveSpaceUI::Refresh() {
     // generateAll must happen bfore updating displays
     if(scheduledGenerateAll) {
-        GenerateAll(Generate::DIRTY, /*andFindFree=*/false, /*genForBBox=*/false);
+		// Clear the flag so that if the call to GenerateAll is blocked by a Message or Error, 
+		// subsequent refreshes do not try to Generate again.
         scheduledGenerateAll = false;
+        GenerateAll(Generate::DIRTY, /*andFindFree=*/false, /*genForBBox=*/false);   
     }
     if(scheduledShowTW) {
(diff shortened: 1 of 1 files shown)


ruevs:
More related discussion here https://github.com/solvespace/solvespace/pull/914

phkahler:
I don't have the problem on Linux, but the patch doesn't seem to break anything.

PR #1336, opened 2023-01-29
Treat group as mesh only if a linked file has a mesh and no shell.

### LINKED ISSUE #1269: Solid model is missing for an assembled part that was forced to triangle mesh at design time
### System information

- **SolveSpace version:** 3.1~d6e1b230
- **Operating system:** Windows 7 x64

### Expected behavior

Triangle-mesh assemblies should have a visible solid model

### Actual behavior

Assembled part that had beed restricted to triangle mesh at design time loads with no visible solid model as an assembly

### Additional information

Using the model attached in #1268 except with ticking "restrict to triangle mesh" on the cylinder group to avoid the naked edges issue, loading it as an assembly there is no solid model of it visible. Also ticking the "restrict to triangle mesh" option in the group created by loading this model as an assembly makes the solid model appear. This is, however, extremely counter-intuitive until one finally figures out this needs to be done to see the solid model.
Solid Model Error 1a
Solid Model Error 1b

phkahler:
Fixes #1269

### DIFF
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -412,7 +412,9 @@ bool Group::IsForcedToMesh() const {
 }
 
 bool Group::IsTriangleMeshAssembly() const {
-    return type == Type::LINKED && linkFile.Extension() == "stl";
+    if (type != Type::LINKED) return false;
+    if (!impMesh.IsEmpty() && impShell.IsEmpty()) return true;
+    return false;
 }
 
 std::string Group::DescriptionString() {


phkahler:
The commit message should say "mesh-only" to avoid confusion. We should probably remove the test for "stl" file extension and verify the new logic covers that case too, as it could eliminate a string comparison.

ruevs:
> The commit message should say "mesh-only" to avoid confusion.
How about:

"Treat linked file groups as mesh, only when they have no shell."

> We should probably remove the test for "stl" file extension and verify the new logic covers that case too

You are probably right. For example the web edition may break this.

Is it enough if
`if (!impMesh.IsEmpty() && impShell.IsEmpty()) return true;`
becomes
`if (impShell.IsEmpty()) return true;`
are there other possibilities?

(Just philosophizing from the phone ;-) )

phkahler:
@ruevs  you can link a 2d sketch which should have no mesh and no shell.. we don't want to force to mesh if there isn't one. 

The IDF (.emn) linker creates a shell and no mesh. The logic should be that the group has a mesh and does not have a shell. That's the only time we know to force it. This also changes the checkbox for the group to text instead.

ruevs:
Makes sense and with this in mind the commit message should be something like yours.
"Treat linked file groups as mesh, only when they have a mesh but no shell."

So why did you not remove the comparison to `stl`?

phkahler:
> So why did you not remove the comparison to `stl`? No `stl` to test with? (just export one)

I thought of it after I had already tested the code as-is and made a PR ;-)

PR #1340, opened 2023-02-01
eliminate possible use before set warnings

phkahler:


phkahler:
I tried refactoring based on (how == PAINT) logic, but even then the compiler can't figure it out and throws a warning. This is the simplest way to make them go away.

PR #1342, opened 2023-02-01
[WIP] Python bindings with pybind11

vespakoen:
Hello all,

I have been looking into merging https://github.com/solvespace/solvespace/pull/493 into SolveSpace, and stumbled upon some things that we might want to change before merging it.

First of all, it is based on the SolveSpace "C bindings", which are a bit minimal, and leave a lot up to the users of the library.
Most notably, in the C bindings there are no helper methods to create Constraints.

In KmolYuan's PR, a couple of classes where added to help with this (see: https://github.com/solvespace/solvespace/pull/493/files#diff-006a24eb227071757e0aed5c8dc2a39625f22957e87562378cfd403b2d5de74fR209-R311)

I really like the helper methods, and they make using the SolveSpace bindings a lot easier, however, I don't like that they are written in cython, because that makes them "python only".

I would much rather have these methods be inside of SolveSpace, so that potential future bindings can also make use of them.

Another issue I have with that PR is that it doesn't use the CMakeLists.txt to compile SolveSpace, but instead uses a custom compile script, this is not a huge problem but I'd rather have less duplication in the build system.

So, I believe the best solution would be to:
- Use those helper methods from the PR, but move them into SolveSpace's C++ codebase.
- Expose SolveSpace by using pybind11 in stead of exposing the C library with cython.
- Use `scikit-build-core` and refactor the CMakeLists.txt so it can also work for the Python library.

Some issues with SolveSpace make the above hard, and the biggest offenders are:
- Our CMakeLists.txt compiles and configures things that are not necessary when NOT compiling the GUI (In the PR, I have wrapped those in `if (NOT ENABLE_PYTHON_LIB) ... endif()`, what I really want to do here is clean up the CMakeLists.txt, and possibly split it up in different modules for all the components it compiles.
- We have an "EntityBase" and "ConstraintBase" class, for use in libraries, but there is no "GroupBase" class available, so I added it in this PR
- When compiling with the "LIBRARY=1" definition, some code in EntityBase / ConstraintBase would still use things from the (non-library variant) Entity / Constraint.

I had to add a couple more `#ifdef LIBRARY`, which I ideally would not have had to use if the code was split up into more files, and I have tried splitting the code up more, but since there are a couple of circular references, it is pretty hard to do this.

While working on this, I have experimented with splitting up the CMakeLists.txt file (see: https://github.com/vespakoen/solvespace/blob/python-cmake/CMakeLists.txt#L301-L332)

And also with "untangling" the core, basically splitting up `solvespace.h` and `dsc.h` into smaller pieces (see: https://github.com/vespakoen/solvespace/tree/experiments/src/core)
This turned out to be very hard, since there are a lot of forward refs and circular dependencies.
I am wondering if this is something we want to reduce / get rid of in the future, or if we just accept it as a part of SolveSpace?

In this PR, I tried to change the least amount possible of SolveSpace, to get the pybind11 bindings working.
Now I want to gather some feedback and thoughts on this before I continue ;)


src/solvespace.h
```diff
@@ -172,16 +188,37 @@ enum class SolveResult : uint32_t {
...
+#endif
+
+#ifndef LIBRARY
 #include "ui.h"
+#endif
 #include "expr.h"
 
-
 // Utility functions that are provided in the platform-independent code.
-class utf8_iterator : std::iterator<std::forward_iterator_tag, char32_t> {
```
vespakoen, line 182: I got a deprecated warning, and fixed it but I didn't actually test this `utf8_iterator` code, so, note to self: revert or test this

python/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+
+find_package(Eigen3 REQUIRED)
+if(Eigen3_FOUND)
+  include_directories(${EIGEN_INCLUDE_DIR})
+endif()
+
+find_package(mimalloc CONFIG REQUIRED)
+if(mimalloc_FOUND)
+  include_directories(${MIMALLOC_INCLUDE_DIR})
+endif()
```
vespakoen, line 17: These should of course be coming from the main CMakeLists.txt

python/rebuild.sh
```diff
@@ -0,0 +1,18 @@
...
+trap cleanup EXIT
+
+cleanup
+cp -r ../src src/
+pybind11-stubgen solvespace --no-setup-py --root-module-suffix "" -o .
+python3.10 -m build .
+python3.10 -m pip install --force-reinstall dist/solvespace-0.0.1-cp310-cp310-macosx_13_0_arm64.whl
+/Applications/Blender.app/Contents/Resources/3.3/python/bin/python3.10 -m pip install --force-reinstall dist/solvespace-0.0.1-cp310-cp310-macosx_13_0_arm64.whl
+pybind11-stubgen solvespace --no-setup-py --root-module-suffix "" -o "/opt/homebrew/lib/python3.11/site-packages/"
+python3.10 tests
```
vespakoen, line 18: This file should be removed ;)

src/solvespace.h
```diff
@@ -499,18 +533,630 @@ class Sketch {
...
+    // param
+    inline Param AddParam(double val) {
+        Param pa = {};
+        pa.val   = val;
+        param.AddAndAssignId(&pa);
+        return pa;
+    }
+
+    // entities
+    inline EntityBase AddPoint2D(double u, double v, EntityBase workplane) {
```
vespakoen, line 600: These functions don't have to be inline

src/system.cpp
```diff
@@ -8,6 +8,16 @@
 //-----------------------------------------------------------------------------
 #include "solvespace.h"
 
+#ifdef LIBRARY
```
vespakoen, line 11: I am not happy with re-defining this here, but the `solvespace.h` header has to "undef" those because otherwise it conflicts with `Style::CONSTRAINT`

phkahler:
@vespakoen Some thoughts:  I think the notion of groups is solvespace specific and should probably stay out of any solver library. It sounds like the solver is basically the algebraic constraint solver, is that correct? By adding entities you end up with a geometric constraint solver. I thought that's where the line was drawn, but it seems I was mistaken! ;-) If the core really is a "C" interface that's good because C++ limits the number of languages you can make binding for.

Maybe it should not be merged, but build as a separate repo that borrows a bunch of files from solvespace? Or just build the solver as a library with limited scope from this repo and add the other stuff on in another repository? Is there even a line where the solver could be used *as a library* by solvespace itself?

PR #1343, opened 2023-02-03
"Stateful" C library + Python bindings (using cython) and JS bindings (using embind)

vespakoen:
So this is another attempt, this time adding helper methods to lib.cpp and by using cython again.

I am considering to use the `Slvs_Entity` struct as the return type and as arguments to the helper functions.
This would remove the overhead of looking up entities in the sketch by their handle.

Would like to hear thoughts again ;)


src/sketch.h
```diff
@@ -459,6 +459,54 @@ class EntityBase {
     // times to apply the transformation.
     int timesApplied;
 
+    inline bool IsFreeIn3D() {
```
vespakoen, line 462: I should probably move these to lib.cpp as these are not used by SolveSpace.

src/lib.pyx
```diff
@@ -0,0 +1,261 @@
+from enum import IntEnum, auto
+
+cdef extern from "slvs.h":
+    ctypedef int Slvs_hEntity
+    ctypedef int Slvs_hGroup
+    ctypedef int Slvs_hConstraint
+    ctypedef int Slvs_hParam
+    ctypedef struct Slvs_SolveResult:
+        int result
```
vespakoen, line 9: I am not happy with `SolveResult.result`, should probably look for a better name here

src/dsc.h
```diff
@@ -521,6 +521,13 @@ class IdList {
         ++n;
     }
 
+    T *FindByHandleV(uint32_t v) {
```
vespakoen, line 524: This doesn't really make things better, I want to get rid of this and looking up entities in the sketch by just returning the full struct of Entities / Constraints from the helper methods, then the lookup can be completely removed.

tests/test.py
```diff
@@ -0,0 +1,252 @@
...
+    slvs.distance(g, p2, p4, 40, wp)
+    slvs.distance(g, p3, p4, 70, wp)
+    slvs.distance(g, p0, p3, 35, wp)
+    slvs.distance(g, p1, p4, 70, wp)
+    line1 = slvs.add_line_2d(g, p0, p3, wp)
+    slvs.angle(g, line0, line1, 45, wp, False)
+
+    result = slvs.solve_sketch(g, 0)
+    print("result", result)
+    # self.assertEqual(result['result'], 0)
```
vespakoen, line 30: Somehow the results in those tests are TOO_MANY_UNKNOWNS, although the sketch seems to have solved correctly, will have to look into what's going on here...

hiaselhans:
Hi @vespakoen 

Looking forward to have solvespace available as a python library.
Did you get any feedback from the solvespace devs? Is there anything missing from this pr? Any place to help?

Thanks for your efforts!

vespakoen:
I think I am alone on this one, and I am pretty happy with how this PR turned out. Minimal changes to the core, also works as a pure C library, and easy to use.
I was going to change my PR over at CAD Sketcher to make use of this solver branch, to try it out in a "real world" scenario, but I got pretty busy with other things.

I should probably look into this again soon and when i feel good about it go ahead and merge this.

I'd love for people to try it out and i'd love feedback about it though!

vespakoen:
[slvs-js.zip](https://github.com/solvespace/solvespace/files/10788115/slvs-js.zip)
Oh my, that worked pretty easily, now we can use solvespace as a library in the browser =)

Aveline67:
> [slvs-js.zip](https://github.com/solvespace/solvespace/files/10788115/slvs-js.zip) Oh my, that worked pretty easily, now we can use solvespace as a library in the browser =)

how did you achieve it ?

Aveline67:
here a build with the missing js file
[bin.zip](https://github.com/solvespace/solvespace/files/10788723/bin.zip)

Aveline67:
how can we generate TypeSvript typings definitions automatically ?

vespakoen:
Automatically will be difficult, but writing them by hand shouldn't be too big of a deal, I'll look into it once I tested this out a bit more.
I will publish this to npm soon to see if it works well in the browser and in node.js

vespakoen:
Just published a test package on NPM (called slvs)

Example: https://observablehq.com/@vespakoen/require-solvespace

vespakoen:
If we are all happy with this I would like to merge this.
@phkahler @ppd @ruevs @rpavlik thoughts / feedback / ideas?

@jwesthues can you create a NPM account (https://www.npmjs.com), create a "classic" token with "write package" permission, and make sure that 2FA is not needed for publishing under "Modify 2FA".

Then add this token to CI as "NPM_TOKEN".

From PyPi (https://pypi.org) (the Python package index), I need a `PYPI_API_TOKEN` and from (https://test.pypi.org) I need a `TEST_PYPI_API_TOKEN`.

Once those are added I can merge this PR and we have improved C bindings, Python bindings and WASM bindings for the solver.

jwesthues:
@vespakoen I've created all three accounts, all with username `solvespace` and email `support@solvespace.com`, and set the corresponding secrets. The npm account uses an "automation" token, which it says should bypass any 2FA. Thanks, and let me know if you see any trouble.

.github/workflows/wasmlib.yml
```diff
@@ -0,0 +1,68 @@
+name: WASM Library
+
+on:
+  push:
+    branches: [ master, python, experiments, stateful-c-lib ]
```
vespakoen, line 5: todo

.github/workflows/python.yml
```diff
@@ -0,0 +1,163 @@
...
+          fetch-depth: 0
+      - run: git submodule update --init extlib/mimalloc extlib/eigen
+      - name: Set version
+        shell: bash
+        run: |
+          if [[ "${GITHUB_REF}" == refs/tags/* ]]; then
+            version="${GITHUB_REF##*/}"
+          else
+            # version="$(git describe --tags).dev${GITHUB_RUN_NUMBER}"
+            version="3.1.0.dev${GITHUB_RUN_NUMBER}"
```
vespakoen, line 112: todo, un-hardcode version

.github/workflows/wasmlib.yml
```diff
@@ -0,0 +1,68 @@
...
+
+          ls artifacts
+          unzip artifacts/slvs-wasmlib/slvs-wasmlib.zip -d js
+          ls js
+
+          # bump version
+          if [[ "${GITHUB_REF}" == refs/tags/* ]]; then
+            version="${GITHUB_REF##*/}"
+          else
+            version="3.1.0-dev.${GITHUB_RUN_NUMBER}"
```
vespakoen, line 58: todo, un-hardcode

pyproject.toml
```diff
@@ -0,0 +1,46 @@
+[build-system]
+requires = ["scikit-build-core", "Cython", "cmake", "ninja"]
+build-backend = "scikit_build_core.build"
+
+[project]
+name = "solvespace"
+version = "0.0.1"
+description="A minimal example package (with pybind11)"
```
vespakoen, line 8: todo update version and description

python/solvespace/solvespace.pyi
```diff
@@ -0,0 +1,276 @@
...
+    PROJ_PT_DISTANCE = 100030
+    WHERE_DRAGGED = 100031
+    CURVE_CURVE_TANGENT = 100032
+    LENGTH_DIFFERENCE = 100033
+    ARC_ARC_LEN_RATIO = 100034
+    ARC_LINE_LEN_RATIO = 100035
+    ARC_ARC_DIFFERENCE = 100036
+    ARC_LINE_DIFFERENCE = 100037
+
+# class EntityType(IntEnum):
```
vespakoen, line 47: todo, uncomment, might be handy for testing types

python/solvespace/solvespace.pyi
```diff
@@ -0,0 +1,276 @@
...
+#     NORMAL_IN_3D = 60000
+#     NORMAL_IN_2D = 60001
+#     DISTANCE = 70000
+#     WORKPLANE = 80000
+#     LINE_SEGMENT = 80001
+#     CUBIC = 80002
+#     CIRCLE = 80003
+#     ARC_OF_CIRCLE = 80004
+
+class ResultFlag(IntEnum):
```
vespakoen, line 59: todo, also add this to wasm library?

js/slvs.d.ts
```diff
@@ -0,0 +1,193 @@
+export interface Entity {
+  h: number;
+  group: number;
+  type: number;
+  wrkpl: number;
+  normal: number;
+  distance: number;
+  param: [number, number, number, number];
+}
```
vespakoen, line 9: point[] is missing?

python/solvespace/solvespace.pyi
```diff
@@ -0,0 +1,276 @@
...
+def distance_proj(grouph: int, e1: Slvs_Entity, e2: Slvs_Entity, value: float) -> Slvs_Constraint:
+    ...
+
+def length_diff(grouph: int, Slvs_entityA: Slvs_Entity, Slvs_entityB: Slvs_Entity, value: float, workplane: Slvs_Entity) -> Slvs_Constraint:
+    ...
+
+def dragged(grouph: int, e1: Slvs_Entity, wp: Slvs_Entity = E_FREE_IN_3D) -> Slvs_Constraint:
+    ...
+
+def clear() -> None:
```
vespakoen, line 269: todo, rename to clear_sketch or rename solve_sketch to just "solve"?

python/solvespace/solvespace.pyi
```diff
@@ -0,0 +1,276 @@
...
+    ...
+
+def solve_sketch(grouph: int, calculateFaileds: bool) -> Slvs_SolveResult:
+    ...
+
+def get_param(e: Slvs_Entity, index: int) -> float:
```
vespakoen, line 275: get_param_value? and add set_param_value

hiaselhans:
Hi @vespakoen 

thanks a lot for pushing this topic! I will give it a try.
Just a side note: i ran the tests (linux) and few were failing with numeric mismatches:
``` python
FAILED test.py::CoreTest::test_crank_rocker - AssertionError: 39.54852 != 20.0 within 4 places (19.548520000000003 difference)
FAILED test.py::CoreTest::test_involute - AssertionError: 12.62467 != 10.0 within 4 places (2.62467 difference)
FAILED test.py::CoreTest::test_jansen_linkage - AssertionError: 18.93036 != 20.0 within 4 places (1.0696399999999997 difference)
```

vespakoen:
Interesting, will take a look at that

hiaselhans:
Thanks!
Out of curiosity: what made you drop the pybind11 implementation and use cython instead?

vespakoen:
The cython bindings are lighter / faster, and most importantly, the pybind11 bindings required a lot of changes to SolveSpace's core.
I am happy that with this PR the changes to the core are minimal, and all the methods that are exposed through the bindings are also available as pure C bindings.

vespakoen:
@hiaselhans

> Hi @vespakoen
> 
> thanks a lot for pushing this topic! I will give it a try. Just a side note: i ran the tests (linux) and few were failing with numeric mismatches:
> 
> ```python
> FAILED test.py::CoreTest::test_crank_rocker - AssertionError: 39.54852 != 20.0 within 4 places (19.548520000000003 difference)
> FAILED test.py::CoreTest::test_involute - AssertionError: 12.62467 != 10.0 within 4 places (2.62467 difference)
> FAILED test.py::CoreTest::test_jansen_linkage - AssertionError: 18.93036 != 20.0 within 4 places (1.0696399999999997 difference)
> ```

Thanks for reporting that, I fixed it in the last commit, some settings on Group were not set (https://github.com/solvespace/solvespace/pull/1343/commits/d95fb0428c407a426d017bc857099cdcf6e3be63#diff-9e41a216e20b54abd85a2494540eaea9810b5d83f4c3efd3db9c494bbadf3cb1R789-R792), and were initialized with random values from memory.

TristeFigure:
Hey, I've started writing Clojure bindings to solvespace (through python interop), potentially to use it alongside OpenScad. My reasoning was that solvespace C bindings didn't offer an easy to use API for state management, so I went to use KmolYuan's python-solvespace package (which is now integrated into this repo if I'm not mistaken). Alas, it doesn't offer the ability to save to slvs files, which is disappointing (I'd like to generate complex scissors meshes structures and then play with them in solvespace's UI to drag some points around and see how it reacts).

How does this PR intersects with the already existing python-solvespace ? Also it doesn't seem to support exporting to slvs files. Could you expose this feature in this API please ? I'm willing to collaborate if needed.

vespakoen:
This PR add's the functionality of python-solvespace to the C library, see this (basic) example: https://github.com/vespakoen/solvespace/blob/stateful-c-lib/exposed/CDemo.c#L30-L58

This doesn't support exporting to a .slvs file, and  adding support for that would be quite difficult since SolveSpace expects the sketch to be in some specific format.
Since the solver library uses base classes for the Entity and Constraint (EntityBase and ContraintBase) it is missing properties that the standard save implementation needs.

I just did a basic test by copying the save functions and removing things that we don't have (Style / Request etc) over here: https://gist.github.com/vespakoen/2713a79fcbf55bcbdb44ff61efb53dad

That prints the file contents to stdout, and to save it as a valid .slvs file the first 3 bytes have to be (octal: 261 262 263) / (hex: B1 B2 B3), I changed those manually for a test by using a HEX editor.

Of course the resulting .slvs files crashed / were invalid, and I don't really know if there is a lot missing to make it work, you're on your own ;)

paulftw:
This is incredible! how can I ~~double~~ triple upvote it?!
On a serious note - I am looking forward to seeing this merged into main, will definitely test the js library in my project!
If there are any tasks pending that do not require in-depth understanding of the solvespace codebase - I'd be happy to help.

phkahler:
@vespakoen  What just happened?  :-)  :-)

I hadn't been following this and suddenly it merged!
Is this being used in some outside projects?

vespakoen:
I was getting regular emails from people that are trying to use the bindings from this PR, and today I looked at it again and it seemed to be complete. So I thought i'd give it a go and see what happens. Of course after all this time I also see potential problems with this approach, for example the copying of the value type Slvs_Entity which is not super small, but I guessed it was a good idea to merge it and go from there.

If this causes any troubles, or you are not happy with the implementation, feel free to revert it ;)

I just invited the npm "solvespace" account to become the maintainer of the "slvs" package, which didn't happen yet and causes the wasmlib-publish step to fail, which should hopefully resolve soon when Jonathan accepts the invite.

jwesthues:
Cool, just accepted the invitation.

phkahler:
@vespakoen  was it really necessary to make all the inline functions to check an entity for specific type? Is it not possible to get the type and have defined values to compare at the next level? It kind of polluted an already large class definitio. But I think I'm OK if there's a legit reason for that.

vespakoen:
I understand the concern, the reason is that it otherwise had to be implemented in both python and in javascript, by writing them in C they only had to be wrapped which deduplicates things a bit. The Slvs_Entity that is returned can / should be smaller than it currently is and I will look into improving that in the future.

ruevs:
I hoped that this would never get merged into the main repository but instead will become a separate project (e.g. https://github.com/solvespace/solverbindings/) including the main project as an external. In my opinion this is just "noise" in the main CAD application project.

But I would have kept quiet if it at leas built properly - but it does not:
With `-DENABLE_CLI=no` and `-DENABLE_GUI=yes` - which is the way I usually build - it does not even compile with the following errors:

```
1>...\SolveSpace\git\src\render\rendercairo.cpp(6,10): fatal error C1083: Cannot open include file: 'cairo.h': No such file or directory
2>...\SolveSpace\git\src\solvespace.h(40,10): fatal error C1083: Cannot open include file: 'Eigen/SparseCore': No such file or directory (compiling source file ...\SolveSpace\git\src\render\gl3shader.cpp)
2>...\SolveSpace\git\src\solvespace.h(40,10): fatal error C1083: Cannot open include file: 'Eigen/SparseCore': No such file or directory (compiling source file ...\SolveSpace\git\src\platform\entrygui.cpp)
2>...\SolveSpace\git\src\solvespace.h(40,10): fatal error C1083: Cannot open include file: 'Eigen/SparseCore': No such file or directory (compiling source file ...\SolveSpace\git\src\render\rendergl3.cpp)
2>...\SolveSpace\git\src\solvespace.h(40,10): fatal error C1083: Cannot open include file: 'Eigen/SparseCore': No such file or directory (compiling source file ...\SolveSpace\git\src\platform\guiwin.cpp)

```

This is somehow caused by the changes in in the [CMakeLists.txt](https://github.com/solvespace/solvespace/pull/1343/files#diff-148715d6ea0c0ea0a346af3f6bd610d010d490eca35ac6a9b408748f7ca9e3f4) files - I presume of the type `if(ENALBE_GUI OR ENABLE_CLI)` but I do not understand why - my CMake knowledge is... limited :-)

More nitpicks:
- Remove the unnecessary change in [system.cpp](https://github.com/solvespace/solvespace/pull/1343/files#diff-d34d69445971e623152a1884dfe7d22e7b59ebc1edf082008dd4b1cc5b5b5aca)
- Remove the unnecessary change in [solvespace.h](https://github.com/solvespace/solvespace/pull/1343/files#diff-8712b70336e60e6bc2d3c49a0793e048d54d99fadbd1bcb3c598a5ec21e6eec3)
- Remove the unnecessary changes in [sketch.h](https://github.com/solvespace/solvespace/pull/1343/files#diff-ad1d0c0e06e0b331a77af373206360a4e9cdb57d2ae461c5e949dcf4f1d2471b) (`ToString`) - they are only used by commented debug code in [lib.cpp](https://github.com/solvespace/solvespace/pull/1343/files#diff-9e41a216e20b54abd85a2494540eaea9810b5d83f4c3efd3db9c494bbadf3cb1)
- Remove the unnecessary changes in [slvs.h](https://github.com/solvespace/solvespace/pull/1343/files#diff-63546514c917421513a9e824cd112c3a1ca3a18a8a14fcdb456b714132e376fe). What is `SLVS_NO_ENTITY`?
- [rebuild.sh](https://github.com/solvespace/solvespace/pull/1343/files#diff-1a10173fe2b8edb74ff22e2da529ec5fa18a4f4d0b137977158a5baf13b7e321) - what is this doing in the root directory? What is it doing in general? At least a short comment at the top perhaps?
- [package.json](https://github.com/solvespace/solvespace/pull/1343/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519)?!
- [pyproject.toml](https://github.com/solvespace/solvespace/pull/1343/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711) ?!

This is so unfinished that I am itching to `git reset hard HEAD~2` `git push origin --force`, but it is considered bad practice **and** it is impolite **and** I am not the maintainer , so I will not :-)

vespakoen:
@ruevs thanks for your feedback.
1. I like the suggestion of keeping the bindings in separate projects, although I am afraid that
  - It will be hardish to create the wasmlib as it needs to compile the source code in library mode which CMake, without changing the CMakeLists in the solvespace project is would probably involve a bunch of hacks to make it work.
  - Having the code in a separate repository will cause it to get out of sync more easily

2. I tried building it here with your given command, I got the same issue, tracked it down to a typo: `ENALBE_GUI` vs `ENABLE_GUI`, will fix.

3. Fair points about the unnecessary changes & rebuild.sh script, they should be removed. the `package.json` and `pyproject.toml` however are needed for the JS & Python package registries.

4. It's always possible to do a git revert, which basically undoes a commit by making a new commit, although I am not sure if we really should. I will now integrate your feedback, and experiment with moving the bindings into an external project.

ruevs:
> 2\. I tried building it here with your given command, I got the same issue, tracked it down to a typo: `ENALBE_GUI` vs `ENABLE_GUI`, will fix.

Oh! I'm blind - dyslexia is not usually a problem :-)

phkahler:
> I understand the concern, the reason is that it otherwise had to be implemented in both python and in javascript, by writing them in C they only had to be wrapped which deduplicates things a bit.

That's a python or javascript problem. Pushing stuff like that "upstream" makes it our problem. I know you plan to stick around so it can mostly be a "your" problem ;-)   Also, don't commit other changes until we know if this is getting reverted.  ;-)

Issue #1339, opened 2023-01-30
Scroll wheel no longer zooms


dannystaple:
### System information

- **SolveSpace version:** Version 3.1 (3.1~70bde63c)
- **Operating system:** macOS Monterey 12.6 (21G115)

### Expected behavior

Scroll wheel should be zooming (as it was in the previous releases)

### Actual behavior

The part is scroll up/down, and the horizontal middle is being placed at the mouse cursor.

### Additional information

This appears to be on multiple files. I was able to reproduce this with https://solvespace.com/2d.pl.

dannystaple:
No sketch needed - can be reproduced just showing the default planes.
I can see this in:
- Version 3.0 (3.0~0e0b0252)
- Version 3.0 (3.0~bcb8cd2c)
And older.  am going to reboot - in case this is me.

dannystaple:
This persisted across reboots. And is seen in older versions. I am suspecting some solvespace setting that I've not understood here.

ruevs:
In my virtual machine with macOS Mojave (10.14.6) under VirtualBox on Windows the scroll wheel still zooms just fine. I am on https://github.com/solvespace/solvespace/commit/d6e970918f5ea77757a6d21da8b11f9c66882422 built from source.

Perhaps this commit https://github.com/solvespace/solvespace/commit/cf4defcd474e0a55cfb5970265b3d1f0e99d16b0 makes a difference? @nanoant @vespakoen ?

nanoant:
@dannystaple Hi. Can you please let me know what kind of mouse you are using?

ruevs:
A generic three button (two buttons and a clicky scroll wheel) PC USB HID mouse. As I said - my macOS is a virtual machine under Windows.

dannystaple:
What is interesting is this HAD worked - my suspicion is that I've managed to modify some solvespace related setting that has changed it's behaviour, since I was able to reproduce it in far earlier versions.
The mouse I am using is a Logitech MX Ergo trackball, connected via Bluetooth.

dannystaple:
Is there a config/settings/preferences file somewhere I could blow away in case that is where this is being stored?

RM250J:
[...]

PR #1344, opened 2023-02-20
mac: Setting to explicitly choose pan or zoom for mouse wheel (#1339)

nanoant:
While auto remains the default, we now support explicit choice of doing pan or zoom when scrolling with mouse wheel or touch pad.

This can be also adapted for other (non-macOS) platforms.

### DIFF
--- a/src/confscreen.cpp
+++ b/src/confscreen.cpp
@@ -135,6 +135,13 @@ void TextWindow::ScreenChangeCanvasSizeAuto(int link, uint32_t v) {
     SS.GW.Invalidate();
 }
 
+void TextWindow::ScreenChangeMouseScrollAction(int link, uint32_t v) {
+    if(link == 'c') {
+        SS.mouseScrollAction = (SolveSpaceUI::MouseScrollAction)v;
+    }
+    SS.GW.Invalidate();
+}
+
 void TextWindow::ScreenChangeCanvasSize(int link, uint32_t v) {
     double d;
     switch(v) {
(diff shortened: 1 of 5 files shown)


vespakoen:
Wouldn't it make more sense to use something like:
```
        AUTO,
        MOUSE,
        TRACKPAD,
```
?

Or maybe I am mistaken and this is not related to https://github.com/solvespace/solvespace/issues/1339 but an extra feature?

phkahler:
I'm with @vespakoen on this. It also should not be platform specific. And pan with scroll wheel doesn't make sense because it only works vertically. Scroll means zoom.

nanoant:
@phkahler @vespakoen I can reiterate on this, what should be then the correct name for the setting?
`Pointing device (*) Auto-detect ( ) Mouse ( ) Trackpad` ?

phkahler:
I should have been clearer, but sometimes I don't want to upset contributors. I don't want scroll wheel to pan even as an option. Won't merge it.

nanoant:
@phkahler No problem for me. And yes this is a follow up for #1339 and #1093 e1b0784b31c157922be8a618a4247b26a9fd642d which started all of this by introducing panning behavior on trackpads implemented inside of `- (void)scrollWheel:(NSEvent *)nsEvent`.

Please note that the event is called "scroll wheel", not "trouch" or "track pad". So its purpose is to handle scrolling not anything else. This implementation broke my Apple Magic Mouse workflow that suddenly started panning instead of zooming. I tried to un-break it by extending too weak `nsEvent.subtype == NSEventSubtypeTabletPoint && kind == Platform::Window::Kind::TOPLEVEL` check, adding `activeTrackpadTouches > 0`. But apparently it stills trips on some weird Logitech mice setups. I tested latest f16c692803396291c12b4c0d9ed8c8bcc7af0ac7 with my two Logitech mice and some 2 Chinese unbranded one, and they all worked fine. But I don't use any Logitech customization software and I can only suspect such a software may make USB HID look like touch pad to `guimac.m` code for whatever reason.

So long story short, I am trying to make this software work for all macOS users, but I am not responsible for having #1339 guys complaining that wheel is panning instead of moving, because I did not contribute that in the first place. (Un?)fortunately I was called up with my GitHub id, because I was last person trying to fix the flawed #1093 implementation.

If #1093 was implemented using gestures maybe we had no such a problem, but the original author decided to do it within `scrollWheel`.

So I hope this is clear. And I am not upset (contributor), but I want to make sure you get the whole picture and as an ex-macOS developer who's now doing something completely else for living and doing 3D printing for fun I am trying to help 🥲

phkahler:
IMHO after trying a couple versions of solvespace on my old MBA, a setting isn't appropriate. We need to use modifier keys to differentiate the mouse button behaviors when using the trackpad. Shift, Control, Super, are all there. I'd even like if one of them were used so click-drag isn't needed since it's wierd on a trackpad. But this is just one opinion ATM.

PR #1361, opened 2023-04-20
Silence compiler warnings

thyssentishman:
When compiling with clang 13.0.0 on OpenBSD the following warning is shown:

```
warning: sprintf() is often misused, please use snprintf()
```

This commit suggests replacing all instances of `sprintf()` with `snprintf()` to silence the above mentioned compiler warning.

### DIFF
--- a/src/textscreens.cpp
+++ b/src/textscreens.cpp
@@ -120,7 +120,7 @@ void TextWindow::ShowListOfGroups() {
             if(dof > 999) {
               strcpy(sdof, "###");
             } else {
-              sprintf(sdof, "%-3d", dof);
+              snprintf(sdof, sizeof(sdof), "%-3d", dof);
             }
         }
         std::string suffix;
@@ -440,7 +440,7 @@ void TextWindow::ShowGroupInfo() {
             &TextWindow::ScreenChangePitchOption,
             g->valB != 0 ? CHECK_TRUE : CHECK_FALSE);
 
-        Printf(false, ""); // blank line    
+        Printf(false, ""); // blank line
     }
 
     if(g->type == Group::Type::EXTRUDE || g->type == Group::Type::LATHE ||
--- a/src/textwin.cpp
+++ b/src/textwin.cpp
@@ -365,7 +365,7 @@ void TextWindow::Printf(bool halfLine, const char *fmt, ...) {
         Printf(halfLine, endString);
         return;
     }
-    
+
     va_list vl;
     va_start(vl, fmt);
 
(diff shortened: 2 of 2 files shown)


src/mesh.cpp
```diff
@@ -8,6 +8,10 @@
 
 #include <set>
 
+#ifdef __OpenBSD__
+#define rand() arc4random()
```
ruevs, line 12: It seems [`arc4random`](https://man.openbsd.org/arc4random) is a "better" `rand` on OpenBSD. But why change to it? SolveSpace uses random numbers for completely "benign" things unrelated to security and I doubt it the quality of the PRNG matters at all.

thyssentishman: Actually `rand()` in OpenBSD is already some kind of wrapper to `arc4random` so really this PR is just to silence the compiler warning. I understand this might be unnecessary and would be okay leaving it out.

phkahler:
If rand() is actually deprecated I think it would be better to replace it in the source rather than do something special just for BSD.  I don't think we care too much about the quality of the random numbers, but speed and thread safety are important.

thyssentishman:
> If rand() is actually deprecated I think it would be better to replace it in the source rather than do something special just for BSD.  I don't think we care too much about the quality of the random numbers, but speed and thread safety are important.

I don't believe its deprecated, just preferred. It seems that OpenBSD modified these functions to provide more secure non-deterministic values. Since I don't know comparable and portable alternatives I think we'd be better sticking to `rand` (which is apparently a wrapper to `arc4random` anyways) and we can just ignore the compiler warning.

phkahler:
I don't want the target specific # stuff in the cpp files like that. @ruevs or @rpavlik what's the best place for that? Or should we just replace use of rand() on all platforms? I'm not sure that's right either - we want reproducible results, so rand actually preferable for us.

ruevs:
@phkahler I wouldn't merge either of these.

rpavlik:
snprintf is better than sprintf, but it's also commonly misused, see https://randomascii.wordpress.com/2013/04/03/stop-using-strncpy-already/ Ideally we do what is suggested there and have a wrapper auto-deduce buffer size and do it right in one place.

Agree that we want determinism so the rand change is not useful here: we are not relying on the random for anything security-related afaik.

neonkingfr:
I'd be curious about real use of determinism here if you're calling a random function then swapping results. sprintf() misuse is even worse than snprintf() misuse also. So code is awesome from a OS portability view if your primary target is windows.

ruevs:
> I'd be curious about real use of determinism here if you're calling a random function then swapping results.

The [`rand`](https://cplusplus.com/reference/cstdlib/rand/) function does not generate a random number. It generates a pseudo random number. If the [seed](https://cplusplus.com/reference/cstdlib/srand/) is the same the sequence of pseudo random numbers will be the same. This allows bugs that may depend on the random sequence to be reproduced.

PR #1380, opened 2023-06-30
UI: Add Ctrl+Shift+S keyboard shortuct for Save As

alufers:
Many desktop programs use this shortcut for the "Save As" dialog, so it makes sense to add it here so that different variants of parts can be quickly created.

Examples of using Ctrl+Shift+S in other programs:

Openscad:

image

LibreOffice:

image


ruevs:
@alufers thank you for the contribution.

Issue #1244, opened 2022-05-14
 Segmentation fault on language change


ktrace:
### System information

- **SolveSpace version:** 3.1~0e0b0252
- **Operating system:** gentoo/Calculate

### Expected behavior
Try to change language:
Help -> Language -> (any language include English (US))

### Actual behavior

Segmentation fault

### Additional information

Try under strace and choose Spanish, strace ended with:

```
openat(AT_FDCWD, "/usr/share/solvespace/locales/es_AR.po", O_RDONLY) = 15
newfstatat(15, "", {st_mode=S_IFREG|0644, st_size=59103, ...}, AT_EMPTY_PATH) = 0
newfstatat(15, "", {st_mode=S_IFREG|0644, st_size=59103, ...}, AT_EMPTY_PATH) = 0
lseek(15, 57344, SEEK_SET)              = 57344
read(15, "\"\nmsgid \"&OK\"\nmsgstr \"&Aceptar\"\n"..., 1759) = 1759
lseek(15, 0, SEEK_SET)                  = 0
read(15, "# Spanish/Argentina translations"..., 57344) = 57344
read(15, "\"\nmsgid \"&OK\"\nmsgstr \"&Aceptar\"\n"..., 4096) = 1759
close(15)                               = 0
write(2, "Missing (empty) translation for "..., 53) = 53
write(2, "\n", 1)                       = 1
write(2, "Missing (empty) translation for "..., 64) = 64
write(2, "\n", 1)                       = 1
write(2, "Missing (empty) translation for "..., 53) = 53
write(2, "\n", 1)                       = 1
write(2, "Missing (empty) translation for "..., 58) = 58
write(2, "\n", 1)                       = 1
write(2, "Missing (empty) translation for "..., 54) = 54
write(2, "\n", 1)                       = 1
--- SIGSEGV {si_signo=SIGSEGV, si_code=SI_KERNEL, si_addr=NULL} ---
+++ killed by SIGSEGV +++

```
bt in gdb:

```
Thread 1 "solvespace" received signal SIGSEGV, Segmentation fault.
0x000055555562c86c in SolveSpace::GraphicsWindow::EnsureValidActives() ()
(gdb) bt
#0  0x000055555562c86c in SolveSpace::GraphicsWindow::EnsureValidActives() ()
#1  0x00007ffff7d3f55e in Gtk::MenuItem_Class::activate_callback(_GtkMenuItem*) () at /usr/lib64/libgtkmm-3.0.so.1
[...]

PR #1409, opened 2023-09-19
Fix language bug

ktrace:
Fix for https://github.com/solvespace/solvespace/issues/1244

### DIFF
--- a/src/graphicswin.cpp
+++ b/src/graphicswin.cpp
@@ -300,7 +300,7 @@ void GraphicsWindow::PopulateMainMenu() {
 
                     SS.UpdateWindowTitles();
                     PopulateMainMenu();
-                    EnsureValidActives();
+                    SS.GW.EnsureValidActives();
                 });
             }
         } else if(Menu[i].fn == NULL) {


phkahler:
@ktrace  Can you offer an explanation for how this fixes the problem but was still compiling and working everywhere else?

BTW I reopened the issue since it's apparently still valid and this will close it properly ;-)

ruevs:
@phkahler @ktrace so how indeed does this change work? (I'm not on a PC so I might be missing something. Is the function called in a different scope?)

ruevs:
(writing from a computer now with the project open)

There is only one instance of the `GraphicsWindow` class which is here https://github.com/solvespace/solvespace/blob/1963a836ef8e3c1f5411d2b301c86259cbdf6fa8/src/solvespace.h#L518 and only one instance of the `SolveSpaceUI` class here https://github.com/solvespace/solvespace/blob/1963a836ef8e3c1f5411d2b301c86259cbdf6fa8/src/solvespace.cpp#L10.

So (unless there is something seriously strange going on) this pull request should not have changed anything.

PR #1411, opened 2023-09-23
Make segment selection more precise

### LINKED ISSUE #1374: Entities outside of the selection box get selected
### System information

- **SolveSpace version:** 3.1~659215d8
- **Operating system:** Win10

### Expected behavior
Only the horizontal line is selected by doing this
image

### Actual behavior
The two lines are selected.
### Additional information
Because only AABB is used for detection, that's rough
https://github.com/solvespace/solvespace/blob/fd25424ab5e5a1c4158c1e65b33b6568b5f9bc2b/src/draw.cpp#L220-L224

vespakoen:
Fixes #1374

### DIFF
--- a/src/draw.cpp
+++ b/src/draw.cpp
@@ -220,6 +220,21 @@ void GraphicsWindow::SelectByMarquee() {
         bool entityHasBBox;
         BBox entityBBox = e.GetOrGenerateScreenBBox(&entityHasBBox);
         if(entityHasBBox && entityBBox.Overlaps(marqueeBBox)) {
+            if(e.type == Entity::Type::LINE_SEGMENT) {
+                Vector p0 = SS.GW.ProjectPoint3(e.EndpointStart());
+                Vector p1 = SS.GW.ProjectPoint3(e.EndpointFinish());
+                if((!marqueeBBox.Contains({p0.x, p0.y}, 0)) &&
+                   (!marqueeBBox.Contains({p1.x, p1.y}, 0))) {
+                    // The selection marquee does not contain either of the line segment end points.
+                    // This means that either the segment is entirely outside the marquee or that
+                    // it intersects it. Check if it does...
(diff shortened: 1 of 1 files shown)


ruevs:
@vespakoen this is actually buggy. Only lines that intersect the bounding box boundaries are selected. Lines that are entirely inside the bounding box get ignored ;-)

[8192LineSegments.zip](https://github.com/solvespace/solvespace/files/12708606/8192LineSegments.zip)

BBWrongSelectionMarquee

BBWrongSelection

ruevs:
@vespakoen @phkahler I fixed it. Please take a look.

phkahler:
OMG fun. I'm reminded of the Separating Axis Theorem. In the case of a box and a line I think testing for separation using the 2 normals of the box (horizontal and vertical) might be equivalent to our bounding box test. If none of those prove it separate it is probably sufficient to check the axis perpendicular to the line segment (dy, -dx). if all 4 points of the bbox projected on that axis are on the same side of the line projected (it is just one point) then they are disjoint given that the other 2 tests were also done. Not sure this makes sense even with some further reading on SAT. Just branching the rabbit hole a bit more ;-)

Evil-Spirit:
It would be good to know that in CAD system exist two selection modes: from left to right and from right to left. One of them selects only fully-contained entities and this is very useful.

Evil-Spirit:
It would be good to know that in CAD system exist two selection modes: from left to right and from right to left. One of them selects only fully-contained entities and this is very useful.

ruevs:
> It would be good to know that in CAD system there are two selection modes exist : from left to right and from right to left. One of them selects only fully-contained entities and this is very useful.

@Evil-Spirit  we discussed it here https://github.com/solvespace/solvespace/issues/1374#issuecomment-1732564533 .

ruevs:
I merged this https://github.com/solvespace/solvespace/commit/0d26ca17f750f62fee917a88101bdb03cf5225dd

PR #1415, opened 2023-10-13
GTK: Eliminate direct references to gdk event struct members in prep …

phkahler:
…for moving to GTK4.

### DIFF
--- a/src/platform/guigtk.cpp
+++ b/src/platform/guigtk.cpp
@@ -517,8 +517,12 @@ protected:
     }
 
     bool on_motion_notify_event(GdkEventMotion *gdk_event) override {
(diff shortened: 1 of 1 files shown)


phkahler:
Not sure how to review this. It seems kind of ugly but works. Some of the functions change for GTK4 anyway, but when I was doing GTK4 changes I ran into this stuff and realized it can be done prior to switching. So here is my attempt.

phkahler:
Just reading the diff for the commit, it seems clear this is correct. The only remaining gdk_event->reference is the window state change event, and I'm not sure gtk4 even has such a thing.

Merging so I can check another box in #835

PR #1419, opened 2023-11-14
Don't copy image entites into new groups

### LINKED ISSUE #1418: Images from a sketch are "extruded" and get in the way
### System information

- **SolveSpace version:** 3.1~36ecb85b 

### Expected behavior

Images in a sketch stay in the sketch, and don't get duplicated when extruded.

### Actual behavior

Sketch images stay visible outside of the sketch, and are duplicated when extruding, making it hard to see anything (especially with an already busy model). Plus they cause z-fighting

image
image

### Additional information
I like to use images for reference and to trace organic shapes in sketches. Sometimes it's helpful to have these images show up outside of the sketch, but most of the time they just get in the way of things. I can just hide the sketch and extrude, but then I can't select any points or lines from them.

My preferred solution would be to have images be togglable (maybe with a "show images from this group" checkbox). Simply not having images show up in extrude groups is also a good solution, since I can then just hide the sketch and all the entities from it are still available from the extrude group.

phkahler:
Fixes #1418

### DIFF
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -1068,6 +1068,7 @@ void Group::CopyEntity(IdList<Entity,hEntity> *el,
 
     switch(ep->type) {
         case Entity::Type::WORKPLANE:
+        case Entity::Type::IMAGE:
             // Don't copy these.
             return;
 


ruevs (changes requested):
A better way in my opinion:
https://github.com/solvespace/solvespace/issues/1418#issuecomment-1812007290
https://github.com/solvespace/solvespace/pull/1420

phkahler:
Oof, this is failing a test...

phkahler:
We're using the other one that makes images construction by default and doesn't copy them if they are construction.

PR #1420, opened 2023-11-15
Do not include construction images in extrude/lathe/revolve/helix

ruevs:
To avoid unnecessary/annoying copies of images added in 2d sketches if they are marked "construction" they will not be copied when creating solids.

Fixes: #1418

### DIFF
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -1180,6 +1180,11 @@ void Group::CopyEntity(IdList<Entity,hEntity> *el,
             break;
 
         default: {
+            if((Entity::Type::IMAGE == ep->type) && (true == ep->construction)) {
+                // Do not copy image entities if they are construction.
+                return;
+            }
+
             int i, points;
             bool hasNormal, hasDistance;
             EntReqTable::GetEntityInfo(ep->type, ep->extraPoints,
--- a/src/mouse.cpp
+++ b/src/mouse.cpp
@@ -1138,6 +1138,7 @@ void GraphicsWindow::MouseLeftDown(double mx, double my, bool shiftDown, bool ct
                     AddToPending(hr);
                     Request *r = SK.GetRequest(hr);
                     r->file = pending.filename;
+                    r->construction = true;
 
                     for(int i = 1; i <= 4; i++) {
                         SK.GetEntity(hr.entity(i))->PointForceTo(v);


phkahler:
@ruevs  Would it be too messy to add a separate case for IMAGE ?  I'm not sure how much of the default stuff applies to them.

ruevs:
> @ruevs Another thing... I tried 3.1 on Windows and can't select an image to make it construction. There's also no way to visually tell if an image is construction or not.
> 
> Oh I see in the video. You have to select the image in the text window and it will highlight. Then you can right-click to select construction. I suppose this is deliberate because people want to trace over an image and don't want it keep highlighting and getting selected...

For cross-reference - here is the video in question https://github.com/solvespace/solvespace/issues/1418#issuecomment-1812007290

> @ruevs Would it be too messy to add a separate case for IMAGE ? I'm not sure how much of the default stuff applies to them.

Well I thought of this, but the "default stuff" is what copies the corner points of the image entity in the 3d groups. And I thought those could always be useful and do not get in the way much. It can easily be changed of course.

jkrei0:
I'm not sure what use case there is for keeping images in the extrude by default (or at all); it seems more reasonable to me to have them hidden by default. And if you *really* want the images, you can always add them back in a new sketch.

I don't dislike the idea of having images be togglable, but it would make more sense to have it either be a checkbox in the text window (not sure how feasable that is to implement), or to have the default be to not copy images.

This approach (of making "construction" images not be copied) feels a bit unintuitive and hidden, especially since (if I understand correctly) you have to select images from the text window when the sketch group is active and then right-click in the graphics area, and since in all other cases, construction entities *are* copied.

phkahler:
@ruevs  I was going to suggest having images be construction entities by default, which is backwards compatible and does what we want by default with your fix.  But I'm with @jkrei0 in thinking this is a huge hack that changes the meaning of "construction" just for images, while also being nearly impossible to find.

I'm not sure how many people use images, but I'm guessing far less want the image to get copied. They can manually add extra copies if they want.

ruevs:
> This approach (of making "construction" images not be copied) feels a bit unintuitive and hidden, especially since (if I understand correctly) you have to select images from the text window when the sketch group is active and then right-click in the graphics area, and since in all other cases, construction entities _are_ copied.

When you are in a 2d (sketch) group the face selection is disabled image by default. If you enable it images can be selected. Here is the code: https://github.com/solvespace/solvespace/blob/e7c0c1665f1684bb3195107147aaf254c852fa44/src/draw.cpp#L448)

After selecting them pressing `G` toggles construction - just like for any other entity. No need to right-click.

I did it through the text window in the video just to make it explicitly obvious that you have to select the original image in the sketch and not one of its copies in the extruded object. For the same reason I showed that when the extrusion is "two sided" you can select the original image from the 3d (extrude) group. And for the same reason I used the right-click menu.

phkahler:
> Whoever has gotten to the point of using images to "trace over" and is annoyed by their replication (if we do not make construction the default) - should read a bit of manual :-)

My understanding is that tracing over is THE reason for images in sketches. The replication must be something people have been annoyed with for years but never complained about. My opinion is that they should be flagged as construction by default regardless, so other parts of the code don't have to filter them now or in the future. But I also think the copies should be unconditionally squashed rather than only suppressed for construction images.

BTW, faces and images should not be selectable in a 2d mode because they'd keep lighting up which would be really annoying. I only brought up the fact that they don't highlight and can't be right-click toggled to construction to point out the difficulty of using their construction flag, not to suggest we change it.

So yes to making them construction by default, and no to letting them get copied at all.

phkahler (commented):
Good enough. I'm still not sure we should ever copy images but you'll have to try real hard to do so with this ;-)

PR #1424, opened 2023-11-22
CTRL-TAB un/hides the toolbar

phkahler:
Becasue I Shift-tab didn't work...

I don't know if we want this, but it fits with TAB hiding the text window.


ruevs:
CTRL-TAB has a "standard" meaning of cycling through tabs in "tabbed" interfaces... But since we do not have any tabs I think it is OK to use it to hide the toolbar (once and forever ;-).

@jwesthues what do you think? I find your judgement on UI very good.

jwesthues:
Seems reasonable to me. Hiding the toolbar seems much less important than hiding the text window (since it occupies less area, and all toolbar items are also accessible via menus and keyboard shortcuts); so a more complex shortcut seems appropriate.

PR #1425, opened 2023-11-25
Dimesions on arcs defaut to radius instead of diameter

phkahler:


phkahler:
This is a fundamental change in behavior, but I think it's probably how it should have been all along.

BKLronin:
I liked the consistency that all round elements are in diameter ( arc, cir, tan etc). This helps to avoid errors while drawing. 

My 0,2 cent

ruevs:
A strong "no" from me on this change as well.

mkyral:
If this happen, I prefer to be configurable.

phkahler:
After consulting with some professional CAD users about this, it is almost always the case that they want circles to have a diameter and arcs to have a radius. It also turns out that drawing standards generally follow that as well.

I'll try to get around to making this a configure option - in which case the default behavior is going to be arcs-use-radius. If not it's gonna go in some time in the near future anyway.

On a slightly related note, issue #1428 opened my eyes. I can imagine "real" CAD users trying to dimension an arc by prefixing the dimension with an R and having Solvespace crash. The fact that we've gone this long without that bug report makes me think a number of things. But it also reinforced my opinion on this PR.

BKLronin:
Option sounds cool with arc as standard. 

-Sincerely, 
a real CAD user :D

phkahler:
I've added a configuration option to allow the old behvior. Also added a bunch of tags to this PR to maybe help find it in the future as an example of how to add simple configuration option with a checkbox to SolveSpace.

phkahler:
Yay! The CI check for macOS passed!  Woot!

ruevs:
Looks good (reading from a phone) I won't be able to test/review it properly before the 26-th. Happy holidays :-)

PR #1426, opened 2023-11-26
Keep old entities when splitting lines/curves

jkrei0:
I mostly did this as an exercise to introduce myself to working with Solvespace's code. I've probably managed to do something weird or overlook something, and I don't claim to know what other peoples' use cases are. No hard feelings if this isn't something you want merged :)

This fixes a couple issues I had with it:
- `Tangent arc at point` marks old entities as construction, and this (new way) is more consistent with that
- Deleting the old entities means you lose their constraints, and lose anything else that depends on them* (you can mark them as construction to keep them as-is, but you still have to replace the coincident constraints on the endpoints. And I didn't even know you could do this until I looked through the code)
- If you don't realize you're deleting the old entities (if you expect it to behave like `tangent arc at point`, or if you don't realize that you lose the old coincident constraints), you risk messing up your sketch/model.

**This is somewhat off topic, but it feels way to easy to accidentally delete a group without realizing it, unless you're paying close attention to all the dialog boxes. Not sure what good solution there would be, though.*

The problems I still have with the new approach:
- There's no way to not keep the old entities - but this isn't really a problem, because you can just delete the old ones afterwards if you don't want them.
- New entities are constrained to the *endpoints* of the old ones, but not constrained to them at the point where they meet. I would have tried to add this except for that there's no "point on spline" constraint, which would mean that only some parts could be constrained.
- If you split the same entity a bunch of times, you get a bunch of extra entities. This becomes pretty annoying unless you keep on top of deleting the ones you don't want, and it becomes *really* messy if you start dragging points around. Idk how many people are going around doing things like this. Perhaps this is why the current behavior is the way it is. 

And a weird thing, that happens (and should probably be its own issue), is that splines get split at their control points, no matter what. Or is that intentional/neccesary?

### DIFF
--- a/src/modify.cpp
+++ b/src/modify.cpp
@@ -461,8 +461,8 @@ hEntity GraphicsWindow::SplitLine(hEntity he, Vector pinter) {
     SK.GetEntity(ei1->point[0])->PointForceTo(pinter);
     SK.GetEntity(ei1->point[1])->PointForceTo(p1);
 
-    ReplacePointInConstraints(hep0, e0i->point[0]);
-    ReplacePointInConstraints(hep1, ei1->point[1]);
+    Constraint::ConstrainCoincident(hep0, e0i->point[0]);
+    Constraint::ConstrainCoincident(hep1, ei1->point[1]);
     Constraint::ConstrainCoincident(e0i->point[1], ei1->point[0]);
     return e0i->point[1];
 }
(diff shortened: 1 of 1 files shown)


ruevs:
> * (you can mark them as construction to keep them as-is, but you still have to replace the coincident constraints on the endpoints. And I didn't even know you could do this until I looked through the code)

This is documented here https://solvespace.com/ref.pl#Splitting but it is not one of the "discoverable" features :-)

Writing from a phone, my inclination would be to perhaps keep the extra coincident constraints you addad, but leave the construction behaviour as it is. I'll look into it from a PC.

jkrei0:
I agree with @ruevs that even if old entities aren't kept by default, they should at least keep their endpoint constraints. (I still think they should be kept by default, but at least then it's two clicks and a keypress either way)

It seems to work to [just do this](https://github.com/jkrei0/solvespace/tree/constrain-split-entities) if you want to keep the old behavior, and let whatever code deletes the old entities remove the redundant constraints when the entities aren't kept?

ruevs:
I need to play with this a little bit when I find some time. I'll try both variants. For now I can not give an opinion.

ruevs:
@jkrei0 @phkahler I tried both. Definitely https://github.com/jkrei0/solvespace/tree/constrain-split-entities for me.
The "piling" construction entities (when I did not intentionally do them) that this one creates truly bother me.

jkrei0:
It seems the constraints only version has a bug :(
If you create a spline, click so the end point is on the start point (so it makes a loop), then try to split it with a line, it crashes.
This doesn't happen with the version that always keeps entities, nor does it happen with solvespace w/o any of these changes.

Issue #1428, opened 2023-11-29
If I write any uppercase letter (for example R15) into dimension input box solvespace crash


dadoprom:
### System information

- **SolveSpace version:** 3.1~727b0153 - Latest from git
- **Operating system:** Lubuntu 22.04.3

### Expected behavior

I sketched a circle and wanted to add dimension as radius with R15 in the dimesion input box, this is not working

### Actual behavior

If I write any uppercase letter (for example R15) into dimension input box solvespace crash

### Additional information

Solvespace does not crash when you write lowercase letters into dimension input box

phkahler:
@dadoprom  This is a known issue and will be fixed when we allow more arbitrary input for #77 Parametric sketches.

If you want to enter a radius instead of a diameter, click on the dimension (just select it, don't double click to edit. The dimension will show in the text window and a checkbox will allow you to change it from diameter to radius.

dgramop:
Right now solve space does crash (like the process dies) when you do this. I was able to replicate

If your dimension starts with a letter, then it crashes. Otherwise it gives you the popup as expected

Action:
Type in `R15` into a dimension box.

Actual behavior: 
Solvespace crashes

Expected behavior: A popup telling you you can't put a letter there
<img width="295" alt="Screenshot 2023-12-06 at 12 57 10 PM" src="https://github.com/solvespace/solvespace/assets/10949560/bd6c53d5-da51-4830-8801-0b2acf27430f">

Issue #80, opened 2016-10-14
Hierarchical sketches


whitequark:
The Holy Grail of version 3.0 would be hierarchical sketches, that is, the ability to link one sketch into another while updating some of the parameters in the linked sketch.

Depends on #77 (since we want to update parameters) and #78 (since we want to generate one sketch while having another open).

hinell:
Wow 6 years later and still not implemented. Sad!

Issue #75, opened 2016-10-14
New file format


whitequark:
SolveSpace would benefit from a file format that suits the following required properties:
  * compact, with a binary representation;
  * has an alternative text representation;
  * fast to read and write;
  * easily extensible, while being backwards compatible;
  * easily supports deeply nested structures, such as for hierarchical sketches.

In addition, an optional but desirable properties would be:
  * easily readable and perhaps even writable from 3rd party software.

The current file format fulfills none of the above (not even the last requirement, since there is no explicit schema). I propose to replace it with a format based on [Google Protocol Buffers](https://developers.google.com/protocol-buffers/docs/overview), which has the following properties:
  * the binary representation is naturally compact--it writes your data and generally does no special tricks during it, except that it defines a variable-length encoding for integers;
  * the text representation is available alongside binary representation, and requires no manual work;
  * has a highly optimized reader and writer implementation in C++;
  * is literally designed because of a desire to handle protocol evolution well;
  * is naturally nested;
  * has a separate schema and very broad library support.

Such a format, with the initial version closely resembling the existing file format, seems like an ideal choice for us.

whitequark:
@jwesthues Any objections? Disregard my earlier thoughts about using LLVM bitcode; after an in-depth discussion with LLVM folks that turned out to be a rather poor choice.

whitequark:
The new file format should also use 64-bit handles (as would the in-memory representation). The rationale is that, once we make large assemblies fast, remapping of entities can rapidly exhaust the 65535-entity limit for one group, if one imports many instances of a large sketch. This migration is an ideal opportunity to add the 32→64 bit expansion code robustly.

jwesthues:
That seems generally reasonable. It would be fashionable to use JSON or some other text representation today; but especially if you optimize assemblies to make better use of the GPU, a binary format's speedup seems material.

whitequark:
Actually, the version 3 of Protocol Buffers has a canonical JSON serialization, so we could just use that. (Google generally recommends version 2, for reasons that I do not fully understand; but 3 is probably fine too.)

whitequark:
[...]

PR #1432, opened 2023-12-06
Address crash described in #1428

dgramop:
If a user enters an uppercase letter as the first character of a constraint, SolveSpace crashes (#1428). A user did this by accident when trying to enter a radius measurement and put a dimension like "R15".

This resulted from some code scaffolding for variable support (#77).

### DIFF
--- a/.github/scripts/install-macos.sh
+++ b/.github/scripts/install-macos.sh
@@ -1,8 +1,8 @@
 #!/bin/sh -xe
 
 if [ "$1" = "ci" ]; then
-    armloc=$(brew fetch --bottle-tag=arm64_big_sur libomp | grep -i downloaded | grep tar.gz | cut -f2 -d:)
-    x64loc=$(brew fetch --bottle-tag=big_sur libomp | grep -i downloaded | grep tar.gz | cut -f2 -d:)
+    armloc=$(brew fetch --bottle-tag=arm64_ventura libomp | grep -i downloaded | grep tar.gz | cut -f2 -d:)
+    x64loc=$(brew fetch --bottle-tag=ventura libomp | grep -i downloaded | grep tar.gz | cut -f2 -d:)
     cp $armloc /tmp/libomp-arm64.tar.gz
     mkdir /tmp/libomp-arm64 || true
     tar -xzvf /tmp/libomp-arm64.tar.gz -C /tmp/libomp-arm64


phkahler:
@dgramop    A few comments. First, we don't allow merge commits. Your PR should address one problem.

As for this specific issue/PR I believe the correct fix is to remove the entire isupper condition, not to make it fail differently. I believe there is also a list (or vector) of variables declared that is never used and should also be removed.  The next issue is that there is a test for variable creation in the test sute, which is failing with this change. And lastly, because this is in the expression parser it would be good to verify that our "parameters" branch can still be easily rebased on master after the PR (it should be possible to easily rebased it on your branch here).

I tried to remove the isupper check once, and then failed to fix the tests, and decided to leave it until we merge the parameters branch. But now I see how easy it would be for someone new to SS to type R50 as a dimmension, crash it, and walk away. I don't want that to happen so I see this as a priority to fix.

dgramop:
Fixing this was really just an excuse for me to take a look under the hood. You can go ahead and close the PR.

 It looks like the parameters branch hasn't been touched since the summer. Parameters is a cool feature that I would love to use in Solvespace. To that end, how can I make myself useful?

Should I go and try to devise a new file format (#75), does anyone have any opinions/draft specs for that. Or should I leave the parameters branch alone and just wait it out? Just looking to help.

phkahler:
> @phkahler My bad about the dirty branch. Fixing this was really just an excuse for me to take a look under the hood. You can go ahead and close the PR.

@dgramop  I really appreciate people taking a look and trying to make SS better!

> It looks like the parameters branch hasn't been touched since the summer. Parameters is a cool feature that I would love to use in Solvespace. To that end, how can I make myself useful?

All I did was rebase the evil-spirit changes on master. That stuff is many years old, but we need more work. Particularly is defining parameters (perhaps via equation constraints similar to comments) and well defined scope rules that can be expanded later (when we eventually get #80 going. My goal was to just use that branch, which adds names to some parameters rather than having a separate list of variables.

> Should I go and try to devise a new file format (#75), does anyone have any opinions/draft specs for that. Or should I leave the parameters branch alone and just wait it out? Just looking to help.

No to the file format. In fact I should close that. I think we can do a lot more without changing the file format, and future changes will require a format change anyway and we can't really predict what that will look like in advance.  Until then we can add new features that won't work with older versions, but new versions can still read the old files.

There are a lot of open issues covering a lot of areas of the code. Is there something you're really interested in? It seems like you've got an interest in the parameters branch. The discussions on that are in a couple places. It's also something I consider important but don't have time for myself ;-) I rebased the existing work primarily to encourage someone to continue it.

Issue #1372, opened 2023-05-30
[Bug] Paste mirror with pt-line distance


liuxilu:
### System information
- **SolveSpace version:** 3.1~659215d8
- **Operating system:** Win10
### Reproduce Steps
Draw this, select all, click paste transformed, select flip, click paste now
![image](https://github.com/solvespace/solvespace/assets/53636947/5bae742b-6d53-404f-95cf-62e898ec6fe3)

### Expected behavior
Mirrored sketch
### Actual behavior
A mess
### Additional information
Attached where I found the bug.
[PasteMirrorTest.txt](https://github.com/solvespace/solvespace/files/11600124/PasteMirrorTest2.txt)

ruevs:
Can not reproduce with the example picture. See attached [1372_PasteTransformedTest.zip](https://github.com/solvespace/solvespace/files/11610917/1372_PasteTransformedTest.zip).

However with your file (https://github.com/solvespace/solvespace/files/11600124/PasteMirrorTest2.txt) the pasted copy does "scramble".

liuxilu:
In your file there is a pt-pt distance. I have replaced a clearer image.

ruevs:
> it is the signed distance.

Aha! Makes sense. Point-line distances are "signed" internally, so they preserve on which side of the line the point is... I can not answer immediately how the paste transformed code works in this case and what is the right thing to do. @phkahler has looked at these parts of the code more than me and may be able to give some insight.

liuxilu:
Seems we can just reverse the sign.

PR #1443, opened 2024-01-03
fix pt-line distance constraint bug on paste transformed

Not4sure:
fixes issue #1372
As mentioned in issue, this bug is caused by using absolute value of `scale` when calculating constraints in paste transformed. 
I removed `fabs()` exactly for pt_line_distance case.
Tested for different geometry and it works ok now.
screen_record

I'm quite new to open-source contributions on GitHub, so excuse me if i missed something or did something wrong.

### DIFF
--- a/src/clipboard.cpp
+++ b/src/clipboard.cpp
@@ -249,8 +249,10 @@ void GraphicsWindow::PasteClipboard(Vector trans, double theta, double scale) {
             case Constraint::Type::COMMENT:
                 c.disp.offset = c.disp.offset.Plus(trans);
                 break;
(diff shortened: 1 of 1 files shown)


phkahler:
> I'm quite new to open-source contributions on GitHub, so excuse me if i missed something or did something wrong.

@Not4sure Looks like you've got a year and a half of heavy github contributions. IMHO Solvespace is one of the more formidable projects to get into since there are several very complex inter-connected pieces and a lot of math. Good find with this one. Clearly a pt-pt distance should always be positive. Even if this caused some unexpected issue it's only in copy/paste and won't break existing sketches.

Issue #1406, opened 2023-09-12
Created a Qt 5 port for solvespace (would like to provide source) 


shingen75:
### System information
- **SolveSpace version:** 3.1~; 
- **Operating system:** Windows ( 10 and 11) / Linux ( CentOS 7 )

I have ported the solvespace user interface to Qt and I have tested all the basic functionality and I believe everything works correctly. 
I have updated the CMakefiles and added an option to build in Qt mode (USE_QT_GUI). An additional library tiny2xml ( added to the 
 extlib and referenced using extlib/tinyxml2/CMakeLists.txt when in Qt mode) was used to save/load the configuration file solvspace.conf (preserve the settings). I was wondering if it would be possible to provide you with the Qt code.

ruevs:
@shingen75  his did you port it? I do not see a fork in your GitHub user? If you provide a link to your fork or create a pull request we can look at your port.

phkahler:
@shingen75  The approach should be to create a new file "guiqt.cpp" in the platform folder.  QT should only be used as a GUI on Linux, and only as an option with the default being GTK.  As ruevs said, create a fork, make a branch for your QT stuff and let us know when it's ready or create a PR.

I'm curious.
1) why did you do this QT version? 
2) Why a new way to save the settings?
3) does it work on X?
4) does it work on Wayland?

ruevs:
> 2\. Why a new way to save the settings?

I was wondering the same, but decided to take it a step at a time.

shingen75:
> @shingen75 how did you port it? I do not see a fork in your GitHub account? If you provide a link to your fork or create a pull request we can look at your port.

@ruevs 
I am sorry, I was not aware of the procedure to do the port on github. I use git internally at work but I have not used github for project development. I will fork the repository and put the port work in it. I will do this over the weekend.

shingen75:
> @shingen75 The approach should be to create a new file "guiqt.cpp" in the platform folder. QT should only be used as a GUI on Linux, and only as an option with the default being GTK. As ruevs said, create a fork, make a branch for your QT stuff and let us know when it's ready or create a PR.
> 
> I'm curious.
> 
> 1. why did you do this QT version?
> 2. Why a new way to save the settings?
> 3. does it work on X?
> 4. does it work on Wayland?

@phkahler 
[...]

PR #1451, opened 2024-03-01
Add Qt 5 GUI backend

WickedSmoke:
This builds upon the work of @shingen75 in issue #1406.  There is a fair amount of cleanup, the scroll bar was implemented, and a few serious problems fixed (commits 1cc39e3a & 470d1107).  It has only been built & run on Linux (KDE).


WickedSmoke:
The code compiles and runs with Qt 6, albeit with many warnings as that requires C++-17.   Most of the warnings seem to be from this iterator in solvespace.h:

```
78% Compile src/textscreens.cpp
In file included from src/undoredo.cpp:8:
src/solvespace.h:182:28: warning: ‘template<class _Category, class _Tp, class _Distance, class _Pointer, class _Reference> struct std::iterator’ is deprecated [-Wdeprecated-declarations]
  182 | class utf8_iterator : std::iterator<std::forward_iterator_tag, char32_t> {
      |                            ^~~~~~~~
In file included from /usr/include/c++/12/bits/stl_construct.h:61,
                 from /usr/include/c++/12/bits/alloc_traits.h:33,
                 from /usr/include/c++/12/ext/alloc_traits.h:34,
                 from /usr/include/c++/12/unordered_map:41,
                 from /usr/include/c++/12/functional:61,
                 from src/resource.h:10,
                 from src/solvespace.h:10:
/usr/include/c++/12/bits/stl_iterator_base_types.h:127:34: note: declared here
  127 |     struct _GLIBCXX17_DEPRECATED iterator
      |                                  ^~~~~~~~
```

CMakeLists.txt
```diff
@@ -34,34 +71,36 @@ endif()
...
 # and instead uncomment the following, adding the complete git hash of the checkout you are using:
 # set(GIT_COMMIT_HASH 0000000000000000000000000000000000000000)
 
 string(SUBSTRING "${GIT_COMMIT_HASH}" 0 8 solvespace_GIT_HASH)
-project(solvespace
+project(solvespaceQt
     VERSION 3.1
     LANGUAGES C CXX ASM)
 
+set(USE_QT_GUI        OFF CACHE BOOL
```
ruevs, line 83: The [CMakeLists.txt](https://github.com/solvespace/solvespace/pull/1451/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a) is not acceptable in this state. As it is all [checks](https://github.com/solvespace/solvespace/pull/1451/checks) fail.

The QT version needs to be an optional _non_ default target on any of the three platforms supported by QT.

Unless this option is set SolveSpace should continue to build just like before on all four platforms using the native/GTK GUI (Windows, Linux, macOS, Web/Emscripten).

src/platform/gui.h
```diff
@@ -382,7 +382,7 @@ FileDialogRef CreateSaveFileDialog(WindowRef parentWindow);
 std::vector<Platform::Path> GetFontFiles();
 void OpenInBrowser(const std::string &url);
 
-std::vector<std::string> InitGui(int argc, char **argv);
+std::vector<std::string> InitGui(int& argc, char **argv);
```
ruevs, line 385: Why `int& argc`?

src/platform/guigtk.cpp
```diff
@@ -1550,7 +1550,7 @@ void OpenInBrowser(const std::string &url) {
 
 Gtk::Main *gtkMain;
 
-std::vector<std::string> InitGui(int argc, char **argv) {
+std::vector<std::string> InitGui(int& argc, char **argv) {
```
ruevs, line 1553: Why `int& argc`?

src/platform/guimac.mm
```diff
@@ -1635,7 +1635,7 @@ - (void)applicationTerminatePrompt {
 
 static SSApplicationDelegate *ssDelegate;
 
-std::vector<std::string> InitGui(int argc, char **argv) {
+std::vector<std::string> InitGui(int& argc, char **argv) {
```
ruevs, line 1638: Why `int& argc`?

src/platform/guinone.cpp
```diff
@@ -128,7 +128,7 @@ std::vector<Platform::Path> GetFontFiles() {
 
 void OpenInBrowser(const std::string &url) {}
 
-std::vector<std::string> InitGui(int argc, char **argv) {
+std::vector<std::string> InitGui(int& argc, char **argv) {
```
ruevs, line 131: Why `int& argc`?

src/platform/guiwin.cpp
```diff
@@ -1688,7 +1688,7 @@ void OpenInBrowser(const std::string &url) {
     ShellExecuteW(NULL, L"open", Widen(url).c_str(), NULL, NULL, SW_SHOWNORMAL);
 }
 
-std::vector<std::string> InitGui(int argc, char **argv) {
+std::vector<std::string> InitGui(int& argc, char **argv) {
```
ruevs, line 1691: Why `int& argc`?

src/platform/guihtml.cpp
```diff
@@ -1432,7 +1432,7 @@ void OnSaveFinishedCallback(const Platform::Path& filename, bool is_saveAs, bool
     EM_ASM(saveFileDone(UTF8ToString($0), $1, $2), filename_str.c_str(), is_saveAs, is_autosave);
 }
 
-std::vector<std::string> InitGui(int argc, char **argv) {
+std::vector<std::string> InitGui(int& argc, char **argv) {
```
ruevs, line 1435: Why `int& argc`?

src/CMakeLists.txt
```diff
@@ -34,7 +43,7 @@ target_link_libraries(slvs_deps INTERFACE
     ${ZLIB_LIBRARY}
     ${PNG_LIBRARY}
     ${FREETYPE_LIBRARY}
-    ${CAIRO_LIBRARIES}
+    #${CAIRO_LIBRARIES}
```
ruevs, line 46: Why?

src/CMakeLists.txt
```diff
@@ -111,68 +120,131 @@ set(every_platform_SOURCES
...
-    srf/boolean.cpp
-    srf/curve.cpp
-    srf/merge.cpp
-    srf/ratpoly.cpp
-    srf/raycast.cpp
-    srf/shell.cpp
-    srf/surface.cpp
-    srf/surfinter.cpp
-    srf/triangulate.cpp)
+if (NOT USE_QT_GUI)
```
ruevs, line 123: `solvespace-core` is cross platform.
If this `if` was actually needed the reason it was needed should be removed.

src/CMakeLists.txt
```diff
@@ -262,128 +334,186 @@ endif()
...
-
-        target_sources(solvespace PRIVATE
-            platform/guihtml.cpp)
-
-        string(REPLACE ";" " " LINK_FLAGS "${LINK_FLAGS}")
-        set_target_properties(solvespace PROPERTIES
-            LINK_FLAGS "${LINK_FLAGS}")
-        set_source_files_properties(platform/guihtml.cpp PROPERTIES
-            OBJECT_DEPENDS ${SHELL})
+    if(NOT USE_QT_GUI)
```
ruevs, line 337: Definitely not like this. The QT conditional things should be integrated "fine grained" instead of a "big" if block and a bunch of copy-paste.
Take a look how `EMSCRIPTEN` was done - it was the lat platform added.

ruevs (changes requested):
The changes in
[CMakeLists.txt](https://github.com/solvespace/solvespace/pull/1451/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a)
and
[src/CMakeLists.txt](https://github.com/solvespace/solvespace/pull/1451/files#diff-148715d6ea0c0ea0a346af3f6bd610d010d490eca35ac6a9b408748f7ca9e3f4)
are pretty intrusive and definitely should be reworked. I'm not particularly skilled with CMake but please refer to how the `EMSCRIPTEN` changes were integrated (look in the log) - it was the latest platform added.

In addition the following files should NOT be modified by this pull request:
[gui.h](https://github.com/solvespace/solvespace/pull/1451/files#diff-1019b80ded2dbac080c4822f9f56c567e0bdbc3593b572001a947e9d5a6a96c9)
[guigtk.cpp](https://github.com/solvespace/solvespace/pull/1451/files#diff-a6e225ca5140d76dab5de13c0865b26db963df9ab932d5dd54691f176e005005)
[guihtml.cpp](https://github.com/solvespace/solvespace/pull/1451/files#diff-f7ffbcfa308037f376f5289cbee7ba9d9fef8467eb49a9bb326e0ff1f6d67098)
[guimac.mm](https://github.com/solvespace/solvespace/pull/1451/files#diff-739d366a89de9497c8a7a4263a760117f153f478f1aa06719f4cf15609b27d92)
[guinone.cpp](https://github.com/solvespace/solvespace/pull/1451/files#diff-d6a26a7591c70e806fe81eda5f573ed810af8a067fb87023faad07d41d90107a)
[guiwin.cpp](https://github.com/solvespace/solvespace/pull/1451/files#diff-27780ee70b90a85bcb6991c38072421d7dceb16c7ba485e506e4e9f5ea02b38c)
Please revert those changes.

ruevs:
> The code compiles and runs with Qt 6, albeit with many warnings as that requires C++-17. Most of the warnings seem to be from this iterator in solvespace.h:
> 
> ```
> 78% Compile src/textscreens.cpp
> In file included from src/undoredo.cpp:8:
> src/solvespace.h:182:28: warning: ‘template<class _Category, class _Tp, class _Distance, class _Pointer, class _Reference> struct std::iterator’ is deprecated [-Wdeprecated-declarations]
>   182 | class utf8_iterator : std::iterator<std::forward_iterator_tag, char32_t> {
>       |                            ^~~~~~~~
> In file included from /usr/include/c++/12/bits/stl_construct.h:61,
>                  from /usr/include/c++/12/bits/alloc_traits.h:33,
>                  from /usr/include/c++/12/ext/alloc_traits.h:34,
>                  from /usr/include/c++/12/unordered_map:41,
>                  from /usr/include/c++/12/functional:61,
>                  from src/resource.h:10,
>                  from src/solvespace.h:10:
> /usr/include/c++/12/bits/stl_iterator_base_types.h:127:34: note: declared here
>   127 |     struct _GLIBCXX17_DEPRECATED iterator
>       |                                  ^~~~~~~~
> ```

This is known. [`std::iterator` was deprecated in C++17](https://timsong-cpp.github.io/lwg-issues/2438), but at this point SolveSopace is strictly C++11. In C++11 it is not deprecated and the warning does not happen.

Can the QT version be compiled in C++11 mode? If not the warning is benign anyway.

CMakeLists.txt
WickedSmoke: I don't know CMake well (all that is from shingen75), so it may take some time to cleanup things.

I'm currently building solvespace by other means as I was unable to get a debug build using cmake (hence https://github.com/solvespace/libdxfrw/pull/14).  I tried this...

    cmake -DOPENGL=3 -DUSE_QT_GUI=ON -DCMAKE_BUILD_TYPE=Debug ..

...but I don't get any symbols in the debugger.

src/platform/gui.h
WickedSmoke: The abstraction layer gets in the way of normal QApplication usage where the main argc and argv are edited to remove any Qt specific options.  See https://doc.qt.io/qt-5/qguiapplication.html#QGuiApplication

WickedSmoke:
> Can the QT version be compiled in C++11 mode? If not the warning is benign anyway.

C++17 is a requirement for Qt 6, not Qt 5.

src/platform/gui.h
ruevs: When I have some time I'll install QT 5 and try to clean this up. It should not be a problem to do it, but I do not want to give suggestions without a compiler :-)

WickedSmoke: There's so little going on in entrygui.cpp that it seems cleaner to just have each platform provide this.   A pure Qt version would be:

```
int main(int argc, char** argv) { 
    QApplication app(argc, argv);
    Platform::Open3DConnexion();
    SS.Init();
    
    if(argc >= 2) {
        if(argc > 2) {
            dbp("Only the first file passed on command line will be opened.");
        }
        SS.Load(argv[1]);
    }

    app.exec();

    Platform::Close3DConnexion();
    SS.Clear();
    SK.Clear();
    return 0;
}
```

This eliminates InitGui(), RunGui(), ClearGui() and the argument string vector.

WickedSmoke:
The "Fix QApplication::exec() crash." commit has been modified to not touch the InitGui() function.  Instead it bypasses the InitGui/RunGui/ClearGui interface and simply provides main itself (as suggested in my post last night).

It looks like there's a bug in entrygui.cpp where the last argument is passed to SS.Load rather then the first.

phkahler:
> > Can the QT version be compiled in C++11 mode? If not the warning is benign anyway.
> 
> C++17 is a requirement for Qt 6, not Qt 5.

I don't have a problem with C++17, the GTK4 version will also require it. The question is if our CI/CD stuff is up to date to allow it on all platforms.

WickedSmoke:
I've gone over (and modified) most of the Qt code now and think it's in a decent shape to merge.

I just wrote a .spec file and can make [RPMs on Copr](https://download.copr.fedorainfracloud.org/results/wickedsmoke/SolveSpace/) (including the -devel package).  The .desktop file is broken as it does not reference the Qt binary, which is currently solvespaceQt.  I'd prefer to rename it to solvespace-qt.  Does that sound OK?

WickedSmoke:
Fedora provides mimalloc (2.1.2 on Fedora 37).   I'm using that on my debug builds and it seems to work.  Is there any particular reason to use the extlib/ version?

phkahler:
> Fedora provides mimalloc (2.1.2 on Fedora 37). I'm using that on my debug builds and it seems to work. Is there any particular reason to use the extlib/ version?

We had some tricky issues and settled on a specific version that worked. If you really want to test a version it's best to turn on address sanitizer and run for a while.

ruevs (changes requested):
@WickedSmoke the Qt port now looks much better. Thank you for your contributions. I tested it on Windows and it compiles and works with a small change. Since you have not enabled maintainer commits in your branch I made a pull request against it here https://github.com/WickedSmoke/solvespace_qt5/pull/2.

Please merge it into your branch so that this pull request gets updated.

Remaining things to do:
- [ ] Try it on macOS - I only have a virtual machine with macOS, which is slow and a bit limited. Perhaps @vespakoen could try it.
- [ ] Check how the settings storage on Windows (in the registry) interacts with the native version. I'll do this. Currently native goes in `HKEY_CURRENT_USER\SOFTWARE\SolveSpace` and Qt ends up in `HKEY_CURRENT_USER\SOFTWARE\SolveSpace\solvespace`
- [ ] On Windows the application icon and window title in the top left corner are missing in the Qt version (but this is very minor and in my opinion not a problem for merging this since Qt will not be the official port) 
image
image
How is it on Linux?
- [ ] Figure out how to statically link Qt. As it is the Qt port requires Qt5Core.dll, Qt5Gui.dll and Qt5Widgets.dll to run - a total of 17.2 MB of runtime. But again - this is not a show stopper.
- [ ] I still need to actually review the code in guiqt.cpp and guiqt.h

WickedSmoke:
> * On Windows the application icon and window title in the top left corner are missing

It shows up on Linux because of these lines:
```
#ifdef __linux
        QIcon icon("/usr/share/icons/hicolor/48x48/apps/solvespace.png");
        setWindowIcon(icon);
#endif
```

> Once I un-dock the property browser I can not figure out how to dock it again

You should be able to drag the window over the left or right sides of the main window to see the docking area appear.

I would give up on the idea of settings compatibility.  The dock settings are stored using custom Qt entries (using `saveGeometry` & `saveState`).

CMakeLists.txt
dgramop: @WickedSmoke To get symbols in the debugger, just making the following changes:

diff --git a/cmake/c_flag_overrides.cmake b/cmake/c_flag_overrides.cmake
index 978b6b65..014d7ada 100644
--- a/cmake/c_flag_overrides.cmake
+++ b/cmake/c_flag_overrides.cmake
@@ -8,3 +8,5 @@ endif()
 if(EMSCRIPTEN)
     set(CMAKE_C_FLAGS_DEBUG_INIT "-g4")
 endif()
+
+set(CMAKE_C_FLAGS_DEBUG_INIT "-g")
diff --git a/cmake/cxx_flag_overrides.cmake b/cmake/cxx_flag_overrides.cmake
index 9c8d15fe..ec43ae18 100644
--- a/cmake/cxx_flag_overrides.cmake
+++ b/cmake/cxx_flag_overrides.cmake
@@ -8,3 +8,5 @@ endif()
 if(EMSCRIPTEN)
     set(CMAKE_CXX_FLAGS_DEBUG_INIT "-g4")
 endif()
+
+set(CMAKE_CXX_FLAGS_DEBUG_INIT "-g")


It's a bit of a hack, and make sure not to commit those changes!

WickedSmoke:
I've created a couple more [Linux packages](https://github.com/WickedSmoke/solvespace_qt5/releases) for the Qt version.  It is now available for Fedora 40 and Debian 12.  The .deb was built with the .spec file and [debbuild](https://github.com/debbuild)).

For Linux users that have Qt installed using these native packages will save a ton of disk space & RAM as compared to the Flatpak which brings in over 760MB of dependencies.

WickedSmoke:
This branch has been rebased on master.  The change to rendergl1.cpp has also been reverted so that fewer files get touched.  Linux uses the GL3 renderer, but did this do anything on Windows?

The Linux version seems to work OK so I would really like to see this work captured on master so I don't have to keep rebasing in the future.  As an alternative, there is now a [Qt branch](https://github.com/WickedSmoke/solvespace_qt5/tree/qt) in my repo which squashes everything into just four commits and omits the tinyxml2 files from history.

ruevs:
Thank you for rebasing.

@WickedSmoke the machine I used to use for SolveSpace development "died" a few months ago and I have not set up Qt on the new one yet. I will do it, debug why opening and saving files fails on Windows and will merge the Qt port.

As for which branch - I was thinking of rebasing and squashing some commits anyway so I the "[qt](https://github.com/WickedSmoke/solvespace_qt5/tree/qt)" would be easier - you did most of the work already. At the same time I am thinking of committing "[karl](https://github.com/WickedSmoke/solvespace_qt5/tree/karl)" as "solvespace_qt5" in the main repository to preserve the detailed history.

dgramop:
@ruevs as far as you know has this branch been tried on MacOS yet? Qt on my Mac always ends up crashy for some reason... but it seems to be just on my builds. Also what are our thoughts on https://nixos.org?

ruevs:
> @ruevs as far as you know has this branch been tried on MacOS yet?

@dgramop as far as I am aware it has not been. As I [mentioned](https://github.com/solvespace/solvespace/pull/1451#pullrequestreview-1943727429) my only macOS is a VirtualBox VM - slow and thus does not stimulate me to use it :-)

WickedSmoke:
Merging "qt" and adding a "solvespace_qt5" branch sounds great, but note that there is one commit hash in a log message that has to be manually updated after rebasing "solvespace_qt5" again.

I see that Wayland breaks a couple things with Qt 5 here (but of course it does).
  1. The OpenGL widgets are not being resized properly, so the rendered area only occupies the lower-left quadrant of the widget.
  2. The TextWindow cannot be re-docked with the main window.  See https://bugreports.qt.io/browse/QTBUG-87332.

WickedSmoke:
These changes can no longer be rebased on master as there are conflicts between the shingen75 changes to CMakeLists.txt and commit a208201c.  Rather than spending time trying to resolve these I'm simply going to remove
all the CMake changes currently in commit ec7a2cbe as most get reverted later.

I'm considering only carrying forward my WickedSmoke:qt branch and leaving WickedSmoke:karl as is.

vthriller:
No idea what branch are you going to merge so I tried all three mentioned here. Nitpicks so far:

- On Wayland with scale of 1.5, text input is rendered at offset. Not reproducible on X11 or when scale is 1.
  
- If I'm not saving sketch on exit, I am asked again whether I want to save it or not.

Other than that, this is pretty impressive.

WickedSmoke:
@vthriller There's a fix for the QTextEdit position in commit 2daac18.

vthriller:
> If I'm not saving sketch on exit, I am asked again whether I want to save it or not.

It's actually a bit more interesting: if I Ctrl-Q in the detached property browser, I get the dialog once, then browser window closes, and the main window stays on the screen. (Ditto for when there are no unsaved changes, property browser just hides instead of terminating the whole thing.)

WickedSmoke:
@vthriller Commit 891e7e1f8 should fix the quit issues so please give it a try.

Issue #855, opened 2020-12-14
Add option to show only SLVS file name in the title of main window (NOT full path)


ghost:
### System information

- **SolveSpace version:** `3.0~` https://github.com/solvespace/solvespace/commit/2939abf5f876a7454e0afbc06b32c888989b6546
- **Operating system:** Debian 10 buster

### Expected behavior

Show only SLVS file name in the title of main window, eg. `<file_name>.slvs`

UPD: Add it as option `[*] show full file path in the window title` (enabled by default) on the *"Property Browser > configuration"* screen.

### Actual behavior

SolveSpace shows full path to SLVS file in the title of main window, eg. `/path_to/<file_name>.slvs`.

### Additional information

![pic.1](https://i.imgur.com/6fG2AQb.png)

ruevs:
Why?!

Symbian9:
> Why?!

For the first, for security reasons! (title should not show user PC name, which is shown actually in path if file stored in home directory)

For the second, if there are few windows opened with different files but in same folder it would be much easier to select needed window on global/system window panel.

ruevs:
Having the full path is extremely useful. As a counter example to yours: if there are many files opened with the same name but in different directories it would be much easier to select...

 The cinvenience/"security" of "tutorial makers" should be completely disregarded. They should know what they are doing.
- Windows: Put the files for your tutorials in "C:\verysecret\..."
- Linux... I don't know, be root, make "/very secret" and give your very secret user name permissions, do the tutorial and delete it.
- Or remove your user name from the screen shot if you are not making a video.

Symbian9:
> Having the full path is extremely useful.

Having the full path is extremely useful *only for geeks*.

@phkahler, as compromise I propose add CLI option `--show-paths` to start SolveSpace with behavior similar to "old school" (aka SolveSpace `2.x`) behaviour.

ruevs:
Whether it is extremely useful only for geeks is hard to know, but it is possible.

On Windows even "common users" put their files all over the place (not only in their user directory). And enen the most recent Windows Explorer has the [“Display the full path in the title bar”](https://www.onmsft.com/how-to/how-to-make-windows-10s-file-explorer-show-the-full-directory-path-in-the-title-bar) option.

On Linux... how many Linux users are not "geeks"? ;-)

As for adding an option - in my opinion adding extra options (and thus code) for such things (fluff) is a path that eventually leads to software that I dont like.

[...]

PR #1453, opened 2024-03-10
Add option to show only SLVS file name in the title of main window

johnathaningle:
This PR is for [issue #855](https://github.com/solvespace/solvespace/issues/855) and allows only the filename to display in the title bar rather than the full file path. A configuration item has been added under "arc default is diameter" for "show full file path in the window title." Toggling the configuration item will show/hide the full path and the setting persists after restart.  
This has only been tested on a Windows 11 machine. Please let me know if any tweaks need to be made.


ruevs:
Merged here 71ad5cbfabb1a6f089eb9c585ef52147de01959c

Issue #1454, opened 2024-03-24
glCreateShader fails on MacOS M1


dgramop:
### System information
MacOS M1 on 87d0e097bbbfb895e622cf970f236bcb108167f6

- **SolveSpace version:** 
- **Operating system:** 

### Expected behavior
Solvespace should launch, in particular, glCreateShader should return a non-zero value

### Actual behavior
Solvespace crashes. Running in lldb, and inspecting the stack after crash:

TL;DR assertion fails when trying to create shaders/edge.vert
```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x000000018b81e0dc libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x000000018b855cc0 libsystem_pthread.dylib`pthread_kill + 288
    frame #2: 0x000000018b761a40 libsystem_c.dylib`abort + 180
    frame #3: 0x0000000100059ee8 SolveSpace`SolveSpace::Platform::FatalError(message="File /Users/dgramop/Projects/solvespace/src/render/gl3shader.cpp, line 104
, function CompileShader:\nAssertion failed: shader != 0.\nMessage: glCreateShader failed.\n") at guimac.mm:90:5
    frame #4: 0x000000010028d3b4 SolveSpace`SolveSpace::AssertFailure(file="/Users/dgramop/Projects/solvespace/src/render/gl3shader.cpp", line=104, function="Co
mpileShader", condition="shader != 0", message="glCreateShader failed") at util.cpp:15:5
    frame #5: 0x000000010002aac4 SolveSpace`SolveSpace::CompileShader(res="shaders/edge.vert", type=35633) at gl3shader.cpp:104:5
    frame #6: 0x000000010002a798 SolveSpace`SolveSpace::Shader::Init(this=0x0000000128841674, vertexRes="shaders/edge.vert", fragmentRes="shaders/edge.frag", lo
cations=size=3) at gl3shader.cpp:135:19
    frame #7: 0x000000010002c96c SolveSpace`SolveSpace::EdgeRenderer::Init(this=0x0000000128841668, a=0x0000000128841610) at gl3shader.cpp:461:12
```

### Additional information
The version of solvespace I have "installed" still works (3.1-70bde63c). I first ran into the issue in the parameters branch, but then checked out master to see if it would happen again. I was able to compile on this mac a few months ago. Perhaps something in my packages changed

dgramop:
[...]

PR #1456, opened 2024-04-04
fix: use macOS Frameworks

dgramop:
We want solvespace to dynamically link with only one OpenGL library at a time, in particular the one that ships with MacOS. For this we use macOS frameworks, and pass the appropriate flag to the linker.

We should not use default cmake logic for finding openGL, since it seems to find the wrong version of libGL, that may have been installed by some other package at some point.

Closes issue #1454


dgramop:
We can close that issue regardless of whether this PR is accepted :)

I think I figured out what's broken and it's unclear if Solvespace should change, find_package in cmake should change, or if whatever package in homebrew or elsewhere that provides /opt/local/lib/libGL.dylib should change.

The case for this PR and making the change in solvespace or CMake's FindOpenGL module is that it makes solvespace more resilient to other weird non-system installs of openGL in MacOS

ruevs:
It's interesting why did CMake [find the "wrong" OpenGL on your system](https://github.com/solvespace/solvespace/issues/1454#issuecomment-2036348552).
According to the [documentation](https://cmake.org/cmake/help/latest/module/FindOpenGL.html#macos-specific) it should have used the framework. Perhaps (as you mentioned) it is a problem with CMake, or is it the way we are using it? My experience with CMake is very limited - @rpavlik and @vespakoen - do you have some time to lreview this?

@dgramop does the new artefact (built by the CI in this pull request) work for you? (I hope you can see it).

dgramop:
I agree with the scare quotes around "wrong", I wonder if some other software that uses cmake relies on its behavior on macOS to find whatever it found on my system instead of just using the framework. 

I can't see the MacOS artifact for the PR. I can see artifacts for commits pushed to master, this one just has "solvespace-x86_64".

ruevs:
> I can't see the MacOS artifact for the PR.

Sorry I forgot - the CI for pull requests does not keep the artefacts.

dgramop:
Hey! Just checking in, did anyone get a chance to try this out on a mac? Maybe there was something I mistinterpreted and was supposed to follow up on?

dgramop:
🎵🎵
"can any-body find meeeeeeeeee"
"somebody toooooooooo"
"~~looooove~~ review my pull request"
🎵🎵

~ in the off chance that somebody with a mac has time to review this, I'd really appreciate it :)

ruevs:
> "can any-body find meeeeeeeeee"
> "somebody toooooooooo"
> "~looooove~ review my pull request"

With such a nice request I'll have to dig out my machine that has a macOS VM on it... it'll  be "deep" digging... hopefully someone on a real Mac takes a look in the meantime :-D

phkahler:
What does it take to set up a build environment on osX? I've got a 2013 MBA but it doesn't have Xcode on it and I don't think the app store is working well on it, and it's got a fairly old OS. It runs Solvespace but I don't build on it.

BKLronin:
Fyi: There is this easy to setup proxmox osx script. I'm using this for the intel builds on an N100. (Runner and gui tests)

https://github.com/luchina-gabriel/OSX-PROXMOX

ruevs:
> What does it take to set up a build environment on osX?

On my macOS 10.14.6 (18G9323) Mojave
all I had to do was install Xcode 11.3.1 (the latest supported) and then follow the official instructions.

My make command is:
`cmake .. -G Xcode -DCMAKE_BUILD_TYPE=Release -DENABLE_OPENMP=ON`
and then I just open the resulting `solvespace.xcodeproj` and build it.

dgramop:
I see. Glad the cmake change worked, but I'm not really sure about Xcode. I didn't even try building with Xcode. Does master usually build correctly through Xcode?

ruevs:
> Does master usually build correctly through Xcode?

It does build just fine.

dgramop:
Hmm, I’m not even making it past the CMake step when i try configuring for XCode

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #1461, opened 2024-05-25
add "tool" to make sentece sensical

daef:


ruevs:
It seems to me that it is a "tradition" to omit "tool" - see the web site - it has been like that for 16 years - as well as the project title here on GitHub.

If one treats CAD as a nown describing a type of tool (I do realize it is not a now) it even makes sence - think of "a parametric 2d/3d shovel" :-)

ruevs:
Up to @phkahler and/or @jwesthues I'm fine with both.

Issue #1152, opened 2021-12-21
Sketching text fails if glyphs overlap


jwebb:
### System information

- **SolveSpace version:** e07b082eb810d2088c92b97d04e864a4ab95816f
- **Operating system:** Ubuntu 20.04

### Expected behavior

Sketching 'Text in TrueType Font' should create a valid contour.

### Actual behavior

The contour may be self-intersecting and therefore unusable if some glyphs overlap e.g. due to kerning. For example, the letters `TY` in DejaVu Sans Bold.

### Additional information

While in general CAD software expects that users should resolve self-intersecting paths manually, this seems an exception: other than picking a different font or text this isn't in the user's control. I think the desired behaviour would always be to take the union of overlapping glyphs in this case.

ruevs:
Yes, if a particular text in a particular font has overlapping glyphs - either due to kerning or "handwriting" fonts for example - then the sketch will be self intersecting and it will not form a valid solid when extruding/revolving/lathe-ing etc.

I don't think there is a general way to fix this.

jwebb:
There is certainly general fix: When two or more glyphs overlap, generate a contour which is their union. That's a standard operation in any 2D vector graphics package, and Solvespace already offers it for 3D solid models. And it's the _right_ solution because it's logically equivalent to how every font rasterizer handles the situation.

Whether it's easy or justified within Solvespace's framework is of course another matter.

ruevs:
I understand. You ment the union of the 2d outlines of the overlapping glyphs. That makes sense. SolveSpace currently does not have this implemented for anything else, so it will be entirely new code. But it is possible.

phkahler:
Split the curves for each glyph against the curves from the previous glyph. Discard the sections from each curve that are inside the outline of the other glyph. The tricky part is the word "inside", but I wouldn't be surprised if there is a lot of code already there that can be used in making that determination. Curve-curve intersection looks like it will be important for #1091 - second issue.

ruevs:
In principle it is "simple" we already do all of it in 3d. The inside/outside classification I am VERY familiar with ;-)

ruevs:
Here is a much older related issue
https://github.com/solvespace/solvespace/issues/285

Issue #285, opened 2017-08-30
Change "split curves at intersection" so it works on text 


DanielO:
### System information

SolveSpace version: 2.3
Operating system: MacOS Sierra 10.12.6

### Expected behavior
It would be nice to be able to "split curves at intersection" on text so I can merge it with another element in the same work plane.

See attached image for what I was creating - in the end I put the circle and text in 2 different work planes and extruded them independently but that was a bit of a pain.

![screen shot 2017-08-30 at 10 34 22](https://user-images.githubusercontent.com/643662/29850732-cd360daa-8d6e-11e7-8e62-a278eeb47755.png)

whitequark:
I'm not sure how feasible is that with our current implementation. Maybe if we add a separate "turn text to entities" option?

DanielO:
Yes that would be great, I actually looked for such a thing as a work around! :)

whitequark:
As a very crude workaround you could export text as DXF and then re-import it back.

DanielO:
I tried that but the result was not good - maybe I did something wrong though - see the attached file.

I created the text, did Export 2d view as DXF, move original text, import DXF and that was the result.

whitequark:
There's no attached file...

DanielO:
Ah oops, I ignored the "we don't support that file type" error..

[issue-285.zip](https://github.com/solvespace/solvespace/files/1262493/issue-285.zip)

whitequark:
Ah, I see. This seems to be related to Bezier curves:
![screenshot_20170831_092624](https://user-images.githubusercontent.com/54771/29916500-82738d0e-8e2e-11e7-856e-7602384f2c67.png)

@Evil-Spirit do you have any quick ideas?

whitequark:
The DXF file seems OK to me so it must be the import feature at fault.

Evil-Spirit:
@whitequark 
Hm. this is my fault. I actually didn't implement it.
Look here
https://github.com/solvespace/solvespace/blob/master/src/importdxf.cpp#L671

Evil-Spirit:
@whitequark
The reason is what we are exported it like beziers degree of 2, but we can import only 3.

Evil-Spirit:
@jwesthues, what do you think about it?

whitequark:
@Evil-Spirit Right, that should have at least had `importer.unknownEntities++;`

jwesthues:
[...]

PR #1462, opened 2024-05-30
Implement basic kerning when rendering text.

### LINKED ISSUE #1460: Font Kerning Not Working
### System information

- **SolveSpace version:** 3.1~flatpak
- **Operating system:** Linux Mint 21.1

### Expected behavior

Sketching curve from text should apply kerning when rendering the text.

### Actual behavior

Kerning not applied, resulting in stilted spacing.

### Additional information

It looks like kerning works correctly on the built-in, default font, but I'm not sure it's working on external fonts (or at least doesn't appear to work on many).

I used FontForge to make some modifications to an existing font and then generated a new font.  I've tried all the options I can find when generating the font that pertain to kerning, but nothing I do seems to make any difference to SolveSpace.  Other applications successfully apply the kerning.

render-comparison

I've tried every combination of:
Format: TrueType, OpenType
Options:
- Old style 'kern'
- Windows-compatible 'kern'
- Prefer native kerning

Any advice on what settings I need in my font in order to get kerning to work correctly in SolveSpace?

kdickerson:
Resolves #1460

Only old-style `kern` tables are supported--modern `GPOS-based` kerning is not supported.

Note, my CPP skills are at least 15 years out of date, so please help me correct anything foolish I've done.


ruevs:
Your C++ skills may be out of date, but you have the right "habits" - you updated the test case - almost no one catches that and does it from the start. I'm very glad someone with your type of "rusty C++ skills" now has the repository cloned and buildable :-)

The code looks fine.

For cross reference - here is my concern with permanently enabled kerning https://github.com/solvespace/solvespace/issues/1460#issuecomment-2138730648

phkahler:
The only issue I have with this is it will move letters in existing designs. They'll probably look better but it will be different. IMHO it's worth doing ;-)  What do you think @ruevs? We should have a 3.2 beta soon and I'd include this.

kdickerson:
Bringing @ruevs comment from #1460 to coalesce conversation into one location:

> Indeed we use FreeType and it seems that "FreeType only supports kerning via the (rather simple) ‘kern’ table."
>
> Until now I was not aware we did not use kerning - never looked at the "text" code, but it is possible that this could have been intentional - kerning can make glyphs overlap and then the resulting sketch is self intersecting and will not form a solid when extruded. See here #1152 and here #285
>
> So the splitting and merging of glyphs should be implemented and/or kerning should be an option - perhaps on each individual "text" entity.

I appreciate those issues.  Also, agree that having existing designs shift around, possibly into a broken state, would be an unpleasant surprise for users.

As a new user, I would expect kerning to be used, so I think the logic would be to leave kerning off when opening an existing project, but default to `on` in new projects.

It sounds like the minimal solution would be to add a per-Text-instance flag for whether kerning will be applied which gets persisted (and default to `false` when loading projects which don't define this flag).

Related:

This comment in the code: 

>       // And we're done, so advance our position by the requested advance
>       // width, plus the user-requested extra advance.
>       dx += fontFace->glyph->advance.x;

Suggests that supporting a user-specified spacing modifier to the text advance was intended at one point (but is not actually implemented) and could probably be implemented at the same time without much additional effort.  This gives users a tool to increase glyph spacing in those situations where glyphs overlap (but doesn't resolve those issues generally).

--

I can certainly take a crack at this, but obviously code that stretches from here to the UI and saving/loading will have significantly more touch points for me to mess up.  I'm a Python and JavaScript developer, so while the principles aren't an issue, I recognize that the nuances of C++ and avoiding memory mistakes are not things I'm well practiced in these days.

ruevs:
All valid points. I'm on a phone now but will look later if one of the (existing) attributes of entities is unused for text entities. In that case we can treat it as a "kerning ON" flag and choose the "off" value such that it matches all existing text entities.

At the same time we make the default "on" for new entities and show the option as a check box in the text window when a text entity is selected.

phkahler:
FYI, for global setting there is a simple example in this commit:
https://github.com/solvespace/solvespace/commit/1f8b3f2589228b5fee3d4d5821917ba537723d76

The issue discussion from that is here:
https://github.com/solvespace/solvespace/pull/1425

If you make a per-item option, it will need to be stored in-entity and save/load are handled differently. The text-screen checkbox will be similar though.

IMHO it'd be best as a global and the main reason to have an option is for compatibility with existing files.

ruevs:
> IMHO it'd be best as a global and the main reason to have an option is for compatibility with existing files.

In my opinion it is better as a per entity flag - in this way an existing file will not change/break when opened in the new version and at the same time new text with kerning can be added or even existing entities changed selectively to have kerning on.

kdickerson:
This needs an additional test case for when kerning is off, but I wanted to get it up to see if I was on the right track when you said "...if one of the (existing) attributes of entities is unused for text entities".   I abused `extraPoints` to hold a boolean flag for whether to apply kerning to the entity.

If that _is_ what you meant I'll add the the appropriate test case.  Of course, if you'd prefer to store it somewhere else I can do that too.

I'll be honest, all of the changes to `describescreen.cpp` are straight-up cargo-culting patterns and functions I found already in the code.  So, they should be looked at carefully to see if I did something totally stupid.

(If abusing an attribute like this is what you mean, it looks like I could also use `aspectRatio` to store a user-specified spacing multiplier which comes in really handy (and, as mentioned previously, looks like was intended, but not implemented); but I'd do that in a separate PR.)

ruevs:
This is exactly what I meant. And your implementation looks fine - reading it superficially from a phone. I'll be away from a computer for at least the next three days, but will pull your branch, review and test it carefully next week.

ruevs:
> Is the value of `extraPoints==false` in existing sketches?

Apparently it is - nice. And you need to revert your change to the test case .PNG now, but I presume you already saw this if you were considering addind a test case for "kerning on text" :-)

https://github.com/solvespace/solvespace/actions/runs/9312968853/job/25635034794?pr=1462#step:4:1073

kdickerson:
Updated to squash into single commit, restore the original test case, add a kerning test case.  And fix aspect-ratio calculation to also account for kerning.

Let me know if anything else is needed to get this merged.

phkahler:
I'm not sure we can abuse extraPoints for this purpose. That's not even a boolean. It is a numeric value indicating a number of "extra points" needed by an entity. I'm not sure what the implications are of setting that to 1. Will it write extra stuff to the file when saved?

kdickerson:
It does add a line in the save file, which works to save/load the value without issue.

I'm also unsure of further ramifications of using it for this purpose, but I was following @ruevs suggestion on cramming it in somewhere rather than adding another field on entities and requests dedicated solely to text rendering.  Doesn't really matter to me one way or the other--I leave that decision to whomever expects to have to deal with it long term for maintenance purposes.

ruevs:
Merged here 6b79e69d60b24d5e9c2dfe55b58edb1ebd4e0aa7

PR #1463, opened 2024-06-02
Add Curvature function for surfaces

phkahler:
I'm not sure this is quite correct even for a numerical approximation. In particular I'm not sure about division by tu.Magnitude and tv.Magnitude.


ruevs:
@phkahler the disk of the machine I usually use to work on SolveSpace died... :-(

Thinking "on paper" - how would the caller choose the `dir` parameter?
Apart from that I need to "parse" more carefully through the `dot` an `cross` products and make some drawings - but at first glance in seems that the function will return  a positive number of the surface "curves towards" the `dir` vector. I think there are some corner cases...

PR #1482, opened 2024-09-12
Use parameters in groups (like repeat)

dgramop:
This pull request adds parameters to groups, such has Linear Repeat, Rotate Repeat, Helix, etc.

Allows expression input into group-level things (usually through the text window)

Related: #77 

Pulling into the parameters branch of solvespace for now.

### DIFF
--- a/src/constrainteq.cpp
+++ b/src/constrainteq.cpp
@@ -255,7 +255,7 @@ void ConstraintBase::Generate(IdList<Param,hParam> *l) {
         size_t eqpos = expression.find_first_of("=");
         ssassert(eqpos == expression.find_last_of("="), "There is at most one equals sign in the relation \"expression\"");
         ssassert(eqpos != std::string::npos, "There is at least one equals sign in the relation \"expression\"");
-        Expr::From(expression.substr(0, eqpos), false, &SK.param, NULL)->Minus(Expr::From(expression.substr(eqpos+1, SIZE_MAX), false, &SK.param, NULL));
+        Expr::From(expression.substr(0, eqpos), false, l, NULL)->Minus(Expr::From(expression.substr(eqpos+1, SIZE_MAX), false, l, NULL));
     } else if(expression != "" && expr_scaling_to_base != 0) {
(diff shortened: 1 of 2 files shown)


dgramop:
couple of TODOs:
- Prevent crash when rotate goes to zero (for example because the solver pushed it there)
- Find similar edge cases if they exist

In a future PR, we should look into adding ceiling and floor operators to expressions.

dgramop:
https://github.com/user-attachments/assets/1d0baa7b-81b7-442a-a9ca-5bb7548dcab9

dgramop:
https://youtu.be/qX3N7cF-32w

dgramop:
TODO: confirm this doesn't break savefile backwards compatibility.

phkahler:
@dgramop  I finally got a look at this. I kind of like the idea of being able to use expressions in the TW, but I don't think this is the right approach. While looking at the Helix Pitch implementation I do think i made an error. The code computes the pitch based on the group parameters for angle and linear displacement and displays that. It uses valB of the group to store the users pitch (which is IMHO a mistake) and if they specify a non-zero value it creates a constraint to enforce that value. I didn't know any better at the time.

What I think should have been done is the group should have created a pitch parameter and a constraint. The new parameter and constraint would result in the same DoF. Then the text window could just show the pitch parameter and offer a way to lock it. I'm not sure how to lock a parameter, but worst case would be to create a simple pitch=value constraint. Notice that I'm talking about a generic parameter and not group->valB. Then we could allow an expression on the right hand side of that constraint equation. I'm not exactly sure where that expression string would live, but I'm pretty sure adding new members to the group is not the way to go.

I do suggest keeping this PR around as an example for now. I'm doing my own stuff in the "parameters" space, but that doesn't mean anything is decided yet ;-)

dgramop:
Yep, I'm reading your comments on other PRs/issues and that approach makes sense. I think it's good for us to stabilize and centralize over some single pattern for parameters. As (i think @ruevs pointed out) there are far too many different models (some of them dead code) in solvespace for this.

PR #1483, opened 2024-09-14
Use `const auto &` in a few for loops that were not doing it.

ruevs:
Could help with compiler optimizations and thus performance.

### DIFF
--- a/src/draw.cpp
+++ b/src/draw.cpp
@@ -69,7 +69,7 @@ void GraphicsWindow::Selection::Draw(bool isHovered, Canvas *canvas) {
         Vector topLeft = camera.UnProjectPoint(topLeftScreen);
 
         auto it = std::unique(refs.begin(), refs.end(),
-                              [](Vector a, Vector b) { return a.Equals(b); });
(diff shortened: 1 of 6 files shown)


rpavlik (approved):
Looks good to me as long as it all builds (meaning the called functions are already const-correct) - nice catch! You might also want to get that std::unique predicate in draw.cpp line 71 with the `const&` wand

src/draw.cpp
```diff
@@ -71,7 +71,7 @@ void GraphicsWindow::Selection::Draw(bool isHovered, Canvas *canvas) {
         auto it = std::unique(refs.begin(), refs.end(),
                               [](Vector a, Vector b) { return a.Equals(b); });
```
rpavlik, line 72: unrelated but more low hanging fruit: 

```suggestion
                              [](const Vector &a, const Vector &b) { return a.Equals(b); });
```

ruevs: Good catch - thank you.

ruevs:
Thanks @rpavlik, added your suggestion and merged.

PR #1486, opened 2024-09-20
Fix saving assemblies when opened with a relative path on the command

ruevs:
... line on Linux.

An extra `Expand` for "filename" is needed on Linux when the file was opened with a relative path on the command line because (I guess) dialog->RunModal() in SolveSpaceUI::GetFilenameAndSave will convert the file name to full path on Windows but not on GTK.


ruevs:
Merged in 5528da7ffbfe511a90c46864da40eec5305a89ae

PR #1491, opened 2024-11-01
Win32: the Property Browser remains visible when going full screen (CTRL-F11)

ruevs:
On Windows when going full screen by pressing CTRL-F11 the Property Browser window remained hidden behind the main window until one ALT-Tab-ed to it or to another application and back.

This commit fixes it.

The fix is not mine. It comes from this Chinese fork: https://gitee.com/zhongzhiSys/SolveSpace
where they mostly took 2.3 and started selectively merging fixes from 3.x with their own commit messages. There are however original fixes like this one interspersed with the merges. I have not finished going through everything.


ruevs:
This is the fork with most commits that I could find:
https://gitee.com/Frankhe303/SolveSpace
nothing particularly interesting in there.

ruevs:
> I wonder why CI is failing on this... it seems spurious and unrelated to your code. If we re-ran it again on the same code, does it still fail?

Flatpack is flaky; I re-ran the job, lets see...

Issue #1492, opened 2024-11-05
Circle tangent to line as construction tool


hofmannu:
### System information

- **SolveSpace version:** 3.1~70bde63c
- **Operating system:** ArchLinux

![image](https://github.com/user-attachments/assets/d03419e5-9810-461a-8cd3-6a2bf9a7dc09)

### Expected behavior

If I add the constraint that the circle should be tangent to the two lines, the sketch is fully defined and the through hole will have ensured equidistance from the two object edges.

### Actual behavior

I get an "Bad selection for tangent/parallel constraint" error.

### Additional information

Attached is the tiny sketch.
[circle_tangent_to_line.zip](https://github.com/user-attachments/files/17627847/circle_tangent_to_line.zip)

phkahler:
@hofmannu  You've found the "correct" way to make a line and circle tangent with what we have.  If someone does implement line/circle tangency I think it would be good to automatically add the point where they meet.

Since you're a new user and the diagram looks a little odd I'm going to guess that maybe your green circle is simply a way to specify a distance (radius) from the center of the white circle. If that's the case, you can just pick the point and one of the lines and press D to set the point-line distance. Do that again for the other point-line distance. The number of constraints available in Solvespace is quite large ;-)  You can even select pt-line and another pt-line and use eQual to set two pt-line distances to be equal while leaving the actual distance unconstrained. Unfortunately that doesn't work when the points are the same (circle center in your case).

hofmannu:
thanks a lot for your answer!

while i am new to this software, i have a background in mechanical engineering and used CAD for more than a decade by now. that tangent circle is quite a common step in the construction process. the two distance approach won't have the two lengths  mapped to each other? so i need to add the number twice or again make two lines, enforce equal length and perpendicularity which is a bit cumbersome even with the nice workflow of the interface and the keyboard shortcuts.

Maybe i will try to implement the constraint type myself as a first contribution ;)

phkahler:
> Maybe i will try to implement the constraint myself as a first contribution

I don't think any of our constraints add sketch entities, so this is unusual - it would be adding both. I'm not even sure what the best way to constrain the point would be - on circle and on line, or on circle and forming a perpendicular to the line and circle center. Then there is the visual representation of this constraint. It's a good project if you want to get involved!

[...]

PR #948, opened 2021-03-03
Fix tangent constraints of curves.

ruevs:
This fixes a regression introduced in  96958f466395148ee7f0a252222a2a27e7345c22 (https://github.com/solvespace/solvespace/pull/833.)
Three `return` statements got "swallowed" by the newly created functions.
(e.g .https://github.com/solvespace/solvespace/commit/96958f466395148ee7f0a252222a2a27e7345c22#diff-49abc03ed071148c0ebae0c64aafb625fa6223135f77aeecdb47dab4cf8b940cL638 )

Because of this it was possible to constrain arcs and cubics tangent to each other or line segments, without them sharing an endpoint.

This kind of worked, but always chose the "starting" points of the curves and lines. In the future this can be turned into a feature. See the discussion in #937.

PR #1493, opened 2024-11-06
Allow tangent on arcs/lines/splines without coincident endpoints.

ruevs:
Implements #937

Test cases:
[ArcsLinesSplinesTangentTests.zip](https://github.com/user-attachments/files/17648175/ArcsLinesSplinesTangentTests.zip)
https://github.com/solvespace/solvespace/pull/833#issuecomment-742082682
https://github.com/solvespace/solvespace/pull/833#issuecomment-744528586

Cross references:
https://github.com/solvespace/solvespace/pull/833
96958f466395148ee7f0a252222a2a27e7345c22
https://github.com/solvespace/solvespace/issues/937
https://github.com/solvespace/solvespace/pull/948
8c101d5f281ba670a3af2663302fe3dc7fb15283
https://github.com/solvespace/solvespace/issues/951
https://github.com/solvespace/solvespace/issues/1492

### DIFF
--- a/src/constraint.cpp
+++ b/src/constraint.cpp
@@ -131,7 +131,8 @@ hConstraint Constraint::ConstrainCoincident(hEntity ptA, hEntity ptB) {
         Entity::NO_ENTITY, Entity::NO_ENTITY, /*other=*/false, /*other2=*/false);
 }
 
-bool Constraint::ConstrainArcLineTangent(Constraint *c, Entity *line, Entity *arc) {
+bool Constraint::ConstrainArcLineTangent(Constraint *c, Entity *line, Entity *arc,
+                                         Entity *arcendpoint) {
     Vector l0 = SK.GetEntity(line->point[0])->PointGetNum(),
            l1 = SK.GetEntity(line->point[1])->PointGetNum();
     Vector a1 = SK.GetEntity(arc->point[1])->PointGetNum(),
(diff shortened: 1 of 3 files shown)


phkahler:
Does this require you to select the end point of the arc? I would hope to automatically figure that out if it's constrained to the line. It's not too hard to search the constraints to figure it out?

ruevs:
It does require one to select the end point(s) of the curve(s) at which the tangency constraint will be applied.

I tried to make the error messages as informative as possible - take a look or play with it.

If an arc/spline already has one point constrained on the line (not on it's end point) it does seem logical to automatically select that point, but it is not the only way - the other end of the curve can be constrained tangent. Auto selecting would constrain the generality of the solution, so I did not attempt to do it. Of cource if one explicitly selected a (the other) point we could override the automatic selection, but I think it becomes confusing.

src/constraint.cpp
```diff
@@ -140,16 +141,32 @@ bool Constraint::ConstrainArcLineTangent(Constraint *c, Entity *line, Entity *ar
...
+    } else if(nullptr != arcendpoint) {
+        Vector p = arcendpoint->PointGetNum();
+        if(a1.Equals(p)) {
+            c->other = false;
+        } else if(a2.Equals(p)) {
+            c->other = true;
+        } else {
+            Error(_("The point you selected does not belong to the arc. "
+                    "The arc and line segment do not share an end point.\n\n"
+                    "Select the end point of the ark at which you want "
```
hofmannu, line 153: arc?

src/constraint.cpp
```diff
@@ -140,16 +141,32 @@ bool Constraint::ConstrainArcLineTangent(Constraint *c, Entity *line, Entity *ar
...
+                    "it to be tangent to the line."));
+            return false;
+        }
     } else {
         Error(_("The tangent arc and line segment must share an "
                 "endpoint. Constrain them with Constrain -> "
-                "On Point before constraining tangent."));
+                "On Point before constraining tangent.\n\n"
+                "Alternatively select the end point of the ark "
```
hofmannu, line 161: arc

src/constraint.cpp
ruevs: :-) Indeed, Noah is not involved :-)

src/constraint.cpp
ruevs: Thanks for the review. I removed the arks and force pushed the branch.

ruevs:
@phkahler so? 2b|!2b ?

phkahler:
> @phkahler so? 2b|!2b ?

I'm just going to trust your judgement on this one. The code looks reasonable ;-)

ruevs:
Merged then :-)

PR #1496, opened 2024-11-19
Add skewed subtype for extrusion groups. Update the textwindow radio …

### LINKED ISSUE #1495: Add a checkbox to "allow skew" of extrusions
### System information

- **SolveSpace version:** ALL
- **Operating system:** ALL

### Expected behavior

It should be easy to make skewed extrusions. Perhaps a checkbox in the group text window to allow it.

### Actual behavior

Extrusions groups have 2 hidden perpendicular constraints to lock the extrusion direction perpendicular to the workplane. 

It turns out you can select a workplane different from the one containing a sketch to extrude the sketch, and this will create a skewed extrusion. This is a little awkward because the direction has to be set in the prior group via a little normal vector rather than using constraints in the extrude group.

### Additional information

I have tried commenting out the 2 perpendicular constraint, and it does result in a 3 DoF extrusion. All we'd really need is a checkbox for extrude groups to "allow skew" or similar which would inhibit creation of the 2 hidden constraint. This might be similar to how a constraint can be enabled via checkbox to lock helix pitch.

Is this a good idea?

phkahler:
…buttons and equation generation.

Closes #1495

### DIFF
--- a/src/sketch.h
+++ b/src/sketch.h
@@ -204,7 +204,8 @@ public:
         //WORKPLANE_BY_FACE          = 6004,
         // For extrudes, translates, and rotates
         ONE_SIDED                  = 7000,
-        TWO_SIDED                  = 7001
+        TWO_SIDED                  = 7001,
+        ONE_SKEWED                 = 7004
     };
     Group::Subtype subtype;
 
--- a/src/sketch.h
+++ b/src/sketch.h
@@ -205,7 +205,8 @@ public:
         // For extrudes, translates, and rotates
         ONE_SIDED                  = 7000,
         TWO_SIDED                  = 7001,
-        ONE_SKEWED                 = 7004
+        ONE_SKEWED                 = 7004,
+        TWO_SKEWED                 = 7005
     };
     Group::Subtype subtype;
 
--- a/src/textscreens.cpp
+++ b/src/textscreens.cpp
@@ -228,6 +228,7 @@ void TextWindow::ScreenChangeGroupOption(int link, uint32_t v) {
     switch(link) {
         case 's': g->subtype = Group::Subtype::ONE_SIDED; break;
         case 'S': g->subtype = Group::Subtype::TWO_SIDED; break;
+        case 'w': g->subtype = Group::Subtype::ONE_SKEWED; break;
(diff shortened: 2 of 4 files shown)


src/sketch.h
```diff
@@ -204,7 +204,8 @@ class Group {
         //WORKPLANE_BY_FACE          = 6004,
         // For extrudes, translates, and rotates
         ONE_SIDED                  = 7000,
-        TWO_SIDED                  = 7001
+        TWO_SIDED                  = 7001,
+        ONE_SKEWED                 = 7004
```
ruevs: Very cunning of you to introduce a new `Group::Subtype` in this way the saving is all taken care of. I like it :-)

But `TWO_SKEWED` is definitely required (as you already suggested).

src/textscreens.cpp
```diff
@@ -377,14 +378,30 @@ void TextWindow::ShowGroupInfo() {
...
 
         bool one = (g->subtype == Group::Subtype::ONE_SIDED);
+        bool two = (g->subtype == Group::Subtype::TWO_SIDED);
+        bool skew = (g->subtype == Group::Subtype::ONE_SKEWED);
+        
+        if (g->type == Group::Type::EXTRUDE) {
+        Printf(false,
+            "%Ba   %f%Ls%Fd%s one-sided%E  "
+                  "%f%LS%Fd%s two-sided%E  "
+                  "%f%Lw%Fd%s skewed%E",
```
ruevs: I think it is better if this is called "perpendicular" and checked by default.

ruevs (changes requested):
I made an extra commit in this branch with my proposed changes.

src/textscreens.cpp
phkahler: > I think it is better if this is called "perpendicular" and checked by default.

Disagree. I think the checkbox should be "skewed" and unchecked by default. The normal/expected/natural behavior should not need to be enabled (even by default) and I kind of dislike the word "prependicular" vs "skewed". Maybe it's too many syllables...  ;-)

phkahler:
You're the best @ruevs!

PR #1500, opened 2024-12-05
Make the animation speed configurable

davidgiven:
This is a minor adjustment to add a configuration setting for the animation speed, including disabling them completely.

image

The animation time equation has been slightly changed to accommodate this, but it should behave mostly the same as before. With the default setting of 800ms you should get approximately the same as before.

### DIFF
--- a/src/graphicswin.cpp
+++ b/src/graphicswin.cpp
@@ -503,12 +503,18 @@ void GraphicsWindow::AnimateOnto(Quaternion quatf, Vector offsetf) {
 
     // Animate transition, unless it's a tiny move.
     int64_t t0 = GetMilliseconds();
-    int32_t dt = (mp < 0.01 && mo < 10) ? (-20) :
-                     (int32_t)(100 + 600*mp + 0.4*mo);
-    // Don't ever animate for longer than 800 ms; we can get absurdly
+    int32_t dt = (mp < 0.01 && mo < 10) ? 0 :
+                     (int32_t)(SS.animationSpeed*0.75*mp + SS.animationSpeed*0.0005*mo);
+    // Apply a minimum animation time, for small moves. This gets overridden by the maximum setting
+    // so setting the animation speed to 0 disables animations entirely.
+    dt = std::max(dt, 100 /* ms */);
+    // Don't ever animate for longer than animationSpeed ms; we can get absurdly
     // long translations (as measured in pixels) if the user zooms out, moves,
     // and then zooms in again.
-    if(dt > 800) dt = 800;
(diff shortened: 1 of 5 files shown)


src/graphicswin.cpp
```diff
@@ -503,12 +503,18 @@ void GraphicsWindow::AnimateOnto(Quaternion quatf, Vector offsetf) {
...
+    // Apply a minimum animation time, for small moves. This gets overridden by the maximum setting
+    // so setting the animation speed to 0 disables animations entirely.
+    dt = std::max(dt, 100 /* ms */);
+    // Don't ever animate for longer than animationSpeed ms; we can get absurdly
     // long translations (as measured in pixels) if the user zooms out, moves,
     // and then zooms in again.
-    if(dt > 800) dt = 800;
+    dt = std::min(dt, SS.animationSpeed);
+    // If the resulting animation time is very short, disable it completely.
+    if (dt < 100) dt = -20;
```
ruevs, line 516: I like the feature as a whole.

The only thing I would remove is this line - it makes any `animation speed` setting below 100 equivalent to zero, which in my opinion is undesirable.

phkahler:
There are several terms in the animation code. Remember you can also do animation via "Step Dimension..." in the Analyze menu. This is one of those times I'm not sure how to test a change to make sure there is no unintended breakage since it touches features I don't use.

ruevs:
> There are several terms in the animation code. Remember you can also do animation via "Step Dimension..." in the Analyze menu. This is one of those times I'm not sure how to test a change to make sure there is no unintended breakage since it touches features I don't use.

@phkahler this pull request is about the speed of the animation when changing/transitioning the view - you touched that code a year ago to speed it up in f399997976d7 .

The step dimension code is completely unrelated - search for `stepDim` to find that code.

src/graphicswin.cpp
davidgiven: The rationale here is that the old code had a hard lower bound of 100ms; and given that the code seems to want to use a `dt` value of -20 to mean 'don't animate' I thought it best to round down. I've changed it to 10ms. Is that better?

davidgiven:
I've never touched the Step Dimension feature before (in fact, I don't recall even seeing the Analyze menu before. I wonder if the version I've been using was very old?) so don't know if anything needs doing there.

ruevs:
Mereged as https://github.com/solvespace/solvespace/commit/33f9cf952207037e54b7f316d0e076c7b5cccc58. Thank you for the nice contribution @davidgiven :-)

PR #1501, opened 2024-12-10
WIP add functions for creating chamfers and fillets to the NURBS kernel

phkahler:
Just an outline at this point (in the code). Hoping to attempt this over the hollidays. Thought people might want to follow along or comment if this goes anywhere.


ruevs:
@phkahler great to see you attempting this. Over the years I've been considering this - most recently Aron the time Dune3D started.

In my head it works similarly to what you started outlining with one more level of abstraction:

A chamfer/fillet is:

- An edge modifier class/module/cpp file goes over the selected adges and splits them in two - this is specific to the fillet/chamfer tool.
- A general purpose patch tool (with conditions as parameters - e.g. G1 continuity), which is a separate class, generates NURBS parches to close the holes between the edges created in the previous step.

In the future the patch class can be used to implement generic patch functionality - e.g. close a "random" hole outlined by a contour of any edges (not necessarily straight) in 3d.

phkahler:
> * in reality each visible edge is already two coincident edges so the code could "just" move them

Woops you're right. Curves are not duplicated, but there are two TrimBy, one for each surface that meets along a curve. These reference the same underlying curve but one is "backward". So I'll still need to generate 2 new curves, and move _both_ Trims rather than creating a new one. Good news - the PWL points are not in the trims!

Assuming all fillets meeting at a point are the same radius, the surface we want there is spherical. It's really hard to find anything (with google) on spherical NURBS patches, never mind triangular ones.  But that's for later. Chamfers first and Fillets with 1 or 2 per vertex. I'm both excited and dreading this...

ruevs:
> I'm both excited and dreading this...

Indeed :-)

BKLronin:
I very welcome the attempt to add this. :)

phkahler:
Getting back to this soon. Here is a sketch to visualize how fillets meet at a corner. Drag the green lines around.
[corner_arcs.zip](https://github.com/user-attachments/files/21923610/corner_arcs.zip)

For equal radius fillets it forms a spherical triangle.

PR #1504, opened 2024-12-11
Allow turntable to use Y axis.

WickedSmoke:
As normal math conventions and OpenGL have the Y axis pointing up I would like an option to have turntable rotation use this axis.

This code currently just changes Z up to Y up.  What would be the best way to add an option to the UI?  Another checkbox would be easiest to implement, but a combo of (Free, Turntable Y, Turntable Z) would be less controls for the user to futz with.  Is there combo box support with the current UI?


WickedSmoke:
I see many years ago @vespakoen tried to implement this in commit 97630bd36.

phkahler:
Checkboxes and radio buttons work exactly the same way. Here is an example:

https://github.com/solvespace/solvespace/blob/30e42ca889b20b6af65d601524085fd812c91662/src/textscreens.cpp#L419

We just added the checkbox for "skewed" extrusions, but it's functionally identical to the radio buttons - just looks different.

WickedSmoke:
Yeah, I saw the addition of "skewed" but I couldn't understand how to use that feature.

A proper combo widget would be more future proof as any number of rotation options could be added without taking up more screen space.

WickedSmoke:
> So 4 options: Off, X, Y, Z ?

Yeah, that would work.

PR #1505, opened 2024-12-12
Add an option for GraphicsWindow zoom to stay centered in view.

WickedSmoke:
Since the mouse is always moved in order to rotate, constantly mixing zooming and rotating will cause the view to drift away from the model. The new option prevents this drift.

### DIFF
--- a/src/confscreen.cpp
+++ b/src/confscreen.cpp
@@ -105,6 +105,10 @@ void TextWindow::ScreenChangeCameraNav(int link, uint32_t v) {
     SS.cameraNav = !SS.cameraNav;
 }
 
+void TextWindow::ScreenChangeZoomNav(int link, uint32_t v) {
+    SS.zoomCenterNav = !SS.zoomCenterNav;
+}
+
 void TextWindow::ScreenChangeImmediatelyEditDimension(int link, uint32_t v) {
     SS.immediatelyEditDimension = !SS.immediatelyEditDimension;
     SS.GW.Invalidate(/*clearPersistent=*/true);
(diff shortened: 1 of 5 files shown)


ruevs:
@WickedSmoke please consider this https://github.com/solvespace/solvespace/pull/1221#issuecomment-1060627406 in context of your pull request. I have not played with yours yest but they are definitely related.

WickedSmoke:
I use the mouse to rotate and zoom, but I don't want the position of the pointer to change what the camera is looking at when zooming.  After each view rotation the pointer position will have moved to some essentially random location on screen.  A subsequent zoom will then move the model towards/away from that random location and also change the offset.  Each round of this tends to move the point of rotation farther and farther from the model so that rotation quickly puts the model out of view.  At this point the view needs to manually be reset to continue working.

This option keeps zooming adjustments independent of offset (which has it's own control via the RMB panning).  Without it I feel like I'm fighting the program just to keep my stuff in view.

WickedSmoke:
I agree that both methods of zooming need to be supported.  I don't think the different input methods should be mapped to different zooms however.  So far I've never used the keyboard controls for this, so the mouse needs to operate to my preferences.

There could be a "Default Zoom Mode" option (To Pointer / Centered) which both key and mouse input would use.  Then, a key modifier could be added to enable the alternate zoom method on-demand.

If mapping inputs to mode is desired then two options would be needed.
-  Mouse Zoom Mode:  [To Pointer / Centered ]
-  Key Zoom Mode: [To Pointer / Centered ]

ruevs:
I like the `CTRL` key inverting the configured  zoom behaviour.
Would it be appropriate to change the text of the option to:
`zoom keeping view centered (hold CTRL while zooming to invert behaviour)`
Or something similar?

I'm not on a computer the next few days (and am thus too lazy ot check), but if the `+` and `-` zooming is implemented through the standard keyboard accelerators it may be a bit harder to add the `CTRL` modifier for them.

WickedSmoke:
@ruevs you are correct about the CTRL modifer.  I left a note in commit 099042e9 which was pushed a couple days ago.

WickedSmoke:
That description is a bit verbose.  Really tool tips are needed on the checkbox items for that sort of thing.

The whole method of defining the GUI through Print statements is a bit funky.  I was looking at adding Combo boxes (similar to the Buttons), but I'm not sure that can work interleaved with the Printed stuff.

ruevs:
> That description is a bit verbose.

You are right. And there are many hard to discover features already.

We'll update the documentation - let people read :-)

As for combo boxes and "UI through Printf" - it is weird - when I started tinkering with SilveSpace on 2014 I was also rather... surprised by it. But in a very "twisted" way I like it a lot - it removes UI as a "thing" to "do" in the software. This may be part of the reason SolveSpace is so "immediate" and "pure" - every time Jonathan, Whitequark, whoever wanted to do some "fency" UI they "bumped" into the "Printf UI 'framework' " and though of a better way to do it "without classic UI". This is one of a the best aspects of the SolveSpace UI - there is no UI for the sake of UI - it is only there in the absolutely minimal form needed to " do the job".

PR #1511, opened 2025-01-12
add handling for STL file instead of only stl file

ccppi3:
If the file name is written in capital letter solvespace crashes on my system.

Some applications do export with "STL" instead of "stl".

This would solve this.


src/file.cpp
```diff
@@ -718,6 +718,8 @@ bool SolveSpaceUI::LoadEntitiesFromFile(const Platform::Path &filename, EntityLi
         return LinkIDF(filename, le, m, sh);
```
ruevs, line 718: The same should happen for the "emn" or IDF file.

ccppi3: Yes you are right.
Sorry i never done a PR can i add a second commit to an existing PR?
Thank you in regard.

ruevs: Yes you can. Just commit in the same branch.

Issue #1515, opened 2025-01-26
Translation updates


ruevs:
[Another](https://github.com/solvespace/solvespace/issues/890) two years have passed since our [previous full translation update](https://github.com/solvespace/solvespace/pull/1331) . We have seven new or modified messages - and the corresponding new functionality :-)

I've [updated the translations](https://github.com/solvespace/solvespace/pull/1514) and this time merged them directly in master.
The commit message of https://github.com/solvespace/solvespace/commit/613f5bbb7c4da349c1453fad78ffdc7302145b23 contains a summary of the messages that need translations.

- [x] _English_ is fine.
- [x] _German_ and _Russian_: updated to the best of my (and the LLMs :-) abilities. @Evil-Spirit, @rurban, @carelinus can you please review them?
- [x] _Spanish_ should be fine - it includes the changes by @andesfreedesign from his pull request https://github.com/solvespace/solvespace/pull/1396 and a clean up from me. @andesfreedesign would you please review it?
- [ ] _All other languages_: the three new messages will be in English until translated. For the changed messages I kept the translated parts and appended the new text in English. They need to be updated:
   * constraint.cpp:158
   * constraint.cpp:193
   * constraint.cpp:244
   * constraint.cpp:887  

@strzinek, @progval @verylowfreq @mhalil @lomatus @Not4sure  would you please take a look at your respective languages if you have time? You can post the updated .po files here or - if you prefer - open pull requests.

strzinek:
Czech language translations updated: https://github.com/solvespace/solvespace/pull/1521

PR #1516, opened 2025-01-26
Fix inconsistent punctuation in French translation and translate new strings.

progval:
cc @ruevs (https://github.com/solvespace/solvespace/issues/1515)

### DIFF
--- a/res/locales/fr_FR.po
+++ b/res/locales/fr_FR.po
@@ -7,15 +7,15 @@ msgstr ""
 "Project-Id-Version: SolveSpace 3.0\n"
 "Report-Msgid-Bugs-To: phkahler@gmail.com\n"
 "POT-Creation-Date: 2025-01-26 21:04+0200\n"
-"PO-Revision-Date: 2022-10-15 17:32+0200\n"
+"PO-Revision-Date: 2025-01-26 21:48+0100\n"
 "Last-Translator: whitequark <whitequark@whitequark.org>\n"
 "Language-Team: none\n"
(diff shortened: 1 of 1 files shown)


ruevs (changes requested):
@progval thank you for the fast reaction. Please also translate the messages labeled:
constraint.cpp:158
constraint.cpp:193
constraint.cpp:244
constraint.cpp:887
where I added the new English text to the existing (already translated) French message.

progval:
I see, done.

Three of those I had to fix were just above the lines you mentioned, so I thought you made a mistake.

Next time, could you either use msgmerge (which should detect similarities and merge them accordingly), or mark these strings as `fuzzy`? this way they show up as needing attention in po editor; while comments are not always visible (are rarely searchable).

ruevs:
> I see, done.

  Thank you!

> Three of those I had to fix were just above the lines you mentioned, so I thought you made a mistake.

Indeed. When I added the English sentences in the translations I thought it may cause confusion but was not sure what is the proper way to do it.

> Next time, could you either use msgmerge (which should detect similarities and merge them accordingly), or mark these strings as `fuzzy`? this way they show up as needing attention in po editor; while comments are not always visible (are rarely searchable).

I'm pretty "amateurish" with translations - I've only used a .po editor once, perhaps 10 years ago. Currently I just use the make target that updates the translations and then use a diff tool to fix things by hand. I was not ware of the [fuzzy](https://poeditor.com/kb/fuzzy-translations) flag - I'll keep it in mind in the future.

PR #1517, opened 2025-01-27
Translated new messages to Ukrainian

Not4sure:
@ruevs  #1515


ruevs:
Thank for the fast response you @Not4sure.

PR #1520, opened 2025-01-28
Try to use macos-latest

vespakoen:
@ruevs I think the action getting stuck might be from the macos-11.0 image, I'd like to give this a try.
We can also configure it to only run on tags so that it doesn't run so often if the build credits become a problem.


ruevs:
It there a way to force the Python CI to run on this before merging it - so that we can check if it helps?
Perhaps merge it in your master and see what happens?

ruevs:
> Perhaps merge it in your master and see what happens?

Anyway I'll try it on my branch.

By the way @vespakoen why do you make branches in the main repository instead of using [your branch](https://github.com/vespakoen/solvespace)?

ruevs:
Works. Rebased and merged here 70e79332ade47de1c608291e025b50b5d66684f2 with a proper commit message.

vespakoen:
Cool! glad it worked.
I pushed to the wrong remote, I made a new clone and didn't swap origin and upstream yet, sorry for that ;)

PR #1523, opened 2025-02-01
Update ja_JP translation.

verylowfreq:
Update translation of ja_JP. https://github.com/solvespace/solvespace/issues/1515

Added translations for new messages and updated for existing translated messages.


ruevs:
Thank you @verylowfreq !

PR #1525, opened 2025-02-04
Add showing 'only unconstrained' groups

nobodyinperson:
This PR adds a new menu entry `only unconstrained` between `show all` and `hide all` that hides all groups besides those that still have degrees of freedom.

One of Solvespace's powers is that every single sketch entity stays draggable and reference-able all the time. But this can get overwhelming as in the below video. You can manually hide individual sketches, but this is tiring as you need to untick the right groups every time.

This PR adds a quick way to do this automatically.

https://github.com/user-attachments/assets/b17a383d-98a6-4499-a9b3-9e078fc63c76

### DIFF
--- a/src/textscreens.cpp
+++ b/src/textscreens.cpp
@@ -57,11 +57,13 @@ void TextWindow::ScreenToggleGroupShown(int link, uint32_t v) {
     SS.GenerateAll();
 }
 void TextWindow::ScreenShowGroupsSpecial(int link, uint32_t v) {
-    bool state = link == 's';
     for(hGroup hg : SK.groupOrder) {
         Group *g = SK.GetGroup(hg);
-
-        g->visible = state;
+        switch(link) {
+        case 's': g->visible = true; break;
+        case 'c': g->visible = g->solved.dof != 0; break;
+        case 'h': g->visible = false; break;
+        }
     }
     SS.GW.persistentDirty = true;
 }
@@ -159,7 +161,8 @@ void TextWindow::ShowListOfGroups() {
         backgroundParity = !backgroundParity;
     }
 
-    Printf(true,  "  %Fl%Ls%fshow all%E / %Fl%Lh%fhide all%E",
+    Printf(true, "  %Fl%Ls%fshow all%E / %Fl%Lc%fonly unconstrained%E / %Fl%Lh%fhide all%E",
+        &(TextWindow::ScreenShowGroupsSpecial),
         &(TextWindow::ScreenShowGroupsSpecial),
         &(TextWindow::ScreenShowGroupsSpecial));
     Printf(true,  "  %Fl%Ls%fline styles%E /"


ruevs:
@nobodyinperson this is beautiful! And the code is a minimal as it gets :-)
I'm seriously tempted to merge it immediately but would like to hear from @phkahler and @jwesthues first.

phkahler:
@nobodyinperson and @ruevs  I like it. The clicky text is a little long but I can't think of a way to shorten it without losing clarity.


On a related note, it would be great if hovering over a surface could highlight its group in the text window.  I just checked and the right-click menu can take you to the group a surface belongs to, but that switches to the group info which is not always what we want.   There is also desire for the reverse - to highlight a group in the sketch when the mouse is over the group in the text window, but I think that's a bigger task.

nobodyinperson:
> The clicky text is a little long but I can't think of a way to shorten it without losing clarity.

Exactly. At least the width still doesn't exceed the icons at the top, so it' fine.

> it would be great if hovering over a surface could highlight its group in the text window

Funny, I'm thinking about the same for some days now. 😁 How about:

- mouseover an element in the model (point, line, surface...) --> very subtle highlight in the group list
- double-click such element: **activate** that group, so the element can be constrained. Currently one has to go on an annoying hunt: click the elememt, find out its group name, back to the group list, manually find that group, activate it, now constrain it. Double-clicks are unused so far AFAIKT and I for one would need this feature regularly. Wasn't totally obvious how to do that in the code, maybe an entry in the right click context menu that activates AND shows the group might ne helpful already and easier to implement.

But all of thtat is for another PR 😃

ruevs:
Rebased and merged here f09bbb96168120dd7a91802d5a0d246312df35dd.

phkahler:
> * mouseover an element in the model (point, line, surface...) --> very subtle highlight in the group list

Why subtle?  I'm not sure how the highlighting would be implemented, but I think this is a really good idea.

> * double-click such element: **activate** that group, so the element can be constrained. Currently one has to go on an annoying hunt: click the elememt, find out its group name, back to the group list, manually find that group, activate it, now constrain it. Double-clicks are unused so far AFAIKT and I for one would need this feature regularly. Wasn't totally obvious how to do that in the code, maybe an entry in the right click context menu that activates AND shows the group might ne helpful already and easier to implement.

The annoying hunt would be helped a lot by the proposed highlighting. Adding something the the right-click menu to activate the group is probably a good idea.

If your goal is to make it easy to "go back" and constrain things in the sketch, you might start running into the whole click vs drag ambiguity. When you click (select) an entity, the one in the current group takes priority. But when you mouse-drag Solvespace knows certain entities are actually constrained and can not be dragged, so it discards them from selection allowing you to drag an entity from a previous group that maybe [can be dragged](https://github.com/solvespace/solvespace/blob/f09bbb96168120dd7a91802d5a0d246312df35dd/src/drawentity.cpp#L197). Usually these would need to be constrained in the previous group. Didn't someone recently ask for keys to navigate up and down the group list? Hmmm.

> But all of that is for another PR 😃

I look forward to it!

nobodyinperson:
> Why subtle?

Because one very often hovers over entities and each hover causing an aggressive highlight in the group list would make SolveSpace's typically very calm UI rather hectic. But we'd have to test that.

> If your goal is to make it easy to "go back" and constrain things in the sketch

Exactly, this is what I mean. Interesting with the auto-discard of constrained elements while dragging. Maybe SolveSpace could then auto-add the constraint to the right underlying group if one wants to contrain the "wrong" one in the top group? 🤔 But probably then one doesn't know the exact mapping of base to ambiguous element...

> Didn't someone recently ask for keys to navigate up and down the group list?

I for one would love PgUp and PgDown for example to move through groups, have been thinking about this for quite a while.

I'll add more PRs when I find the time. 🙂

PR #1530, opened 2025-02-21
Fix issues with type annotations in the Cython bindings

iscgar:
For some reason the Cython bindings contain invalid type annotations, even though the accompanying `.pyi` file has them correctly. This causes warnings to be emitted  during the Python library build, so I changed them to match.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,8 @@ classifiers = [
   "Programming Language :: Python :: 3.9",
   "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
 ]
 
 [tool.scikit-build]
--- a/src/lib.pyx
+++ b/src/lib.pyx
@@ -190,19 +190,19 @@ cpdef tuple make_quaternion(double ux, double uy, double uz, double vx, double v
     return qw, qx, qy, qz
 
 # entities
-def add_point_2d(grouph: int, u: double, v: double, workplane: Slvs_Entity) -> Slvs_Entity:
+def add_point_2d(grouph: int, u: float, v: float, workplane: Slvs_Entity) -> Slvs_Entity:
     return Slvs_AddPoint2D(grouph, u, v, workplane)
 
-def add_point_3d(grouph: int, x: double, y: double, z: double) -> Slvs_Entity:
+def add_point_3d(grouph: int, x: float, y: float, z: float) -> Slvs_Entity:
     return Slvs_AddPoint3D(grouph, x, y, z)
 
 def add_normal_2d(grouph: int, workplane: Slvs_Entity) -> Slvs_Entity:
     return Slvs_AddNormal2D(grouph, workplane)
 
-def add_normal_3d(grouph: int, qw: double, qx: double, qy: double, qz: double) -> Slvs_Entity:
+def add_normal_3d(grouph: int, qw: float, qx: float, qy: float, qz: float) -> Slvs_Entity:
     return Slvs_AddNormal3D(grouph, qw, qx, qy, qz)
 
-def add_distance(grouph: int, value: double, workplane: Slvs_Entity) -> Slvs_Entity:
+def add_distance(grouph: int, value: float, workplane: Slvs_Entity) -> Slvs_Entity:
     return Slvs_AddDistance(grouph, value, workplane)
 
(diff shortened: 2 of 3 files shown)


ruevs:
@vespakoen is your guy ;-)

pyproject.toml
```diff
@@ -20,6 +20,8 @@ classifiers = [
   "Programming Language :: Python :: 3.9",
   "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
```
vespakoen, line 24: Please also add new versions to the `cd.yml` matrix

src/lib.pyx
```diff
@@ -361,13 +361,13 @@ class EntityType(IntEnum):
     CIRCLE = _SLVS_E_CIRCLE
     ARC_OF_CIRCLE = _SLVS_E_ARC_OF_CIRCLE
 
-def solve_sketch(grouph: int, calculateFaileds: int):
+def solve_sketch(grouph: int, calculateFaileds: bool):
```
vespakoen, line 364: is `bool` supported just like that?

iscgar: Yes, and it actually doesn't work without it. When I run the tests, I get the following output:

```
======================================================================
ERROR: test_crank_rocker (test.CoreTest.test_crank_rocker)
Crank rocker example.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/isaacg/projects/cad-sketch/solvespace/python/tests/test.py", line 27, in test_crank_rocker
    slvs.angle(g, line0, line1, 45, wp, False)
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Argument 'inverse' has incorrect type (expected int, got bool)

======================================================================
ERROR: test_involute (test.CoreTest.test_involute)
Involute example.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/isaacg/projects/cad-sketch/solvespace/python/tests/test.py", line 54, in test_involute
    slvs.perpendicular(g, line0, line1, wp, False)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Argument 'inverse' has incorrect type (expected int, got bool)

======================================================================
ERROR: test_jansen_linkage (test.CoreTest.test_jansen_linkage)
Jansen's linkage example.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/isaacg/projects/cad-sketch/solvespace/python/tests/test.py", line 103, in test_jansen_linkage
    slvs.angle(g, line0, line_base, 45, wp, False)
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Argument 'inverse' has incorrect type (expected int, got bool)

======================================================================
ERROR: test_nut_cracker (test.CoreTest.test_nut_cracker)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/isaacg/projects/cad-sketch/solvespace/python/tests/test.py", line 140, in test_nut_cracker
    result = slvs.solve_sketch(g, False)
TypeError: Argument 'calculateFaileds' has incorrect type (expected int, got bool)

======================================================================
ERROR: test_pydemo (test.CoreTest.test_pydemo)
Some sample code for slvs.dll.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/isaacg/projects/cad-sketch/solvespace/python/tests/test.py", line 224, in test_pydemo
    g1result = slvs.solve_sketch(g1, False)
TypeError: Argument 'calculateFaileds' has incorrect type (expected int, got bool)

----------------------------------------------------------------------
```

pyproject.toml
iscgar: Done.

iscgar:
The build seems to be failing for Python 3.12 and 3.13 due to an outdated version of the `pypa/cibuildwheel` action. I'll open a separate PR to address it (maybe we should change it so that the pipeline is run on the PR before the merge to `master`?).

vespakoen:
I cannot trigger a python build from the UI but perhaps you can add your branch to the list of triggers for the CD pipeline and then it might be possible to run it before merging

iscgar:
I'll do this temporarily, though I don't think specifying private branches every time is the way to go. I'll add a `workflow_dispatch` trigger to allow triggering it in the future.

PR #1531, opened 2025-02-21
Corrected ZX Plane Rotation.

spy653:
ZX Plane is incorrectly oriented resulting in H/V constraints being swapped and having the planes label on the wrong edge. 

This has annoyed me as a user for years.

### DIFF
--- a/src/generate.cpp
+++ b/src/generate.cpp
@@ -393,7 +393,7 @@ void SolveSpaceUI::ForceReferences() {
     } Quat[] = {
         { Request::HREQUEST_REFERENCE_XY, { 1,    0,    0,    0,   } },
         { Request::HREQUEST_REFERENCE_YZ, { 0.5,  0.5,  0.5,  0.5, } },
-        { Request::HREQUEST_REFERENCE_ZX, { 0.707,0.707,0,    0,   } },
+        { Request::HREQUEST_REFERENCE_ZX, { 0,    0.707,0.707,0,   } },
     };
     for(int i = 0; i < 3; i++) {
         hRequest hr = Quat[i].hr;
--- a/src/generate.cpp
+++ b/src/generate.cpp
@@ -393,7 +393,7 @@ void SolveSpaceUI::ForceReferences() {
     } Quat[] = {
         { Request::HREQUEST_REFERENCE_XY, { 1,    0,    0,    0,   } },
         { Request::HREQUEST_REFERENCE_YZ, { 0.5,  0.5,  0.5,  0.5, } },
-        { Request::HREQUEST_REFERENCE_ZX, { 0,    0.707,0.707,0,   } },
+        { Request::HREQUEST_REFERENCE_ZX, { 0,    0, 0.707, 0.707,   } },
     };
     for(int i = 0; i < 3; i++) {
         hRequest hr = Quat[i].hr;
--- a/src/generate.cpp
+++ b/src/generate.cpp
@@ -393,7 +393,7 @@ void SolveSpaceUI::ForceReferences() {
     } Quat[] = {
         { Request::HREQUEST_REFERENCE_XY, { 1,    0,    0,    0,   } },
         { Request::HREQUEST_REFERENCE_YZ, { 0.5,  0.5,  0.5,  0.5, } },
-        { Request::HREQUEST_REFERENCE_ZX, { 0.5, -0.5, -0.5, -0.5, } },
+        { Request::HREQUEST_REFERENCE_ZX, { 0.707,0.707,0,    0,   } },
     };
     for(int i = 0; i < 3; i++) {
         hRequest hr = Quat[i].hr;


spy653:
Test failures appears unrelated to file change.

ruevs:
@spy653 would you please explain what you mean by "incorrectly oriented" and "H/V constraints being swapped"? Perhaps some annotated screen shots?

I think your change makes the coordinate system "[left handed](https://en.wikipedia.org/wiki/Right-hand_rule)" which is certainly not "standard".

By the way the tests are fail because of the change. The tests compare reference PNG files with corresponding slvs files and check whether the slvs files generate pixel perfect matches for the PNG files (look here https://github.com/solvespace/solvespace/tree/master/test).

spy653:
> @spy653 would you please explain what you mean by "incorrectly oriented" and "H/V constraints being swapped"? Perhaps some annotated screen shots?
SS-ZX-Normal
SS-ZX-Mine
As you can see, in the first image the vertical lines are constrained with horizontal, and vice versa. This is the issue I was attempting to solve, also noticing that the frame title appears in the bottom right rather than the bottom left.
Vs in the second image the constraints and title are correct.

> I think your change makes the coordinate system "[left handed](https://en.wikipedia.org/wiki/Right-hand_rule)" which is certainly not "standard".
Ah yes I see now that I am most used to left hand rule and didn't notice that my quaternion viewer has an unusual axis configuration. I can see that my change unintentionally inverts an axis, when I was only intending to rotate the normal.

> By the way the tests do fail because of the change. They compare reference PNG files with corresponding slvs files and check whether the slvs files generate pixel perfect matches for the PNG files (look here https://github.com/solvespace/solvespace/tree/master/test). You change of course changes what they look like.
I see now what you mean, I apologize for not knowing about the tests, first time trying to contribute.

ruevs:
> As you can see, in the first image the vertical lines are constrained with horizontal, and vice versa. This is the issue I was attempting to solve, also noticing that the frame title appears in the bottom right rather than the bottom left.
> Vs in the second image the constraints and title are correct.

I think I understand now.

You can create a workplane coinciding with but rotated at any multiple of 90° relative to the ZX (or any other) reference plane:

1. Rotate the view so that ZX is roughly parallel to the screen.
2. `F2`
3. Rorate the view in the plane of the screen with `CTRL+RightButtonDrag` so that the ZX plane is rotated closest to the (90°) orientation you want.
4. `F2`
5. Select the origin.
6. `Shift+W` (or Sketch -> Workplane if you want a new workplane in the current group)

With the above you can create a workplane coinciding with ZX but rotated any way you wish.
The horisontal and vertical constraints are "relative" to the active workplane and not the reference planes(s). 

Also `W` aligns the view to the active workplane (unlike F2 which aligns it to the nearest orthogonal view)

I think this "solves" your problem without needing to change the orientation of the reference quaternions.

>> I think your change makes the coordinate system "[left handed](https://en.wikipedia.org/wiki/Right-hand_rule)" which is certainly not "standard".

> Ah yes I see now that I am most used to left hand rule and didn't notice that my quaternion viewer has an unusual axis configuration. I can see that my change unintentionally inverts an axis, when I was only intending to rotate the normal.

So do you think that the ZX plane normal/quaternion is not properly rotated? I'll have to think a lot more carefully to imagine how the three reference planes are rotated and if that rotation is "correct" - my mind totally lacks a quaternion visualization logic :-)

> I see now what you mean, I apologize for not knowing about the tests, first time trying to contribute.

No need to apologize. The tests are an obscure thing and not documented anywhere. We can adjust the test cases if/when we decide to make a change that breaks them. And I am always very glad when someone new decides to look at the SolveSpace source and contribute.

If what I described above is not what you had in mind or not what your problem was please do not hesitate to explain/argue further :-)

spy653:
> I think this "solves" your problem without needing to change the orientation of the reference quaternions.

The reason I believe this behavior is an error is that I created this workplane by simply selecting the ZX plane, effectively duplicating it, 

> So do you think that the ZX plane normal/quaternion is not properly rotated? I'll have to think a lot more carefully to imagine how the three reference planes are rotated and if that rotation is "correct" - my mind totally lacks a quaternion visualization logic :-)

I do believe the quaternion is not correctly rotated, I have a background in game development so have a fairly solid understanding of them, I believe that even though it is oriented correctly, it's rotation is 90 degrees off, in its kind of 4th dimension.
 
SS-Mine
SS-Old
look at the rotation of the highlighted arrow in the original build vs mine

I initially incorrectly assumed the order of axes involved but believe I have corrected that now. I am still getting the hang of tests though.

spy653:
> ... believe I have corrected that now. 

I had not.

spy653:
Having determined the required axes to modify, I have updating the PR and whilst the tests fail, looking at the diff between expected and actual test results it can be seen that all functionality is the same but the pixel difference is from the intentionally different change to axis orientation. Also having been testing this locally it's worth mentioning that as this code is only called on generation of a new file, it's non-breaking for existing files. I believe it is now ready.

ruevs:
OK I refreshed my mental model of [how quaternions work](https://eater.net/quaternions/video/intro) (for the Nth time in my life) and carefully considered your suggestion.

In my opinion the orientations are correct as they are. And I still can not reproduce your problem with horisontal and vertical being swapped. In your first screen shot [here](https://github.com/solvespace/solvespace/pull/1531#issuecomment-2676998786) the screen view is simply rotated 90° CCW.

Here is how the current quaternions rotate the XY (ij) plane to create the three reference planes:
1. XY {1,    0,    0,    0} - no rotation 
image
2. YZ { 0.5,  0.5,  0.5,  0.5} - the vector is 0.5, 0.5, 0.5 in other words - pointing "in the middle" of the first octant, rotate 120° (60° in the quaternion) (cos(60)=sin(60)=0.5):
image
2. ZX { 0.5, -0.5, -0.5, -0.5, } ([equivalent to](https://eater.net/quaternions/video/doublecover) { -0.5, 0.5, 0.5, 0.5, }) - same vector but rotate -120° ((60° in the quaternion):
image

The above is "right" because the rotation of the planes is consistent:
- from X1Y1 to Y2Z2: X1 becomes Y2, Y1 becomes Z2
- From Y2Z2 to Z2X2: Y2 becomes Z3, Z2 becomes X2 (if you substitute the above X1 becomes Z3, Y1 becomes X3)
- and if one continues rotationg: Z3X3 to X1Y1 - Z3 becomes  X1, X3 becomes Y1 - or substituting X1 becomes X1, Y1 becomes Y1 - and it becomes a loop.

In other words the coordinate system can be continuously rotated through 120° degree and each plane "becomes" the next one.

Another advantage is that 1 and 0.5 are exactly representable in `float` while your proposed 0.70710678119 (cos(45) in the quaternion, rotation by 90°) is not and leads to rounding errors form the very beginning of the model.

For comparison here is what your proposed quaternion does:
image
this does not have the property that you can endlessly rotate the coordinate system and have the planes and their orientations line up.

While I am pretty sure that I am right and the above is correct I would still like to hear what @phkahler and @jwesthues think of this.


P.S. Here is "the vector in the middle of the first octant" that we rotate the coordinate system around that I mentioned above:
image

spy653:
I agree with what you're saying, however *personally*: I view the consistency of the constraints in the default plane as more important than cleanliness of avoiding floating point errors. 

I'm curious about you not being able to reproduce the issue in question (shown below) as this screenshot is taken from an unmodified install. Just for clarity I'll repeat that I 

`Ctrl-N` open a new file, 
select the ZX plane (g001 - r003), Shift+W to create a new workplane from selected, and R to create rectange.
SS-ZX-Normal

jwesthues:
You can always view the model from any orientation, so it's always possible to rotate a plane so that lines constrained horizontal look vertical on your display (or diagonal, or invisible because they're on edge, etc.). I think if you choose View -> Align View to Workplane then the horizontal lines should look horizontal though?

The quaternion part is just an implementation detail. A workplane has a horizontal axis u, a vertical axis v, and a normal n = u &times; v. The are 3! = 6 ways to assign {x, y, z} to {u, v, n}, of which three satisfy that cross product relationship. Those three are the planes we're using now. To use any other would break the symmetries that @ruevs notes and many others.

Perhaps it feels bothersome that x is the horizontal axis in the XY plane, but the vertical axis in the ZX? But symmetry implies that must be so--there are three workplanes and three axes, so each of {x, y, z} must be the horizontal axis in exactly one workplane.

This is independent of whether the base coordinate system is drawn left- or right-handed. For completeness, you could choose n = -u &times; v if you wanted workplanes with opposite handedness to your base coordinate system. That's symmetric but doesn't fix the bother, since x is now horizontal in the XZ plane but vertical in the YX.

ruevs:
> I'm curious about you not being able to reproduce the issue in question (shown below) as this screenshot is taken from an unmodified install. Just for clarity I'll repeat that:

I can reproduce the picture exactly. But as soon as I press `W` (View -> Align View to Work plane) the view rotates 90° clockwise, vertical looks vertical, horizontal looks horizontal, the X axis points up and the Z axis points right. And I just like @jwesthues consider this normal/logical/expected.

spy653:
> ... so each of {x, y, z} must be the horizontal axis in exactly one workplane.

I can see where you're coming from, the idea of X not being a horizontal axis just feels wrong to me, perhaps 'X is always horizontal' could be available as a setting. 
(Personally I think I will continue to use my fork until the next major version comes out.)

I won't try to persuade you if this is how you think, I merely thought this was unintentional and had been overlooked. Would you like to see what the other maintainers think or close it now?

phkahler:
> > ... so each of {x, y, z} must be the horizontal axis in exactly one workplane.
> 
> I can see where you're coming from, the idea of X not being a horizontal axis just feels wrong to me, perhaps 'X is always horizontal' could be available as a setting. (Personally I think I will continue to use my fork until the next major version comes out.)
> 
> I won't try to persuade you if this is how you think, I merely thought this was unintentional and had been overlooked. Would you like to see what the other maintainers think or close it now?

I just came to this issue.  When I start a new sketch, rotate the reference axis a little so I can select the Z-X plane and press Shift-W to sketch in that workplane it works as I expect and the view rotates to bring that workplane into alignment with the screen. In that orientation H and V constraints work as expected.  The workplanes are labled XY, YZ, and ZX,   with the first letter of each pair being the horizontal axis and the second being vertical. Different industries sometimes use weird coordinate systems ( a common vehicle Csys uses X as the forward direction, Y to the left and Z toward the sky). I'm not sure this should be changed, and it would probably break existing sketches.

This does remind me of a related issue. By default Solvespace creates a sketch in the XY plane which initially can not be deleted. But if you select a different reference plane and create a new sketch-in-workplane it is then possible to delete the XY one. But then your groups will be g001 and g003. That's a little odd but not a big deal.

jwesthues:
> and it would probably break existing sketches.

Indeed, redefining any coordinate plane would break any sketch that had constrained against it using the prior definition. It would be possible to define additional coordinate planes, though I don't see what makes XZ more special than all the other possibilities (six using only the positive coordinate axes, twenty-four including negative) already available through `WORKPLANE_BY_POINT_ORTHO`.

I introduced those planes primarily so their normals could serve visually as the axes of our base coordinate system. The XZ plane would give us (x, -y, z), which seems surprising.

spy653:
From what I've seen whilst making this change: existing files are actually uneffected. 

As this change is only called during the initial generation of the file, older files already have their basis vectors so this isn't called.

I did expect this to be a breaking change but when opening old files it ended up being fine.

PR #1534, opened 2025-02-22
Fix PyPi publishing

vespakoen:
Attempt to fix the python publishing to PyPi by disabling attestations and enabling verbose output for debugging potential further issues.

I pushed this branch to the solvespace repository and ran it using the new workflow trigger option to try it out (with the solvespace secrets), it's running over here: https://github.com/solvespace/solvespace/actions/runs/13476643225


jwesthues:
Thanks for looking at this, and I've just set up 2FA. Let me know if you need anything else--I see there's already an API token, but "last used" shows as "never".

vespakoen:
Thanks! I just re-ran the CI but it failed again with the same error.
Perhaps the 2FA has to be setup for https://test.pypi.org separately?

jwesthues:
> Perhaps the 2FA has to be setup for https://test.pypi.org/ separately?

Seems so, just repeated the process there. That one also has an API token, also never used.

vespakoen:
It got further now, but apparently it doesn't like the name `solvespace` for the package, trying `slvs` now...
if that doesn't work I guess we have to open an issue upstream (https://peps.python.org/pep-0541/#how-to-request-a-name-transfer)

The API tokens seem to work, they just were never (successfully) used yet.

jwesthues:
In the test environment, I think we're a maintainer for `slvs` but not `solvespace`? I think you just successfully used the token for `slvs`, cool.

In the production environment, neither seems to exist yet. It would be nice to resolve the name inconsistency (user `solvespace` with project `slvs`) if possible first, though not vital if that's hard.

vespakoen:
Indeed, it worked for me as well, I can install the `slvs` package from test pypi, and run the tests:

```
❯ python3 -m venv ./venv
❯ source ./venv/bin/activate
❯ python3 -m pip install --index-url https://test.pypi.org/simple slvs
❯ cd python
❯ python3 tests
Crank rocker
.Involute
result {'result': 0, 'dof': 6, 'rank': 0, 'bad': 0}
.Jansen's linkage
result {'result': 0, 'dof': 10, 'rank': 0, 'bad': 0}
.Nut cracker
result {'result': 0, 'dof': 6, 'rank': 0, 'bad': 0}
x 1.3907598174588294
..
----------------------------------------------------------------------
Ran 5 tests in 0.018s

OK
```

ruevs:
@vespakoen while you are at it the WASM library publishing fails as well:
https://github.com/solvespace/solvespace/actions/runs/13484704162/job/37674426136#step:6:126

```
npm notice 
npm ERR! code E403
npm ERR! 403 403 Forbidden - PUT https://registry.npmjs.org/slvs - You cannot publish over the previously published versions: 3.1.0-dev.20.
npm ERR! 403 In most cases, you or one of your dependencies are requesting
npm ERR! 403 a package version that is forbidden by your security policy.
```

vespakoen:
@ruevs The wasmlib library cannot be published because apparently I have been publishing them before from my own fork.

Currently:
- the version number is still the same
- the build number (github action run number) is lower on this repo (20) than the one on mine (22)

This will automatically solve itself when bumping the solvespace version and / or after 2 more failing builds

ruevs:
> @ruevs The wasmlib library cannot be published because

Thank you very much for the explanation. Makes sense.

ruevs (commented):
Since you requested a review @vespakoen - this looks OK to me...
That said (as is plainly obvious) my CI/Python knowledge is almost non-existent :-)

ruevs:
It worked :-)
https://github.com/solvespace/solvespace/actions/runs/13485993704

vespakoen:
done ;)

Issue #1186, opened 2022-01-09
Performance testing models (slow speed benchmarks)


ruevs:
This is a list of SolveSpace models that have shown performance bottlenecks in the past. Some of them were resolved, some were improved but still are slow.

I have marked all related issues and pull requests (that I could find or remember) with the [`performance`](https://github.com/solvespace/solvespace/issues?q=label%3Aperformance+sort%3Acreated-desc) label.

Links to the `.slvs` models reproducing the problems and the fixes may be useful for performance testing.
If the "Fixed on" column is empty the issue is probably still open (or I did not update this list :-)

Source | Model | Author | Fixed on | Fixed in
------ | ----- | ------ | -------- | --------
https://github.com/solvespace/solvespace/issues/131 <br/> https://github.com/solvespace/solvespace/issues/225 | [constrainfreeze.zip](https://github.com/solvespace/solvespace/files/645378/constrainfreeze.zip) | @whitequark | 2020-09-16 workaround | https://github.com/solvespace/solvespace/commit/615708440f97570cd7055ec65ec0a3bbe6f8f18d
https://github.com/solvespace/solvespace/issues/386#issuecomment-544301008 <br/> https://github.com/solvespace/solvespace/issues/386#issuecomment-998587512 <br/> https://github.com/solvespace/solvespace/issues/386#issuecomment-1000371505https://github.com/solvespace/solvespace/issues/386#issuecomment-1000371505 <br/>  | [profiling.zip](https://github.com/solvespace/solvespace/files/3748504/profiling.zip)<br/>[keyb_case_MAX_UNKNOWNS_2048.zip](https://github.com/solvespace/solvespace/files/7750846/keyb_case_MAX_UNKNOWNS_2048.zip)<br/>[keyb_case_MAX_UNKNOWNS_4096.zip](https://github.com/solvespace/solvespace/files/7770072/keyb_case_MAX_UNKNOWNS_4096.zip)<br/>[]()| @bmorel<br/>@ruevs<br/>@ruevs<br/> | 2021-12-23 | https://github.com/solvespace/solvespace/pull/1159
https://github.com/solvespace/solvespace/pull/594#issuecomment-624983188 <br/>  <br/>  <br/>  | [intersect_ct1.zip](https://github.com/solvespace/solvespace/files/4590263/intersect_ct1.zip) | @phkahler | 2020-05-08 | https://github.com/solvespace/solvespace/pull/594
[...]

PR #1536, opened 2025-02-25
Minor refactoring of surface curve data into its own struct

iscgar:
This allows keeping its functionality together and gives us the ability to copy it separately when needed, avoiding a copy of the entire surface data.

This was the case in `MakeFirstOrderRevolvedSurfaces()` as well as in `SplitInHalf()` -- both of which have been converted to work on the curve data alone.

### DIFF
--- a/src/srf/raycast.cpp
+++ b/src/srf/raycast.cpp
@@ -125,14 +125,18 @@ void SSurface::BlendRowOrCol(bool row, int this_ij, SSurface *a, int a_ij,
         }
     }
 }
-void SSurface::SplitInHalf(bool byU, SSurface *sa, SSurface *sb) {
-    sa->degm = sb->degm = degm;
-    sa->degn = sb->degn = degn;
-
+void SSurface::SplitInHalf(bool byU, SSurface *sa, SSurface *sb) const {
     // by de Casteljau's algorithm in a projective space; so we must work
     // on points (w*x, w*y, w*z, w) so create a temporary copy
-    SSurface st;
-    st = *this;
+    SSurface st = {};
+
+    // Only copy the data we need into the new object
+    memcpy(st.ctrl, ctrl, sizeof(ctrl));
+    memcpy(st.weight, weight, sizeof(weight));
+
+    st.degm = sa->degm = sb->degm = degm;
+    st.degn = sa->degn = sb->degn = degn;
+
     st.WeightControlPoints();
 
     switch(byU ? degm : degn) {
--- a/src/srf/shell.cpp
+++ b/src/srf/shell.cpp
@@ -524,21 +524,26 @@ void SShell::MakeFirstOrderRevolvedSurfaces(Vector pt, Vector axis, int i0) {
             if(fabs(d0 - d1) < LENGTH_EPS) {
                 // This is a cylinder; so transpose it so that we'll recognize
                 // it as a surface of extrusion.
-                SSurface sn = *srf;
 
                 // Transposing u and v flips the normal, so reverse u to
-                // flip it again and put it back where we started.
-                sn.degm = 2;
-                sn.degn = 1;
-                int dm, dn;
-                for(dm = 0; dm <= 1; dm++) {
-                    for(dn = 0; dn <= 2; dn++) {
-                        sn.ctrl  [dn][dm] = srf->ctrl  [1-dm][dn];
-                        sn.weight[dn][dm] = srf->weight[1-dm][dn];
+                // flip it again and put it back where we started (work on
+                // a copy and copy back the result in order to make the loop
+                // simpler).
+                decltype(SSurface::ctrl) ctrl = {};
+                decltype(SSurface::weight) weight = {};
+
+                srf->degm = 2;
+                srf->degn = 1;
+                for(int dm = 0; dm <= 1; dm++) {
+                    for(int dn = 0; dn <= 2; dn++) {
+                        ctrl  [dn][dm] = srf->ctrl  [1-dm][dn];
+                        weight[dn][dm] = srf->weight[1-dm][dn];
                     }
                 }
 
-                *srf = sn;
+                // Copy back the transposed data
+                memcpy(srf->ctrl, ctrl, sizeof(ctrl));
+                memcpy(srf->weight, weight, sizeof(weight));
                 continue;
             }
         }
--- a/src/srf/surface.h
+++ b/src/srf/surface.h
@@ -317,7 +317,7 @@ public:
     void BlendRowOrCol(bool row, int this_ij, SSurface *a, int a_ij,
                                               SSurface *b, int b_ij);
     double DepartureFromCoplanar() const;
-    void SplitInHalf(bool byU, SSurface *sa, SSurface *sb);
+    void SplitInHalf(bool byU, SSurface *sa, SSurface *sb) const;
     void AllPointsIntersecting(Vector a, Vector b,
                                List<SInter> *l,
                                bool asSegment, bool trimmed, bool inclTangent);


ruevs:
This may avoid copying a few bytes here and there (as in [`SplitInHalf`](https://github.com/solvespace/solvespace/blob/a5d40f81e3c2fccf9fd1cf28de6b4d4485fde068/src/srf/raycast.cpp#L135)) but in my opinion it is not worth it. The many `curve.` and to a lesser extent `Curve::` that appear are ugly and cause "churn" see my comment about it [here](https://github.com/solvespace/solvespace/pull/1537#pullrequestreview-2643864964).

However if you have an example model where profiling shows that this change causes a significant speed up or is on the critical path we should consider it.

phkahler:
@iscgar   Are you saying that SplitInHalf is doing a deep copy of all the trims? That should not be happening, but if it is we should fix it for performance reasons.  BTW we don't care about copying the color or handle, that's very minor.

My biggest with this issue is that names are important. In Solvespace a curve is a one-dimensional curve either in a plane or in 3d space. A surface is a 2-D surface in 3-D space. We don't want to use curve terminology for a surface data structure - that's just confusing.

Second complaint is that this just takes more text to specify a piece of data. Almost every line of the diff is just adding the extra colons and the word "curve" to the code.

iscgar:
> This may avoid copying a few bytes here and there (as in [`SplitInHalf`](https://github.com/solvespace/solvespace/blob/a5d40f81e3c2fccf9fd1cf28de6b4d4485fde068/src/srf/raycast.cpp#L135)) but in my opinion it is not worth it. The many `curve.` and to a lesser extent `Curve::` that appear are ugly and cause "churn" see my comment about it [here](https://github.com/solvespace/solvespace/pull/1537#pullrequestreview-2643864964).
> 
> However if you have an example model where profiling shows that this change causes a significant speed up or is on the critical path we should consider it.

Right now this is indeed only copying a few additional bytes because copying a `List<T>` is a shallow operation which does not copy the stored items, as `List<T>` requires an explicit call to `.Clear()` in order to destroy the items and free the memory.

But as I noted in my reply to your comment there, I'm working on reworking the way `List<T>` and `IdList<T>` are implemented in order to speed them up and reduce memory usage. This includes making `List<T>` an owner of its memory and responsible for tracking and releasing it, and therefore a copy of it performs a deep copy of its underlying storage instead of just copying a pointer.

As part of that change I also made the following conversions to code that copied `List<T>`:
* Places that were copying a `List<T>` and clearing the source were changed into performing an `std::move()`
* Places where items were copied one-by-one using `.Add()` (in order to make a deep copy) were changed to a simple assignment

In that context, copying the entire `SSurface` struct can become a heavy operation if there are many edges or trims, and neither of the aforementioned conversions could work in the two cases that are mentioned in the commit message, so I resorted to refactoring the curve data into its own struct in order to allow working on it without running into the potential performance issue.

> @iscgar Are you saying that SplitInHalf is doing a deep copy of all the trims? That should not be happening, but if it is we should fix it for performance reasons. BTW we don't care about copying the color or handle, that's very minor.

As I explain above, no deep copy is performed with the existing code.

> My biggest with this issue is that names are important. In Solvespace a curve is a one-dimensional curve either in a plane or in 3d space. A surface is a 2-D surface in 3-D space. We don't want to use curve terminology for a surface data structure - that's just confusing.
> 
> Second complaint is that this just takes more text to specify a piece of data. Almost every line of the diff is just adding the extra colons and the word "curve" to the code.

Both valid points, and I fully accept your reservations (especially in light of @ruevs 's comment in the other PR about avoiding disturbing the existing code as much as possible). However, assuming my reason for making this change seems reasonable, I'm open to suggestions for a better term, or to thinking of other ways to achieve the same result without introducing another nesting level.

ruevs:
> I'm working on reworking the way `List<T>` and `IdList<T>` are implemented

You will not be the first one ;-)
I rewrote IdList four years ago
https://github.com/solvespace/solvespace/pull/1002

with help, ideas and inspiration from:
@pjanx 
https://github.com/solvespace/solvespace/pull/992
https://github.com/solvespace/solvespace/pull/973

@rpavlik 
https://github.com/solvespace/solvespace/pull/433
https://github.com/solvespace/solvespace/pull/429

@Evil-Spirit 

Discussions:
https://github.com/solvespace/solvespace/issues/939
https://github.com/solvespace/solvespace/issues/932

In general keep in mind these when working on performance:
https://github.com/solvespace/solvespace/labels/performance

Test models to profile with:
https://github.com/solvespace/solvespace/issues/1186

Last time I profiled (about half an year ago, but nothing performance related has changed since) neither List nor IdList were a bottleneck. For example on this https://github.com/solvespace/solvespace/issues/759#issuecomment-713442468 most of the time was spent on axis aligned bounding boxes :-) A torture test model for IdList is here:
https://github.com/solvespace/solvespace/issues/932#issuecomment-778506906
and one more:
https://github.com/solvespace/solvespace/issues/841#issuecomment-742153838

More wisdom from @rpavlik https://github.com/solvespace/solvespace/issues/932#issuecomment-779157700

Test and profile with various models in _release_ mode (with optimisations on and debug info for profiling). A debug build is no indication at all about the real performance and bottlenecks.

In short - I'm very exited that you have started looking deeply into SolveSpace and am very curious what will come out of it. 

What platform are you on?

iscgar:
@ruevs thanks for the links to past discussions and ideas! I've seen some of them, but definitely not all of them. I'll make sure to take the information in them into account before I submit my work.

I'm on Linux, but I'm also running on Windows every once in a while.

iscgar:
I added a different approach to achieve the same result, which hopefully addresses the concerns raised (assuming such a change seems reasonable despite not having an impact on the existing implementation).

I intentionally did it by adding a revert commit and then the new approach, so that the two approaches could be compared easily. If the change is acceptable, I'll remove the first two commits.

phkahler:
One possible reason to put the raw Rational Bezier Surface data into a separate struct within SSurface would be to abstract the underlying surface type (later - much much later) so someone could implement other types of surfaces, or just more generalized NURBS surfaces.

So if we allowed that, there must be a mathematical word for "the surface that the trim polygon defines a subset of". We'd want to name it that instead of curve (It must not be curve! ;)  and maybe use an abbreviation.

iscgar:
I only used `Curve` because I didn't want to add to the confusion by having a a type named `Surface` nested inside `SSurface`, but if an abbreviation is acceptable, the type can just be `RationalBezierSurface` and the member abbreviated to `rbs`.

But that's assuming we indeed want to go this way instead of just changing the two places where this will become relevant with my other changes.

ruevs:
> However if you have an example model where profiling shows that this change causes a significant speed up or is on the critical path we should consider it.

Here is a model
[SplineIntersectedWith128ConeLikeThings.zip](https://github.com/user-attachments/files/19129104/SplineIntersectedWith128ConeLikeThings.zip)
that (on Windows, 32 bit release build, with LTO, with MSVC 2019) spends 31% of its time in:

```
SSurface::AllPointsIntersecting
 and then in:
   SSurface::AllPointsIntersectingUntrimmed
      ClosestPointTo(p, &u, &v);
        Vector tryp = PointAt(tryu, tryv);
```
image

As you can see currently the bottleneck is not `SplitInHalf` but in `SSurface::ClosestPointTo` -> `SSurface::PointAt`. Unfortunately nothing can be done about `SSurface::PointAt` (at least that I can think of).
image

 The recursive `AllPointsIntersectingUntrimmed` that does the `SplitInHalf` is more interesting. Comenting out the 5 debug lines starting with this one:

https://github.com/solvespace/solvespace/blob/bd7b62785a612e03c54769626193b3aa3208cdc9/src/srf/raycast.cpp#L215

Reduces the loading time from 15.1s to 13.9s on my machine. The critical path is still the same, but a bit faster. @phkahler perhaps we should #ifdef DEBUG those lines?

phkahler:
From a performance perspective, I think optimizing low level data structures is a waste of time. Even the OpenMP stuff is kind of a hack (but a really good one). The real performance problems come from algorithms that are O(n^2) and higher. The fix for those things will be a spatial index to reduce the number of things tested against other things. I've got an index in mind but using that would make the code more complex and harder to fix the bugs we still have in the NURBS code. We're still in the "make it work right" phase but some "make it fast" changes are still OK.

BTW a major bottleneck for a lot of things is the silhouette edges being rendered. Turn those of in the GUI  ;-)

Even the first item in [the old wishlist](https://github.com/solvespace/solvespace/blob/master/wishlist.txt) is related to algorithmic complexity. That is still an O(n^2) algorithm.

PR #1541, opened 2025-03-04
render: remove a call to `AllocTemporary()` from the OpenGL 1 renderer

iscgar:
This needlessly uses the global temporary arena, when the lifetime can be managed by replacing `GLU_TESS_COMBINE` with `GLU_TESS_COMBINE_DATA` instead.

While the use of the global temporary arena is mostly harmless, having it around needlessly prevents restricting the arena's scope to the sketch in a possible future where the global sketch is removed. Additionally, it means that calls to `FreeAllTemporary()` must synchronise with the renderer, which is a needless restriction on the users of the arena given the fact that none of the other renderers use it.

### DIFF
--- a/src/render/rendergl1.cpp
+++ b/src/render/rendergl1.cpp
@@ -557,34 +557,39 @@ void OpenGl1Renderer::DrawPoint(const Vector &o, Canvas::hStroke hcs) {
 #endif
 typedef void(SSGL_CALLBACK *GLUCallback)();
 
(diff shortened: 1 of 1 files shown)


phkahler:
Just want to ping @ruevs since he is the main supporter of OpenGL 1 which is primarily for older hardware. Looks fine to me.

ruevs:
Merged. I wanted to be on a PC to look though it carefully and make sure that the smart pointer going out of scope is what frees the memory.

PR #1542, opened 2025-03-06
Expose the solver's `SolveResult::REDUNDANT_OKAY` result in the slvs library

iscgar:
Currently there's no way to distinguish between a successful solve and a failed one in the case of an overly constrained sketch, as the return value is the same in both cases (`SLVS_RESULT_INCONSISTENT`).

Allow users to distinguish between the two by giving a successful solve a distinct return value, so that users of the solver could use the solution and deal with the redundant constraints in any way they like.


ruevs:
I'm perfectly fine with this (but I never use the library or the Python bindings). @vespakoen, @phkahler what do you think?

vespakoen (approved):

phkahler:
> I'm perfectly fine with this (but I never use the library, the Python bindings or the WASM library). @vespakoen, @phkahler what do you think?

I've never used that stuff either. Sometimes I'm also annoyed that the library exists because it complicates things a bit, but as long as someone is taking care of it I'm good ;-)  Thank you @vespakoen and @iscgar for that.

PR #1543, opened 2025-03-06
Update the vendored FreeType version to 2.13.3

### LINKED ISSUE #1539: macOS build instructions in README don't work
OS: macOS Sequoia 15.3.1 arm64

```text
$ brew install git cmake libomp
$ git submodule update --init
$ cd build
$ cmake .. -DCMAKE_BUILD_TYPE=Release -DENABLE_OPENMP=ON
-- The C compiler identification is AppleClang 16.0.0.16000026
-- The CXX compiler identification is AppleClang 16.0.0.16000026
-- The ASM compiler identification is AppleClang
-- Found assembler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at /opt/homebrew/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
  Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES)
Call Stack (most recent call first):
  /opt/homebrew/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
  /opt/homebrew/share/cmake/Modules/FindOpenMP.cmake:616 (find_package_handle_standard_args)
  CMakeLists.txt:123 (find_package)


-- Configuring incomplete, errors occurred!

$ cmake .. -G Xcode
CMake Error: Error: generator : Xcode
Does not match the generator used previously: Unix Makefiles
Either remove the CMakeCache.txt file and CMakeFiles directory or choose a different binary directory.
```

iscgar:
The existing version (2.7.1) is using an ancient zlib version, which fails to build with modern Xcode tools on macOS.

Fixes #1539

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,7 +7,7 @@
 	url = https://github.com/glennrp/libpng
 [submodule "extlib/freetype"]
 	path = extlib/freetype
-	url = https://git.savannah.nongnu.org/r/freetype/freetype2.git/
+	url = https://git.savannah.nongnu.org/git/freetype/freetype2.git
 [submodule "extlib/libdxfrw"]
 	path = extlib/libdxfrw
 	url = https://github.com/solvespace/libdxfrw.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -252,10 +252,9 @@ if(ENABLE_GUI OR ENABLE_CLI)
         list(APPEND PNG_PNG_INCLUDE_DIR ${CMAKE_BINARY_DIR}/extlib/libpng)
 
         find_vendored_package(Freetype freetype
-            WITH_ZLIB               OFF
-            WITH_BZip2              OFF
-            WITH_PNG                OFF
-            WITH_HarfBuzz           OFF
+            FT_DISABLE_BZIP2        ON
+            FT_DISABLE_HARFBUZZ     ON
+            FT_DISABLE_BROTLI       ON
             FREETYPE_LIBRARY        freetype
             FREETYPE_INCLUDE_DIRS   ${CMAKE_SOURCE_DIR}/extlib/freetype/include)
 


iscgar:
Had to force push because I didn't have a way to test this before pushing, and apparently the name of the configuration variables to disable dependencies changed between the versions (and Brotli was added as well). What's weird though, is that we were already using `WITH_ZLIB=OFF`, but for some reason FreeType code was including its vendored header file of zlib regardless.

Because we build our vendored zlib and libpng anyway, there's no point in disabling their build, so I didn't include them in the list of dependencies to disable. Since they're included and found earlier, CMake picks our versions and links them statically.

.gitmodules
```diff
@@ -7,7 +7,7 @@
 	url = https://github.com/glennrp/libpng
 [submodule "extlib/freetype"]
 	path = extlib/freetype
-	url = https://git.savannah.nongnu.org/r/freetype/freetype2.git/
+	url = https://git.savannah.nongnu.org/git/freetype/freetype2.git
```
ruevs, line 10: I did the same yesterday (locally to test).

Why did you change the URL? Both of them apparently point to the same git repository.

nobodyinperson: Oooh please please keep the new one with the `git` in the url, that one supports shallow cloning (`--depth=1`) and the current one does not, causing a lot of pain when building in nix and otherwise.

Good find! 👌

CMakeLists.txt
```diff
@@ -252,10 +252,9 @@ if(ENABLE_GUI OR ENABLE_CLI)
         list(APPEND PNG_PNG_INCLUDE_DIR ${CMAKE_BINARY_DIR}/extlib/libpng)
 
         find_vendored_package(Freetype freetype
-            WITH_ZLIB               OFF
-            WITH_BZip2              OFF
-            WITH_PNG                OFF
-            WITH_HarfBuzz           OFF
+            FT_DISABLE_BZIP2        ON
+            FT_DISABLE_HARFBUZZ     ON
+            FT_DISABLE_BROTLI       ON
```
ruevs, line 257:
```suggestion
            FT_DISABLE_BROTLI       ON
            FT_DISABLE_PNG          ON
            FT_DISABLE_ZLIB         ON
```

We should disable PNG - as it was before.
And I do not see why not do `FT_DISABLE_ZLIB=ON` - it disables the system ZLIB and uses the internal one anyway.

.gitmodules
ruevs: OK this makes sense.

ruevs:
> What's weird though, is that we were already using `WITH_ZLIB=OFF` with the older version, but for some reason FreeType code was including its vendored header file of zlib regardless.

I think it is not weird because regardless of its description:

```
# . Set the `WITH_ZLIB', `WITH_BZip2', `WITH_PNG', and `WITH_HarfBuzz'
#   CMake variables to `ON' or `OFF' to force or skip using a dependency.
#   Leave a variable undefined (which is the default) to use the dependency
#   only if it is available.  Example:
#
#     cmake ... -DWITH_ZLIB=ON -DWITH_HarfBuzz=OFF ...
#
# . Installation of FreeType can be controlled with the CMake variables
#   `SKIP_INSTALL_HEADERS', `SKIP_INSTALL_LIBRARIES', and `SKIP_INSTALL_ALL'
#   (this is compatible with the same CMake variables in zlib's CMake
#   support).

```

it seems that the `WITH_ZLIB` define is not used anywhere in version 069083ccc ("* Version 2.7.1 released.") (or am I not finding it?).

The new option disables using the system ZLIB and uses the internal one instead:
image

iscgar:
> > What's weird though, is that we were already using `WITH_ZLIB=OFF` with the older version, but for some reason FreeType code was including its vendored header file of zlib regardless.
> 
> I think it is not weird because regardless of its description:
> 
> [snip]
> 
> it seems that the `WITH_ZLIB` define is not used anywhere in version 069083ccc ("* Version 2.7.1 released.") (or am I not finding it?).

It is a bit obscure, but it is used in the CMake loop [here](https://github.com/freetype/freetype/blob/069083cccd73d1d68da68116c8d050bb62cdfe0e/CMakeLists.txt#L168) to find the package.

CMakeLists.txt
iscgar: I'm fine with disabling libpng (which disables support of PNG compressed OpenType embedded bitmaps), although we build it anyway as a direct dependency of SolveSpace, so I'm not sure why we would want to prevent FreeType from using it.

As for zlib, "system" is a bit confusing here. When disabling zlib it does not use the zlib that we build as a direct dependency of SolveSpace, but rather the copy of zlib vendored in FreeType, so we just end up with two copies of zlib in the binary, each from a different version. Without disabling it, CMake ends up using the SolveSpace zlib that we introduce into the build before adding FreeType, so we have only a single version of zlib in the binary.

.gitmodules
iscgar: I'm glad it solves an issue, but TBH I just did it because the URL looked a bit weird to me with the slash at the end, and when I looked up the URL on the FreeType project page I couldn't find it, so I assumed it was perhaps an old repository and used the up to date URL instead.

iscgar:
> > What's weird though, is that we were already using `WITH_ZLIB=OFF` with the older version, but for some reason FreeType code was including its vendored header file of zlib regardless.
> 
> I think it is not weird because regardless of its description:
> 
> ```
> # . Set the `WITH_ZLIB', `WITH_BZip2', `WITH_PNG', and `WITH_HarfBuzz'
> #   CMake variables to `ON' or `OFF' to force or skip using a dependency.
> #   Leave a variable undefined (which is the default) to use the dependency
> #   only if it is available.  Example:
> #
> #     cmake ... -DWITH_ZLIB=ON -DWITH_HarfBuzz=OFF ...
> #
> # . Installation of FreeType can be controlled with the CMake variables
> #   `SKIP_INSTALL_HEADERS', `SKIP_INSTALL_LIBRARIES', and `SKIP_INSTALL_ALL'
> #   (this is compatible with the same CMake variables in zlib's CMake
> #   support).
> ```
> 
> it seems that the `WITH_ZLIB` define is not used anywhere in version 069083ccc ("* Version 2.7.1 released.") (or am I not finding it?).
> 
> The new option disables looking for a system ZLIB and uses the internal one instead: image
> 
> ```
> if (NOT FT_DISABLE_ZLIB)
>   if (FT_REQUIRE_ZLIB)
>     find_package(ZLIB REQUIRED)
>   else ()
>     find_package(ZLIB)
>   endif ()
> endif ()
> ```

iscgar:
Accidentally closed this. Sorry for the noise.

CMakeLists.txt
ruevs: > so I'm not sure why we would want to prevent FreeType from using it.

I'm not sure what "PNG compressed OpenType embedded bitmaps." are but I presume that since they are bitmaps they would be useless for the text feature of SolveSpace, which uses only the curves from the fonts.

However enabling the option causes the EXE (Windows 32 bit build with MSVC 2019 with LTO) size to **_decrease_** for me:

FT_DISABLE_PNG          ON:
5.93 MB (6,218,752 bytes)

FT_DISABLE_PNG          OFF:
5.92 MB (6,208,512 bytes)

WTF?!

The `FT_DISABLE_ZLIB` option does **_not_** influence the EXE size. So let us leave them as you configured them.

iscgar: Weird. On Linux I see the reverse (when building with vendored zlib, libpng, and freetype):

```
$ ls -ln --time-style=+"" bin/solvespace
-rwxr-xr-x 1 1000 1000 52871080  bin/solvespace
$ ls -ln --time-style=+"" ../solvespace-png
-rwxr-xr-x 1 1000 1000 52884800  ../solvespace-png
```

I'll try to boot into Windows and check what's happening there when I have some time.

iscgar: Same result, though to a bit lesser effect, with `-DCMAKE_BUILD_TYPE=Release`:

```
$ ls -ln --time-style=+"" bin/solvespace
-rwxr-xr-x 1 1000 1000 5290248  bin/solvespace
$ ls -ln --time-style=+"" ../solvespace-png
-rwxr-xr-x 1 1000 1000 5294640  ../solvespace-png
```

iscgar: @ruevs I'm having trouble reproducing your results on Windows (Visual Studio 2022, with `-DCMAKE_BUILD_TYPE=Release-DENABLE_LTO=1`):

```
>dir /c bin\solvespace.exe ..\solvespace-png.exe
8,433,152 solvespace.exe
               1 File(s)      8,433,152 bytes

8,436,224 solvespace-png.exe
               1 File(s)      8,436,224 bytes
```

Given the fact that as you noted embedded bitmaps support is unneeded, and that on both Windows and Linux I see an increase in file size of 3-4KB, it's probably best if we disable libpng support in FreeType. Should I submit a PR?

ruevs: Done.

PR #1544, opened 2025-03-07
Remove the unused and unset rank parameter from `System::Solve()` and fix it in `System::SolveRank()`

iscgar:
It wasn't set inside `System::Solve()`, and it's not useful anyway when rank testing is fully or partially suppressed.

OTOH, in `System::SolveRank()` it is useful, and although it wasn't set, it was actually checked in `Constraint::TryConstrain()` (through the call to `SolveSpaceUI::TestRankForGroup()`), so fix it by letting `System::TestRank()` set it if provided.

### DIFF
--- a/js/slvs.d.ts
+++ b/js/slvs.d.ts
@@ -27,7 +27,6 @@ export interface Constraint {
 export interface SolveResult {
   result: number;
   dof: number;
-  rank: number;
   bad: number;
 }
 
--- a/src/generate.cpp
+++ b/src/generate.cpp
@@ -536,8 +536,7 @@ void SolveSpaceUI::SolveGroup(hGroup hg, bool andFindFree) {
     Group *g = SK.GetGroup(hg);
     g->solved.remove.Clear();
     g->solved.findToFixTimeout = SS.timeoutRedundantConstr;
-    SolveResult how = sys.Solve(g, NULL,
-                                   &(g->solved.dof),
+    SolveResult how = sys.Solve(g, &(g->solved.dof),
                                    &(g->solved.remove),
                                    /*andFindBad=*/!g->allowRedundant,
                                    /*andFindFree=*/andFindFree,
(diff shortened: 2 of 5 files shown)


iscgar:
ping @phkahler @ruevs

ruevs:
Yes. This makes sense and actually helps to remove a redundant V/H constraint when a sketch is already over constrained by allowing [the second condition in the if statement in `TryConstrain`](https://github.com/solvespace/solvespace/blob/27d1d119064f2a2741efc3eca83bf541dcf5c764/src/constraint.cpp#L121) to work.

It is a different question that this can never be hit by a normal user since drawing a line when the sketch is already "red" immediately causes the solve to fail and "finishes" the line segment with two overlapping endpoints before the user has a chance to drag it close to vertical or horizontal.

This change can be tested by doing the following with a debugger:
- opening this sketch [1544_SolveRank_ReproductionCase.zip](https://github.com/user-attachments/files/19761913/1544_SolveRank_ReproductionCase.zip)
- "allow redundant constraints" (or "suppress DOF calculation")
- putting a break point here https://github.com/solvespace/solvespace/blob/27d1d119064f2a2741efc3eca83bf541dcf5c764/src/constraint.cpp#L114
- Drawing a third line coinciding with the two already vertical ones
- "lying" about `suppressDofCalculation` or `allowRedundant` before the two `TestRankForGroup` calls

The `rankBefore` and `rankAfter` are calculated the same, the second condition becomes `true` and the automatic constraint is deleted.

So I am merging this.

ruevs:
@iscgar I do not know how you stumbled upon this (presumably through the library). But thank you for another good contribution.

iscgar:
> @iscgar I do not know how you stumbled upon this (presumably through the library)

Indeed. I started looking into it because of https://github.com/hlorus/CAD_Sketcher/pull/504#issuecomment-2706391831.

PR #1545, opened 2025-03-17
added text size scale configuration. Its used to render the dimension…

parceval:
added configuration parameter to adjust the font scaling for displaying the dimensions and constraints in the drawing window.

### DIFF
--- a/src/confscreen.cpp
+++ b/src/confscreen.cpp
@@ -451,11 +451,16 @@ bool TextWindow::EditControlDoneForConfiguration(const std::string &s) {
         case Edit::GRID_SPACING: {
             SS.gridSpacing = (float)min(1e4, max(1e-3, SS.StringToMm(s)));
             SS.GW.Invalidate();
-            break;
+            break; 
         }
         case Edit::EXPLODE_DISTANCE: {
             SS.explodeDistance = min(1e4, max(-1e4, SS.StringToMm(s)));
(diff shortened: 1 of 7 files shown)


ruevs:
@parceval Why is this needed?

Are you aware of the "[styles](https://solvespace.com/ref.pl#LineStyles)" function of SolveSpace?
image

parceval:
Yes i know about styles, but this is a global scaling parameter applied to all styles on top of it.

ruevs:
> Yes i know about styles, but this is a global scaling parameter applied to all styles on top of it.

But why is it needed/useful?

parceval:
For me it is useful because i use multiple screens with different resolutions, one is 4k and the other is hd, and its easier to adjust one value, than changing the size in all styles.

phkahler:
@parceval so this is a global size applied *in addition to* the style specific sizes, is that correct?

parceval:
> @parceval so this is a global size applied _in addition to_ the style specific sizes, is that correct?

Exactly.

phkahler:
@parceval  why does this use MmtoString and StringToMm?  This scale factor should not depend on what units are selected. Or am I confusing those functions with those that convert units to/from mm?

That'd be weird to do float<-->string operations in a function and claim mm is relevant in the function name if it's not.

BTW, sorry about the delay looking at this - it seems reasonable.

PR #1557, opened 2025-04-02
Fix mimalloc CMake compatibility issue in CI workflows

GlassOnTin:
This PR addresses the CI build failures in the feature branches by adding a patch for the mimalloc library.

## Issue
The CI builds for all feature branches were failing with the following error:
```
CMake Error at extlib/mimalloc/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

## Solution
1. Created a patch file that updates the minimum CMake version in mimalloc's CMakeLists.txt from 3.0 to 3.5
2. Added an apply-patches.sh script to apply this patch during the build process
3. Modified the build scripts for Ubuntu, Windows, and macOS to run the patch script before building

This ensures the CI can successfully build the project without modifying the actual submodule content in the repository.

### DIFF
--- a/.github/patches/mimalloc-cmake-version.patch
+++ b/.github/patches/mimalloc-cmake-version.patch
@@ -1,0 +1,20 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 8127e096..7bb6108a 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -1,4 +1,4 @@
+-cmake_minimum_required(VERSION 3.0)
++cmake_minimum_required(VERSION 3.5)
+ project(libmimalloc C CXX)
+ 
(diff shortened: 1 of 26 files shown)


phkahler:
@GlassOnTin  Please stop. This PR says it fixes a mimalloc issue but contains 12 commits.  You've also recently tried to push that "implementation plans for major features" file, which is not in line with the goals of the project. It'll take time to review all these changes and I'm sure some will not be merged.

PR #1558, opened 2025-04-02
Fix: Update mimalloc CMake minimum version to 3.5

GlassOnTin:
This PR fixes the CI build failures in all feature branches by patching the vendored mimalloc library's CMake version requirement.

## Issue
The CI builds were failing with the following error:
```
CMake Error at extlib/mimalloc/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.
```

## Solution
- Added a patch file that updates mimalloc's CMakeLists.txt version from 3.0 to 3.5
- Created a simple apply-patches.sh script to apply patches during builds
- Updated the build scripts to apply this patch before configuring

This is a minimal, focused change that ensures CI can build without changing the submodule itself.

### DIFF
--- a/.github/patches/mimalloc-cmake-version.patch
+++ b/.github/patches/mimalloc-cmake-version.patch
@@ -1,0 +1,20 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 8127e096..7bb6108a 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -1,4 +1,4 @@
+-cmake_minimum_required(VERSION 3.0)
++cmake_minimum_required(VERSION 3.5)
+ project(libmimalloc C CXX)
+ 
+ set(CMAKE_C_STANDARD 11)
+diff --git a/test/CMakeLists.txt b/test/CMakeLists.txt
+index bb8dc97e..f756c6b2 100644
+--- a/test/CMakeLists.txt
++++ b/test/CMakeLists.txt
+@@ -1,4 +1,4 @@
+-cmake_minimum_required(VERSION 3.0)
++cmake_minimum_required(VERSION 3.5)
+ project(mimalloc-test C CXX)
+ 
+ set(CMAKE_C_STANDARD 11)
--- a/.github/scripts/apply-patches.sh
+++ b/.github/scripts/apply-patches.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+set -e
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+REPO_ROOT="$(dirname "$(dirname "$SCRIPT_DIR")")"
+PATCHES_DIR="$SCRIPT_DIR/../patches"
+
+# Apply mimalloc CMake version patch
+echo "Applying mimalloc CMake version patch..."
+patch -p1 -d "$REPO_ROOT/extlib/mimalloc" < "$PATCHES_DIR/mimalloc-cmake-version.patch"
--- a/.github/scripts/build-macos.sh
+++ b/.github/scripts/build-macos.sh
@@ -1,5 +1,8 @@
 #!/bin/sh -xe
 
+# Apply patches
+.github/scripts/apply-patches.sh
+
 ENABLE_SANITIZERS="OFF"
 if [ "$1" = "release" ]; then
     BUILD_TYPE="RelWithDebInfo"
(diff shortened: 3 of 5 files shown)


iscgar:
CI builds are not failing on master. You likely updated CMake on your system to version 4.0.0, which dropped support for versions older than 3.5. However, for CMake 4.0.0 this change isn't enough, since there are similar issues with the vendored versions of ANGLE, Cairo, libpng, pixman, and zlib.

GlassOnTin:
My local Ubuntu install reports cmake 3.3.  This issue only showed up for the GitHub build actions.

iscgar:
Yes, I'm seeing this now in one of my branches (likely some of the GitHub runners are simply running older images, and I just lucked out and had my jobs running on them).

This currently impacts only the `windows-2019` image, which updated to CMake 4.0.0 yesterday (https://github.com/actions/runner-images/commit/5826942237eb4ad1f40963b2014f5bf294360880), but is likely to impact all of the images in CI in the future because none are pinned to a specific version (except for the Python workflow which uses `ubuntu-22.04` for the Linux build, we use `ubuntu-latest` and `macos-latest` everywhere).

My point still stands though: since this currently only breaks on Windows, which uses only vendored libraries, changing mimalloc isn't enough and all of the other vendored libraries that I mentioned need to be fixed as well.

iscgar:
Here's my ugly hack to fix it by pinning to a specific version on Windows, which should fix this and hopefully prevent this in the future: https://github.com/iscgar/solvespace/commit/45da9cfd8c3c9d91fa7f2c978b9365f9c4d84696

This hack only deals with Windows, though, which is why I'm not submitting it as a PR. We likely want to pin our Ubuntu images to specific versions and maybe use Homebrew to forcefully install the pinned version in a similar manner on macOS. @ruevs @vespakoen, what are your thoughts on this?

ruevs:
My thoughts - update the (CMake configurations of the) vendored libraries individually [libdxfrw](https://github.com/solvespace/libdxfrw), [angle](https://github.com/solvespace/angle), [pixman](https://github.com/solvespace/pixman), [cairo](https://github.com/solvespace/cairo).

Update to newer revisions of the submodules that hopefully have updated CMake rules and do not break anything else.

For [libdxfrw](https://github.com/solvespace/libdxfrw) I have a local fork where I updated it to the latest upstream and had just a few of our patches left to merge. But I'll not have time to work on it or any of the above for the next few days.

This PR - as it is - with the bash script patching mimalloc is a no-go.

iscgar:
CMake deprecations are a moving target, and chasing them in third party dependencies is a waste of time IMO (CMake 3.27 was the first to start warning about versions < 3.5 on the 18th of July 2023, and less than two years later it's now a hard error. Version 3.31, which was released on the 6th of November 2024, started warning about versions < 3.10, and it's likely that that compatibility will be removed as well in the not so distant future, but even a recently updated dependency like FreeType only specifies 3.5), and I also remeber seeing maintainer comments in some SolveSpace issues expressing hesitation regarding updating core dependencies like mimalloc, for example, without extensive testing. That is why I think pinning the CI build is the best approach at least as a stopgap measure, so that PRs could continue to be tested and potential releases cut, while updates to individual depenencies made and tested separately.

However, as a quick and dirty test, I updated mimalloc and zlib in [this branch](https://github.com/iscgar/solvespace/tree/iscgar/update-extlib-cmake) on my fork, and the resulting binary doesn't seem to have any obvious regressions with the few models that I tested. Since you're already working on libdxrfrw, I decided not to touch any of the SolveSpace forks of the other dependencies for now in order to avoid duplicated work. If you want me to, I'm willing to tackle them once your work on libdxrfrw is complete.

iscgar:
The build forces the use of the vendored mimalloc, so if there are issues with newer versions we have no way to know (although as I said, I didn't notice any obvious issues with the updated version in my branch). The fact that it's forced to the vendored version on all platforms is precisely why CMake 4.0.0 bails with an error about mimalloc first.

I'm 100% in favour of updating the vendored dependencies and against applying patches. I was only arguing aginst doing so for almost all the dependencies at once just to solve a CMake deprecation warning that became an error. As I noted, I already updated in my branch the vendored dependencies that weren't forked, but that alone doesn't solve the issue, because the forked dependencies also require updates to their CMake config. That's why I propose to pin the version of the CMake binary in CI as a stopgap measure until the dependencies are updated, because right now some of the Windows builds in CI are failing due to the updated `windows-2019` image, and the chances of hitting this will only grow as more GitHub runners are updated to the latest image.

PR #1561, opened 2025-04-03
slvs: various minor cleanups and improvements

iscgar:
This PR makes the following enhancements to `slvs` library code and its build:

### Extract the `slvs`-specific code and build config into its own directory
This is done in order to allow managing the various `slvs` build types without cluttering the main CMake config, and in order to separate the `slvs`-specific code from the SolveSpace program code.

This change also includes minor cleanups to the way the Python and WASM modules are built.

### Remove the unused and unset `rank` member from the `Slvs_SolveResult` type
As it wasn't set, it always simply contained 0, and it's not really useful anyway on its own, especially given the existence of the `dof` member.

### Remove the `dragged` member from the `Slvs_System` type
It was limited from the beginning to only 4 params, which is mostly fine for a single dragged point, but doesn't cover more complex use cases. Since the `WHERE_DRAGGED` constraint exists (and is used in the sketch-related API which was introduced by @vespakoen), there's no reason not to use it instead in the System-related API in order to populate the dragged params list.

### Enable a commented out example
While fixing the dragged example in `CDemo` as part of the previous change, I noticed that for some reason the call to the 3D example (of which the dragged example is part) was commented out at some point, even though it compiles and is fully functional. This change simply enables it.

### DIFF
--- a/exposed/CDemo.c
+++ b/exposed/CDemo.c
@@ -290,7 +290,8 @@ int main()
     sys.failed  = CheckMalloc(50*sizeof(sys.failed[0]));
     sys.faileds = 50;
 
-    /*Example3d();*/
+    Example3d();
+    sys.params = sys.constraints = sys.entities = sys.ndragged = 0;
     for(;;) {
         Example2d();
         sys.params = sys.constraints = sys.entities = sys.ndragged = 0;


vespakoen:
I am fine with this change and happy to get some help on this topic =)

iscgar:
Sorry for the noise with the force pushes. This now builds correctly on all platforms, and is ready for review.

iscgar:
The `WHERE_DRAGGED` locks a param in place and prevents it from moving, whereas marking it as dragged merely asks the solver to try to move it as little as possible., so I've reverted that changed and replaced it with a change that allows dragging more than 4 parameters.

phkahler:
@iscgar  I recently noticed how the where dragged constraint works. Remember it is a constraint you can apply to a point like any other constraint.  When generating equations this constraint creates an expression (param - value) where value is the current numeric value of the parameter. Actually I think it's the numeric value of the point, not the parameter itself. That's a crude way of holding a parameter constant while solving.

For the named parameters stuff (in progress) I still don't understand how parameters from previous groups are held constant. It may be as simple as substituting a constant into an expression if the parameter is not from the current group, but I haven't actually spotted the specific code.  I seem to recall such a substitution happening in one of the "copy" operations in setting up the system. I don't *actually* need to know right now since I'm creating named parameters as part of a request so that's all handled automatically.

iscgar:
> [snip] When generating equations this constraint creates an expression (param - value) where value is the current numeric value of the parameter. Actually I think it's the numeric value of the point, not the parameter itself.

It actually generates two or three equations for a point (depending on if the point is in a workplane or in free space, respectively) in the form `expr - value`, where `expr` is a coordinate expression (for a point in 3D it's simply a parameter, but that's not the case for every type of point), and `value` is indeed the evaluated numeric value of the coordinate expression (see [here](https://github.com/solvespace/solvespace/blob/cc871ab0aea7928defeb7e33158211696312bec1/src/constrainteq.cpp#L1036-L1050)).

> For the named parameters stuff (in progress) I still don't understand how parameters from previous groups are held constant. It may be as simple as substituting a constant into an expression if the parameter is not from the current group, but I haven't actually spotted the specific code. I seem to recall such a substitution happening in one of the "copy" operations in setting up the system.

Indeed. See the code in `Expr::DeepCopyWithParamsAsPointers()`: https://github.com/solvespace/solvespace/blob/cc871ab0aea7928defeb7e33158211696312bec1/src/expr.cpp#L319-L325

When solving a group, all of the system's parameters are initially set as `known = false` here: https://github.com/solvespace/solvespace/blob/cc871ab0aea7928defeb7e33158211696312bec1/src/generate.cpp#L525-L527

Then, after the system is successfully solved, each of the parameters is set as `known = true` here: https://github.com/solvespace/solvespace/blob/cc871ab0aea7928defeb7e33158211696312bec1/src/system.cpp#L495-L506

So when `Expr::DeepCopyWithParamsAsPointers()` is called from `System::WriteJacobian()`, the aforementioned code simply substitutes known parameters (which are all of the parameters except those in the group that's currently being solved) with constants.

phkahler:
@iscgar  Thank you for the contribution. It is good to have another person looking this deep into solvespace!

Issue #1247, opened 2022-05-20
SolveSpace fails to solve solvable constraints


tc7000:
### System information

- **SolveSpace version:** 3.0~ab00823a
- **Operating system:** Ubuntu 20.04.4 LTS 64-bit

### Expected behavior

There should be no error solving constraints.

### Actual behavior

unsolvable constraints error (distance and angle)

### Additional information

Extract attached zip file, open test.slvs, then modify the angle constraint to 60°. Change again the angle to 30°.
[test.slvs.zip](https://github.com/solvespace/solvespace/files/8741298/test.slvs.zip)

phkahler:
@tc7000  Thank you for such a simple test case.

That's really annoying. It can handle a jump from 60 to 40, but not 60 to 30. This seems similar to another solver failure issue but also much simpler and more likely for people to run into.

ruevs:
Nice test case! Changing the angle from 60 to anything between 27-32 causes the solve to fail.
59.1 -> 28. fails as well, 59.0 -> 28.0 does not.  Interesting.

phkahler:
This is among the things #1149 might address. Need a mathematician for this stuff  ;-)

ruevs:
There is some nice debug code from @Evil-Spirit here https://github.com/Evil-Spirit/solvespace-master/commits/solver-debug I have not pulled it into master in years but it is a rather interesting tool. No time now but I'll do it... some day.

ruevs:
I rebased
https://github.com/Evil-Spirit/solvespace-master/commits/solver-debug
onto master here:
https://github.com/ruevs/solvespace/tree/EvilSpirit_solver-debug
it is interesting to watch the solve fail, and see the equations but no immediate help. I need to think/debug much more...

ruevs:
Issue https://github.com/solvespace/solvespace/issues/21 seems to me somewhat related. Even though the details differ.

zwn:
I have also a somewhat simple reproduction case:
- create a block (extruded rectangle) and save file
- create a new file to be used as assembly
- imported the block into the assembly
- constrain the orientation of the imported block (_constrain normals in same orientation_) on the Z axes - works as expected
- undo the constrain
- import the same block for a second time
- try to add the same constrain that worked before on the firs block - unsolvable
- try to add the same constrain on the newly imported block - works as expected

[unsolvable.zip](https://github.com/solvespace/solvespace/files/14201184/unsolvable.zip)

PR #1563, opened 2025-04-08
Solver: clean up and optimise `SolveBySubstitution()` and `WriteJacobian()`

iscgar:
This PR cleans up and optimises various parts of the solver, especially around `SolveBySubstitution()` and `WriteJacobian()`. The latter can result in noticeable speedups on some models (see initial numbers in my comment [here](https://github.com/solvespace/solvespace/pull/1563#issuecomment-2807422537)).

-----
 #### Original PR description (superseded)

~While trying to look into #1247 I was having trouble reproducing the issue, and the model was failing in weird ways (such as multiple constraints not satisfied, causing horizontal and vertical lines to become diagonal).~

~After staring hard at the code for a long time, I finally traced the issue down to commit 708a08f04b910337965859249b87684696369219 (merged in #1159) which made changes in order to speed up solving by substitution. However, the changes inadvertently  caused substitution to persist across solve runs and across groups, which broke multiple cases.~

~There are multiple options for fixing it, and I chose to do it by removing the `substd` member from `Param` (in order to prevent
such breakage in the future), and replacing the logic with similar runtime complexity logic that doesn't require modifying `Param` itself.~

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -400,16 +400,11 @@ Expr *Expr::PartialWrt(hParam p) const {
     ssassert(false, "Unexpected operation");
 }
 
-void Expr::ParamsUsedList(std::vector<hParam> *list) const {
+void Expr::ParamsUsedList(ParamSet *list) const {
     if(op == Op::PARAM || op == Op::PARAM_PTR) {
         // leaf: just add ourselves if we aren't already there
         hParam param = (op == Op::PARAM) ? parh : parp->h;
-        if(list->end() != std::find_if(list->begin(), list->end(),
-                                       [=](const hParam &p) { return p.v == param.v; })) {
-            // We found ourselves in the list already, early out.
-            return;
-        }
-        list->push_back(param);
+        list->insert(param);
         return;
     }
 
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -6,6 +6,7 @@
 //
 // Copyright 2008-2013 Jonathan Westhues.
 //-----------------------------------------------------------------------------
+#include <limits>
 #include "solvespace.h"
 
 ExprVector ExprVector::From(Expr *x, Expr *y, Expr *z) {
@@ -542,7 +543,7 @@ void Expr::Substitute(const SubstitutionMap &subMap) {
 // If multiple params are referenced, then return MULTIPLE_PARAMS.
 //-----------------------------------------------------------------------------
 const hParam Expr::NO_PARAMS       = { 0 };
-const hParam Expr::MULTIPLE_PARAMS = { 1 };
+const hParam Expr::MULTIPLE_PARAMS = { std::numeric_limits<decltype(hParam::v)>::max() };
 hParam Expr::ReferencedParams(ParamList *pl) const {
     if(op == Op::PARAM) {
         if(pl->FindByIdNoOops(parh)) {
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -307,8 +307,7 @@ Expr *Expr::DeepCopy() const {
 }
 
 Expr *Expr::DeepCopyWithParamsAsPointers(IdList<Param,hParam> *firstTry,
-    IdList<Param,hParam> *thenTry) const
-{
+                                         IdList<Param, hParam> *thenTry, bool foldConstants) const {
     Expr *n = AllocExpr();
     if(op == Op::PARAM) {
         // A param that is referenced by its hParam gets rewritten to go
@@ -328,8 +327,17 @@ Expr *Expr::DeepCopyWithParamsAsPointers(IdList<Param,hParam> *firstTry,
 
     *n = *this;
     int c = n->Children();
-    if(c > 0) n->a = a->DeepCopyWithParamsAsPointers(firstTry, thenTry);
-    if(c > 1) n->b = b->DeepCopyWithParamsAsPointers(firstTry, thenTry);
+    if(c > 0) {
+        n->a = a->DeepCopyWithParamsAsPointers(firstTry, thenTry, foldConstants);
+        bool hasConstants = n->a->op == Op::CONSTANT;
+        if(c > 1) {
+            n->b = b->DeepCopyWithParamsAsPointers(firstTry, thenTry, foldConstants);
+            hasConstants |= n->b->op == Op::CONSTANT;
+        }
+        if(hasConstants && foldConstants) {
+            n = n->FoldConstants(false, 0);
+        }
+    }
     return n;
 }
 
@@ -433,13 +441,15 @@ bool Expr::IsZeroConst() const {
     return op == Op::CONSTANT && EXACT(v == 0.0);
 }
 
-Expr *Expr::FoldConstants() {
-    Expr *n = AllocExpr();
-    *n = *this;
-
-    int c = Children();
-    if(c >= 1) n->a = a->FoldConstants();
-    if(c >= 2) n->b = b->FoldConstants();
+Expr *Expr::FoldConstants(bool allocCopy, size_t depth) {
+    Expr *n;
+    
+    if(!allocCopy) {
+        n = this;
+    } else {
+        n = AllocExpr();
+        *n = *this;
+    }
 
     switch(op) {
         case Op::PARAM_PTR:
@@ -452,6 +462,11 @@ Expr *Expr::FoldConstants() {
         case Op::TIMES:
         case Op::DIV:
         case Op::PLUS:
+            if(depth > 0) {
+                n->a = a->FoldConstants(allocCopy, depth - 1);
+                n->b = b->FoldConstants(allocCopy, depth - 1);
+            }
+    
             // If both ops are known, then we can evaluate immediately
             if(n->a->op == Op::CONSTANT && n->b->op == Op::CONSTANT) {
                 double nv = n->Eval();
@@ -490,6 +505,10 @@ Expr *Expr::FoldConstants() {
         case Op::COS:
         case Op::ASIN:
         case Op::ACOS:
+            if(depth > 0) {
+                n->a = a->FoldConstants(allocCopy, depth - 1);
+            }
+
             if(n->a->op == Op::CONSTANT) {
                 double nv = n->Eval();
                 n->op = Op::CONSTANT;
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -500,15 +500,21 @@ Expr *Expr::FoldConstants() {
     return n;
 }
 
-void Expr::Substitute(hParam oldh, hParam newh) {
+void Expr::Substitute(const SubstitutionMap &subMap) {
     ssassert(op != Op::PARAM_PTR, "Expected an expression that refer to params via handles");
 
-    if(op == Op::PARAM && parh == oldh) {
-        parh = newh;
+    if(op == Op::PARAM) {
+        auto it = subMap.find(parh);
+        if(it != subMap.end()) {
+            parh = it->second->h;
+        }
+    } else {
+        int c = Children();
+        if(c >= 1) {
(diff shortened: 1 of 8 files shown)


iscgar:
CI is failing due to spurious 502 errors when cloning the FreeType repository. The job should be retired, but I have no permission to do so. cc @ruevs.

iscgar:
No, it does not fix #1247. I only encountered the problem this PR is intended to fix while trying to debug that other issue.

However, please hold off integrating this PR for now, because I'm currently unable to reproduce the issue. Also, in hindsight my description of it doesn't make much sense, becuase `System::SolveBySubstitution()` only works on copies of the parameters that are held by the system, so they should not persist across solve runs, and most definitely not across groups.

While I still think this PR brings some measurable improvement (such as reducing memory usage by getting rid of the `substd` member of `Param`, which is only ever used by substitution, and better algorithmic complexity for `Expr::UsedParams()` through the usage of `std::unordered_set`, as well as better algorithmic complexity by avoidance of traversing the substitution linked list), I want to try and reproduce the problem that I encountered in order to understand what caused it and ensure that it is actually fixed.

iscgar:
I'm still unable to reproduce the issues I was having, so I'll have to chalk it up to a broken build on my machine.

This PR's description is now incorrect, as it turned into a bunch of small cleanups and optimisations, with a focus on `SolveBySubstitution` and `WriteJacobian`. I'll update the description to match the new state of things soon, but I'm marking it as ready to review in the mean time, so please take a look and let me know what you think.

Depending on the model used, different parts of the code are exercised, so the overall speedup will vary as well. The following list tallies the improvement in runtime (in release mode with LTO) for two operations: generating groups (with DoF calculation suppressed; depending on model this can exercise either of `SolveBySubstitution()` and `WriteJacobian()`, or both), and marking free points (which exercises `WriteJacobian()`):

* 8040_finished_constrained
  * Generate
    * Before: ALL(BB)=57ms, ALL=85ms, DIRTY=33ms
    * After: ALL(BB)=52ms, ALL=79ms, DIRTY=31ms
    * Change: insignificant
  * Mark free
    * Before: 927ms
    * After: 710ms
    * Change: -24%
* case-vents
  * Generate
    * Before: ALL(BB)=36ms, ALL=1247ms, DIRTY=1132ms
    * After: ALL(BB)=33ms, ALL=1243ms, DIRTY=1118ms
    * Change: insignificant
  * Mark free
    * Before: 1323ms
    * After: 1290ms
    * Change: insignificant
* Chapiteau vue de dessus
  * Generate
    * Before: ALL(BB)=120ms, ALL=129ms, DIRTY=not measured
    * After: ALL(BB)=118ms, ALL=126ms, DIRTY=not measured
    * Change: insignificant
  * Mark free
    * Before: 4985ms
    * After: 3970ms
    * Change: -20%
* constrainfreeze
  * Generate
    * Before: ALL(BB)=60ms, ALL=61ms, DIRTY=not measured
    * After: ALL(BB)=55ms, ALL=55ms, DIRTY=not measured
    * Change: insignificant
  * Mark free
    * Before: 245ms
    * After: 125ms
    * Change: -49%
* LeliaManya2
  * Generate
    * Before: ALL=not measured, DIRTY(BB)=493ms, DIRTY=836ms
    * After: ALL=not measured, DIRTY(BB)=491ms, DIRTY=830ms
    * Change: insignificant
  * Mark free
    * Before: 719ms
    * After: 706ms
    * Change: insignificant

All of my tests were done with DoF calculation suppression, and yet `SolveBySubstitution()` has essentially no impact, and the changes which were originally motivated by what I perceived as a bug are now merely a cleanup which may only benefit models with many substituable equations (I was having trouble coming up with such a model, so I'm not sure this is actually useful).

phkahler:
@iscgar  These look pretty good. I'm not sure I like 
[this one](https://github.com/solvespace/solvespace/pull/1563/commits/a6a4da833572af071f7ad4df660a087b55ce2263) because it adds extra complexity and the solver is rarely a performance bottleneck. How has it been for you to learn the solver code and work with it? Would that change make such work harder?


Unrelated, there has been discussion in the past about solving more than one group at a time. Maybe by adding a checkbox to a group "solve with previous group" which could be chained to solve several together. This is not an existing feature, but I'm wondering how possible you think this might be, and if maybe the unused "tag" was related?

iscgar:
> @iscgar These look pretty good. I'm not sure I like [this one](https://github.com/solvespace/solvespace/pull/1563/commits/a6a4da833572af071f7ad4df660a087b55ce2263) because it adds extra complexity and the solver is rarely a performance bottleneck. How has it been for you to learn the solver code and work with it? Would that change make such work harder?

Yeah, I can't say I like it either, but without it `SolveBySubstitution()` has `O(n^2)` complexity in case there are many independent substitutions, and I just couldn't let it stand considering the original Evil-Spirit commit message contained the words "linear complexity" :)

I find the solver code quite easy to work with, since it's mostly a direct implementation of the textbook Gauss-Newton algorithm. I initially started looking into `SolveBySubstitution()` because it was the only exception, and it took me a moment to figure out what it was doing with dragged parameters and cycle breaking. That specific change is still easier to understand IMO than the original implementation, as everything directly points to the target instead of traversing a linked list, but it is definitely more complex than my initial change in the commit preceding it (which is why I added a bit too many comments to explain why the code works).

As I was unable to come up with a model where `SolveBySubstitution()` would be the bottleneck in the solver, I'm fine with dropping that change if you feel that it isn't worth the added complexity.

> Unrelated, there has been discussion in the past about solving more than one group at a time. Maybe by adding a checkbox to a group "solve with previous group" which could be chained to solve several together. This is not an existing feature, but I'm wondering how possible you think this might be, and if maybe the unused "tag" was related?

Solving groups together should be easy enough to implement in the solver, as it's just a matter of using more than a single value in filtering constraints and entities to use for writing the equation system. However, from a UI standpoint it would probably be harder to reason about, with rank test suppression and e.g. bad constraints from multiple groups. I'm curious though: what's the advantage of solving multiple groups together? Is there a specific use case where it can potentially result in a noticeable speedup?

The `tag` is only an internal solver parameter for separating subsystems, so it's unlikely to have any connection with solving multiple groups, and is just a leftover from the pre-Eigen solver (originally the Jacobian was written inside `NewtonSolve()`, and this was changed in bab13b821f5fc874463eeb8d0b7c438100373aba).

iscgar:
> One big advantage would be simulating assemblies (mechanisms). Each part in an assembly is a separate group which means you can drag the first part to drive a machine, but you can't drag a later part to move a machine. Well you can, but only the later parts can move. A machine with 1 DoF must be dragged by the first part in the assembly, but it'd be better to grab it wherever is convenient.
> 
> Another use would be designing something where an extrusion length may not be known until later parts of the geometry are constructed. Or when designing something out of 8020 extrusions - it'd be great if an entire assembly could be resized later. As it is today you need to make a sketch-in-3D as a first group with key dimensions and then build everything on top of that. It's actually a pretty good way to do it, but a bit abstract I think for new users.

I see. The problem with solving multiple groups together in these cases is that it can lead to an explosion of unknowns, slowing down the solver, and making convergence less likely given that the solver is already struggling today with some simple cases like #1247 or when points are dragged too quickly. Also, we artificially limit the number of unknowns to 2048 in order to prevent a slowdown, and when multiple groups are solved together it would be less obvious how to fix the issue if a "too many unknowns" error is encountered.

So while letting the solver handle multiple groups would be easy, I'm not sure it would be useful, and I think we need to at least improve convergence when solving a single group first. Then we need to think about the UI for solving multiple groups (especially in the error cases) before enabling it.

phkahler:
@iscgar  A linked sketch creates a group with 7 unknowns regardless of the number of entities in the linked sketch.  The upper limit of 2048 is arbitrary and can be dropped but things will definitely slow down with huge numbers of constraints. I agree that fixing those solve failures is higher priority.

src/expr.cpp
```diff
@@ -400,16 +400,11 @@ Expr *Expr::PartialWrt(hParam p) const {
     ssassert(false, "Unexpected operation");
 }
 
-void Expr::ParamsUsedList(std::vector<hParam> *list) const {
```
Evil-Spirit, line 403: Have you really tested it? There is no "large" param sets, linear search is the best suitable for this kind of data.

src/expr.cpp
```diff
@@ -400,16 +400,11 @@ Expr *Expr::PartialWrt(hParam p) const {
...
+void Expr::ParamsUsedList(ParamSet *list) const {
```
Evil-Spirit, line 412: Have you really tested it? There is no "large" param sets actually exist, linear search is the best suitable for this kind of data.

src/solvespace.h
```diff
@@ -229,7 +230,7 @@ class System {
 
     // A list of parameters that are being dragged; these are the ones that
     // we should put as close as possible to their initial positions.
-    List<hParam>                    dragged;
+    ParamSet                        dragged;
```
Evil-Spirit, line 233: Please, carefully test it. For small dragging lists std::vector is preferrable, but depending on size of the list, custom container can be created (using linear search for small amount of data and switch to hastable when data is huge). The point of using such things - you can drag single point of huge object and it is more often way to drag, but also you can drag a big number of objects, so it is worth to optimize both cases.

src/expr.cpp
```diff
@@ -452,6 +462,11 @@ Expr *Expr::FoldConstants() {
         case Op::TIMES:
         case Op::DIV:
         case Op::PLUS:
+            if(depth > 0) {
+                n->a = a->FoldConstants(allocCopy, depth - 1);
+                n->b = b->FoldConstants(allocCopy, depth - 1);
+            }
+    
             // If both ops are known, then we can evaluate immediately
             if(n->a->op == Op::CONSTANT && n->b->op == Op::CONSTANT) {
```
Evil-Spirit, line 472: The best way to do folding - is using optimization druring writing equations (test on constants inside Plus/Minus/Times/Etc). So no need to call FoldConstants or OptimizeExpression(or how it named?). Please, see NoteCAD source code (Exp) as reference.

src/expr.cpp
iscgar: Admittedly no expression currently contains more than a handful of parameters, and even less duplicated ones in the largest of expressions, but in my testing it made no difference in performance one way or the other even with models that generate a huge amount of equations, and the reason I went with an unordered set was simply because it made the code simpler since it doesn't require an explicit check for duplicates. I'm fine with dropping this change if the maintainers feel uncomfortable with it, but as noted above this was done as a cleanup rather than as an optimisation.

src/solvespace.h
iscgar: This is again more of a cleanup than an optimisation. I did test it, and while there can be a potential gain with an extremely large amount of dragged parameters, in practice it doesn't make a meaningful difference in the grand scheme of things the way SolveSpace uses it, especially given the fact it's only used when solving by substitution and scaling the parameters, while when solving the bulk of the time is spent inside Eigen when solving and rank testing anyway.

src/expr.cpp
iscgar: This is precisely what I'm doing inside DeepCopyWithParamsAsPointers(), indirectly using FoldConstants() with a depth of 1, since the expression is copied from the leaves to the root (thus not missing either existing constants, nor constants introduced during the copy by parameters being resolved as known).

I did take a look at your code in NoteCAD (as well as the Adjacent port to C++) and it is an impressive work, but as you undoubtably know it's not directly applicable to the way expressions are allocated and constructed in SolveSpace. I have a branch where I created an expression system that does optimise during construction (with more simplification rules than just constant evaluation), but it's not complete yet, and in my testing there didn't seem to be many more opportunities for optimisations of the existing expressions in a way that would meaningfully affect performance further than what is currently achievable with my approach in this PR.

PR #1564, opened 2025-04-10
emscripten: fix build issues and make the interface functional

iscgar:
The Emscripten port is currently broken with the latest Emscripten tools due to the way the toolchain is defined, so this PR gets rid of the bespoke toolchain files in favour of the official Emscripten one (it's not needed if an emsdk environment is active).

In addition to the build issues, some of the hacks that were made in the port broke with the latest toolchain, causing runtime issues and crashing. This PR fixes just enough issues to make the port usable again so that it's at least able to run and have a functional interface. Additional work is still needed to bring it to feature parity with the native ports.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,11 @@ project(solvespace
     LANGUAGES C CXX ASM)
 
 if(EMSCRIPTEN)
+    # Enable optimization. Workaround for "too many locals" error when runs in the browser.
+    if(NOT CMAKE_BUILD_TYPE OR CMAKE_BUILD_TYPE STREQUAL "Debug")
+        add_compile_options(-O1)
+    endif()
+
     set(INCLUDE_NATIVE_COMPONENTS OFF)
 else()
     set(INCLUDE_NATIVE_COMPONENTS ON)
@@ -192,10 +197,6 @@ endif()
 
 if(APPLE)
     set(CMAKE_FIND_FRAMEWORK FIRST)
-endif()
-
-if(EMSCRIPTEN)
-    set(M_LIBRARY "" CACHE STRING "libm (not necessary)" FORCE)
 endif()
 
 if(ENABLE_GUI OR ENABLE_CLI)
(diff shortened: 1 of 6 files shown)


ruevs:
@iscgar This pull request causes a problem for me with emsdk [ce1d281ce](https://github.com/emscripten-core/emsdk/commit/ce1d281ce65b5c7398b17ed6484efc6a3901df6f).

Before it I could build and run the web version with the following commands (on Windows with MSVC2019):

```
emsdk activate latest

"C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars32.bat"

emcmake cmake .. -DCMAKE_BUILD_TYPE=Release -DENALBE_GUI="ON" -DENABLE_LTO="ON" -DENABLE_TESTS="OFF" -DENABLE_CLI="OFF" -DENABLE_COVERAGE="OFF"

ninja

emrun bin\solvespace.html
```

After this commit I get the following error after the `emcmake` command:

```
ninja: error: build.ninja:876: multiple rules generate bin/libz.a [-w dupbuild=err]
```

I will now pull the latest emsdk and see if it makes a difference.

ruevs:
Reverting the changes to the CMake files and keeping only the change in src/platform/guihtml.cpp it configures, builds and runs fine with emsdk 4.0.6 while the previous version does crash with:

```
Aborted(Assertion failed: allocate no longer takes a type argument)
   [solvespace.js:1:25001](http://localhost:6931/solvespace.js)
Uncaught RuntimeError: Aborted(Assertion failed: allocate no longer takes a type argument)
   [solvespace.js:1:25175](http://localhost:6931/solvespace.js)
```

So the change in guihtml.cpp is good, but the CMake changes do not work - at least for me. My CMake is `cmake version 3.29.5`

ruevs:
@iscgar this is great! Your knowledge of CMake is sliiiiiightly better than mine :-)
It now builds with a small tweak. I'll comment in your [pull request](https://github.com/solvespace/solvespace/pull/1566).

PR #1565, opened 2025-04-11
ci: fix the broken Python and WASM library builds

iscgar:
Similar to #1558, the CI Python build has been failing because the CMake dependency in `pyproject.toml` was not pinned to a specific version and it was pulling version 4.0.0 with the compatibility issue.

Additionally, I inadvertently broke the CI build in #1561 by introducing more granular targets without adjusting the installation of those targets accordingly. This PR fixes it by introducing CMake components and using them in the Python build, and properly specifying the WASM target in the build script.

### DIFF
--- a/.github/scripts/build-wasmlib.sh
+++ b/.github/scripts/build-wasmlib.sh
@@ -10,7 +10,7 @@ source ../emsdk/emsdk_env.sh
 mkdir -p build-wasmlib
 cd build-wasmlib
 emcmake cmake .. \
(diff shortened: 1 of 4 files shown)


ruevs:
Let's see...

vespakoen (commented):
LGTM

ruevs:
It worked:
https://github.com/solvespace/solvespace/actions/runs/14473581871
https://github.com/solvespace/solvespace/actions/runs/14473581893

PR #1568, opened 2025-04-16
slvs: support getting the list of bad constraints in the sketch solving API

### LINKED ISSUE #1546: Missing Indices of Bad Constraints in Python Binding
Description:
When solving a sketch using the Python bindings for the SolveSpace solver, the result includes a bad field indicating the number of bad constraints. However, the indices of these bad constraints are not provided, making it difficult to diagnose which constraints are causing issues.

Expected Behavior:
The solver should return the indices of the failed constraints instead of the count, allowing users to identify and fix problematic constraints.

Would be great to get feedback on whether this is feasible or if there's already a way to access this information that I may have missed. Thanks!

SolveSpace version: 3.1.0.dev26

iscgar:
This functionality was only available in the system solving API until now, and that made the sketch solving API less usable, even though it's the only API available through the Cython and Emscripten bindings.

In this initial implementation there's a redundant allocation and copy of the bad constraint list, but it was done this way in order to make a minimal change that would be the most ergonomic to use from both the C interface as well as the FFI binding interfaces.

Fixes #1546

### DIFF
--- a/exposed/CDemo.c
+++ b/exposed/CDemo.c
@@ -43,7 +43,7 @@ void ExampleStateful()
     // Slvs_Entity c1 = Slvs_AddCircle(g, n, center, radius, wp);
 
     Slvs_Vertical(g, p1, wp, p2);
-    Slvs_SolveResult res = Slvs_SolveSketch(g, 0);
+    Slvs_SolveResult res = Slvs_SolveSketch(g, NULL);
     printf("res: %i\n", res.result);
     printf("dof: %i\n", res.dof);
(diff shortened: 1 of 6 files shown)


ruevs:
@vespakoen @phkahler it seems this holds up https://github.com/hlorus/CAD_Sketcher/pull/504
What do you think of it?

iscgar:
@vespakoen @phkahler ping.

phkahler:
Looks OK and I can appreciate making things easier for CAD sketcher.

Issue #1569, opened 2025-04-17
Crash with/without "Handle isn't unique" error message


blinkenlight:
Solvespace version: 3.1~bde63c (openMP, 32-bit)
OS: Windows 7
Sample file: well, it's complicated. It's kinda large...

I have a moderately complex file of assemblies that teeters precariously on the edge of crashing. Change the number of translated copies up by a few? Crash. Import ANY new assembly? Crash. You get the idea. Most of the time there is NO error message at all whatsoever - only a "ding" then I have to kill Solvespace form the task manager. But sometimes there is, and this is what it says:

```
File D:\a\solvespace\solvespace\src\dsc.h, line 495, function
SolveSpace::IdList <class Solvespace::Entity, class
SolveSpace::hEntity>::Add:
Assertion failed: FindByIdNoOops(t->h) == nullptr.
Message: Handle isn't unique.

Generate debug report?
```

This seems to be an ancient issue if I'm not mistaken, as mentioned in [this old post](https://github.com/solvespace/solvespace/issues/932#issuecomment-808785294) - some kind of 16-bit resource depletion, apparently. While it's nice that we all know about it, this creates a huge problem for me, being unable to continue doing, well, basically ANYTHING with my design, even after enduring the obligatory seconds-per-frame punishment at each action. Please advise, what do? Any workarounds...? Heavens forbid, a patch...?

ruevs:
This is indeed a known limitation.
https://github.com/solvespace/solvespace/issues/932#issuecomment-808785294
https://github.com/solvespace/solvespace/issues/1227

Removing it requires a change in the file format, and all the hurdles related to keeping backward compatibility with old files and converting between the two.

Machines, have gotten faster, we've made SolveSpace much faster and these days it is possible to hit this limitation...

blinkenlight:
...so is the conclusion simply "too hard, fuggedaboutit in any kind of future that could be considered near"...? Sigh. I suppose it's true that the only thing preventing Solvespace from being the best 3D CAD there is and conquering the world is... Solvespace.

iscgar:
> ...so is the conclusion simply "too hard, fuggedaboutit in any kind of future that could be considered near"...? Sigh.

[...]

PR #1574, opened 2025-05-01
entity: fix a bug in the way equations are created from entity handles

iscgar:
The `hEntity::equation()` creates an equation handle from an entity handle, but it does so in a destructive way that could lead to duplicate equation handles because the equation index is simply or-ed with the entity handle value.

This only affects the TTF text entity which generates equations with a non zero index, so that's probably why this bug was rarely hit, if at all.

Note that this fix doesn't affect compatibility because equations are never saved into a file.

### DIFF
--- a/src/sketch.h
+++ b/src/sketch.h
@@ -959,7 +959,7 @@ inline hEntity hGroup::entity(int i) const
 inline hParam hGroup::param(int i) const
     { hParam r; r.v = 0x80000000 | (v << 16) | (uint32_t)i; return r; }
 inline hEquation hGroup::equation(int i) const
-    { hEquation r; r.v = (v << 16) | 0x80000000 | (uint32_t)i; return r; }
+    { hEquation r; r.v = (v << 2) | 0x80000000 | (uint32_t)i; return r; }
 
 inline bool hRequest::IsFromReferences() const {
     if(*this == Request::HREQUEST_REFERENCE_XY) return true;
@@ -979,21 +979,21 @@ inline hRequest hEntity::request() const
 inline hGroup hEntity::group() const
     { hGroup r; r.v = (v >> 16) & 0x3fff; return r; }
 inline hEquation hEntity::equation(int i) const
-    { hEquation r; r.v = v | 0x40000000 | (uint32_t)i; return r; }
+    { hEquation r; r.v = (v << 2) | 0x40000000 | (uint32_t)i; return r; }
 
 inline hRequest hParam::request() const
     { hRequest r; r.v = (v >> 16); return r; }
 
 
 inline hEquation hConstraint::equation(int i) const
-    { hEquation r; r.v = (v << 16) | (uint32_t)i; return r; }
+    { hEquation r; r.v = (v << 2) | (uint32_t)i; return r; }
 inline hParam hConstraint::param(int i) const
     { hParam r; r.v = v | 0x40000000 | (uint32_t)i; return r; }
 
 inline bool hEquation::isFromConstraint() const
     { if(v & 0xc0000000) return false; else return true; }
 inline hConstraint hEquation::constraint() const
-    { hConstraint r; r.v = (v >> 16); return r; }
+    { hConstraint r; r.v = (v >> 2); return r; }
 
 // The format for entities stored on the clipboard.
 class ClipboardRequest {


ruevs:
:-D I see you have been looking at the handles :-) So have we, @phkahler has documented his digging a bit more than I have:

```
https://github.com/solvespace/solvespace/blob/master/developer_docs/IdLists_Entities_and_Remap.txt
https://github.com/solvespace/solvespace/issues/1436#issuecomment-1870468829
https://github.com/solvespace/solvespace/issues/1436#issuecomment-2438966553
https://github.com/solvespace/solvespace/issues/1436#issuecomment-1892440724
```

Presumably [this](https://github.com/solvespace/solvespace/issues/1569#issuecomment-2845607275) is what prompted you to look :-)

As for this pull request - I have no idea if if is correct - I'll have to look carefully on a PC and debug :-)

phkahler:
I don't think this is right.  As I recall, the lower 16 bits of a handle are the main identifier. The upper 16 bits determine what created the thing with a handle. The top 2 bits determine what kind of thing created the handle (and thing associated with it).  For example, all entities have a 16bit handle, but the upper 16 bits encode the request that the entity belongs to.  Turns out entities can also be created by a group (see repeat groups, link groups, extrude, lathe, revolve, helix). When a group creates an entity the upper bit (I think) is set to indicate that the entity belongs to a group and the next 15 bits are the group handle. Likewise, parameters can be created by a request or a group. I think parameters can also be created by a constraint, and so the top 2 bits are needed to indicate from what type it came. And then there are handles for equations too...

Looks like equation handles put their parent (group, constraint) handle in the upper 16 bits and use the lower 16 for the index. That allows way more equations than we should ever need, but only reserving 2 bits seems like too few. Also when right shifting only 2 bits, you're preserving a bunch of bits of "parent" information that are not meant to be returned.

I recall seeing what looked like a mistake in one of the handles code, but I didn't think it was actually used and if so I didn't know how.

What we need is a short document showing the bits and fields for all the handle types. Then we'll be able to compare code to those definitions to determine correctness. It will also be extremely valuable for moving to 64 bit handles.

I'm in favor of expanding to 64bit handles only when we have a number of other changes that will brake compatibility, and only after we've got hierarchical sketches working (or as part of that implementation).

BTW this is also important to completely understand if relation constraints are allowed to create local variables (named parameters) on the fly. I still don't know if that's possible but it seems like it should be. And that reminds me what I need to do soon...

iscgar:
> Presumably [this](https://github.com/solvespace/solvespace/issues/1569#issuecomment-2845607275) is what prompted you to look :-)

I actually have been digging into the handles since I started working on my `List` and `IdList` changes, because when testing I hit the "handle isn't unique" assertion far too many times. I have at least four branches with failed attempts to make the conversion, and they all got nowhere because the code was too hacky and I wanted to do it right in order to avoid issues such as the one this PR is trying to address (there is a similar issue with the way contraints generate params, which is done in the same way entities generate equations, and only works because constraints generate at most a single param, so it's always or-ed with 0).

> I don't think this is right. As I recall, the lower 16 bits of a handle are the main identifier. The upper 16 bits determine what created the thing with a handle. The top 2 bits determine what kind of thing created the handle (and thing associated with it). For example, all entities have a 16bit handle, but the upper 16 bits encode the request that the entity belongs to. Turns out entities can also be created by a group (see repeat groups, link groups, extrude, lathe, revolve, helix). When a group creates an entity the upper bit (I think) is set to indicate that the entity belongs to a group and the next 15 bits are the group handle. Likewise, parameters can be created by a request or a group. I think parameters can also be created by a constraint, and so the top 2 bits are needed to indicate from what type it came.

This is mostly correct. Handles for params, entities, groups, requests, constraints and equations all behave slightly differently. When working on my 64-bit conversion trials I mainly divided the handle types into two types: regular handles, and "ownable" handles. Regular handles essentially use the entire 32 bit space of the handle value, whereas ownable handles encode their parent handle in the upper 16 bits, potentially with 1 or 2 bits taken for identifying the type of parent handle.

The only regular handles that we care about are essentially `hRequest`, `hGroup`, `hConstraint`, and `hStyle` (there are other types used as semi handles which also fall under the regular handle definition, such as `hStroke`, `hFill`, `hSSurface`, and `hSCurve`, but since they aren't directly used by the sketch I'm ignoring them in this classification).

All other handles are ownable handles:

* Params can be created by requests, constraints, and groups. In all three cases, the parent hadnle is encoded in the upper 16 bits, but only groups and constraints set one of the top two bits to indicate the type of parent (0x80000000 or 0x40000000, respectively).
* Entities can be generated by requests and groups. In both cases the parent handle is encoded in the upper 16 bits, but only a group sets the top bit (0x80000000) to indicate that it came from a group.
* Equations can be created by groups, constraints, and entities. When a group or a constraint is the parent, the top 16 bits are used to encode it, with the top bit (0x80000000) being set for a group parent. For entities this is more complicated, since to identify the entity you must preserve its top 16 bits as well (otherwise entities from different request might have the same handle value. The same goes from entities created by different groups). What the code is currently doing in order to keep this is to not shift the entity handle value at all, and just or it with the equation index and set the second most significant bit (0x40000000) which for entities isn't set by groups or requests.

The issue with TTF text entities, is that they create 3 equations, and as such, the or with the 1 and 2 indices simply sets bits which could already be set, leading to potentially duplicate entities.

phkahler:
@iscgar and @ruevs  BTW if handles are extended to 64bits we will need to update the remap tables to support that as well. In the future there may be reasons to expand remap to allow new things to be created in new ways - for example it would be desirable to create an entite remapped from 2 faces so that lines (or in some cases curves) created from booleans could get an associated entity.

ruevs:
> @iscgar and @ruevs BTW if handles are extended to 64bits we will need to update the remap tables to support that as well.

Indeed. And a few other things - see my old "mental plan" (that I finally typed today) here: https://github.com/solvespace/solvespace/issues/1569#issuecomment-2847200162

> @iscgar I'm really glad someone new is taking a deep dive into this stuff, and I hope you'll stick around ;-)

Me too :-) By the way @iscgar - don't take my slow reaction to your pull requests personally - they touch parts of the code I've rarely looked at and never modified - so it takes me time, a fresh mind and a desire to sit in front of a PC after doing it all day 5 days a week - all in short supply :-)

> BTW, what the heck are these about?
> `struct IsHandleOracle<hGroup> : std::true_type {};`

@phkahler They are type traits that are used in dsc.h to "enable" defining operator overloading functions (`==`, `!=`, `<`) for the handles. See here:

https://github.com/solvespace/solvespace/blob/30da776ae9b9ffc8bdec0337ffabd362f1c8e783/src/dsc.h#L13

And finally "ack." to both of you for the descriptions of how handles work - mostly matches my memories from the last time I looked at them when I rewrote IdList (2020?). Thank you. I'm labelling this "docummentation".

iscgar:
> I suspect the equation handles are mostly irrelevant, since there is no user interaction with equations and they are AFICT always generated (and not regenerated requiring a remap to preserve the handle). There really isn't a reason to query what the parent of an equation is, is there?

For equation coming from constraints there actually is a use for querying the parent handle, since they are used for indicating bad constraints when the solver fails to converge.

> BTW, what the heck are these about? `struct IsHandleOracle<hGroup> : std::true_type {};`
> 
> That's just a little odd thing I've never understood.

As @ruevs said in his comment, this serves to mark the `hGroup` type as a handle, so that the templated (in)equality and less-than operators defined at the top of `dsc.h` can participate in overload resolution for them, allowing you to compare handles directly instead of having to reach intro them and compare their `v` member.

It works by providing what's called an explicit specialisation of the `IsHandleOracle` templated type (for `hGroup`, in this instance), which by default is implemented (in `dsc.h`) to inherit from `std::false_type` (`std::true_type` and `std::false_type` are defined in the standard header [`<type_traits>`](https://en.cppreference.com/w/cpp/types/integral_constant). They are simply types providing a `bool` member named `value` whose value is `true` and `false`, respectively). This is a common pattern in C++ template metaprogramming to attach properties to types, in order to enable/disable functionality based on the result of querying the value of instantiating a template for a specific type.

> BTW if handles are extended to 64bits we will need to update the remap tables to support that as well. In the future there may be reasons to expand remap to allow new things to be created in new ways - for example it would be desirable to create an entite remapped from 2 faces so that lines (or in some cases curves) created from booleans could get an associated entity.

I don't see how the remap tables are relevant to 64-bit handle support. A remap is just a simple mapping of an entity handle and a "copy number" to an index, which can then be combined with the group handle to create an new entity handle with the group as its parent, as I decribed above. If we need to allow an entity to be remapped from more than a single entity, all you need is to allow the remap table key to contain more than a single entity (presumably the copy number will be 0 in that case, unless multiple entities can be created from a boolean operation).

> By the way @iscgar - don't take my slow reaction to your pull requests personally - they touch parts of the code I've rarely looked at and never modified - so it takes me time, a fresh mind and a desire to sit in front of a PC after doing it all day 5 days a week - all in short supply :-)

No worries. As you can see, while I have times with a flurry of activity on SolveSpace, I also have weeks where I don't have time to do anything. Life gets us all :)

iscgar:
After thinking about it some more, I think there's no bug here, and my approach in this PR is actually the buggy one.

Just as with constraint-generated param, which I mentioned above, entity-generated equations actually don't break when they're or-ed like this with the equation index, but the reason is different. While for constraint-generated params it's not a problem because they're only ever or-ed with 0, since constraints generate at most a single param, for entity-generated equations the reason is that only image and TTF text entities generate equations with non-0 indices, and they are always the first entity generated by their respective request (so the entity index in the request is always 0). This means that no other bits are set in the entitiy index when it's or-ed with the equation indices.

They also don't clash with other entity-generated equations, because the only other entities to generate equations are arc of a circle (which is never part of their request) and normal in 3D (which is always index 32 in the request, or 0x20 in hex, so it has a different distinguishing bit set).

My approach in this PR, though, does introduce an issue because for entity-generated equations it chops off the top two bits of an entity handle (the top bit is used to distinguish between request-generated entities and group-generated ones) and unconditionally sets the now second most significant bit, potentially leading to a collision which wasn't there before.

So while the existing use of bit storage in the handles is a bit too clever for my own taste, and just barely escapes problems, it does work and should not be changed. I think for 64-bit handles it would be better if we clearly segment the bits in the handles, so that developers will not need to hold this entire context in their head when they're introducing changes to the way handles are constructed and used.

Given the above, I'm closing this PR.

PR #1575, opened 2025-05-01
Minor cleanup: use the ParamList and EntityList type aliases

iscgar:
For some reason, even though the `ParamList` and `EntityList` aliases exist, parts of the code still spell the full type names (`IdList<Param, hParam>` and `IdList<Entity, hEntity>`, respectively). This makes the code verbose, less readable, and inconsistent. This PR changes it so that they use the aliases for better readability and consistency.

### DIFF
--- a/src/slvs/lib.cpp
+++ b/src/slvs/lib.cpp
@@ -1000,7 +1000,7 @@ void Slvs_Solve(Slvs_System *ssys, uint32_t shg)
 
         SK.entity.Add(&e);
     }
-    IdList<Param, hParam> params = {};
+    ParamList params = {};
     for(i = 0; i < ssys->constraints; i++) {
         Slvs_Constraint *sc = &(ssys->constraint[i]);
         ConstraintBase c = {};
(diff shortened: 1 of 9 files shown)


iscgar:
CI is again failing on 502 when cloning FreeType. We should probably just switch from the nongnu Sahannah instance to the upstream freedesktop [GitLab repository](https://gitlab.freedesktop.org/freetype/freetype) or the official [GitHub mirror](https://github.com/freetype/freetype). I suppose using the GitHub mirror would be better because it would be less likely to experience a downtime when our CI is running. @ruevs, WDYT?

iscgar:
Force-pushed to rebase on master after the merging of #1578, in order to check that FreeType is indeed cloned successfully without hitting 502 errors.

src/slvs/lib.cpp
```diff
@@ -860,7 +860,7 @@ Slvs_SolveResult Slvs_SolveSketch(uint32_t shg, int calculateFaileds = 0)
     }
 
     // add params from constraints
-    IdList<Param, hParam> constraintParams = {};
+    ParamList constraintParams = {};
```
ruevs, line 863: Where did this come from?

ruevs:
To me this is a no-brainer. Rebased on master.
Removed a mysterious `IdList<Param, hParam> constraintParams = {};` - see comment above.

Issue #206, opened 2017-03-04
Feature Request: Exporting solid STEP


urmilparikh:
The STEP file exported by SolveSpace contains only bounding surface.
Please add an option to export solid model.

whitequark:
@Jonimoose can you please verify that the issue here is that STEP files exported by SolveSpace do not share edges between different bounding surfaces in the same solid?

Jonimoose:
I just imported an exported solvespace model into 3 different commercial CAD and CAM software packages and they all imported it as a "solid". 

The software I used were Autodesk PowerShape, Autodesk FeatureCAM, Solidworks, and Vero Software EdgeCAM.

I did notice that FreeCAD seems to import the model differently after I've exported it from PowerShape as a step than it would exported straight from solvespace but that is the only software I've found that doesn't import the model as a "solid."

diggit:
I have same problem with non solid STEP models in KiCAD which uses OCE for STEP model handling. (as well as FreeCAD)

urmilparikh:
Autodesk 123D Design imports it as hollow model.

![image](https://cloud.githubusercontent.com/assets/10322815/25087860/9793d5a4-238f-11e7-974a-e95362dfe35b.png)

Source and step files: [cube.zip](https://github.com/solvespace/solvespace/files/925447/cube.zip)

Jonimoose:
[10mmcube.zip](https://github.com/solvespace/solvespace/files/925840/10mmcube.zip)
I've attached a a 10mm cube model that has been exported from PowerShape which does not exhibit the issue in FreeCAD, other than being about half the number of lines the main difference I can see is on line 42 and 43 of the powershape export:
`#28=SHAPE_REPRESENTATION_RELATIONSHIP('','',#114,#30);
#29=SHAPE_REPRESENTATION_RELATIONSHIP('','',#114,#115);`

and further down on lines 146 and 147:
`#114=SHAPE_REPRESENTATION('master',(#116),#186);
#115=SHAPE_REPRESENTATION('',(#117,#118),#186);`

I don't see any equivalent lines in the solvespace export but those lines are also not referenced anywhere else so I'm not sure if that is what is doing it.

Jonimoose:
I have tracked down the issue to being what I and @whitequark initially thought, the step file output not reusing the EDGE_CURVE objects along shared edges. 

[...]

Issue #1579, opened 2025-05-04
macOS build instructions in README don't work (again)


rcarmo:
I read through #1539 and did a little experimenting, but am stuck with a `cairo` issue:

```bash
/Users/rcarmo/Build/solvespace/extlib/cairo/src/cairo-mutex-impl-private.h:262:3: error: "XXX: No mutex implementation found.  Cairo will not work with multiple threads.  Define CAIRO_NO_MUTEX to 1 to acknowledge and accept this limitation and compile cairo without thread-safety support."
  262 | # error "XXX: No mutex implementation found.  Cairo will not work with multiple threads.  Define CAIRO_NO_MUTEX to 1 to acknowledge and accept this limitation and compile cairo without thread-safety support."
```

I think this is due to `cmake` not setting the right values, since I am trying to build with:

```bash
cmake .. -DCMAKE_BUILD_TYPE=Release -DENABLE_OPENMP=ON -DCMAKE_POLICY_VERSION_MINIMUM=3.5
```

This is because if I do it without the policy version I get this:

```bash
...
-- Using in-tree mimalloc
CMake Error at extlib/mimalloc/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.

-- Configuring incomplete, errors occurred!
```

My `cmake` is 4.0.1, by the way (and I have all the latest `brew` tooling, running on Sequoia)

iscgar:
CMake >= 4.0.0 introduced an incompatibility with existing CMake projects due to dropping support for versions < 3.5 (see the discussion in #1558). Some of our dependencies (such as zlib) still haven't released a version that's compatible with the new CMake >= 4.0.0 policies, which is why we didn't bother updating any of our dependencies yet.

In any case, the CMake recommendation to set back the policy to 3.5 is bad, and breaks our CMake rules. The best course of action is to downgrade your local CMake to 3.31.6, but if you can't do that, try setting `-DCMAKE_POLICY_VERSION_MINIMUM` to `3.13`, which worked for me (although note that this hack is not a supported in any way, and isn't guaranteed to work).

rcarmo:
[...]

PR #1580, opened 2025-05-04
Improve STEP export

rcarmo:
After a little fiddling with the STEP exporter, I managed to take the cube from #206 and export it correctly (select all, export).

This PR adds better float precision, a bit of clamping when outputting coordinates, and a couple of iterations at navigating the surface tree for export to try to ensure correct ordering.

### DIFF
--- a/src/exportstep.cpp
+++ b/src/exportstep.cpp
@@ -347,7 +347,7 @@ void StepFileWriter::ExportSurface(SSurface *ss, SBezierList *sbl) {
         // And create the face inner boundaries from any inner loops that
         // lie within this contour.
         loop = sbls->l.NextAfter(loop);
-        for(; loop; loop = sblss.l.NextAfter(loop)) {
+        for(; loop; loop = sbls->l.NextAfter(loop)) {
             int fib = ExportCurveLoop(loop, /*inner=*/true);
             listOfLoops.Add(&fib);
         }
--- a/src/exportstep.cpp
+++ b/src/exportstep.cpp
@@ -6,6 +6,9 @@
 #include "solvespace.h"
 #include <string>
 #include <unordered_map>
+#include <unordered_set>
+#include <vector>
+#include <functional>
 
 // Helper function to normalize coordinate values for STEP export
 // This avoids tiny floating point errors like 14.9999999993 that lead to open edges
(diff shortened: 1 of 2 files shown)


ruevs:
@phkahler I agree - as you said - what belongs to what is available/deducible from the model.

ruevs:
The "correct" STEP export is neither this nor #1589 but some "nonlinear combination" of both may be :-)

rcarmo:
Yeah, well, this was a way for me to get familiar with the code, so the SShell thing wasn't immediately obvious. Plus I've been held back by #1579 and had to do a "blind" build on Linux to work around it, etc.

The key blocker, though, is that I think I need a reference test file. I don't see one for #1589, but am happy to pull that in and see if I can merge them - I just think this needs some sort of formal test case to validate it.

ruevs:
https://github.com/solvespace/solvespace/issues/206#issuecomment-955550719

Here is the "test file" that #1589 used.

rcarmo:
Much obliged. I'll try to have another go this weekend.

ruevs:
Your solution exports unnecessary edges vertexes and even one extra surface.
See:
https://github.com/solvespace/solvespace/pull/1589#issuecomment-2887032507
https://github.com/solvespace/solvespace/pull/1589#issuecomment-2887037398

src/exportstep.cpp
```diff
@@ -4,6 +4,339 @@
...
 #include "solvespace.h"
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+#include <functional>
+
+// Helper function to normalize coordinate values for STEP export
+// This avoids tiny floating point errors like 14.9999999993 that lead to open edges
+static double normalizeCoordinate(double val) {
```
ruevs, line 15: The rounding that this function does can be achieved much easier and the function becomes redundant.

All you need to do is round the values in the `fprintf` statements to five decimal places (search for "%." in the file). This should solve the problem since LENGTH_EPS = 1e-6 (defined [here](https://github.com/solvespace/solvespace/blob/master/src/solvespace.h#L129)) is used everywhere as a threshold for numerical solutions. Therefore all numbers (like your 14.9999999993 example) should be either ".00000" or ".99999" up to the fith decimal place and will round to the "intended" value.

For example instead of :

```
fprintf(f, "#%d=CARTESIAN_POINT('',(%.10f,%.10f,%.10f));\n"`
```

use:

```
fprintf(f, "#%d=CARTESIAN_POINT('',(%.5f,%.5f,%.5f));\n"`
```

PR #1581, opened 2025-05-06
build: silence "maybe uninitialised" warnings

iscgar:
These pop up with GCC in the LTO build, and are easily worked around by adding initialisation or asserting, as needed.

### DIFF
--- a/src/clipboard.cpp
+++ b/src/clipboard.cpp
@@ -93,8 +93,8 @@ void GraphicsWindow::CopySelection() {
             if(!e->h.isFromRequest()) continue;
             Request *r = SK.GetRequest(e->h.request());
             if(r->type != Request::Type::DATUM_POINT) continue;
-            EntReqTable::GetEntityInfo((Entity::Type)0, e->extraPoints,
-                &req, &pts, NULL, &hasDistance);
+            ssassert(EntReqTable::GetEntityInfo((Entity::Type)0, e->extraPoints,
+                &req, &pts, NULL, &hasDistance), "No entity info");
         }
         if(req == Request::Type::WORKPLANE) continue;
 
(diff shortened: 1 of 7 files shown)


phkahler:
@iscgar  I'm not sure about enclosing in an assert to quiet compiler warnings.  @ruevs what do you think?

iscgar:
While some of them are only useful for silencing the compiler, I think the ones in `request.cpp` and `group.cpp` are useful in general in order to catch cases where new requests or entities (respectively) are added but their request<->entity mapping is missed.

ruevs:
I would've used (void) in front of the function calls. But I don't mind either way.

iscgar:
> I would've used (void) in front of the function calls. But I don't mind either way.

Note that this wouldn't resolve the maybe uninitialised issue (it only takes care of `-Wunused-value`, which I don't think is active, because I don't remember seeing it). The maybe uninitialised issue stems from the fact that `GetEntityInfo()` iterates over the request->entity mapping, but not all entities are generated by a request, in which case the function returns `false` and its output parameters are never set. Because they don't have an initialisation value, they may be used uninitialised if the function gets an entity type that isn't in the map (and while the compiler can see that due to LTO, it's apparently unable to look into the mapping table and see that only supported entity types are provided in all cases where the function is called and its return value is not checked). The assertion takes care of it, because it calls `abort()` on failure, preventing the code from using the uninitialised variables.

ruevs:
Oh sorry, you are right. I was not reading carefully :-)

phkahler:
If the return value is never checked, we could get rid of it.

iscgar:
> If the return value is never checked, we could get rid of it.

The return value is indeed never checked for `GetRequestInfo()`, so I can make that change there. But that's not the case for `GetEntityInfo()`, which is checked in exactly one place, here: https://github.com/solvespace/solvespace/blob/8042810d66de8072219d7e27dc2ace7e39af7738/src/clipboard.cpp#L90

PR #1583, opened 2025-05-09
Fix a bug in IDF arc creation. When swapping points we need to make t…

### LINKED ISSUE #1582: Circuit board (Linked IDF file) has incorrect arc direction
The attached .emn file links properly except the inside (concave) curves of the outline. The NURBS surfaces are correct but the sketch entities are go around the wrong way. See picture:

Image

[Snap.Timer.PCB.zip](https://github.com/user-attachments/files/20107205/Snap.Timer.PCB.zip)

This issue was originally reported by @blinkenlight in issue #1569 which is for an unrelated issue.

Additional info:

The entire IDF linker is in [importidf.cpp](https://github.com/solvespace/solvespace/blob/master/src/importidf.cpp)
Looking at the code comments [here](https://github.com/solvespace/solvespace/blob/e8f6ea3977553e4b07858b0a17cd74843bceeed5/src/importidf.cpp#L209) and the image above, it looks like the problem is a failure in calculating the circle center after reversing the end points. Solvespace always makes an arc counterclockwise from the first to second point, going around the center point. The center of this arc is clearly not in the correct location.

Also notable is that the NURBS surface is obviously not created from the arc. Somebody took a shortcut when they wrote this code...

phkahler:
…he arc angle positive as well. Also simplified the math for center distance calculation - the new formula does not need a special case for 180 degree arcs.

Closes #1582

### DIFF
--- a/src/importidf.cpp
+++ b/src/importidf.cpp
@@ -211,7 +211,7 @@ static Vector ArcCenter(Vector p0, Vector p1, double angle) {
 // indicates a circle centered at x1,y1 passing through x2,y2 and is a complete loop.
 static void CreateEntity(EntityList *el, int *id, hEntity h0, hEntity h1, hEntity hnorm,
                     Vector p0, Vector p1, double angle, bool keepout) {
-    if (angle == 0.0) {
+    if (fabs(angle) < 0.1) {
         //line
         if(p0.Equals(p1)) return;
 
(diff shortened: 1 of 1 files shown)


extlib/angle
ruevs, line 1: Was this intentional?

src/importidf.cpp
```diff
@@ -211,7 +211,7 @@ static Vector ArcCenter(Vector p0, Vector p1, double angle) {
 // indicates a circle centered at x1,y1 passing through x2,y2 and is a complete loop.
 static void CreateEntity(EntityList *el, int *id, hEntity h0, hEntity h1, hEntity hnorm,
                     Vector p0, Vector p1, double angle, bool keepout) {
-    if (angle == 0.0) {
+    if (fabs(angle) < 0.1) {
```
ruevs, line 214: "Length epsilon" perhaps?

src/importidf.cpp
```diff
@@ -228,16 +228,14 @@ static void CreateEntity(EntityList *el, int *id, hEntity h0, hEntity h1, hEntit
...
         Vector m = p0.Plus(p1).ScaledBy(0.5);
         Vector perp = Vector::From(p1.y-p0.y, p0.x-p1.x, 0.0).WithMagnitude(1.0);
-        double dist = 0;
-        if (angle != 180) {
-            dist = (p1.Minus(m).Magnitude())/tan(0.5*angle*3.141592653589793/180.0);
-        } else {
-            dist = 0.0;
-        }
+        // half angle in radians
+        double theta = 0.5*angle*3.141592653589793/180.0;
```
ruevs, line 237: We have Pi defined.

ruevs (changes requested):
@phkahler please clean this up a bit.

extlib/angle
phkahler: No, not at all. Ugh...

PR #1585, opened 2025-05-09
build: fix link error when building on Windows with Clang (not clang-cl)

iscgar:
With CMake >= 3.22, the "-Xlinker /MANIFEST:EMBED" flag is provided even when no manifest files are added, and since we add our own manifest file through a resource file, this fails with a duplicate resource error.

Unfortunately there's no easy way to avoid it, other than special casing for CMake >= 3.22 to add the manifest as a source for non-MINGW WIN32, and exclude it from the resources, but otherwise add it to the resource file.

However, this solution requires conditionally reaching into `res/` from the `src/` directory build, which is really bad, so for now hack out the offending part from `CMAKE_CXX_LINK_EXECUTABLE`, since we actually don't want it.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -356,9 +356,24 @@ if(ENABLE_GUI)
             ${FONTCONFIG_LIBRARIES})
     endif()
 
-    if(MSVC)
-        set_target_properties(solvespace PROPERTIES
-            LINK_FLAGS "/MANIFEST:NO /SAFESEH:NO /INCREMENTAL:NO /OPT:REF")
+    if(WIN32)
+        if(MSVC)
+            set_target_properties(solvespace PROPERTIES
+                LINK_FLAGS "/MANIFEST:NO /SAFESEH:NO /INCREMENTAL:NO /OPT:REF")
+        else(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND MSVC_VERSION)
+            # Hack to fix broken build with clang (not clang-cl) on Windows with CMake >= 3.22
+            # which unconditionally adds "-Xlinker /MANIFEST:EMBED" to the command line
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -356,9 +356,24 @@ if(ENABLE_GUI)
             ${FONTCONFIG_LIBRARIES})
     endif()
 
-    if(MSVC)
-        set_target_properties(solvespace PROPERTIES
-            LINK_FLAGS "/MANIFEST:NO /SAFESEH:NO /INCREMENTAL:NO /OPT:REF")
```
iscgar, line 361: Ugh, I'm only now noticing that I used `else()` instead of `elseif()`. This doesn't actually break anything, because MINGW simply doesn't have that flag set, so it would be a no-op. But this should really be fixed to use `elseif()` instead at some point, at least for clarity.

src/CMakeLists.txt
```diff
@@ -356,9 +356,24 @@ if(ENABLE_GUI)
...
+    if(WIN32)
+        if(MSVC)
+            set_target_properties(solvespace PROPERTIES
+                LINK_FLAGS "/MANIFEST:NO /SAFESEH:NO /INCREMENTAL:NO /OPT:REF")
+        else(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND MSVC_VERSION)
```
iscgar, line 363: Ugh, I'm only now noticing that I used `else()` instead of `elseif(). This doesn't break anything, because MINGW and clang-cl simply don't have that flag set, so this would be a no-op. But this should really be changed to `elseif()` at some point in the future, at least for clarity.

ruevs: Done

PR #1586, opened 2025-05-10
group: add an explicit assertion for too many entities

iscgar:
This makes it clearer to users who are faced with a SolveSpace crash, instead of getting the "handle isn't unique" error from `IdList` much later in a different code-path (prompted by the discussion in #1569).

While at it, add a minor fix to the documentation of `EntityMap`, which wrongly stated that entities are remapped from requests instead of from groups.


iscgar:
@ruevs @phkahler, I don't know what the timeline for the 3.2 release, and even if I'll be able to finish and submit my 64-bit handle works before then, I'm not sure it would be wise to rush its inclusion right before a release, so can we at least merge this in the mean time in order to make it easier for people t understand what the actual error is instead of crashing later with the `IdList` uniqueness error?

Issue #628, opened 2020-06-06
Deleting a workplane with an arc locked to it crashes SolveSpace


johreh:
### System information

SolveSpace version: 3.0~225f82a2 

Operating system: Arch 

### Actual behavior
After constructing a new workplane inside an existing group and constructing an arc on that workplane, SolveSpace crashes when deleting the workplane.

### Additional information
Example file with conditions already set up. Deleting r0004-workplane crashes SolveSpace.
[workplane_arc.zip](https://github.com/solvespace/solvespace/files/4740623/workplane_arc.zip)

phkahler:
I can reproduce the crash, but I can not figure out how you created that file. If I use Sketch->Workplane to create a new workplane it says I need to start a new sketch-in-3d first. If I create a sketch-in-3d first, I get a new group in the text window and can then add a new workplane. Your file does not have a sketch-in-3d group. What were the steps used to create that file?

ruevs:
It is easy:
- `Ctrl+N`
- `3` ("Sketch | Anywhere in 3d")
- Rotate view with middle button drag
- "Sketch | Workpalne"
- Click on origin to create the workplane
- Select the created "workplane"
- `2` ("Sketch | In Workplane")
- `A` ("Sketch | Ark of a Circle")
- Make an ark

Done. Now delete the workplane:

- `ESC`
- Click `g002-sketch-in-plane`
- Click `r004-wokplane1`
- `Del`

Crash.

It's a stack overflow in (the recursive) `SolveSpaceUI::GenerateAll` but I have not had time to debug it. Probably it is caused by the ark remaining an "orphan" when the workplane it is located in is deleted, but how exactly?...

Call stack:
```
>	solvespace.exe!malloc(unsigned int size) Line 23	C++
 	[External Code]	
 	solvespace.exe!SolveSpace::IdList<SolveSpace::Entity,SolveSpace::hEntity>::Add(SolveSpace::Entity * t) Line 424	C++
 	solvespace.exe!SolveSpace::Group::Generate(SolveSpace::IdList<SolveSpace::Entity,SolveSpace::hEntity> * entity, SolveSpace::IdList<SolveSpace::Param,SolveSpace::hParam> * param) Line 471	C++
 	solvespace.exe!SolveSpace::SolveSpaceUI::GenerateAll(SolveSpace::SolveSpaceUI::Generate type, bool andFindFree, bool genForBBox) Line 243	C++
[...]

PR #1587, opened 2025-05-10
generate: fix `PruneRequests()` to actually prune requests

iscgar:
Commit 86f20cc7e5061d14323fdc0d814980922425e152 (merged in #428) was marked NFC, even though it didn't just convert loops to ranged-for, but also introduced a functional change in the request pruning logic to prune entities instead.

This doesn't make much sense, since the requests will stay, and their entities will simply be regenerated, leading to an infinite recursion and resulting in a crash with stack overflow. I'm not sure how it slipped past code review, but this was the cause of #628 and is the reason for the additional workplane deletion code that whitequark added in 586b0477d2be31b2ac24c351f5eeeb90f86a7c2b, so this PR also reverts that code as well.

### DIFF
--- a/src/generate.cpp
+++ b/src/generate.cpp
@@ -86,46 +86,53 @@ bool SolveSpaceUI::PruneGroups(hGroup hg) {
     return true;
 }
 
-bool SolveSpaceUI::PruneRequests(hGroup hg) {
+bool SolveSpaceUI::PruneRequestsAndConstraints(hGroup hg) {
+    auto entityRequestExists = [](hEntity he, bool checkEntity = false) {
+        if(he == Entity::NO_ENTITY) {
+            return true;
+        }
+
+        if(he.isFromRequest()) {
+            if(SK.request.FindByIdNoOops(he.request()) != nullptr) {
+                return true;
+            }
+        } else if(!checkEntity || SK.entity.FindByIdNoOops(he) != nullptr) {
+            return true;
+        }
+
+        return false;
+    };
+
     const int requests = SK.request.n;
-    for(Entity &e : SK.entity) {
-        if(e.group != hg) {
+    for(Request &r : SK.request) {
+        if(r.group != hg) {
             continue;
         }
 
-        if(!e.h.isFromRequest()) {
+        if(entityRequestExists(r.workplane)) {
             continue;
         }
 
-        if(EntityExists(e.workplane)) {
-            continue;
-        }
-
-        Request *r = SK.GetRequest(e.h.request());
-        r->tag = 1;
+        r.tag = 1;
     }
     SK.request.RemoveTagged();
     deleted.requests += requests - SK.request.n;
 
-    return requests > SK.request.n;
-}
-
-bool SolveSpaceUI::PruneConstraints(hGroup hg) {
     const int constraints = SK.constraint.n;
     for(Constraint &c : SK.constraint) {
-        if(c.group != hg) continue;
+        if(c.group != hg)
+            continue;
 
-        if(EntityExists(c.workplane) &&
-           EntityExists(c.ptA) &&
-           EntityExists(c.ptB) &&
-           EntityExists(c.entityA) &&
-           EntityExists(c.entityB) &&
-           EntityExists(c.entityC) &&
-           EntityExists(c.entityD)) {
+        if(entityRequestExists(c.workplane, true) &&
+           entityRequestExists(c.ptA, true) &&
+           entityRequestExists(c.ptB, true) &&
+           entityRequestExists(c.entityA, true) &&
+           entityRequestExists(c.entityB, true) &&
+           entityRequestExists(c.entityC, true) &&
+           entityRequestExists(c.entityD, true)) {
             continue;
         }
 
-        (deleted.constraints)++;
         if(c.type != Constraint::Type::POINTS_COINCIDENT &&
            c.type != Constraint::Type::HORIZONTAL &&
            c.type != Constraint::Type::VERTICAL) {
@@ -134,10 +141,10 @@ bool SolveSpaceUI::PruneConstraints(hGroup hg) {
 
         c.tag = 1;
     }
-
     SK.constraint.RemoveTagged();
+    deleted.constraints += constraints - SK.constraint.n;
 
-    return constraints > SK.constraint.n;
+    return (requests > SK.request.n) || (constraints > SK.constraint.n);
 }
 
 void SolveSpaceUI::GenerateAll(Generate type, bool andFindFree, bool genForBBox) {
@@ -215,8 +222,7 @@ void SolveSpaceUI::GenerateAll(Generate type, bool andFindFree, bool genForBBox)
     // Remove any requests or constraints that refer to a nonexistent
     // group; can check those immediately, since we know what the list
     // of groups should be.
-    while(PruneOrphans())
-        ;
+    PruneOrphans();
 
     // Don't lose our numerical guesses when we regenerate.
     IdList<Param,hParam> prev = {};
@@ -254,7 +260,7 @@ void SolveSpaceUI::GenerateAll(Generate type, bool andFindFree, bool genForBBox)
 
         // The requests and constraints depend on stuff in this or the
         // previous group, so check them after generating.
-        if(PruneRequests(hg) || PruneConstraints(hg))
+        if(PruneRequestsAndConstraints(hg))
             goto pruned;
 
         // Use the previous values for params that we've seen before, as
--- a/src/generate.cpp
+++ b/src/generate.cpp
@@ -87,14 +87,27 @@ bool SolveSpaceUI::PruneGroups(hGroup hg) {
 }
 
 bool SolveSpaceUI::PruneRequests(hGroup hg) {
-    auto e = std::find_if(SK.entity.begin(), SK.entity.end(),
-                          [&](Entity &e) { return e.group == hg && !EntityExists(e.workplane); });
-    if(e != SK.entity.end()) {
-        (deleted.requests)++;
-        SK.entity.RemoveById(e->h);
-        return true;
+    const int requests = SK.request.n;
+    for(Entity &e : SK.entity) {
+        if(e.group != hg) {
+            continue;
+        }
+
+        if(!e.h.isFromRequest()) {
+            continue;
+        }
+
+        if(EntityExists(e.workplane)) {
+            continue;
+        }
+
+        Request *r = SK.GetRequest(e.h.request());
+        r->tag = 1;
     }
-    return false;
+    SK.request.RemoveTagged();
+    deleted.requests += requests - SK.request.n;
+
+    return requests > SK.request.n;
 }
 
 bool SolveSpaceUI::PruneConstraints(hGroup hg) {
--- a/src/generate.cpp
+++ b/src/generate.cpp
@@ -30,24 +30,24 @@ void SolveSpaceUI::MarkGroupDirty(hGroup hg, bool onlyThis) {
 }
 
 bool SolveSpaceUI::PruneOrphans() {
-
-    auto r = std::find_if(SK.request.begin(), SK.request.end(),
-                          [&](Request &r) { return !GroupExists(r.group); });
-    if(r != SK.request.end()) {
-        (deleted.requests)++;
-        SK.request.RemoveById(r->h);
-        return true;
+    const int requests = SK.request.n;
+    for(Request &r : SK.request) {
+        if(!GroupExists(r.group))
+            r.tag = 1;
     }
+    SK.request.RemoveTagged();
+    deleted.requests += requests - SK.request.n;
 
-    auto c = std::find_if(SK.constraint.begin(), SK.constraint.end(),
-                          [&](Constraint &c) { return !GroupExists(c.group); });
-    if(c != SK.constraint.end()) {
-        (deleted.constraints)++;
-        (deleted.nonTrivialConstraints)++;
-        SK.constraint.RemoveById(c->h);
-        return true;
+    const int constraints = SK.constraint.n;
+    for(Constraint &c : SK.constraint) {
+        if(!GroupExists(c.group))
+            c.tag = 1;
     }
-    return false;
+    SK.constraint.RemoveTagged();
+    deleted.constraints += constraints - SK.constraint.n;
+    deleted.nonTrivialConstraints += constraints - SK.constraint.n;
+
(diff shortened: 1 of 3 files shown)


phkahler:
@iscgar  I'm still testing things and thinking about this.  The first problem is a real problem - pruning requests should actually prune requests, not entities. The entities need to go too, but I'm assuming that was working prior to the bug.  The second piece of code that tags and removes entities in a workplane that has been tagged for deletion should stay - I think.  If a workplane is going to be deleted, any requests living in that workplane will need to be deleted as well.

Most of the time when we sketch in a workplane it is by doing NewGroup->Sketch in New Workplane.  Which means deleting the workplane will delete the group, and hence all entities in the group/workplane.  However, you can do Sketch in 3D, then create a workplane (just Sketch->Workplane), select that workplane and then Sketch->In Workplane and start drawing. Nothing in this scenario creates a new group, and you get a bunch of entities in a workplane that is just another part of the current group. If that plane is deleted we need to delete all the requests and entities belonging to it. Are we sure this is handled somewhere else?

iscgar:
> The first problem is a real problem - pruning requests should actually prune requests, not entities. The entities need to go too, but I'm assuming that was working prior to the bug. The second piece of code that tags and removes entities in a workplane that has been tagged for deletion should stay - I think. If a workplane is going to be deleted, any requests living in that workplane will need to be deleted as well.
>
> [snip]
>
>If that plane is deleted we need to delete all the requests and entities belonging to it. Are we sure this is handled somewhere else?

Both pieces of code are doing the exact same thing after my changes.

In general, all entities are created by requests. Group entities (`remap` entities) are simply copies of other entities, that may themselves be copies of other entities (through other `remap`s), but the chain of copies always ends in an entity that was created by a request.

During regeneration all of the entities are thrown away and regenerated from the existing requets and the groups. So when a request is removed, all of its entities go away automatically and there's no need to prune individual entities.

Requests generate all of their entitites on their own, and don't generally share entities, so when a request is removed, in theory we don't need to worry about other requests at all. However, the only exception to that rule is a workplane entity, which is shared by the request that created it and all of the entitities that are sketched in that workplane, and that's exactly why `PruneRequests()` exists.

This PR restores the logic of `PruneRequests()` to be this: iterate all of the entities, and if we encounter an entity that was generated by a request, and its workplane is from a request that doesn't exist, the request from which the current entity was generated is also removed (this doesn't need to be done recursively because only workplace entities are shared between requests, and workplane requests themselves don't depend on other requests, since they have their workplane set to `FREE_IN_3D`).

The other piece is now redundant, since it implements the same logic, and was only needed because the changes in https://github.com/solvespace/solvespace/commit/86f20cc7e5061d14323fdc0d814980922425e152 broke that logic by pruning entities instead of requests, so that's why I reverted it.

iscgar:
> For extrusions, remap create entities from prior entities. The new ones are not AFICT associated with any request. Also when linking files remap is used to assign new handles to entities in the other file, and these are considered to have the group as their parent. Similarly, when linking STL or IDF files, entities are created using remap _from nothing_ they are not even descended from a request.

Right, sorry. What I meant to highlight is that the entities themselves are thrown away and regenerated during `GenerateAll()`, and that's true for linked groups as well (except instead of being regenerated from request entities, the `impEntities` is used as a source). For extrusions, the new entities are copied from existing request entities though, and while they have the group set as their parent instead of a request, the extrude group is associated with a workplane, and the code in the `PruneGroups()` function is responsible for finding and pruning groups whose workplane or origin are gone.

In any case, there's code to handle pruning of both group-generated entities and request-generated entities, and pruning the entities themselves is useless because they are always regenerated.

> The remap table allows re-created entities to have the same handle so that subsequent constraints can be preserved. BTW In the future I'd prefer if remapped entities are not deleted and recreated if possible so that we may assign other attribute to them that get preserved ( I'm thinking fillet or chamfer, but right now we can't even assign a different style).

The best way to do that would be to extend the remap table to contain a mini version of request that stores metadata as well. That way, during regeneration the metadata could be attached to the regenerated entities too.

> I guess I'm arguing to preserve the code that whitequark added. OTOH if you're sure its removing exactly the same entities then the code can be removed. But one function is removing group-specific entities and the other is removing anything that's been tagged from the global sketch. Or is the one being called after tagging by the previous one?

It's indeed removing exactly the same entities. I'm not sure I follow your question though. Neither piece of code is removing group-specific entities. In fact, there's no code to do that, only a workplace can be depended upon by other entities, so when a workplane request is removed, we only need to remove specific requests that depend on it, or an entire group if it depends on it for its workplane (group pruning is done recursively using `PruneGroups()`, because groups can depend on previous groups that get pruned). If you're referring to pruning of `remap` entities, then those are never pruned directly (they can only be pruned as a result of their entire group being pruned).

ruevs:
@phkahler I am with @iscgar on this one. 86f20cc7e5061d14323fdc0d814980922425e152 did introduce the bug (#628) by removing only the first entity in the group without a workplane. The old code found all such entities and removed their requests. 586b0477d2be31b2ac24c351f5eeeb90f86a7c2b was an ugly hack to fix it.

src/generate.cpp
```diff
@@ -87,14 +87,27 @@ bool SolveSpaceUI::PruneGroups(hGroup hg) {
...
-        (deleted.requests)++;
-        SK.entity.RemoveById(e->h);
-        return true;
+    const int requests = SK.request.n;
+    for(Entity &e : SK.entity) {
+        if(e.group != hg) {
+            continue;
+        }
+
+        if(!e.h.isFromRequest()) {
```
ruevs, line 96: @iscgar Why did you decide to `continue` instead of:

```
        ssassert(e->h.isFromRequest(), "Only explicitly created entities can be pruned");
```

like [the old code did](https://github.com/solvespace/solvespace/commit/86f20cc7e5061d14323fdc0d814980922425e152#diff-d5fcd20826c6010f2f22908bf70fa6dfe7f2b186d5e151cb7fb4511adf41811bL101)?

ruevs:
And while on the topic. I think the change in [`SolveSpaceUI::PruneConstraints`](https://github.com/solvespace/solvespace/commit/86f20cc7e5061d14323fdc0d814980922425e152#diff-d5fcd20826c6010f2f22908bf70fa6dfe7f2b186d5e151cb7fb4511adf41811bR100) is also not functionally equivalent - `std::find_if` exists as soon as it finds the first matching item - so at most one item will be removed. This is not the same as the `for` loop it replaced. 

And the two `std::find_if` in [`SolveSpaceUI::PruneOrphans`](https://github.com/solvespace/solvespace/commit/86f20cc7e5061d14323fdc0d814980922425e152#diff-d5fcd20826c6010f2f22908bf70fa6dfe7f2b186d5e151cb7fb4511adf41811bR32) are "wrong" for the same reason.

@rpavlik are we missing something or did we miss it at the time?

src/generate.cpp
phkahler: @ruevs  You can't assert since not all entities are from requests. It's searching for ones that are from requests, so we don't want to assert if they are not.

phkahler:
> not functionally equivalent - `std::find_if` exists as soon as it finds the first matching item

Oh that's bad.

iscgar:
> And while on the topic. I think the change in [`SolveSpaceUI::PruneConstraints`](https://github.com/solvespace/solvespace/commit/86f20cc7e5061d14323fdc0d814980922425e152#diff-d5fcd20826c6010f2f22908bf70fa6dfe7f2b186d5e151cb7fb4511adf41811bR100) is also not functionally equivalent - `std::find_if` exists as soon as it finds the first matching item - so at most one item will be removed. This is not the same as the `for` loop it replaced.

The original for loop also returned after removing a single request, so the conversion to `std::find_if()` by @rpavlik is correct in that case. However, I intentionally changed it to tag all bad requests and remove them at once,, because otherwise we recurse into `GenerateAll()` for each request that is removed, which is wasteful.

> And the two `std::find_if` in [`SolveSpaceUI::PruneOrphans`](https://github.com/solvespace/solvespace/commit/86f20cc7e5061d14323fdc0d814980922425e152#diff-d5fcd20826c6010f2f22908bf70fa6dfe7f2b186d5e151cb7fb4511adf41811bR32) are "wrong" for the same reason.
> 
> @rpavlik are we missing something or did we miss it at the time?

No, the original code behaved that way for `PruneOrphans()` and `PruneConstraints()` as well. I have a branch where I changed them too to remove all of the requests and constraints that match the criteria instead of just the first one, and I planned to submit it as a separate PR later. I can submit those changes as part of this PR if you prefer that, or restore the code of `PruneRequests()` to prune a single request each time and submit the bulk removal change as a separate PR.

iscgar:
> Could this be the cause for #1239?

Unlikely, since that message is shown only in the code path that doesn't prune any objects, while the combination of https://github.com/solvespace/solvespace/commit/86f20cc7e5061d14323fdc0d814980922425e152 and https://github.com/solvespace/solvespace/commit/586b0477d2be31b2ac24c351f5eeeb90f86a7c2b just means that `PruneRequests()` never actually prunes anything (because all of the relevant requests are already pruned by the code introduced in https://github.com/solvespace/solvespace/commit/586b0477d2be31b2ac24c351f5eeeb90f86a7c2b), allowing the code to continue to try pruning constraints and groups without recursing back into `GnenerateAll()` through the `pruned` path.

ruevs:
> > And while on the topic. I think the change in [`SolveSpaceUI::PruneConstraints`](https://github.com/solvespace/solvespace/commit/86f20cc7e5061d14323fdc0d814980922425e152#diff-d5fcd20826c6010f2f22908bf70fa6dfe7f2b186d5e151cb7fb4511adf41811bR100) is also not functionally equivalent - `std::find_if` exists as soon as it finds the first matching item - so at most one item will be removed. This is not the same as the `for` loop it replaced.
> 
> The original for loop also returned after removing a single request, so the conversion to `std::find_if()` by @rpavlik is correct in that case. However, I intentionally changed it to tag all bad requests and remove them at once,, because otherwise we recurse into `GenerateAll()` for each request that is removed, which is wasteful.
> 
> > And the two `std::find_if` in [`SolveSpaceUI::PruneOrphans`](https://github.com/solvespace/solvespace/commit/86f20cc7e5061d14323fdc0d814980922425e152#diff-d5fcd20826c6010f2f22908bf70fa6dfe7f2b186d5e151cb7fb4511adf41811bR32) are "wrong" for the same reason.
> > @rpavlik are we missing something or did we miss it at the time?
> 
> No, the original code behaved that way for `PruneOrphans()` and `PruneConstraints()` as well. I have a branch where I changed them too to remove all of the requests and constraints that match the criteria instead of just the first one, and I planned to submit it as a separate PR later. I can submit those changes as part of this PR if you prefer that, or restore the code of `PruneRequests()` to prune a single request each time and submit the bulk removal change as a separate PR.

OK I read more carefully (not from the phone) this time and you are right about everything.


1. Leave `PruneOrphans` alone. This:
https://github.com/solvespace/solvespace/blob/2c5bb5d475e03a73da685649d2009539bfa96a7e/src/generate.cpp#L206
may be a bit stupid, but it is harmless :-)

2. If we are going to change `PruneRequests` then we should change `PruneConstraints` to prune everything at once as well. After all both decide
https://github.com/solvespace/solvespace/blob/2c5bb5d475e03a73da685649d2009539bfa96a7e/src/generate.cpp#L245
whether we will `goto pruned;` and call `GenerateAll` again.

3. This will be a very nice optimization and a significant speed up in some cases.

4. My only lingering doubt about doing it is whether pruning the requests and constraints one by one and running `GenerateAll`  each time - an thus solving "a little bit at a time" (everything already pruned) - somehow helps with the model not rearranging itself drastically when something is deleted... This may be stupid. This is the first time I am looking at this code.

phkahler:
@ruevs  So do you want to push the green button?  ;-)   I'm satisfied.

iscgar:
> 2. If we are going to change `PruneRequests` then we should change `PruneConstraints` to prune everything at once as well. After all both decide
> https://github.com/solvespace/solvespace/blob/2c5bb5d475e03a73da685649d2009539bfa96a7e/src/generate.cpp#L245
>        
> whether we will `goto pruned;` and call `GenerateAll` again.

As I said, I already have that ready in a private branch. I can just add it to this PR if you prefer to tackle them both here.

> 4. My only lingering doubt about doing it is whether pruning the requests and constraints one by one and running `GenerateAll`  each time - an thus solving "a little bit at a time" (everything already pruned) - somehow helps with the model not rearranging itself drastically when something is deleted... This may be stupid. This is the first time I am looking at this code.

As long as things get pruned, we simply recurse into `GenerateAll()` again, and we don't try to solve the current group. So it doesn't matter if we do it one by one or all at once, since the solver will only be called to rearrange things after we've pruned everything we could in the current group.

ruevs:
> As I said, I already have that ready in a private branch. I can just add it to this PR if you prefer to tackle them both here.

@iscgar ooops I committed a change to `PruneOrphans` and `PruneConstraints`, to prune all items, before reading your comment - and I forgot you already mentioned it earlier. Sorry. Feel free to revert it and take your versions.

But yes - let us do it all at once in this PR.

ruevs:
> As long as things get pruned, we simply recurse into `GenerateAll()` again, and we don't try to solve the current group. So it doesn't matter if we do it one by one or all at once, since the solver will only be called to rearrange things after we've pruned everything we could in the current group.

Reading even more carefully :-) you are right again - all it does is `Generate` stuff before the pruning. So there is not problem, but the optimization will not be as large as I initially (incorrectly) thought.

iscgar:
> @iscgar ooops I committed a change to `PruneOrphans` and `PruneConstraints`, to prune all items, before reading your comment - and I forgot you already mentioned it earlier. Sorry. Feel free to revert it and take your versions.
> 
> But yes - let us do it all at once in this PR.

No worries, your code is almost exactly what I had in my branch. I added another minor optimisation from my branch, but feel free to drop it if you think that it doesn't belong in this PR.

src/generate.cpp
```diff
@@ -86,46 +86,49 @@ bool SolveSpaceUI::PruneGroups(hGroup hg) {
     return true;
 }
 
-bool SolveSpaceUI::PruneRequests(hGroup hg) {
-    const int requests = SK.request.n;
-    for(Entity &e : SK.entity) {
-        if(e.group != hg) {
```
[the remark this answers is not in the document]
iscgar: Right, sorry. My private branch has a different order of operations for prunning, so this check was removed by accident when I cherry-picked into this PR's branch. I force-pushed to restore the check.

ruevs:
Thank you @iscgar .

Issue #866, opened 2020-12-19
Discuss approach to FEA in solvespace


phkahler:
Someone has recently been trying to display FEA results using SolveSpace by linking a mesh and modifying our rendering code. This invites the idea of doing more FEA related stuff in SolveSpace so I thought we could discuss that idea here.

As I see it, the renderer needs to be modified to interpolate an additional (heat, stress, etc...) parameter across triangles and use that value to index a color map as you can see in any search for [FEA results.](https://www.google.com/search?q=FEA+result&client=firefox-b-1-d&source=lnms&tbm=isch&sa=X&ved=2ahUKEwig9onN3trtAhUUK80KHV4aDk4Q_AUoAnoECAoQBA&biw=1175&bih=1008)
The other thing it needs to do is add a displacement to each vertex to show deformation. This displacement is often exaggerated (or even animated) to help visualize what is happening.

My thought on rendering is that we could add those features and leave them enabled full-time. The color map could be an RGBA map that we just make transparent in normal use. The displacement vectors would be zeroed and their multiplier would also be zero. The SolveSpace renderer seems fast enough that this would make no visible difference.  If displacements were done in a shader we probably couldn't even measure a difference, but that may bump our OpenGL version.

Beyond rendering is the question, how far should we go? Should FEA meshes be generated in SolveSpace? We could borrow code for 2d and 3d meshing from Gmsh which is also under GPL. Should we then add the ability to apply load conditions (or heat or whatever) in SolveSpace and just use an external FEA solver?

My main concern in heading this direction is maintaining the quality CAD experience SolveSpace is becoming known for, as well as keeping the code small and manageable with few resources. This would be a very large undertaking, but we might be able to go in smallish steps that each have some (limited) utility.

@ceanwang @jwesthues Thoughts?

jwesthues:
No strong feelings myself. I see the appeal of the feature, and e.g. of using the existing constraint/relation UI to apply boundary conditions. It's a big effort, though.

One possibility might be to add the UI for specifying conditions and viewing results, but keep the meshing and FEA solver itself as an external program, invoked from SolveSpace but appearing in a separate command prompt window? I'm thinking sort of analogous to invoking a compiler from within an IDE.

phkahler:
[...]

PR #1589, opened 2025-05-14
Export volumes to STEP files

ilguido:
A simple workaround to export valid volumes in STEP format. Quick description:

- all code is contained in _exportstep.cpp_;
- duplicated entities are treated as aliases, so that only one copy for each entity is used in all the file;
- if two points are within the same distance accuracy, they are treated as duplicated.

I tested this with a good number of drawings already and it works.

Screenshot_20250514_204207

### DIFF
--- a/src/exportstep.cpp
+++ b/src/exportstep.cpp
@@ -4,47 +4,11 @@
 // Copyright 2008-2013 Jonathan Westhues.
 //-----------------------------------------------------------------------------
 #include "solvespace.h"
-#include <vector>
-#include <iostream>
 
 const double PRECISION = 2*LENGTH_EPS;
 
 //-----------------------------------------------------------------------------
-// Structs to keep track of duplicated entities
-//-----------------------------------------------------------------------------
-// Basic alias.
-typedef struct {
-    int reference;
-    std::vector<int> aliases;
-} alias_t;
-
-// Cartesian points.
-typedef struct {
-    alias_t alias;
-    alias_t vertexAlias;
-    Vector v;
-} pointAliases_t;
-
-// Curves.
-typedef struct {
-    alias_t alias;
-    std::vector<int> memberPoints;
-} curveAliases_t;
-
-// Edges.
-typedef struct {
-    alias_t alias;
-    int prevFinish;
-    int thisFinish;
-    int curveId;
-} edgeAliases_t;
-
-std::vector<pointAliases_t> pointAliases;
-std::vector<edgeAliases_t> edgeAliases;
-std::vector<curveAliases_t> curveAliases;
-
-//-----------------------------------------------------------------------------
-// Functions for STEP export
+// Functions for STEP export: duplication check.
 //-----------------------------------------------------------------------------
 // Check if this point was already defined with a different ID number.
 // inputs:
@@ -141,11 +105,11 @@ bool StepFileWriter::HasBSplineCurveAnAlias(int number, std::vector<int> points)
         if(points.size() != c.memberPoints.size()) {
             continue;
         } else {
-            uint matches = 0; // is this the same curve?
+            size_t matches = 0; // is this the same curve?
             // FIXME: this hack should work _most_ of the times
-            for(uint i = 0; i < points.size(); i++) {
-                for(uint j = 0; j < points.size(); j++) {
-                    if(points.at(i) == c.memberPoints.at(j)) {
+            for(size_t i = 0; i < points.size(); i++) {
+                for(size_t j = 0; j < points.size(); j++) {
+                    if(points[i] == c.memberPoints[j]) {
                         matches++;
                     }
                 }
@@ -231,6 +195,9 @@ int StepFileWriter::InsertOrientedEdge(int number) {
     return -1;
 }
 
+//-----------------------------------------------------------------------------
+// Functions for STEP export: print to file.
+//-----------------------------------------------------------------------------
 void StepFileWriter::WriteHeader() {
     fprintf(f,
 "ISO-10303-21;\n"
--- a/src/exportstep.cpp
+++ b/src/exportstep.cpp
@@ -5,6 +5,238 @@
 //-----------------------------------------------------------------------------
 #include "solvespace.h"
 
+const double PRECISION = 0.001;
+
+/*
+ * Alias struct.
+ */
+typedef struct {
+    int reference;
+    List<int> aliases = {};
+} alias_t;
+
+/*
+ * Cartesian points.
+ */
+typedef struct {
+    alias_t alias;
+    alias_t vertexAlias;
+    Vector v;
+} pointAliases_t;
+
+List<pointAliases_t> pointAliases;
+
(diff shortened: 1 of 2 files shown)


iscgar:
I don't know the first thing about STEP files, so I have no comments about the logic, but please don't use `List<T>`. It's just a bespoke implementation of `std::vector<T>` which doesn't offer any benefits other than implicit shallow copy for code that needs it (at the cost of requiring users to manually free the memory). The STEP code isn't such a case, so please switch to `std::vector<int>` and use ranged for loops.

src/exportstep.cpp
```diff
@@ -5,6 +5,235 @@
 //-----------------------------------------------------------------------------
 #include "solvespace.h"
 
+const double PRECISION = 0.001;
```
ruevs, line 8:
```suggestion
const double PRECISION = 2*LENGTH_EPS;
```

ruevs: Defined [here](https://github.com/solvespace/solvespace/blob/master/src/solvespace.h#L129) used everywhere as a threshold for numerical solutions. `2*LENGTH_EPS` should guarantee (I would think) matching points.

0.001 is too big - there are people (very rarely) doing (for whatever reason) very precise things.

src/exportstep.cpp
```diff
@@ -5,6 +5,235 @@
...
+
+//-----------------------------------------------------------------------------
+// Structs to keep track of duplicated entities
+//-----------------------------------------------------------------------------
+// Basic alias.
+typedef struct {
+    int reference;
+    List<int> aliases = {};
```
ruevs, line 16: (As @iscgar pointed out) use `std::vector` unless `List` is actually needed.

src/exportstep.cpp
```diff
@@ -60,8 +289,8 @@ void StepFileWriter::WriteHeader() {
 "#171=DIRECTION('',(0.,0.,1.));\n"
 "#172=DIRECTION('',(1.,0.,0.));\n"
 "#173=CARTESIAN_POINT('',(0.,0.,0.));\n"
-"\n"
-    );
+"\n",
+    PRECISION);
```
ruevs, line 293:
```suggestion
    LENGTH_EPS);
```

src/exportstep.cpp
```diff
@@ -5,6 +5,235 @@
...
+// Functions for STEP export
+//-----------------------------------------------------------------------------
+// Check if this point was already defined with a different ID number.
+// inputs:
+//        number -> id of the cartesian point
+//        v -> position of the cartesian point
+//        vertex -> id of the vertex linked to this point (<1 if none)
+// return:
+//        true, if the cartesian point is already defined
+bool HasCartesianPointAnAlias(int number, Vector v, int vertex) {
```
ruevs, line 54:
```suggestion
bool StepFileWriter::HasCartesianPointAnAlias(int number, Vector v, int vertex) {
```

Ad so on for all newly added functions.

ruevs (commented):
I'm starting to think that this approach (discovering matching/duplicate points/lines/curves numerically at this level instead of trying to deduce it from the structure) may not be so bad... I need to try the other way ("logically") myself to see how hard it would be.

rcarmo:
@ilguido would you mind sharing the file that you're using so I can use it as a reference?

phkahler:
> I'm starting to think that this approach (discovering matching/duplicate points/lines/curves numerically at this level instead of trying to deduce it from the structure) may not be so bad...

@ruevs   I'm also in favor of doing it "the right way", but this would be the 3rd attempt by non-core developers to resolve the issue (of step file quality). When there is that much interest I think it's time to do something, and if you or I don't have the time to work through a "more correct" solution I think it's time to accept one or more of these others.  It may be worth pointing out that these "fixes" to the existing exporter do not impact the rest of the application - they are limited to the step file exporter.

Based on the image at the top of this PR it looks like this one handles the PWL curves as well, and to me that's important. I kind of like the snapping of #1580 but that's secondary and I'm not sure if that one handles PWL curves yet.

@ilguido   I agree with @iscgar comments (above) on this PR and using standard data structures instead of the solvespace Lists where possible and not introducing globals.

src/exportstep.cpp
```diff
@@ -37,9 +39,9 @@ typedef struct {
     int curveId;
 } edgeAliases_t;
 
-List<pointAliases_t> pointAliases;
-List<edgeAliases_t> edgeAliases;
-List<curveAliases_t> curveAliases;
+std::vector<pointAliases_t> pointAliases;
```
ruevs, line 42: Members of `StepFileWriter` please.

src/exportstep.cpp
```diff
@@ -136,25 +135,25 @@ int InsertVertex(int number) {
...
+bool StepFileWriter::HasBSplineCurveAnAlias(int number, std::vector<int> points) {
     // Look for this curve in the alias list.
-    for(curveAliases_t *c = curveAliases.First(); c; c = curveAliases.NextAfter(c)) {
-        if(points.n != c->memberPoints.n) {
+    for(curveAliases_t &c : curveAliases) {
+        if(points.size() != c.memberPoints.size()) {
             continue;
         } else {
-            int matches = 0; // is this the same curve?
+            uint matches = 0; // is this the same curve?
```
ruevs, line 144: `size_t`

src/exportstep.cpp
```diff
@@ -136,25 +135,25 @@ int InsertVertex(int number) {
...
             // FIXME: this hack should work _most_ of the times
-            for(int i = 0; i < points.n; i++) {
-                for(int j = 0; j < points.n; j++) {
-                    if(points.Get(i) == c->memberPoints.Get(j)) {
+            for(uint i = 0; i < points.size(); i++) {
```
ruevs, line 146: `size_t`

src/exportstep.cpp
```diff
@@ -136,25 +135,25 @@ int InsertVertex(int number) {
...
+                for(uint j = 0; j < points.size(); j++) {
```
ruevs, line 147: `size_t`

src/exportstep.cpp
```diff
@@ -136,25 +135,25 @@ int InsertVertex(int number) {
...
+                    if(points.at(i) == c.memberPoints.at(j)) {
```
ruevs, line 148: `if(points[i] == c.memberPoints[j])`

ruevs (changes requested):

ruevs:
You solution does "the right thing" when exporting the two samples I tried:
https://github.com/solvespace/solvespace/issues/206#issuecomment-955550719
[1mmCube.zip](https://github.com/user-attachments/files/20255813/1mmCube.zip)

image

unlike #1580

ilguido:
The refactored code should be OK now. I further tested it, but I invite you to do your checks anyway.

phkahler:
@ilguido  you imported into gmsh. Do you do FEA work? If so you might have something to say under issue #866

phkahler:
I've merged the other version of this PR.

PR #1591, opened 2025-05-17
split: fix line splitting for lines intersecting at their endpoints

iscgar:
When two line segments only intersect at their endpoints, the splitting logic creates four line segments instead of the original two. Two of the line segments have their endpoints constrained such that they are fully intersecting and cannot be moved as separate lines, and the original intersecting endpoint is forced to jump to 0,0,0. This is caused by the logic not finding an intersection point, but failing to indicate this case, causing the following code to use the zero vector as the intersection point.

Since the lines intersect on their endpoints, splitting them doesn't make much sense, even if the actual intersection point was found, because it would create zero length line segments. With the created line segments fully intersecting, and the original point forced to 0, this is even less useful, so just refuse to split in such a case.

### DIFF
--- a/src/modify.cpp
+++ b/src/modify.cpp
@@ -637,12 +637,7 @@ void GraphicsWindow::SplitLinesOrCurves() {
 
     Entity *ea = SK.GetEntity(ha),
            *eb = SK.GetEntity(hb);
-    SPointList inters = {};
-    SBezierList sbla = {},
-                sblb = {};
     Vector pi = Vector::From(0, 0, 0);
-
-    SK.constraint.ClearTags();
 
     // First, decide the point where we're going to make the split.
     bool foundInters = false;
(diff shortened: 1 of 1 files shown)


ruevs:
For cross reference: https://github.com/solvespace/solvespace/issues/1320#issuecomment-1374220862

iscgar:
I have a branch where I also added code to avoid creating the zero length line segment, but that's not what this PR is about.
Here's a video demonstrating the jump to 0:

https://github.com/user-attachments/assets/2436f3e6-b6c7-4657-bafd-a08c29533fb7

ruevs:
I'll take a look at both before we release 3.2 final.

ruevs:
Thank you @iscgar for the fix. Merged.

P.S. I like it how you clean up the areas you touch :-) In this case the fix was just moving `foundInters = true;` inside `if(!inters.l.IsEmpty())` but you changed the for loop to a range-based one and moved the locals to the minimal needed scope :-)

PR #1592, opened 2025-05-18
split: fix line splitting to avoid creating zero length line segments

iscgar:
When an endpoint of a line intersects another entity, line splitting replaces the line with two segments, one of them having zero length, and unnecessarily discarding all constraints on it in the process as well.

This is likely not what users expect when they split such a line, especially since they need to move the intersection endpoint in order to be able to do anything useful with the zero length segment. So this PR changes the behaviour to not split the line in that case, and instead just use the intersecting endpoint for the new coincident constraint.

Originally reported as part of #1320. While debugging #1239, I stumbled upon it and decided to fix it.

Before:

https://github.com/user-attachments/assets/8ce24f52-2b9b-4867-b612-0c0f2a7c00a3

After:

https://github.com/user-attachments/assets/7684d133-eb16-4ff5-8e66-3cab44629491

### DIFF
--- a/src/modify.cpp
+++ b/src/modify.cpp
@@ -448,6 +448,14 @@ hEntity GraphicsWindow::SplitLine(hEntity he, Vector pinter) {
     Vector p0 = SK.GetEntity(hep0)->PointGetNum(),
            p1 = SK.GetEntity(hep1)->PointGetNum();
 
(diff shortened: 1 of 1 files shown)


ruevs:
Merged. Thanks @iscgar!

Issue #1573, opened 2025-04-23
Bad STEP generation


DPlFnAz5J91kerf:
SolveSpace version 3.1~ed5e0b43

I created a cylinder with a diameter and a height of 30mm, exported it in STEP and opened it in PrusaSlicer, which showed the error: "744 Open Edges".

In cylinder.step, I replaced all substrings "14.9999999993" with "15.00000000" and "0.0000000007" with "0.0000000000" and saved as cylinder_fixed.step: the vertical "seams" disappeared in the slicer and the number of errors decreased.

Can it be a floating point crap? https://0.30000000000000004.com/

![Image](https://github.com/user-attachments/assets/ff434a37-22c9-4ac9-aa0d-f53037be2b8d)

[bad_step.zip](https://github.com/user-attachments/files/19873862/bad_step.zip)

iscgar:
Likely a duplicate of #206.

phkahler:
> In cylinder.step, I replaced all substrings "14.9999999993" with "15.00000000" and "0.0000000007" with "0.0000000000" and saved as cylinder_fixed.step: the vertical "seams" disappeared in the slicer and the number of errors decreased.

For issues like that I believe OCCT has shape healing functions. Prusa Slicer should apply those functions to imported STEP files, but they probably don't do that yet. Submitting a STEP file like this to them might be helpful.  But this is also a duplicate issue here - Solvespace could output better step files.

Until one or the other program gets a fix, you should be exporting your models as .stl  for 3D printing. This will also allow you to set the chord tolerance in solvespace to minimize the visibility of facets on your prints. It looks like Prusa isn't offering this setting on their STEP to triangle operation yet, so you'll have more control with .stl.

ilguido:
Using the method from #1589 I exported a cylinder which was good for PrusaSlicer: no error shown.

PR #1594, opened 2025-05-20
Experimental STEP export

ilguido:
Same as #1589, plus the possibility to use colours to define different volumes. Quick description:

- `StepFileWriter::exportParts` is the switch to enable the new feature, if **false**, everything is the same as #1589;
- instead, when **true** , coincident surfaces and edges can be distinguished by (surface) colour too, which is useful when treating assemblies with many parts in contact.

This way, the exported STEP file of an assembly with many parts of different colours should assign a distinct volume to each part.

Please, test this.


ruevs:
@ilguido @phkahler I added a commit to this pull request that rounds all numbers to the 5-th decimal point. It should solve/avoid the problem described in #1573 entirely.

phkahler:
@ruevs I'm not sure I like the rounding. The approach taken in #1580 did not round arbitrary numbers, it rounded things that were near integer values to integers on the assumption that they were supposed to be integers.  In particular we don't want to round weights to less precision when every 90 degree arc has a point with a weight of sqrt(2)/2 for example.

ruevs:
> The approach taken in #1580 did not round arbitrary numbers, it rounded things that were near integer values to integers on the assumption that they were supposed to be integers.

OK but who is to say that these {0, 0.5, 0.25, 0.75, 1/3, 2/3}

https://github.com/rcarmo/solvespace/blob/a2a2b3a9af200f97dc333f3a329c9e92983e7124/src/exportstep.cpp#L22

are the only numbers that were supposed to be "nice and round"? What if I wanted 1/5? This is very weird to me.

And we do so many numerical calculations down to `ChordTolMm` or `LENGTH_EPS` (1e-6) that I do not see the problem in rounding to 5 decimal places.

If fact searching for `ChordTolMm` (24 places) and `LENGTH_EPS` (160) and `Equals(` (185) and then narrowing down to the places where they are mixed (e.g. boolean.cpp) is a promising way to look for our chord tolerance dependent NURBS problems - time consuming though. But this is a different topic...

ruevs:
We should also keep in mind this https://steptools.com/stds/stp_aim/html/t_uncertainty_measure_with_unit.html

When we exported in master it is

```
#167=UNCERTAINTY_MEASURE_WITH_UNIT(LENGTH_MEASURE(0.001),#158,
'DISTANCE_ACCURACY_VALUE',
```

and in this pull request it currently is:

```
#167=UNCERTAINTY_MEASURE_WITH_UNIT(LENGTH_MEASURE(0.000002),#158,
'DISTANCE_ACCURACY_VALUE'
```

I do not know exactly what it does to different programs opening STEP files but it definitely has some relevance. All the "big guns" are discussing it:

https://community.ptc.com/t5/3D-Part-Assembly-Design/Is-there-a-way-to-define-accuracy-for-STEP-export/td-p/91286
https://community.sw.siemens.com/s/question/0D54O000061xGtnSAE/find-out-tolerance-used-for-step-file
https://forums.autodesk.com/t5/inventor-forum/exporting-to-step-accuracy-problem/td-p/12698909
https://discourse.mcneel.com/t/how-is-document-tolerance-calculated-when-you-open-a-step-file/149482
https://www.ibm.com/support/pages/apar/HD13614
https://dev.opencascade.org/content/how-read-units-step-file

I'm neither a mechanical engineer nor a STEP expert, but @ilguido chose to set this value to the tolerance he uses for deciding whether two points (and thus line segments, curves, surfaces) are the same (currently `2*LENGTH_EPS` on my suggestion).  And I thought 5 times (0.00001) lower precision for the actual values/numbers should be OK...

phkahler:
@ruevs  It's one thing to specify an accuracy, and another to deliberately round a number to some accuracy. Right?  I'm also curious about all those links. There is a whole area of CAD called Geometric Dimensioning and Tolerancing (GDT) which has to do with required accuracy of things when they are machined/manufactured (not how they are stored in a file). It's all outside my experience, but I still don't see a need to round floating point values to lesser precision. GD&T are parameters that can be called out on a drawing, so probably not what's being discussed in those links.

ruevs:
@ilguido I rebased this on current master.

phkahler:
@ruevs We should not apply FP to weights. I'd prefer FP to be more than 5, but regardless I don't think it should be applied to weights. I see you've added a "help wanted" tag I presume to get some input from someone with deeper understanding of STEP? I was really hoping to merge this ;-)

ruevs:
> but regardless I don't think it should be applied to weights.

Oh, now I understood - you don't like it for the `fprintf(f, FP, sb->weight[i]);` and `fprintf(f, "%.10f", ss->weight[i][j]);`. OK. And now your previous comment makes sense. I agree, I'll change it when I'm on a PC.

> I'd prefer FP to be more than 5

Initially I was going to make it 6 (same as LENGTH_EPS) but then I thought of the possibility of ending up with values like 13.00000099999 that will get rounded to 13.000001... so I went with 5. The errors in the cylinder from #1573 are much smaller, however I have not even tried to figure out if they will be smaller than 4e-7 (so that the rounding to 6 works) in all cases.

You are right - I added the "help wanted" hoping someone may shed some light on the UNCERTAINTY_MEASURE_WITH_UNIT value - is it "simply" a default precision/tolerance for the whole model (as in "machine this part to ±0.01mm") or does it somehow interact with the numerical coordinates in the file? At least for Prusa Slicer it seems it does not have anything to do with rounding the numbers - otherwise 0.001 would have been more than enough to round out the 14.99999999999s that are in that file.

If this is just a tolerance we should perhaps set it to the " export chord tolerance".

phkahler:
> OK. And now your previous comment makes sense.

Sometimes I chain different thoughts together too closely  ;-)

You're the best ruevs!

PR #1596, opened 2025-06-02
Experimental Gmsh GEO file export

ilguido:
The GEO file format is just a script file with a sequence of instructions for Gmsh. It can be used to define geometric objects. This PR adds a function to export points, curves and surfaces from Solvespace to a GEO file.

Exporting volumes requires some more work.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -117,6 +117,7 @@ if(ENABLE_GUI OR ENABLE_CLI)
         drawconstraint.cpp
         drawentity.cpp
         export.cpp
+        exportgeo.cpp
         exportstep.cpp
         exportvector.cpp
         file.cpp
--- a/src/exportgeo.cpp
+++ b/src/exportgeo.cpp
@@ -337,7 +337,11 @@ void GeoFileWriter::ExportSurfacesTo(const Platform::Path &filename) {
     curves = {};
     loops = {};
     surfaces = {};
+    // A special list of all the reference numbers of those loops that cannot be on NURBS surfaces:
+    // NURBS surfaces must have 2 < number of edges < 5.
     std::vector<int> planeSurfaces = {};
+    // Another special list with all the edges of each surface.
+    std::vector<geoEl_t> surfacesWithEdges = {};
 
     for(SSurface &ss : shell->surface) {
         if(ss.trim.IsEmpty())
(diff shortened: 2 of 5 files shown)


ilguido:
@phkahler As far as I have tested it, it works. Compared to exporting STEP files, it has advantages and disadvantages.

The advantages are (almost) all related to the use of Gmsh. It is the native format of Gmsh scripts, so it can be easily integrated into Gmsh projects, which use the Gmsh scripting language (it is also possible to use the Gmsh API for C++, Python etc., which is, by the way, much more powerful, but less immediate); it is a much simpler format, so it can be easily parsed by other programs (e.g. a GNU Octave script).

The disadvantages are that it is not an ISO standard and it is far less powerful than the STEP format. E.g. there are curves that the GEO format cannot describe (the Gmsh API can, though), it has no CAM information etc.

At the end it is just a small plus, but I think it is also quite an unobstrusive one. The choice is yours, mine is just a proposal.

`What does that mean? A plane is a surface of degree 2x2. Any surface can have a variable number of perimeter curves.`

NURBS surfaces have just four edges, or three if they are degenerate. So curved surfaces are usually represented by four-sided patches: to my knowledge, Solvespace follows this rule. However NURBS surfaces are only really  useful for curved surfaces: plane surfaces can have any number of edges. 

solvespace2step
As an example, I made a quick drawing in solvespace, exported it to the STEP file format, and then imported it into Gmsh (I could have used any program, actually): the hexagonal head has six edges, because it is a plane surface, while the circles at both the ends of the shank were split into four curves by Solvespace itself (the STEP format just reflects that). Gmsh requires to explicity declare a surface as plain or not, hence that line of code.

phkahler:
@ilguido   I'm guessing you currently import STEP in gmsh and have it create a geo file. This is just a shortcut? That's fine with me, since it's most self contained in a separate .cpp file.  As for surfaces, I think you already noticed how solvespace does things. Every surface consists of a SSurface object which is a NURBS surface of degree 3 in each direction (u,v). You are correct that a cone, or the tip of a sphere forms degenerate NURBS patches where all 3 control points along one edge are coincident. But then each surface has a set of trim curves. When extruding sketches, all the sides of the extrusion the trim curves (3D NURBS curves) exactly match the control points of the underlying surface. The top and bottom surfaces will get trim curves for every (non construction) entity in the sketch. That can be 2 bezier curves, or any number of shapes and cut-outs.

I have no way of testing this PR because I don't use gmsh. Let us know when you consider it "done" and we can merge it. By done, I would expect it to be useful to other gmsh users.

PR #1600, opened 2025-06-18
Allow dragging extrusions by dragging the end faces. Add the face poi…

phkahler:
…nt to the dragged list.

This does not work yet. It seems that ChooseFromHoverToSelect() and ChooseFromHoverToDrag() never recieve face entites. If they don't I'm not sure how faces get selected.

### DIFF
--- a/src/draw.cpp
+++ b/src/draw.cpp
@@ -502,6 +502,9 @@ void GraphicsWindow::HitTestMakeSelection(Point2d mp) {
             uint32_t v = m->FirstIntersectionWith(mp);
             if(v) {
                 sel.entity.v = v;
+                Hover hov = {};
+                hov.selection = sel;
+                hoverList.Add(&hov);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


phkahler:
We did a lot of face selection in #1258.  I'll look through some of those areas...

phkahler:
Looking at some other code for  [hover and selection](https://github.com/solvespace/solvespace/blob/925a52735cb2b04fed4831a3a65d2a570e2c0e20/src/draw.cpp#L475)  faces are never added to the hoverlist. A variable called "hover" will be set near the end of the function but that is distinct from hoverList. Dragging grabs something from the hoverList, and faces won't be in there.

So it looks like HitTestMakeSelection() uses a local "sel" to hold the one item that will be selected on click, but at the end it is transferred to "hover" which is part of the class.  hoverList holds *everything* that is under the mouse pointer so slightly different logic can be used for dragging. This could be cleaned up a bit... Hmmm did I split out the function ChooseFromHoverToDrag() ??   I know I added the "CanBeDragged" function to get click-through working right.

PR #1601, opened 2025-06-18
STEP export: fix broken export - `ORIENTED_EDGE`s are properly oriented.

ruevs:
After bb6cc1a0..d1b52cd6 in the exported STEP files "half" of the
`ORIENTED_EDGE`s were in the wrong direction. This caused some tools (e.g.
Kisters 3DViewStation) to render them incorrectly (this is understandable).

Fixed by flipping edges when required, keeping an `orientedEdgeAliases`,
list and reusing them.


ruevs:
@phkahler Merged. The risk of regressions is very low.

ilguido:
I tested it and I found no regressions so far. If there is some strange corner case, where it fails, users will submit a bug report.

PR #1603, opened 2025-06-20
Win32: Make SolveSpace grea...static again

ruevs:
For some reason the [/MT options in the `CMAKE_C_FLAGS_RELEASE_INIT`](https://github.com/solvespace/solvespace/blob/0df1ba74089b41316630270c1a44070672ab1966/cmake/c_flag_overrides.cmake#L4) etc. CMake targets stopped working at some point and SolveSpace became linked to the DLL versions of the C/C++ runtime (MSVCP140.DLL, VCRUNTIME140.DLL).

We do not like this. Therefore:

Use the proper "modern" CMake incantation to make SolveSpace static again ;-)
https://cmake.org/cmake/help/latest/prop_tgt/MSVC_RUNTIME_LIBRARY.html

The OpenMP build still depends on VCOMP140.DLL but this seems unavoidable.
https://stackoverflow.com/questions/37037733/statically-link-with-microsoft-crt-and-openmp
https://github.com/godotengine/godot/issues/14935

image


ruevs:
@rpavlik any idea why these stopped working?

https://github.com/solvespace/solvespace/blob/0df1ba74089b41316630270c1a44070672ab1966/cmake/c_flag_overrides.cmake#L4

PR #1605, opened 2025-06-26
emscripten: fix resource paths during packaging

iscgar:
The current hack of trying to fix the relative path in order to fix the "file not found" error on Windows broke the Linux build, because the working directory is different (originally the reverse was true--the Linux build worked, but the Windows build was broken).

I tried to use absolute paths in solvespace#1566, but that triggered the infamous "command too long" on error, because Emscripten uses wrapper batch scripts, and CMD has a limit of 8k on the length of the command line, and trying to use a response file didn't work because internally `file_packager` was called without using a response file.

With https://github.com/emscripten-core/emscripten/pull/24107 this was fixed since Emscripten 4.0.8, so we can now use that to pass absolute paths and fix it once and for all, but we unfortunately still need to resort to use some arcane CMake hacks to ensure that the dependencies are correctly specified.

@ruevs I don't currently have access to a Windows machine, so while I'm pretty sure this should work, I'd appreciate if you could test it.


iscgar:
Tested on my Windows machine using the following commands and it is indeed working:

```
emsdk install latest
emsdk activate latest
emcmake .. -GNinja
ninja solvespace
```

iscgar:
My claim about the portability of using `OBJECT_DEPENDS` was wrong, as it too is only supported by the Ninja and Makefile generators. So I changed the whole thing to use `LINK_DEPENDS`, and added a note about the portability caveat.

ruevs:
Thank you @iscgar this works. Merged.

PR #1607, opened 2025-07-02
general: rework the way files are included for better maintainability

iscgar:
Before this change, every header file was included in `solvespace.h`, inside the `SolveSpace` namespace. This meant that no header file could itself include a header file, unless it was included from `solvespace.h` as well, as otherwise standard library symbols would be injected into the `SolveSpace` namespace and used from there, causing compilation and linking issues.

It also meant that no header file could be included on its own, because its dependencies were only satisfied by the inclusion order in `solvespace.h`, which means that the only way to include anything was to just include `solvespace.h`.

An additional annoyance for developers was that code completion tools broke because of this inclusion pattern, making working on SolveSpace a tad more difficult.

This PR introduces as minimal changes as possible to make it so that the headers files are properly included at the top of `solvespace.h`, outside the `SolveSpace` namespace. As a consequence of this change, each header file needs its dependencies to be explicitly satisfied, so do that. This also required shuffling the order of a few definitions in some files, as well as extracting some small definitions into their own files, so they could be depended upon without causing a circular inclusion.

This is the latest iteration of the change that I mentioned [here](https://github.com/solvespace/solvespace/pull/1537#issuecomment-2698500961), after working with it in my local branches and tweaking it.

### DIFF
--- a/test/analysis/contour_area/test.cpp
+++ b/test/analysis/contour_area/test.cpp
@@ -1,3 +1,5 @@
+#include "solvespace.h"
+
 #include "harness.h"
 
 TEST_CASE(normal_roundtrip) {
--- a/test/constraint/pt_on_circle/test.cpp
+++ b/test/constraint/pt_on_circle/test.cpp
@@ -1,3 +1,5 @@
+#include "solvespace.h"
+
 #include "harness.h"
 
 TEST_CASE(normal_roundtrip) {
--- a/test/core/expr/test.cpp
+++ b/test/core/expr/test.cpp
@@ -1,3 +1,5 @@
+#include "expr.h"
+
 #include "harness.h"
 
 #define CHECK_PARSE(var, expr) \
--- a/test/core/locale/test.cpp
+++ b/test/core/locale/test.cpp
@@ -1,3 +1,5 @@
+#include "ui.h"
+
 #include "harness.h"
 
 TEST_CASE(parseable) {
--- a/test/group/translate_asy/test.cpp
+++ b/test/group/translate_asy/test.cpp
@@ -1,3 +1,5 @@
+#include "solvespace.h"
+
 #include "harness.h"
 
 TEST_CASE(normal_roundtrip) {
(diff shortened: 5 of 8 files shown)


iscgar:
Yes, it's mostly for the benefit of the library, but it's also useful in general to avoid conflicts with global symbols defined by dependencies of SolveSpace. The need to put even source files that are not used by the library into the namespace was mainly for consistency, as otherwise we'd still need to add `using namespace SolveSpace` in the source files (the existing approach of putting that in the `solvespace.h` header was wrong, as it ends up polluting the global namesapce of anything that includes it).

Regarding the handles, I did it mainly to reduce the dependance on `dsc.h` in places that don't need it. Extracting the functions is a bit more comlex due to interdependecies between the handle types, so I left most of them intact (that's why they are defined at the end of `solvespace.h`, after all of the handle types have been defined). As part of my 64-bit handles work I refactored this as well, and handles are a bit more self-contained, but I wanted to keep the diff for this PR as small as possible, just to achieve the stated goal. We can continue to iterate from there as needed afterwards.

phkahler:
@rpavlik  can you weigh in on this namespace stuff?

iscgar:
Rebased on master to resolve conflicts after the merging of #1567 and #1613.

iscgar:
> @phkahler I'm fine with this 99%. The only thing I do not understand it the large rearrangement of code in `sketch.h`. @iscgar what necessitated it?

I don't remember why I originally did it this way, but I think it was in order to reduce the number of forward declarations needed to get the code to work properly and to simplify the definition dependency, since `Entity` is defined in the same file and there's noting requiring it to be defined later.

Additionally, this branch is also the base of my (still WIP) `List<T>` and `IdList<T>` branch. I added some compile time assertions to simplify the code, but they require a complete type to work, so `EntityList` couldn't be instantiated until the definition of `Entity` was visible, but `Group` needs this definition for linked sketches, so the easiest fix for me was to move the definition of `Entity` to be before `Group`. This of course isn't a reason for pushing this change to master, though (at least not before my `IdList` work is ready for submission).

Issue #1498, opened 2024-11-27
Keyboard Shortcut for switching through group stack


nobodyinperson:
### System information

- **SolveSpace version:** latest master + some unrelated changes (https://github.com/nobodyinperson/solvespace), built like [this](https://gitlab.com/nobodyinperson/nixconfig/-/blob/main/pkgs/solvespace.nix?ref_type=heads).
- **Operating system:** NixOS
 
### Expected behavior

- e.g. PgUp and PgDown (currently unused keys in SolveSpace, right?) could switch through the group stack

### Actual behavior

nothing 🙃

### Additional information

Navigating through groups is essential for the solvespace workflow. Currently AFAIK the only way is to use the attribute browser window. But how about adding a keyboard shortcut to go up/down one group?

PR #1610, opened 2025-07-29
Add PageUp/PageDown keyboard shortcuts for group navigation

killerdevildog:
Implements Issue #1498 - adds keyboard shortcuts to navigate between groups in the group stack using PageUp (previous group) and PageDown (next group).

Changes:
- Added GROUP_PREVIOUS and GROUP_NEXT commands to ui.h
- Added menu entries and handlers in graphicswin.cpp
- Extended GTK keyboard handling for PageUp/PageDown recognition
- Updated accelerator descriptions to show proper key names
- Added translation strings for 'Previous Group' and 'Next Group'

The shortcuts are available both as keyboard shortcuts and via the Edit menu. Navigation stops at first/last group and maintains UI consistency.

### DIFF
--- a/KEYBOARD_GROUP_NAVIGATION_TEST.md
+++ b/KEYBOARD_GROUP_NAVIGATION_TEST.md
@@ -1,0 +1,46 @@
+# Keyboard Group Navigation Test
+
+## What was implemented
+
+Added PageUp and PageDown keyboard shortcuts to navigate between groups in SolveSpace.
+
+## Files modified
+
+1. `src/ui.h` - Added `GROUP_PREVIOUS` and `GROUP_NEXT` command enums
+2. `src/graphicswin.cpp` - Added menu entries and command handlers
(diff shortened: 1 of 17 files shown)


phkahler:
@ruevs  should this be in the menu? I'm thinking not. It can be documented somewhere else.

@killerdevildog Does this work on Apple and Windows, or just GTK?

ruevs:
> @ruevs should this be in the menu? I'm thinking not. It can be documented somewhere else.

I think it should _not_ be in the menu. Just like "ESC", '+', '-' etc.

I've not tried this PR yet but it apparently was LLM generated (at least partially) - and at least the two new large .md files are completely unnecessary and should not be part of the commit.

The translation updates should be a separate commit - if they are needed at all - they will not be if there are no new menu items.

phkahler:
This raises a question for me. What is the "correct way" to get keys to do things if they're not command shortcuts defined in the menu?

Issue #1595, opened 2025-05-28
Build failure on macOS 15.5


phansel:
Clang 17 incompatibility:

```
[ 17%] Building CXX object extlib/libdxfrw/CMakeFiles/dxfrw.dir/intern/dwgreader21.cpp.o
In file included from /Users/user/solvespace/extlib/zlib/zutil.c:8:
/Users/user/solvespace/extlib/zlib/zutil.h:163:11: warning: 'OS_CODE' macro redefined [-Wmacro-redefined]
  163 | #  define OS_CODE 19
      |           ^
/Users/user/solvespace/extlib/zlib/zutil.h:134:11: note: previous definition is here
  134 | #  define OS_CODE  7
      |           ^
In file included from /Users/user/solvespace/extlib/zlib/zutil.c:10:
In file included from /Users/user/solvespace/extlib/zlib/gzguts.h:21:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/stdio.h:61:
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/_stdio.h:318:7: error: expected identifier or '('
  318 | FILE    *fdopen(int, const char *) __DARWIN_ALIAS_STARTING(__MAC_10_6, __IPHONE_2_0, __DARWIN_ALIAS(fdopen));
      |          ^
```

Same issue as https://github.com/ahrm/sioyek/issues/1361.

-------------------

XCode incompatibility: fp.h is no longer present.

```
[ 19%] Building C object extlib/pixman/pixman/CMakeFiles/pixman.dir/pixman-matrix.c.o
In file included from /Users/user/solvespace/extlib/libpng/scripts/intprefix.c:22:
/Users/user/solvespace/extlib/libpng/scripts/../pngpriv.h:527:16: fatal error: 'fp.h' file not found
  527 | #      include <fp.h>
      |                ^~~~~~
[ 19%] Built target mimalloc-static
1 error generated.
CMake Error at scripts/genout.cmake:78 (message):
  Failed to generate
  /Users/user/solvespace/build/extlib/libpng/scripts/intprefix.out.tf1
```

Seems to be related to this: https://community.aseprite.org/t/build-failed-on-macos-15-4-due-to-fp-h-not-found/25192/2

ruevs:
This can be solved by updating zlilb and LibPNG.

Here are some instructions how to do it from someone who did it recently.
[...]

PR #1611, opened 2025-08-09
extlib: update zlib, libpng, and mimalloc to their latest releases

iscgar:
This takes care of the outdated libraries that use the upstream repositories directly, and should fix the build issues on macOS reported in #1579 and #1595. I have a separate branch that also updates pixman and cairo, but they require updating the SolveSpace forks first, and I'm not sure what the best way to tackle it is, given that both of them have a non default branch where the SolveSpace patches are added, so I can't just open a PR against them. I also want to update ANGLE while I'm at it, and it's taking longer than expected, so I wanted to get the low hanging fruit out of the way first.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -243,9 +243,22 @@ if(ENABLE_GUI OR ENABLE_CLI)
         set(FORCE_VENDORED_PNG      ON)
         set(FORCE_VENDORED_Freetype ON)
 
-        find_vendored_package(ZLIB zlib
-            ZLIB_LIBRARY            zlibstatic
-            ZLIB_INCLUDE_DIR        ${CMAKE_SOURCE_DIR}/extlib/zlib)
+        # HACK: This is essentially what `find_vendored_package()` does, except we also
+        # define the ZLIB::ZLIB target to be an alias for zlibstatic before calling
+        # `find_package()` in order to avoid having it defined as an imported target
+        # by the built-in FindZLIB.cmake module, which breaks libpng's png_static
+        # target that relies on ZLIB::ZLIB instead of on zlibstatic.
+        # Once zlib releases its develop branch post 1.3.1, the built-in module will not
+        # generate an imported target since we'll have proper targets, but libpng would
+        # still need to be updated to use ZLIB::ZLIBSTATIC instead of ZLIB::ZLIB, so this
+        # ugly hack will have to stay until both libraries release fixed versions.
+        add_vendored_subdirectory(extlib/zlib)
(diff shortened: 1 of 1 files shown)


iscgar:
The macOS build issue seems unrelated to this change, and I'm not sure what's causing it. Interesting that we run into an ASan issue on Ubuntu. I'll try to debug it and report back.

iscgar:
The macOS issue indeed isn't related to this change (encountered 4 months ago by this HN commenter: https://news.ycombinator.com/item?id=43552607), but I have no idea what's causing it and how to fix it becuase I don't have access to a macOS setup.

Based on [this CMake issue](https://gitlab.kitware.com/cmake/cmake/-/issues/24966), I temporarily pushed a commit with the diff below, and it seems to work, so I opened #1614 to fix that, as it's not related to the changes in this PR. I'll rebased this and the other PRs on it after it's merged.

```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,7 +136,7 @@ if(ENABLE_OPENMP)
     find_package( OpenMP REQUIRED )
     if(OPENMP_FOUND)
         add_library(slvs_openmp INTERFACE)
-        target_compile_options(slvs_openmp INTERFACE ${OpenMP_CXX_FLAGS})
+        target_compile_options(slvs_openmp INTERFACE "SHELL:${OpenMP_CXX_FLAGS}")
         target_link_libraries(slvs_openmp INTERFACE
             ${OpenMP_CXX_LIBRARIES})
         target_include_directories(slvs_openmp SYSTEM INTERFACE
```

ruevs:
Offff mimalloc is a PITA, always has been https://github.com/solvespace/solvespace/actions/runs/16865637438/job/47772106561#step:4:155 :-(

iscgar:
The Python issue is an [upstream mimalloc issue](https://github.com/microsoft/mimalloc/issues/1064), which was fixed in 2.2.4, that we cannot use due to the issue mentioned above. I'll try to find a version that works later today, and report back if I find a good solution that doesn't require reverting the update altogether.

The WASM issue seems to be caused by Emscripten support in mimalloc, which layers it on top of emmalloc (see https://github.com/emscripten-core/emscripten/pull/20651). Choosing to specify `-sMALLOC=emmalloc` would slow down allocations of everything that doesn't use the temporary heap, so I'm not sure it's the right choice. An alternative might be to just not build mimalloc with Emscripten, and instead specify `-sMALLOC=mimalloc`, so that the Emscripten mimalloc is used for regular allocations as well.

iscgar:
#1616 should take care of both issues.

PR #1613, opened 2025-08-09
tests: actually randomise the order of tests

iscgar:
`std::random_shuffle` can use `std::rand` (and does so on Windows), and without intialising with `srand()` the order will always be the same, so we don't get the expected randomisation.

Switch to using `std::shuffle` with a proper random number generator in order to fix this and get random order execution.

### DIFF
--- a/test/harness.cpp
+++ b/test/harness.cpp
@@ -3,6 +3,8 @@
 //
 // Copyright 2016 whitequark
 //-----------------------------------------------------------------------------
+#include <cstdio>
+#include <random>
 #include <regex>
 #include <cairo.h>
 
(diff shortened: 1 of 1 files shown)


iscgar:
The [Mersenne Twister](https://en.wikipedia.org/wiki/Mersenne_Twister) is still a PRNG, just seeded from what might be a true random (not required by the standard). The original code always had a deterministic order because the RNG wasn't seeded, so the shuffle did nothing. If we're after reproducibility, I can just print the seed and add an option to supply a seed for running the tests to make it reproducible. Would that be acceptable?

test/harness.cpp
```diff
@@ -337,19 +339,28 @@ int Test::Case::Register(Test::Case testCase) {
...
 
-    std::regex filter(".*");
+    std::string filterPattern = ".*";
+    unsigned int seed = std::random_device{}();
     if(args.size() == 1) {
     } else if(args.size() == 2) {
-        filter = args[1];
+        filterPattern = args[1];
+    } else if(args.size() == 3) {
+        seed = std::stoul(args[2]);
```
ruevs, line 348:
```suggestion
        filterPattern = args[1];
        seed = std::stoul(args[2]);
```

PR #1618, opened 2025-08-17
Enable building for `nix` & `NixOS` based systems

grenewode:
* Support building using the unstable but defacto-standard "nix flakes"
* Support building using the stable but defacto-depricated "nix-build"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /CMakeCache.txt
 /build*/
+/result
 /test/**/*.diff.*
 /test/**/*.curr.*
 *.trace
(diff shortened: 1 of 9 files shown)


grenewode:
This doesn't really add upstream packages for `solvespace` in this repo; that's actually already available via [the nixpkgs repository](https://github.com/NixOS/nixpkgs/blob/nixos-25.05/pkgs/by-name/so/solvespace/package.nix#L96) (essentially a `ports` tree for NixOS). 

It's more similar to [`meson`'s subprojects/wrap-dependencies](https://mesonbuild.com/Wrap-dependency-system-manual.html). With some additional tooling/capabilities from the NixOS project, such as easy to use development environments (for Very Good But Inconvenient Reasons, NixOS makes it almost impossible to develop anything without also packaging that something).

In any case, it seems like putting this in this repo isn't the best option, so I'll move it to a dedicated repo elsewhere.

PR #1620, opened 2025-08-29
WIP Named Parameters

phkahler:
HELP WANTED - C++ Language Lawyer please.

This is my latest attempt to integrate named parameters. The first commits adds the text window stuff for creating requests for named parameters. These are unitless free parameters, though their value can be changed in the text window. They can be used in constraint expressions on the sketch.

I've partly implemented a way to handle units. It's a little stupid, and currently broken. I can fix that but...

My current roadblock is way down in Expr::From() the last function in expr.cpp where we create an expression from a string. I've temporarily added code in there to create a list of name/parameter-handle pairs. This is then passed into the ExprParser::Parse function, which assigns them to member variable of the parse so we don't have to keep passing optional parameters all around the parser functions.

The problem is that I need to use the new, commented-out function header to allow these two new variables to be optionally passed in. If they're left off it should behave as before. The idea is to create the parameter dictionary at an even higher level (group level) to enforce proper scope rules. I also need to know in GraphicsWindow::EditControlDone() the number of named parameters that were used in the expression (really just if that's non-zero). That will determine how the expression is stored and is critical to getting units handled correctly.

Anyway, when I used the commented out function header in Expr.cpp and also in Expr.h, I get build errors. More specifically when building the library, which I have little familiarity with.

A bit of help here and I can get the concept done so we can decide if it's good.

How do I get things to build with those two optional parameters in the last function in expr.cpp?

### DIFF
--- a/src/clipboard.cpp
+++ b/src/clipboard.cpp
@@ -244,6 +244,7 @@ void GraphicsWindow::PasteClipboard(Vector trans, double theta, double scale) {
         c.reference = cc->reference;
         c.disp = cc->disp;
         c.comment = cc->comment;
+        c.expression = cc->expression;
         bool dontAddConstraint = false;
         switch(c.type) {
             case Constraint::Type::COMMENT:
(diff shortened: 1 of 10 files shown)


phkahler:
Never mind.  I had an extra Expr:: in the function prototype

PR #1623, opened 2025-09-12
add missing else statement in Slvs_MarkDragged_check

yiman0919:
I notice in the Slvs_MarkDragged_check function, the FatalError function will be called at the end in any case. It is obvious incorrect, because it means once you call this method, the program will quit right now.I add the missing else statement, and only quit when parameter type is incorrect.


iscgar:
Nice catch! Thanks for finding and fixing this!

PR #1629, opened 2025-10-18
ci: fix the macOS CI build

iscgar:
The build is currently failing due to the removal of the Hombrew Ventura bottle for OpenMP, because Ventura is EOL since the release of macOS Tahoe.

The oldest supported macOS version is currently Sonoma. However, it'll be EOL in a year, so for arm64 use Sequoia instead, as most users are likely on it or Tahoe anyway (can't use Sequoia for x86_64, because the only supported version for x86_64 is Sonoma, so we'll likely need to get rid of the x86_64 build in a year, or build without OpenMP support).

While at it, remove the patch for libpng, which doesn't apply and isn't needed after the last update of the dependency.


ruevs:
Thanks @iscgar. Merged.

Interesting whether Sequoia is really not supported on x64 [Wikipedia thinks it is](https://en.wikipedia.org/wiki/MacOS_version_history) and so does [Apple](https://support.apple.com/en-in/120282) . Or do you mean Homebrew? It is harder to figure out what it supports on x64 but it looks to me that [14/Sonoma or higher are supported](https://docs.brew.sh/Installation#macos-requirements).

ruevs:
FWIW my ancient Mojave 10.14.6 in a virtual machine offers me to update to Sequoia when I check for updates.

PR #1631, opened 2025-10-23
WIP: create arc entities from revolved points. The direction of the arcs d…

phkahler:
…oes not flip when the revolve angle changes direction. Not sure how to fix that.

### DIFF
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -617,9 +617,15 @@ void Group::Generate(EntityList *entity, ParamList *param)
                        h.param(1), h.param(2), h.param(3), h.param(4), h.param(5),
                        h.param(6), NO_PARAM, CopyAs::N_ROT_AA);
 
-                // Arcs are not generated for revolve groups, for now, because our current arc
-                // entity is not chiral, and dragging a revolve may break the arc by inverting it.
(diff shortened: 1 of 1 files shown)


phkahler:
> That said (but without ever having looked closely at the <360 degree lathe feature), I'm surprised that swap would be required. An arc entity should always run counterclockwise from its start point to its end point when viewed with its normal facing us. This defines any arc angle in [0, 360) degrees, which seems like what you want. Perhaps the trouble is in the normal?

@jwesthues  The revolve group can be dragged either side of the originating sketch plane. So +270 degrees is not the same as -90 degrees, so the valid range is +/-360.  The angle can technically go way beyond that and is also used in the Helix groups where it can cover many turns with +/- pitch.  Anyway, I tried swapping the point handles and also tried pointing the normal in the opposite direction - both based on the sign of the angle - and neither works here. When I originally made the revolve group I concluded that we'd need an arc entity defined in a different way to create these things, but now I see it's just a matter of getting the existing ones to "go the other way" when needed. The points will not actually be swapped, just their handles withing the arc entity.

ruevs:
> duplicate extrude lines

What do you mean?

PR #1633, opened 2025-10-25
guigtk: make double click handling consistent with Windows and macOS

iscgar:
GTK sends a press event for the 2nd click before synthesising a double click event. This is inconsistent with the GUI code on Windows and macOS and causes double clicks to be interpreted as triple clicks by various receivers.

Fix it by saving the last press event, and eating the current event when it would be considered a double click by GTK and a receiver is set up, since a double click event will follow immediately.

### DIFF
--- a/src/platform/guigtk.cpp
+++ b/src/platform/guigtk.cpp
@@ -21,6 +21,7 @@
 #include <gtkmm/menu.h>
 #include <gtkmm/menubar.h>
 #include <gtkmm/messagedialog.h>
+#include <gtkmm/settings.h>
 #include <gtkmm/scrollbar.h>
 #include <gtkmm/separatormenuitem.h>
 #include <gtkmm/tooltip.h>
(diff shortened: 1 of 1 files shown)


iscgar:
> @iscgar I hate to say it but I don't even know what double clicking does in solvespace, so I can't really comment on how Windows and Linux differ in response to it. Is there some use for double clicks?

I don't think there's any part of the code that handles double clicks in any special way. Rather, the existing code treats the double click event as a regular click AFAICT. The problem with the GTK double click handling is that when a double click occurs, it behaves as if there was a triple click (because the events received are: click, click, double-click; whereas on Windows and macOS the events are: click, double-click). This is especially annoying in my named parameters branch, when I want to quickly create multiple parameters and double-click the "add" button, but it results in three parameters being added.

iscgar:
> [snip]  Some solutions:
> 
>     1. what you're doing with the timer to eliminate the second click event.
> 
>     2. We could just ignore double click events entirely - simpler?
> 
>     3. We could ignore button down events (in the text window) and only respond to release events.
> 
> 
> If your intent is for fast clicking to perform the same operation rapidly, I'd go with option 2 for simplicity since we don't use double click anyway. Or would that create inconsistent behavior between GTK and Windows/mac?

I looked at the code more carefully, and apparently while the text window doesn't handle a double click in any special way, the graphic window does use it for constraint editing. So ignoring the double click event in general is not really possible.

We could theoretically just ignore the double click event in the text window, but because on Windows and macOS the second click is only delivered as a double click event, that would lead to double clicks being interpreted as a single click on those platforms, which is still inconsistent.

So this leaves 1 and 3, and of those I think 1 is the simplest fix, as it simply aligns the behaviour with the other platforms for consistency.

PR #1635, opened 2025-11-10
Name fix

rpavlik:
Looks like I failed to mailmap and stuff here, oops!

Would be good to make sure the metadata file is updated before the next release so we don't have to patch it in flathub too.


ruevs:
This is the first time I've heard about this git feature. Interesting.

https://git-scm.com/docs/gitmailmap

Is it necessary?

PR #1641, opened 2025-12-01
Add Qt GUI for Linux

WickedSmoke:
This replaces the #1451 request and limits the scope to support Linux only.

### DIFF
--- a/src/platform/guiqt.cpp
+++ b/src/platform/guiqt.cpp
@@ -302,7 +302,7 @@ public:
     }
 
     void SetAccelerator(KeyboardEvent accel) override {
-        int key;
+        int key = 0;
 
         if (accel.key == KeyboardEvent::Key::CHARACTER) {
             if (accel.chr == '\t') {
@@ -318,13 +318,15 @@ public:
             key = Qt::Key_F1 + accel.num - 1;
         }
 
-        if (accel.controlDown) {
-            key |= Qt::CTRL;
+        if (key) {
+            if (accel.controlDown) {
+                key |= Qt::CTRL;
+            }
+            if (accel.shiftDown) {
+                key |= Qt::SHIFT;
+            }
+            action.setShortcut(key);
         }
-        if (accel.shiftDown) {
-            key |= Qt::SHIFT;
-        }
-        action.setShortcut(key);
     }
 
     /*


WickedSmoke:
> @WickedSmoke Why so much code in a header file? qglwindow.h

Not sure what you're referring to.  Raed's header was qglmainwindow.h and that was converted to guiqt.h.


> @WickedSmoke you made an RPM package. Is there any hope of getting that into copr or even Fedora repository?

My copr is [wickedsmoke/SolveSpace](https://copr.fedorainfracloud.org/coprs/wickedsmoke/SolveSpace/) but there are some hacks to get that working.  The Source URLs need to pull from Github archives (including submodules).  I would prefer to build with the system `mimalloc` to avoid that submodule.

phkahler:
@WickedSmoke  I pulled this branch. How do I tell it to build with qt?

WickedSmoke:
Here's the BuildRequires line from [my .spec file](https://github.com/WickedSmoke/solvespace_qt5/commit/b51f1a4f1ea):

    BuildRequires: g++, cmake, qt6-qtbase-devel, eigen3-devel, fontconfig-devel, freetype-devel, libpng-devel, zlib-devel, mesa-libGL-devel mesa-libGLU-devel

My [Releases page](https://github.com/WickedSmoke/solvespace_qt5/releases) links to a Fedora 43 package built last night on Copr.

vthriller:
> The text window became unmovable for me when undocked on gnome

This problem is somewhere between Qt and Wayland compositors actually. It is reproducible in sway with or without `QT_QPA_PLATFORM=xcb`, but not in proper X11 (e.g. Xephyr), and is relevant not just to SolveSpace but to other software like LibreCAD as well.

WickedSmoke:
Qt only [officially supports Linux/X11](https://doc.qt.io/qt-6/supported-platforms.html).  The QDockWidget bugs are a Wayland issue and it will be good for programs to be out in the wild that exhibit them so that they get attention.

I prefer giving users options so I was pleased to introduce the use of a docking widget.

phkahler:
@WickedSmoke  and  @shingen75   Thank you for this excellent contribution to Solvespace.

PR #1645, opened 2025-12-06
CMakeLists: Rename Qt binary to solvespace-qt.

WickedSmoke:


WickedSmoke:
A solvespace-qt.desktop file is now built using the same template.

There will still be conflicts with the icons and such if a Gtk package were installed, but I'm OK with that for this pull request.

PR #1655, opened 2025-12-25
Fix: macOS points not draggable

7408647n:
On macOS 26, interactive dragging could appear “stuck” because the one-shot NSTimer was registered only for the run loop’s default mode. During mouse drags AppKit runs the main run loop in an event-tracking mode, where default-mode timers don’t fire. Register the timer in NSRunLoopCommonModes so it continues firing during tracking and interactive updates keep running.

### DIFF
--- a/src/platform/guimac.mm
+++ b/src/platform/guimac.mm
@@ -176,8 +176,9 @@ public:
         if(timer != NULL) {
             [timer invalidate];
         }
-        timer = [NSTimer scheduledTimerWithTimeInterval:(milliseconds / 1000.0)
+        timer = [NSTimer timerWithTimeInterval:(milliseconds / 1000.0)
             invocation:invocation repeats:NO];
+        [[NSRunLoop currentRunLoop] addTimer:timer forMode:NSRunLoopCommonModes];
     }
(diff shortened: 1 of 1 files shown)


ruevs:
@7408647n is this behaviour change of AppKit in macOS 26/Tahoe documented somewhere or did you figure it out by debugging?

phkahler:
It sounds like we're using a timer to trigger a refresh/repaint or similar. There's nothing really time related in a click and drag scenario. Might there be a different event we should be using?

BTW im ok changing the timer stuff, just questioning why its used at all for this.

phkahler:
@7408647n  and @vespakoen  any idea why the macOS CI builds are failing (again)? It'd be nice to have a nightly produced when this is merged so everyone can easily try it on mac.  Also so we can push out version 3.2 final once this is in.

7408647n:
> @7408647n and @vespakoen any idea why the macOS CI builds are failing (again)?

I created a commit #1658

ruevs:
This works (does no cause a regression) on my macOS 10.14.6 (Mojave) VM.

Issue #389, opened 2019-03-26
Feature - New Group: Helix


phkahler:
Several people have wanted the Lathe operation to limit the angle of revolution to less than 360 degrees. This feature would work like Lathe but would allow sweeping the sketch though arbitrary angles while also sliding it along the axis of revolution to produce helical objects. Because NURBS can not exactly represent helical surfaces (the angular vs axial position is not perfectly linear) there should be adjustable number of segments to allow the user to minimize that distortion. When the axial displacement is 0.0 this will produce the same geometry as Lathe but with adjustable sweep angle. All surfaces created this way would be pure NURBS without doing boolean operations.

Evil-Spirit:
This is excactly how I made it for NoteCAD.
![Unity 2018 3 2f1 Personal - NoteCAD unity - NoteCAD - WebGL_ _DX11 on DX9 GPU_ 2019-03-27 13 27 34](https://user-images.githubusercontent.com/1980537/55054950-1fc69280-5094-11e9-965c-832333b27e1e.png)

phkahler:
I started working on this last night and have some questions about internals.

Getting the surface patches (control points) correct was easy but the trim curves are tricky. I've copied SShell::MakeFromRevolutionOf() as a starting point for SShell::MakeFromHelicalRevolutionOf(). The part where trim curves are created is very confusing due to me not understanding all the requirements of the internal data structures. Is there any explanation of what Remap is doing? Would it be OK to trim a surface patch using curves constructed from the same control points that define the surface? That would seem to be an easy function to implement for surface patches that don't have holes. Is there some kind of linkage needed between patches that are part of the same surface? Are the curves reused just to save memory or is it important that shared edges not be replicated?

From a code point of view I also copied SSurface::FromRevolutionOf() to SSurface::FromHelicalRevolutionOf and include two additional distance parameters to specify how far along the axis the start and end of the revolved surface are. Should this stay a separate function or should older code use the new version and pass zeros for these new parameters? It should produce the exact same results in that case with less new code.

Evil-Spirit:
Only @jwesthues can answer this questions.

jwesthues:
[...]

PR #1657, opened 2025-12-26
Enable face dragging for translate groups. Add the necessary handles …

### LINKED ISSUE #1113: SketchUp-like moving by grab faces of solid
### System information

- **Operating system:** all supported

### Expected behavior

Allow moving linked solid parts by selected face from this solid group (moving should follow orthogonal to selected flat face).

Also, allow extend "Extrude"/"Lathe"/"Helix" by grabbing face (based on previous sketch).

### Actual behavior

Actually inside assembly its possible to move linked solid parts *only* by selected *entities* (points, lines, origins, ) from linked parts.

Same limitation is for "Exrude"/"Lathe"/"Helix" operations.

### Additional information

Actually this may be limited just to selectable flat faces.

**NOTE:** More details and possible solution TBA.

phkahler:
…for face dragging linked sketches, but don't enable that feature because existing sketches will crash when dragged. In mouse.cpp StartDraggingByEntity needs some face types uncommented to enable different face dragging options. All rotational movements are disabled because the points used for those faces are on-axis(I think), this needs to be changed for those to be more useful.


phkahler:
I'd like to merge this prior to 3.2 so people will start creating files with point handles in the face entities. That is a prerequisite for dragging linked objects by their faces - which works fine but is currently disabled to prevent crashes.

Oh, maybe we can check for a valid handle and only drag if one is present.

phkahler:
Looks like I need to update the files in 4 tests because they are no longer identicle. This adds a point handle to some faces.

ruevs:
@phkahler to summarize your work on dragging faces so far:

1. https://github.com/phkahler/solvespace/tree/face-drag The previous attempt from Nov 30, 2025 with some debugging stuff from 2023 and the beginning of the story.)
2. https://github.com/phkahler/solvespace/tree/drag-faces The one you already merged here https://github.com/solvespace/solvespace/commit/e3f21de93d2d8a1dd254292c196f53eaae2ba4a5. It works and is non-intrusive - I like it. The branch should be deleted)
3. https://github.com/phkahler/solvespace/tree/faces This PR

So back to this PR:

- Adding the extra `en.point[0]` is a bit suspect to me. I need to think more.
- You wrote in the first commit message "Add the necessary handles for face dragging linked sketches" - but linked **2**d sketches are not draggable by their "face" - not surprising but the commit message is misleading.
- Dragging linked 3d files created with "old" versions by the faces works with this version. Why?! (I did not debug)
- Being able to drag by the faces is not important to me at all - in fact I keep the face selection disabled most of the time faces

Overall I think this is not ready.

phkahler:
> @phkahler to summarize your work on dragging faces so far:
> 
>     1. https://github.com/phkahler/solvespace/tree/face-drag The previous attempt from Nov 30, 2025 with some debugging stuff from 2023 and the beginning of the story.)
> 
>     2. https://github.com/phkahler/solvespace/tree/drag-faces The one you already merged here [e3f21de](https://github.com/solvespace/solvespace/commit/e3f21de93d2d8a1dd254292c196f53eaae2ba4a5). It works and is non-intrusive - I like it. The branch should be deleted)

Branch deleted, thanks!

>     3. https://github.com/phkahler/solvespace/tree/faces This PR
> 
> So back to this PR:
> 
>     * Adding the extra `en.point[0]` is a bit suspect to me. I need to think more.

When we drag a line, we add the end points to the drag list. It makes sense for dragging faces to add their point to the drag list. The problem is that when linking, all entites are copied with copy type N_ROT_TRANS, which creates a new entity using only numeric values (no free parameters). You can move them around because it's possible to get expressions for the points [which are numeric values modified by the group parameters](https://github.com/solvespace/solvespace/blob/fdb80e3ea7a47fbfe73fce308fd9d65faab5aa58/src/entity.cpp#L786). Some [face types just return the expressions for their underlying point](https://github.com/solvespace/solvespace/blob/fdb80e3ea7a47fbfe73fce308fd9d65faab5aa58/src/entity.cpp#L778) which is probably what they all should do.

The problem: Some faces never set the entity handle for their point, which is clearly an oversite. This was never a problem because nothing ever tried to access that point by its handle - like adding it to the drag list.

>     * You wrote in the first commit message "Add the necessary handles for face dragging linked sketches" - but linked **2**d sketches are not draggable by their "face" - not surprising but the commit message is misleading.

2d sketches do not have a "face" entity. Faces are defined by a plane and can be selected by the corresponding triangles in the mesh - yes,  [every triangle has an entity handle!](https://github.com/solvespace/solvespace/blob/fdb80e3ea7a47fbfe73fce308fd9d65faab5aa58/src/polygon.h#L153).

Fun fact. When I wrote the very first Helix code and posted [this picture](https://github.com/solvespace/solvespace/issues/389#issuecomment-477828837) the upper surface would highlight with mouse-over, but it would crash if clicked because there was no actual face entity. The triangles were getting properly tagged because of some code I copied from somewhere. That was a *big* surprise to me because it's a twisted surface - not flat at all.

>     * Dragging linked 3d files created with "old" versions by the faces works with this version. Why?! (I did not debug)

My poor description. Trying to drag old linked files will not crash, it just fails to drag. This is because I added the check for a valid handle before using the face-associated point. It does not "work", it just doesn't crash.

>     * Being able to drag by the faces is not important to me at all - in fact I keep the face selection disabled most of the time faces

Faces tend to light up often so it makes sense to turn them off. You are free to leave them turned off ;-) But when they are turned on it makes sense that they be useful and even draggable. I think this is really important for linked sketches where the main thing we do is position them relative to other parts in an assembly.

> Overall I think this is not ready.

I think it is, but maybe I should update the commit message claiming linking old files work. Edit: Oh it's correct already.

phkahler:
@ruevs Ping!!  And Happy New Year!!

src/group.cpp
```diff
@@ -899,6 +899,10 @@ void Group::MakeExtrusionLines(EntityList *el, hEntity in) {
         en.param[1] = h.param(1);
         en.param[2] = h.param(2);
         en.numPoint = a;
+        // might use the original point, not the remapped one - but surface face uses TOP
+        // using the original point might allow face dragging the sides unconstrained extrusions
```
ruevs: This is the part that I found "suspect" as well. I do not understand the remapping well.
@jwesthues what do you think? (Happy New Yewr!).

src/group.cpp
```diff
@@ -1184,6 +1188,7 @@ void Group::CopyEntity(EntityList *el,
             }
             en.numPoint  = (ep->actPoint).ScaledBy(scale);
             en.numNormal = (ep->actNormal).ScaledBy(scale);
+            en.point[0] = Remap(ep->point[0], remap);
```
ruevs: Same as above.

ruevs (changes requested):

ruevs (changes requested):
Squash 1d709cff309995ab474c2029b3c7f7f6f68d88b3 and e4556fc0f9698bef18db94ad05766b5466fad5bb together.

src/group.cpp
jwesthues: Happy New Year! Looking at `EntityBase::FaceGetPointExprs()`, I see that `FACE_NORMAL_PT` and `FACE_ROT_NORMAL_PT` both simply return `point[0]`, while the others calculate the point from `numPoint`.  I think I did that to simplify the generation of the others. I believe suitable point entities do always exist in a real sketch, but I probably thought it would be simpler to repeat the transform calculation than to find that suitably transformed point.

This change seems to keep that behavior in `FaceGetPointExprs()`, but sometimes additionally set `point[0]` to a suitable point to drag. So `point[0]` provides both definition and dragging for `FACE_NORMAL_PT` and `FACE_ROT_NORMAL_PT`, but only dragging for others. That's inconsistent, though the use of those members is generally inconsistent between different entity types so that seems fine to me. It perhaps deserves a comment explicitly noting that inconsistent overloading.

phkahler:
@ruevs squashed. Some comments cleaned/removed. And I did opt to use the other point for extrusion sides - no idea why it won't drag a rectangle from a previous group by using that, but it will never do it with a remapped point.

src/group.cpp
jwesthues: @phkahler I think you'd have to keep the old codepath to support linked files saved by earlier versions. So I definitely wouldn't change `FaceGetPointXXX()` now, and probably not ever.

ruevs:
> @ruevs squashed.

But you still merged three commits ;-) At least 1cec22f5b6eed88cf20593dc738b045b9cde89c0 and 457aec977f238d14bd15d6d288f12c72f8ef3888 should have been squashed. Ideally also 2ee0cc6bd6d490212afe6e1a8cd70bc3e083b96f.

But it's OK... it's very unlikely someone will ever bisect is such a way that this is a problem.

src/group.cpp
ruevs: > @ruevs I thought I finally understood remapping. Went to verify a few details, and now... It does what I thought, but im not sure how exactly. There's a step that seems to be missing, and the mechanism is IMHO rather broken. But it seems to work anyway. More on that later.

I still don't understand it well enough because I have looked through this part of the code only a few times, superficially. An I've never tried to modify it or debug it to understand better. I should...

But the discussion between you and Jonathan above helped me a lot.

PR #1658, opened 2025-12-27
Fix: macOS test action not work

7408647n:
When run with "ci“, the script now fetches the libomp bottles for arm64_tahoe and sonoma (on my research, [x86]tahoe doesn't exist), locates them in the Homebrew cache, copies them to /tmp, and extracts them.

### DIFF
--- a/.github/scripts/install-macos.sh
+++ b/.github/scripts/install-macos.sh
@@ -1,12 +1,18 @@
 #!/bin/sh -xe
+set -o pipefail
 
 if [ "$1" = "ci" ]; then
-    armloc=$(brew fetch --bottle-tag=arm64_sequoia libomp | grep -i downloaded | grep tar.gz | cut -f2 -d:)
-    x64loc=$(brew fetch --bottle-tag=sequoia libomp | grep -i downloaded | grep tar.gz | cut -f2 -d:)
-    cp $armloc /tmp/libomp-arm64.tar.gz
+    brew_cache=$(brew --cache)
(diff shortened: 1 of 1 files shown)


7408647n:
I tested this change on my Mac, but it should be checked again to make sure it works properly as a GitHub action.

ruevs:
In my opinion there are many, many more Intel/x64 Macs compared to ARM Macs in the world (outside of California :-) ). Therefore we should be building the x64 version in the CI and _not_ the ARM version.

Therefore if the missing x64 `libomp` "bottle" under Tahoe is the problem I would propose that we change this line:

https://github.com/solvespace/solvespace/blob/de2c5feec8824dd988c3ed382795031a58de8a5c/.github/workflows/cd.yml#L48

To point to macOS 15/Sequoia... or whatever available CI runner, where the stupid `brew` has the x64 library.

Overall macOS has been a major PITA. They don't care about backward compatibility at all (compared to Windows and Linux). Just look at this mess: https://github.com/solvespace/solvespace/issues?q=state%3Aclosed%20label%3AmacOS

And this includes the open source `brew`. The stupid Tahoe is supposed to support x64 Macs:
https://en.wikipedia.org/wiki/MacOS#Timeline_of_releases

Then why the hell did they remove the OpenMP x64 "bottle"?!?

And we just discussed this exact topic literally two months ago! https://github.com/solvespace/solvespace/pull/1630

Or am I missing something?

I used to create issues every time the macOS build/CI stopped working but I am fed up with their endless BS by now so I stopped doing it.

Issue #1660, opened 2025-12-28
macOS CI is broken again


ruevs:
https://github.com/solvespace/solvespace/actions/runs/20549355428/job/59025095410#step:5:62

More details here https://github.com/solvespace/solvespace/pull/1658#issuecomment-3694920934

vespakoen:
It looks like the cert is gone (expired probably) i'll make a new one in a couple of days, sorry for the wait, now new year release =(

Happy new year everyone!!!

PR #1661, opened 2025-12-28
Fix: macOS sign action not work

7408647n:
Sets the user's keychain search to "build.keychain" so codesign/security can find the imported cert

Note: both MACOS_APPSTORE_APP_PASSWORD and MACOS_APPSTORE_USERNAME must be set.


ruevs:
Fixed here https://github.com/solvespace/solvespace/issues/1660#issuecomment-3706019164

PR #1662, opened 2025-12-31
QT GUI on Windows

ruevs:
For cross reference see:
https://github.com/ruevs/solvespace/commit/ad9e8fa05813783b295f8bc8cfeb2c6d95252f5a#diff-62f364215497b1852e5bb659ce16d89f3cf36b77dc49cb2774103b92548054e8

Add:
CMAKE_PREFIX_PATH = [path to your Qt]/Qt/Qt5.12.12/5.12.12/msvc2017_64/lib/cmake/Qt5
to CMake to make it work.

##   QT, Win32: The QT GUI now compiles on Windows and loading/saving works...
    
    Fix compilation on Windows
      QT5: Compiles and works, but the property browser window is broken.
      QT6: Compiles but crashes.
    
    Fix file opening and saving on Windows by using `QDir::toNativeSeparators()`
    in `Platform::Path GetFilename()` to retrieve the path returned by
    `QFileDialog` in a platform-independent way and normalize the file
    separators (`/` for Unix-like platforms and `\` for Windows).
    
    Here’s how the returned path differs by platform:
    
    1. **Unix/Linux/macOS**:
       - File path: `"/home/user/example/file.slvs"`
       - Result of `QDir::toNativeSeparators()`: `"/home/user/example/file.slvs"`
    
    2. **Windows**:
       - File path: `"C:/Users/user/example/file.slvs"`
       - Result of `QDir::toNativeSeparators()`: `"C:\Users\user\example\file.slvs"`

## CMake: fix CMake system include directives
    
    The saame as 7377c5003d2906b079558bbff95a8b9f0364615e but for the newly
    added `USE_QT_GUI` section:
    
        The CMake `SYSTEM` modifier of `target_include_directories()` must come
        before the visibility specification. However, it was accidentally in the
        wrong order, so it was simply adding a directory named `SYSTEM` to the
        target's include paths.
    
        Fix it to correctly request the include paths to be treated as system
        include paths.

### DIFF
--- a/src/platform/guiqt.cpp
+++ b/src/platform/guiqt.cpp
@@ -27,6 +27,9 @@
 #include <QSettings>
 #include <QTimer>
 
+#ifdef WIN32
+#    include <windows.h>
+#endif
 
 namespace SolveSpace {
 namespace Platform {
@@ -159,7 +162,7 @@ public:
 
     Platform::Path GetFilename() {
         QStringList files = fileDialogQ.selectedFiles();
-        return Path::From(files.at(0).toStdString());
+        return Path::From(QDir::toNativeSeparators(files[0]).toStdString());
     }
 
     void SetFilename(Platform::Path path) {
@@ -935,6 +938,11 @@ public:
     }
 };
 
+#ifdef WIN32
+// This interferes with our identifier.
+#    undef CreateWindow
+#endif
+
 WindowRef CreateWindow(Window::Kind kind, WindowRef parentWindow) {
     return std::make_shared<WindowImplQt>(kind,
                 std::static_pointer_cast<WindowImplQt>(parentWindow));
@@ -1044,6 +1052,7 @@ int main(int argc, char** argv) {
     // See https://bugreports.qt.io/browse/QTBUG-89812
     QCoreApplication::setAttribute(Qt::AA_ShareOpenGLContexts, true);
 
+#ifndef WIN32   // Allows using OpenGL 1 on Windows
     // Specify the GL version so Qt doesn't use GLES on Wayland, which
     // doesn't handle "#version 120" shaders.
     {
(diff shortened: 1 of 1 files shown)


src/platform/guiqt.cpp
```diff
@@ -27,6 +27,9 @@
 #include <QSettings>
 #include <QTimer>
 
+#if WIN32
```
WickedSmoke: I'd test the macro in a consistent manner (e.g. #ifdef/#ifndef) throughout the file.

src/platform/guiqt.cpp
```diff
@@ -968,7 +976,6 @@ void Request3DConnexionEventsForWindow(WindowRef window) {}
 //-----------------------------------------------------------------------------
 // Application-wide APIs
 //-----------------------------------------------------------------------------
-
```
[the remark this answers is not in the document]
ruevs: Good point. Fixed.

src/platform/guiqt.cpp
```diff
@@ -27,6 +27,9 @@
 #include <QSettings>
 #include <QTimer>
 
+#ifdef WIN32
+#    include <windows.h>
```
WickedSmoke: What symbol is windows.h providing?

ruevs: Essentially allows this:

https://github.com/solvespace/solvespace/blob/fdb80e3ea7a47fbfe73fce308fd9d65faab5aa58/src/platform/guiqt.cpp#L973

to compile.

phkahler:
@ruevs do you want me to check the QT build still works on Linux? The CI builds OK with GTK. Or should I check specific functionality? On GTK or QT?

ruevs:
> @ruevs do you want me to check the QT build still works on Linux? The CI builds OK with GTK. Or should I check specific functionality? On GTK or QT?

@phkahler  Just the QT build on Linux. As @WickedSmoke already said - it is very, very unlikely that this breaks it, but just in case.

WickedSmoke:
@phkahler Yes, only the single built-in font is available.  I've created #1663 to fix this, but note that a few lines overlap with the changes here.

PR #1663, opened 2025-12-31
Qt: Enable fontconfig use.

WickedSmoke:
There are a couple annoying things I see after enabling the font detection.

First, the list in Properties is extremely long as each individual file is shown rather than each font family (as in a typical application).  There are over 560 files on my system.  A combo box widget would really be helpful.

The second issue is that there are now about 100 freetype warnings printed to the console like this:

```
freetype: CID-to-GID mapping for CID 0x0041 in file '/usr/share/fonts/google-noto-vf/NotoSansMeeteiMayek[wght].ttf' failed: no error; using CID as GID
Assuming cap height is the same as requested height (this is likely wrong).
```

### DIFF
--- a/src/platform/guiqt.cpp
+++ b/src/platform/guiqt.cpp
@@ -29,6 +29,8 @@
 
 #ifdef WIN32
 #    include <windows.h>
+#else
+#    include <fontconfig/fontconfig.h>
 #endif
 
 namespace SolveSpace {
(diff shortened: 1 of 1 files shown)


ruevs:
> The second issue is that there are now about 100 freetype warnings printed to the console like this:
> 
> ```
> freetype: CID-to-GID mapping for CID 0x0041 in file '/usr/share/fonts/google-noto-vf/NotoSansMeeteiMayek[wght].ttf' failed: no error; using CID as GID
> Assuming cap height is the same as requested height (this is likely wrong).
> ```

This happens for all fonts that do not have 'A'. For example on my machine `C:\windows\fonts\SegoeIcons.ttf`. This is completely normal. The code is here:

https://github.com/solvespace/solvespace/blob/47f1b0794b30f664fcc4b16f0269c9046bbf809f/src/ttf.cpp#L235

PR #1664, opened 2026-01-01
Update the year to 2026 in the About dialog and Windows version resou…

ruevs:
…rce.


phkahler:
@ruevs we should have a document with all these little tips and procedures. Years ago whitequark had a script to update all the images in the tests unite after some changes altered a bunch of them. There are other things too I cant think of right now.

Issue #1668, opened 2026-01-03
Qt build crashes on opening assembly: assertion failed in FindById


dadoprom:
I’ve just compiled SolveSpace from the latest commits and the Qt version crashes when opening my assembly project.

On startup of the project, SolveSpace-Qt shows this error message:

> File ./src/dsc.h, line 506, function FindById:
> Assertion failed: t != nullptr.
> Message: Cannot find handle.

After clicking OK, the application immediately closes.

Important notes:
The same project opens and works correctly in the GTK version.
It also worked in a previous Qt build (until the latest updates).
The crash happens only when opening this assembly.

Let me know if you need the project file.

> I compiled with this command:
> cmake .. -DCMAKE_BUILD_TYPE=Release -DENABLE_OPENMP=ON -DENABLE_LTO=ON -DUSE_QT_GUI=ON

phkahler:
@dadoprom  does it work if you rebuild the gtk version from the same source, or is that an older gtk version that works?

wolfpackmars2:
I'm getting the same error.

SolveSpace-Qt 3.2~e74c2eae - Linux Ubuntu 24.04

I have some sample files coming - I haven't traced the exact cause yet, but I am suspecting it is either a problem with Revolve extrusions or Copy Rotating of Extrusions.

This doesn't happen with all files, as I have created several files with this version without issue. Just the file that takes the longest to sketch out... After the 3rd time making the same model, I finally got smart and documented all the dimensions via PNG.

wolfpackmars2:
[AssertionError.zip](https://github.com/user-attachments/files/25028379/AssertionError.zip)

This zip contains 3 copies of the same file that were saved at various stages of edit. All 3 have the same error when opening. Included PNGs of the final dimensions. The problem only occurs when opening certain files, like the 3 slvs attached.

Tangs.slvs contains only the single revolve extrusion, prior to duplicating the object. The first edit file is after Step Rotating the tang. The second edit file is after I realized I made a mistake on a dimension and then saved it separately.

phkahler:
@dadoprom  I merged that fix to master, so you should be all set. Not going to close this issue yet because it sprouted another problem with these other files.

@wolfpackmars2  I'm a little confused. Your zip file contains 3 slvs files that all crash when opened with the latest software. OK (confirmed). You indicate that they can be opened with an older version from November. OK (I have no reason to doubt). So my questions:

What version of Solvespace were these 3 files created with?

[...]

PR #1671, opened 2026-02-03
When extruding a line to face, use a point in new group for the face.…

phkahler:
… Using one from the previous group causes subsequent translate groups to have undraggable faces and possibly other problems.


ruevs:
The good thing is [we have users using edge builds or compiling](https://github.com/solvespace/solvespace/issues/1668) and they caught it :-)

I regret [not thinking more](https://github.com/solvespace/solvespace/pull/1657#issuecomment-3694905764) last time... it was just a hunch but I could not immediately rationalize it...

PR #1673, opened 2026-02-04
Update README install and build guides

### LINKED ISSUE #1653: Linux: Update metainfo for SolveSpace after install
after `make install` when building and installing from source, the following two actions need to be performed for the OS (Linux Mint, probably other Debuntus) to load the file associations for thumbnailing and opening .slvs files in the desktop environment.

```bash
# Update application metainfo so user can open .slvs files natively (by double click) in file manager
$ sudo update-desktop-database /usr/local/share/applications/
```

```bash
# Update mime data for .slvs files to be properly recognized; enables thumbnailer to work
sudo update-mime-database /usr/local/share/mime/
```

As an aside, SolveSpace is the only application on my system which uses /usr/local/share/ for these files. However, my research has shown that this is not wrong and may actually be most correct to spec.

On my system (Linux Mint) `/usr/share/thumbnailers` is what is used by default for thumbnailers and `/usr/share/applications` for the mime type information.

This seems to be a problem/feature the cmake files. I have only some experience with make - very little with actually writing or modifying the make system. I also acknowledge that cmake/make is not meant to handle these types of operations - such as updating system associations after an install.

anyone experienced with cmake to weigh in

### LINKED ISSUE #1654: Investigate sudo requirement for make install
Does SS need to be installed with sudo privileges on *nix?

I believe it isn't necessary, and if this is the case, the cmake files can be updated to choose the install paths by default based on the user context it is run under.

EDIT: Removed comments not relevant to the specific request.

Related:  #1653

wolfpackmars2:
Added instructions for installing via unofficial distro packages and additional information when building from source for various platforms.

Changed header from "Via Source Code" to "Getting the source code". Added a target to not break existing links.

Fix other small typos and grammar.

Resolves #1654 and resolves #1653

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,15 @@ automatically built by the SolveSpace maintainers for each stable release.
 
 [rel]: https://github.com/solvespace/solvespace/releases
 
+### Via _Unofficial_ Distro Packages
+
+Some Linux distributions may provide packages for install from the distro
+specific app repositories. For example, SolveSpace can be installed on
+Debian (and derivatives like Ubuntu and Mint) with `sudo apt install solvespace`.
+While these packages may be behind the current release and will not include the
+latest features in development, the system packages offer a simple alternative for
+installing SolveSpace which should be stable for your operating system.
+
 ### Via Flathub
 
 Official releases can be installed as a Flatpak from Flathub.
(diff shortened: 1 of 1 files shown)


wolfpackmars2:
I apologize for my presumptuousness.

PR #1674, opened 2026-02-04
Modify GNU thumbnailer

wolfpackmars2:
This edit aligns the thumbnailer feature with PR #1673. Specifically, it modifies the solvespace.thumbnail file to point to the installed executable. This means that if a user installs solvespace to ~/.local, the thumbnailer will work for that user.

This is necessary because ~/.local/bin is not normally included in the default PATH environment variable.

### DIFF
--- a/res/freedesktop/solvespace.thumbnailer.in
+++ b/res/freedesktop/solvespace.thumbnailer.in
@@ -1,4 +1,4 @@
 [Thumbnailer Entry]
-TryExec=solvespace-cli
-Exec=solvespace-cli thumbnail --view isometric --size %sx%s --output %o %i
+TryExec=${CMAKE_INSTALL_FULL_BINDIR}/solvespace-cli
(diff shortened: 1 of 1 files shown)


wolfpackmars2:
This PR aligns the .thumbnailer with the .desktop file.

> I just don't like the idea of saving a file from email or off the internet and having software open it by default.

Is SS so insecure? I am confused by this. So don't use thumbnails then. Every thumbnail operates the same way. I'm so confused I don't even know where to start.

> I would never run `make install` if compiling from source

The alternative being... software compiled by others and installed with root privileges... ?

phkahler:
> Is SS so insecure? I am confused by this. So don't use thumbnails then. Every thumbnail operates the same way. I'm so confused I don't even know where to start.

No, thumbnails do not all operate the same way. You're asking Solvespace to run every time the OS creates a thumbnail for a .slvs file. No thumbnails do not operate the same way - other file types don't run Solvespace to make a thumbnail. I can not speak to the security of Solvespace so I won't encourage this. It is also an issue for packagers, not upstream.

PR #1688, opened 2026-03-15
generate: make GenerateAll non recursive and optimise pruning logic

iscgar:
This results in a reduction of up to 50% in regeneration duration since it only prunes sequentially. Pruning should be further optimised so that it avoid traversing the full request and constraint list for every group again.

I had this change sitting in my fork for a few months alreay and just forgot to submit it. Sorry for the delay.

### DIFF
--- a/src/generate.cpp
+++ b/src/generate.cpp
@@ -31,35 +31,20 @@ void SolveSpaceUI::MarkGroupDirty(hGroup hg, bool onlyThis) {
     ScheduleGenerateAll();
 }
 
-bool SolveSpaceUI::PruneOrphans() {
-    const int requests = SK.request.n;
-    for(Request &r : SK.request) {
-        if(!GroupExists(r.group))
-            r.tag = 1;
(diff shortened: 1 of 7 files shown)


phkahler:
@iscgar   I'm in favor of this change, and now is a good time to make it - after a release, not just before. @ruevs what do you think?

ruevs:
@phkahler I have not looked at this PR closely - sorry.  I'll try to do it next week.

Evil-Spirit:
> This results in a reduction of up to 50% in regeneration duration since it only prunes sequentially. Pruning should be further optimised so that it avoid traversing the full request and constraint list for every group again.
> 
> I had this change sitting in my fork for a few months alreay and just forgot to submit it. Sorry for the delay.

Good Idea since I was not very impressed by double regeneration (but it was me who did it). 
But it is better to make your changes separated (optimize pruning and removing double regeneration for chord tolerance calculation should be different commits.)

PR #1695, opened 2026-03-25
Windows CI path fixes

vespakoen:
Ah damn, pushed it to the solvespace repo instead of my fork...
Anyways, I think this should fix the path issue for the Windows release.

I wonder if it makes sense to add a "_32" (bit) postfix or something because now I think some people might prefer downloading `solvespace.exe` over `solvespace_x64.exe` as `solvespace.exe` looks more like the "main one"?

It also looks like [actions/upload-release-asset](https://github.com/actions/upload-release-asset) is not maintained anymore and we should probably migrate it to something newer. (apparently [this](https://github.com/softprops/action-gh-release))

### DIFF
--- a/.github/scripts/build-windows.sh
+++ b/.github/scripts/build-windows.sh
@@ -48,6 +48,8 @@ if [ "$3" = "x64" ]; then
 	fi
 else
 	if [ "$2" != "openmp" ]; then
-		mv bin/$BUILD_TYPE/solvespace.exe bin/$BUILD_TYPE/solvespace_single_core.exe
+		mv bin/$BUILD_TYPE/solvespace.exe bin/$BUILD_TYPE/solvespace_single_core_x86.exe
+    else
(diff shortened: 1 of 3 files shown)


vespakoen:
@phkahler ok, I downloaded the binaries from the 3.2 run [from here](https://github.com/solvespace/solvespace/actions/runs/23559277610), which succeeded except for uploading the binaries to the release.
I attached them to the 3.2 release manually.

I also created a "test_release" release to see if these changes worked, that is still running over here: https://github.com/solvespace/solvespace/actions/runs/23619924004

I will remove that release and tag (test_release) once it's done.

vespakoen:
It worked except for 2 missing single core files, for which I added a fix in #1698 (together with some other version bumps)

The source tarball run also failed because the tag has to start with a `v` for it to work.

This is what the release looked like:

Screenshot 2026-03-26 at 23 17 40

PR #1707, opened 2026-04-01
fix copy paste error in file.cpp

g19fanatic:
There is a copy paste error in file.cpp.

Simple quick fix.


phkahler:
I searched group.cpp  generate.cpp and textscreens.cpp and see no reference to it. Must currently be unused so I'll just merge this so it doesn't confuse us if we do use valC some day.

PR #1714, opened 2026-04-08
Change macOS shortcuts that use Function keys to use Command

iffy:
On the standard macOS wireless keyboard, the function keys are a little annoying to press (needs 2 hands). This changes shortcuts keys to use Command instead of Function. I'm not offended if you reject this.


ruevs:
In its present form - a definite no from me as well. I totally agree with what @phkahler wrote above.

This should _only_ modify `guimac.mm`.

PR #1724, opened 2026-06-17
QT: Add missing file extension on sketch save

### LINKED ISSUE #1689: QT: Solvespace not appending .slvs file extension by default
The current QT build of Solvespace is not adding .slvs to the end of the file name by default when saving a sketch.

This is compounded when trying to open a sketch, as the solvespace interface doesn't allow any file filter other than the .slvs extension.

The user should at least have the option to select "All Files" available as a fall back for the open file dialog box.

wolfpackmars2:
This fix should append .slvs to sketch files, regardless of the UI used, if no file extension was given.

If a different file extension exists, it will be retained and .slvs will not be added.

Resolves #1689

Tested in Ubuntu Noble for Qt and GTK interfaces. That is, it fixes the Qt problem and doesn't break the GTK interface.

### DIFF
--- a/src/solvespace.cpp
+++ b/src/solvespace.cpp
@@ -574,6 +574,15 @@ void SolveSpaceUI::AddToRecentList(const Platform::Path &filename) {
     GW.PopulateRecentFiles();
 }
 
+static Platform::Path EnsureSketchExtension(Platform::Path filename) {
+    std::string basename = filename.FileName();
+    if(basename.find('.') == std::string::npos) {
+        filename.raw += ".";
+        filename.raw += SolveSpaceUI::SKETCH_EXT;
+    }
+    return filename;
+}
+
 bool SolveSpaceUI::GetFilenameAndSave(bool saveAs) {
     Platform::SettingsRef settings = Platform::GetSettings();
     Platform::Path newSaveFile = saveFile;
@@ -594,6 +603,7 @@ bool SolveSpaceUI::GetFilenameAndSave(bool saveAs) {
             dbp("Calling FreezeChoices()...");
             dialog->FreezeChoices(settings, "Sketch");
             newSaveFile = dialog->GetFilename();
+            newSaveFile = EnsureSketchExtension(newSaveFile);
         } else {
             return false;
         }
--- a/src/solvespace.cpp
+++ b/src/solvespace.cpp
@@ -574,13 +574,37 @@ void SolveSpaceUI::AddToRecentList(const Platform::Path &filename) {
     GW.PopulateRecentFiles();
 }
 
-static Platform::Path EnsureSketchExtension(Platform::Path filename) {
-    std::string basename = filename.FileName();
+bool EnsureSketchExtension(Platform::Path *filename, bool *shouldConfirmOverwrite) {
+    std::string basename = filename->FileName();
     if(basename.find('.') == std::string::npos) {
-        filename.raw += ".";
-        filename.raw += SolveSpaceUI::SKETCH_EXT;
+        filename->raw += ".";
+        filename->raw += SolveSpaceUI::SKETCH_EXT;
+        if(shouldConfirmOverwrite != NULL) {
+            *shouldConfirmOverwrite = FileExists(*filename);
+        }
+        return true;
     }
-    return filename;
+    if(shouldConfirmOverwrite != NULL) {
+        *shouldConfirmOverwrite = false;
+    }
+    return false;
+}
+
+static bool OkayToOverwriteFile(const Platform::Path &filename) {
+    Platform::MessageDialogRef dialog = CreateMessageDialog(SS.GW.window);
+
+    using Platform::MessageDialog;
+    dialog->SetType(MessageDialog::Type::QUESTION);
+    dialog->SetTitle(C_("title", "Overwrite File?"));
+    dialog->SetMessage(ssprintf(C_("dialog", "The file “%s” already exists."),
+                                filename.raw.c_str()));
+    dialog->SetDescription(C_("dialog", "Do you want to replace it?"));
+    dialog->AddButton(C_("button", "&Overwrite"), MessageDialog::Response::YES,
+                      /*isDefault=*/false);
+    dialog->AddButton(C_("button", "&Cancel"), MessageDialog::Response::NO,
+                      /*isDefault=*/true);
+
+    return dialog->RunModal() == MessageDialog::Response::YES;
 }
 
 bool SolveSpaceUI::GetFilenameAndSave(bool saveAs) {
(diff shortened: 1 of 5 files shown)


wolfpackmars2:
I realized that changing the filename outside of the usual GUI mechanisms could introduce filename conflicts, so the new commit fixes that.

It's no secret that I am working with Codex, but the code, and the oversight, are mine.

To apologize for my oversight, I added some tests for my changes since it was recently mentioned that test coverage could be improved.

ruevs:
@wolfpackmars2 please describe the problem/bug that you are trying to fix. At least on Windows the file extensions are handled normally.

After that we can discuss the code that the LLM wrote.

wolfpackmars2:
😮‍💨

My own LLM provides a better response to your "concerns" than I could come up with myself, so I'll share it verbatim since there is nothing to disagree with.


> The bug is #1689: in the Qt build on Ubuntu Noble, saving a sketch without an extension produces a file without .slvs, which then is not visible through the default open dialog filter. The GTK build and Windows may behave differently, which is why I tested both Qt and GTK on Ubuntu.<br/>
> The intent of this PR is to make extension handling consistent before the file is saved, while avoiding conflicts with existing filenames. I’m happy to walk through the code or adjust the approach if there’s a preferred place for this logic.


But really, I spend a lot of time on every comment and especially every PR I create, so it is understandably and reasonably frustrating when my hard work is dismissed so offhandedly.

src/solvespace.cpp
ruevs, line 618: Writing from a phone I did not notice that you linked #1689 to the PR - I apologize.

## Regarding the missing extension on QT:

This is the line that adds the extension(s) to the dialog.

Compare the Windows implementation of `AddFilter` [here ](https://github.com/solvespace/solvespace/blob/08fdb3596922dc93837578a8cd8cda32e3c83f23/src/platform/guiwin.cpp#L1630) to the QT implementation [here](https://github.com/solvespace/solvespace/blob/08fdb3596922dc93837578a8cd8cda32e3c83f23/src/platform/guiqt.cpp#L183)

It seems the QT version is adding the extension(s) to the dialog. Therefore the question remains why does the file dialog not append the extension to the entered file name. A quick Google search leads to this: https://stackoverflow.com/questions/1953631/qfiledialog-adding-extension-automatically-when-saving-file perhaps the solution is in there? Perhaps in [`GetFileName`](https://github.com/solvespace/solvespace/blob/08fdb3596922dc93837578a8cd8cda32e3c83f23/src/platform/guiqt.cpp#L165)?

Maybe [`QFileDialog::setDefaultSuffix()`](https://doc.qt.io/qt-6/qfiledialog.html#defaultSuffix-prop) should be called somewhere?

As you can see by his comment [here](https://github.com/solvespace/solvespace/blob/08fdb3596922dc93837578a8cd8cda32e3c83f23/src/platform/guiqt.cpp#L146) @WickedSmoke was already thinking about this when he worked on the QT port. 

## Regarding the code the LLM generated

The fix should be in `guiqt.cpp` and not here (in `solvespace.cpp`) - and this is the only thing I'll say - in order to maintain civility.

wolfpackmars2:
Final edit. ChatGPT has betrayed me. Looks like I'm on my own...

I appreciate the apology. I also need to apologize for my earlier reply; it was inappropriate and unhelpful and doesn't show professionalism. I was frustrated because I spent a lot of time making this PR (and this response apparently! Thx for nothing damn klank.... ) "Professionalism" ☺️

The "LLM generated" title is not true, not fair and in my opinion not warranted. `"the code, and the oversight, are mine."` Full. Stop. Every single line of code that I posted, was written by me. I have seen the change to CONTRIBUTING.md, and I have not violated the new terms.

"Having said that, feel free to use AI to understand the code and **help you to find a solution that works, is minimal and complete.**" My use falls within these terms.

## I made the decision to use solvespace.cpp

My reasoning was:

- SolveSpace currently relies on the file extension to recognize sketch files. `file.cpp`
- If the user provides a filename without an extension, the final filename passed to `SaveToFile()` should be a valid SolveSpace sketch filename regardless of which UI provided it.
- I wanted a solution which solved the current problem, and avoids the same problem reoccurring in the future.
- Maybe putting it in file.cpp would have aligned better with my goals and intent. Not that I am trying to muddy this further, just stating only **one** of my observations/considerations.

Ultimately, my my goal is to close #1689 because it is small, stupid, simple, straightforward, and I just want to move on. If you still want it handled in guiqt.cpp, I will make the change.

Having my efforts dismissed so offhandedly hurts. The "LLM generated" tag is incredibly  offensive and not at all a reflection of the truth, as I have tried to make clear. I absolutely regret even bringing it up. But w/e, I want to move on to bigger and more interesting things.

P.s. My previous edits are lon... more detailed. Reading them may provide useful detail and context. But they were too... detailed. The second edit is posted as sarc/humor. Nothing I posted is or was intended to be negative or inflammatory.

ruevs:
> Ultimately, my my goal is to close #1689 because it is small, stupid, simple, straightforward, and I just want to move on. If you still want it handled in guiqt.cpp, I will make the change.

Please do.

I understand your explanation for putting the change in solvespace.cpp. However, personally - I hate it. Since you wrote the code I'll explain why (initially I understood, that you "prompted it", and I have no desire to comment on /bad/ LLM generated code):

In no particular order:
- Five changed files (including a new testcase) instead of a (most probably) one liner in a single file.
- A bunch of new code affecting all (five) platforms.
- Duplicated logic for checking file overwriting.
- A new dialog box!
- New text to translate in [10 languages](https://github.com/solvespace/solvespace/tree/master/res/locales).
- Two new functions that do not "belong" - logically and aesthetically - in this file.

P.S. Commenting on bad human written code is worth it because: it may trigger a useful discussion, "bad" may be a matter of taste and someone may learn something. Arguing with an LLM - on the other hand - may be worth it only in the session, in which it generated the slop.

wolfpackmars2:
Fair enough, I'll make the changes. Would you like me to remove the tests as well? I will update the tests based on the changes or I can remove them altogether.

I will try to get this done tonight, but I do not share upstream changes without compiling and testing them first.

I'm prepared to explain or discuss any changes or ideas. It's difficult to defend against changes dismissed out of hand. Thank you for your time and guidance and taking another look at this.

wolfpackmars2:
I am cancelling this PR. I thought I was addressing the root issue by ensuring the file extension gets appended properly, regardless of which interface is used. However, this PR actually ignores the root problem exposed by the issue it was intended to resolve.

To move forward, I think more discussion is necessary with more experienced stakeholders in order to identify what the root cause is, and what the best fix would be. I lack experience in this domain (Linux GUI apps) to confidently address this by myself, at this time.

PR #1727, opened 2026-06-24
Added suffix handling for Qt SAVE file dialogs

### LINKED ISSUE #1689: QT: Solvespace not appending .slvs file extension by default

wolfpackmars2:
Resolves #1689

Not much else to say, the code is straightforward and matches how guigtk.cpp handles file extensions as closely as possible.

This also fix how file extensions are handled for other export file types, like .stl etc.

I have functionally tested this in Ubuntu 24.04.

### DIFF
--- a/src/platform/guiqt.cpp
+++ b/src/platform/guiqt.cpp
@@ -137,20 +137,27 @@ class FileDialogImplQt final : public FileDialog {
 public:
     QFileDialog fileDialogQ;
     QStringList filters;
+    QStringList defaultSuffixes;
     bool filtersMod;
+    bool isSaveDialog;
 
-    FileDialogImplQt(QWidget* parent, bool save) : filtersMod(false) {
+    FileDialogImplQt(QWidget* parent, bool save)
+        : filtersMod(false), isSaveDialog(save) {
         fileDialogQ.setParent(parent);
         fileDialogQ.setWindowFlags(Qt::Dialog); // Reset after setParent.
 
         // NOTE: The file extension is automatically set by the KDE native
-        // dialog but not with the Qt one.
+        // dialog and now with the Qt one.
 #if 0
         fileDialogQ.setOption(QFileDialog::DontUseNativeDialog);
 #endif
 
         if (save) {
             fileDialogQ.setAcceptMode(QFileDialog::AcceptSave);
+            QObject::connect(&fileDialogQ, &QFileDialog::filterSelected,
+                             [this](const QString &filter) {
+                                 updateDefaultSuffix(filter);
+                             });
         }
     }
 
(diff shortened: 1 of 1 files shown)


phkahler:
@wolfpackmars2   This looks much better than the last attempt. Simple and localized to the one file!  Unfortunately my Linux box is still broken so I'll have to let someone else test this. I might just trust that you tested it enough since it's so simple and restricted to QT  ;-)

@WickedSmoke  can you give it a look, if not an actual build/test?

src/platform/guiqt.cpp
```diff
@@ -137,20 +137,27 @@ class FileDialogImplQt final : public FileDialog {
 public:
     QFileDialog fileDialogQ;
     QStringList filters;
+    QStringList defaultSuffixes;
```
WickedSmoke, line 140: There's no need to add new data members.  I edited this patch to remove them on my [qt-ext3 branch](https://github.com/WickedSmoke/solvespace_qt5/commit/2efb82c3e18f74).

src/platform/guiqt.cpp
```diff
@@ -137,20 +137,27 @@ class FileDialogImplQt final : public FileDialog {
...
 
-    FileDialogImplQt(QWidget* parent, bool save) : filtersMod(false) {
+    FileDialogImplQt(QWidget* parent, bool save)
+        : filtersMod(false), isSaveDialog(save) {
         fileDialogQ.setParent(parent);
         fileDialogQ.setWindowFlags(Qt::Dialog); // Reset after setParent.
 
         // NOTE: The file extension is automatically set by the KDE native
-        // dialog but not with the Qt one.
+        // dialog and now with the Qt one.
```
WickedSmoke: This is a comment on the behavior of the platform dialogs, not about this class.  I've updated the comment to include Gtk on my qt-ext3 change.

wolfpackmars2: > This is a comment on the behavior of the platform dialogs, not about this class. 

this was just a last minute "cheeky" edit.

I will get these tested sometime this week and then get this closed out if no concerns are identified. I'm sure it's fine, just haven't had a chance to look at it yet.

Thanks for reviewing this for me @WickedSmoke.

wolfpackmars2:
@WickedSmoke Without the defaultSuffixes and related code, the Qt dialog would not automatically add the file suffix, unless the user first modified the Filters combo box. 

I did update the comments slightly.

Long story short, the edits in the qt-ext3 branch will force a file extension, when the user initiates the change in the dialog box, but it didn't set the "default" file suffixes.

I considered dropping the added bool, but using it didn't seem any less efficient than querying the .acceptMode of the dialog.

ruevs:
@WickedSmoke I tested your [https://github.com/WickedSmoke/solvespace_qt5/](https://github.com/WickedSmoke/solvespace_qt5/tree/qt-ext3) branch at commit 2efb82c3e18f7478adc27094921f9e5d54e8a193 with Qt5 on Windows and it works fine. The .slvs. extension is added. Of course with the [`#if 0`](https://github.com/WickedSmoke/solvespace_qt5/blob/2efb82c3e18f7478adc27094921f9e5d54e8a193/src/platform/guiqt.cpp#L145) changed to `#if 1`.

My suggestion is to merge that branch into master and close this PR. @phkahler, @wolfpackmars2  what do you think?

phkahler:
> My suggestion is to merge that branch into master and close this PR. @phkahler, @wolfpackmars2 what do you think?

But we don't do merges? Is that even compatible? The fix is simple enough, but merge the whole branch? Can @WickedSmoke rebase it on our master?  I'm all for the fix.

WickedSmoke:
> Long story short, the edits in the qt-ext3 branch will force a file extension, when the user initiates the change in the dialog box, but it didn't set the "default" file suffixes.

@wolfpackmars2 setDefaultSuffix does get called (inside RunModal) even if the filterSelected signal is never emitted.  I've just tested this again.  Please re-verify the behavior on your system.

> But we don't do merges? Is that even compatible?

@phkahler I'm not sure what you're talking about here.  qt-ext3 is after 6f816057, which is after master.  I'd prefer to keep both commits to give the appropriate authors credit, but squash them if you must.

wolfpackmars2:
> My suggestion is to merge that branch into master and close this PR. @phkahler, @wolfpackmars2 what do you think?

I'm fine with this.

ruevs:
I merged the [https://github.com/WickedSmoke/solvespace_qt5/](https://github.com/WickedSmoke/solvespace_qt5/tree/qt-ext3) branch is master. Closing this.

Issue #1619, opened 2025-08-17
NURBS failure when ClassifyEdge fails (was: Difference fail depending on compilation options (was: Difference fail with OpenMP))


phkahler:
This was first [reported on the forum](https://solvespace.com/forum.pl?action=viewthread&parent=6129&tt=1755262509) using the attached file.

[Issue_OpenMP.zip](https://github.com/user-attachments/files/21822164/Issue_OpenMP.zip)

You can make the problem go away by:
1) Dragging the original extrusion the other direction
2) Using a version of Solvespace built without OpenMP
3) Open it using the web version (which does not use OpenMP)

Treating it as a concurrency problem I have been able to make the issue go away by commenting out the following line in SShell::CopySurfacesTrimAgainst()
```
#pragma omp parallel for
```
That function calls MakeCopyTrimAgainst() and inside of that we can "fix" things by adding #pragma omp critical  to either of the two big loops [1 here](https://github.com/solvespace/solvespace/blob/74a13ef61c490a400eaa49c5335622d91ada12fc/src/srf/boolean.cpp#L603) or the one under it [2 here](https://github.com/solvespace/solvespace/blob/74a13ef61c490a400eaa49c5335622d91ada12fc/src/srf/boolean.cpp#L637).

I don't understand how it can be fixed by making a critical section for either of two loops. At this point I don't have any idea what's causing it.

iscgar:
Adding debug prints and comparing the state between a broken build and a working one shows the following diff:

```diff
$ diff -Nura ok2.txt bad2.txt
--- ok2.txt	2025-08-26 21:30:11.798839788 +0300
+++ bad2.txt	2025-08-26 21:30:04.274076415 +0300
@@ -30,7 +30,7 @@
  enout = { 0.000000, 0.000000, -0.080000 }
  surfn = { -4744.000000, 0.000000, 0.000000 }
   --------------------------
- classified = 0, indir_shell = 0, outdir_shell = 0, indir_orig = 100, outdir_orig = 200
+ classified = 0, indir_shell = 39, outdir_shell = 39, indir_orig = 100, outdir_orig = 200
   --------------------------
  orig.chain[0] = { tag = 0, auxA = 6, auxB = 1, a = { 1.000000, 1.000000, 0.000000 }, b = { 1.000000, 0.000000, 0.000000 } }
[...]

Issue #1091, opened 2021-08-14
NURBS failure with countersunk holes


tomsci:
### System information

- **SolveSpace version:** master @ 56719415dea4de7af750227653362f4c8cad78f8 (3.0)
- **Operating system:** macOS 11.5.1 (Big Sur)

### Expected behavior

Opened [turc-nurbs-1.slvs.zip](https://github.com/solvespace/solvespace/files/6986416/turc-nurbs-1.slvs.zip)
Expected no NURBS issues

### Actual behavior

The 'board-1-inset' group shows "The Boolean operation failed", there are angry red lines around some of the geometry from this group and Analyze->Naked Edges reports "The mesh has naked edges (NOT okay, invalid). 10 problematic edges, bad." On the console I get:

```
failed: I=20, avoid=7
print 4 edges
failed to assemble polygon to trim nurbs surface in uv space
```

### Additional information

The problematic model has a couple of holes through a solid board (called "top-plate" in the model), modelled as a difference extrude ("board-1-holes" being the extrude of the "board-1" sketch). These holes are countersunk by adding a second sketch of slightly-larger holes and a second difference extrude that doesn't go all the way through the board ("board-1-inset-sketch" and "board-1-inset"). I've simplified board-1-inset-sketch to reduce the number of holes, and only countersink the top-left hole, to hopefully make the test case easier. When I tried removing some more construction geometry from "board-1" the problem went away and I'm not sure why.

The overall project that this model is part of quickly degenerated into having to tick "Force NURBS surfaces to triangle mesh" early on and is pretty complex and incredibly slow to recalculate (Generate::ALL takes 30s). I read a comment on another issue suggesting that the triangle meshing might be contributing to the performance issues, hence why I'm raising this to see if I can get the whole thing to work in NURBS.

Symbian9:
JFTR, Increasing `chord tolerance (in percents)` to `0,01 %` prevents this NURBS issue.

![pic](https://i.imgur.com/fFH3Uv8.png)

tomsci:
The correctness of NURBS operations is predicated on the chord tolerance setting? 🤨 I didn't realise that the chord tolerance would affect such things, I assumed it was only when you forced to triangle mesh (and thus started approximating everything) that the chord tolerance became a factor. My mistake.

[...]

PR #1728, opened 2026-07-06
Fix an OpenMP use-after-free on into->curve in AddExactIntersectionCurve()

BoykoNeov:
# Fix an OpenMP use-after-free on into->curve in AddExactIntersectionCurve()

**Branch:** `fix-surfinter-openmp-race` (one commit, `3b3a12b4`, off master)

## Summary

`SSurface::AddExactIntersectionCurve()` (src/srf/surfinter.cpp) searches
`into->curve` for an existing identical exact curve inside
`#pragma omp critical(into)`, but keeps the raw pointer `existing` and
dereferences it (iterating `existing->pts`) **after** leaving the critical
section. Meanwhile other threads in the surface-intersection loop call
`into->curve.AddAndAssignId()`, which can grow the underlying `std::vector`
and reallocate its storage — so the reader walks freed memory: a classic
use-after-free.

The bug is pre-existing on master and independent of any classification
logic; it just needs a model whose Boolean produces many *identical* exact
intersection curves (so the `existing` path is taken often) plus OpenMP
scheduling luck. Coincident/overlapping faces do exactly that.

Fix: copy the piecewise-linear points into the local curve while still
inside the critical section. The later `sc.pts.Reverse()` then operates on
the local copy only.

## How it was found and verified

All on Windows 11, MSVC 19.44 x64, `-G Ninja`, `ENABLE_OPENMP=ON`, using the
`NURBS_ThreePrisms.slvs` model from #1091 (its coplanar prism faces generate
many identical exact curves):

- **Reproduced**: `solvespace-cli export-mesh` on that model segfaulted in
  ~5–7 of 60 Debug runs; **0 of 60** with `OMP_NUM_THREADS=1`.
- **Pinpointed with ASan** (`-DENABLE_SANITIZERS=ON`, `/fsanitize=address`):
  heap-use-after-free — read at the `existing->pts` iteration
  (surfinter.cpp:51 area) of memory freed by `_Emplace_reallocate` reached
  from `AddAndAssignId` in another thread's `AddExactIntersectionCurve`
  (surfinter.cpp:111 area). Full report captured.
- **After the fix**: 0 crashes in 100 Debug runs, 60 ASan runs, and
  40 Release runs of the same model; repeated OpenMP runs produce
  byte-identical triangle sets. Full test suite passes in Debug and Release.

## Notes

- The pointer was already obtained under `critical(into)` and the writers in
  boolean.cpp use a different (unnamed) critical, but those run in a
  different phase; the racing writer is the *same* function on other
  threads, whose `AddAndAssignId` is outside any critical section's
  protection of the read.
- Related but not touched: `SSurface::cached` (ratpoly.cpp
  `ClosestPointTo`) is also shared mutable state across threads; it only
  degrades a Newton starting guess, so it's left alone here.
- The classification PR (#1619 / #1091 knife edge) adds a regression test
  whose model is the heaviest known trigger of this race — with that test in
  the suite but this fix absent, the full test suite segfaults in roughly
  half of Debug OpenMP runs. The classification branch is therefore stacked
  on this one; **this PR should merge first**.

### DIFF
--- a/src/srf/surfinter.cpp
+++ b/src/srf/surfinter.cpp
@@ -44,12 +44,17 @@ void SSurface::AddExactIntersectionCurve(SBezier *sb, SSurface *srfB,
                 }
             }
         }
+        if(existing) {
+            // Copy the pwl points while still in the critical section;
+            // existing points into into->curve, which reallocates when
+            // another thread adds a curve.
+            SCurvePt *v;
+            for(v = existing->pts.First(); v; v = existing->pts.NextAfter(v)) {
+                sc.pts.Add(v);
+            }
+        }
     }// end omp critical
     if(existing) {
-        SCurvePt *v;
-        for(v = existing->pts.First(); v; v = existing->pts.NextAfter(v)) {
-            sc.pts.Add(v);
-        }
         if(backwards) sc.pts.Reverse();
         split = sc;
         sc = {};


src/srf/surfinter.cpp
```diff
@@ -44,12 +44,17 @@ void SSurface::AddExactIntersectionCurve(SBezier *sb, SSurface *srfB,
...
+                sc.pts.Add(v);
+            }
+        }
     }// end omp critical
     if(existing) {
-        SCurvePt *v;
-        for(v = existing->pts.First(); v; v = existing->pts.NextAfter(v)) {
-            sc.pts.Add(v);
-        }
         if(backwards) sc.pts.Reverse();
```
phkahler, line 58: Should the next line also be moved inside the critical section?

if(backwards) sc.pts.Reverse();

phkahler (commented):
Looks good except for my one question.

src/srf/surfinter.cpp
phkahler: I think the answer is no.   sc is created outside the critical section, so reversing its points after the section should be OK. It was copying the points from se that was problematic because se is set inside the section and should only be used there.

phkahler:
@BoykoNeov  Thank you for this contribution!

BoykoNeov:
> @BoykoNeov Thank you for this contribution!

My contribution is practically only the idea to unleash the LLM here and see what it can do :) Thanks for maintaining this great piece of software!

Issue #1452, opened 2024-03-07
Naked edges with only rectangles and extrusion


kristianlm:
Hi, and thanks for this great "little" piece of software with so much punch. SolveSpace is the first and only CAD I've learned to use.

I've got a situation that requires me to force everything to triangle meshes which doesn't seem right. Every object is 90 degrees as I'm only using rectangles. I've managed to reproduce this on the Arch Linux AUR package and on my GNU Guix system. I don't think it's a packaging issue.

### System information

- **SolveSpace version**: 3.1~70bde63c
- **Operating system:** Reproduced on Arch Linux and GNU Guix

### Expected behavior

![image](https://github.com/solvespace/solvespace/assets/259445/a2213a8a-1943-4ac4-aa81-b343da9a5a29)

I've forced `NURBS to triangle mesh` is on to get that screenshot

### Actual behavior

![image](https://github.com/solvespace/solvespace/assets/259445/71459e71-9096-4517-8830-a13981da4c18)

When `Force NURBS to triangle mesh` off, I'm missing an entire side surface.

### Additional information

[bug-report-naked-edges-with-only-cuboids-extrude-and-unions.zip](https://github.com/solvespace/solvespace/files/14521863/bug-report-naked-edges-with-only-cuboids-extrude-and-unions.zip)

The model is easily produced as follows:

![image](https://github.com/solvespace/solvespace/assets/259445/36c2d650-4da7-4ec2-9a83-d14f3cf8b7c6)

- Make rectangle A
- Extrude it downwards
- Make rectangle B in new sketch using previous sketch workplane or P
- Extrude it upwards
- Make rectangle C in new sketch using B's equivalent of point P (selecting its two connected lines)
- Extrude it upwards
- Make rectangle D in new sketch using P and its two perpendicular lines (or point on the other two cuboids)
- Extrude it right

The extrusion on A is in the opposite direction from B's extrusion. I cannot reproduce this bug if both are extruded in the same direction. The bug depends on the height of D: If D touches just A and B everything is ok. What's very strange is that if D if just _a little bit_ above C, it still works:
![image](https://github.com/solvespace/solvespace/assets/259445/d050cf02-b89f-4db5-90b4-25f98d1e8bbe)
Here, D touches C just a tiny bit and `force NURBS to triangle mesh` is off!

phkahler:
@kristianlm can you provide a simple .slvs file demonstrating this?

kristianlm:
[...]

PR #1410, opened 2023-09-22
Fix the logic in KeepRegion() and change inSame to coincSame for clar…

phkahler:
…ity.

This makes some real improvement on @rpavlik bug_repro file from issue #297. Still not fixed, but fewer missing surface areas.

Issue #1401, opened 2023-09-01
Simple extrusion of workplane on face of solid results in error


henrebotha:
### System information

- **SolveSpace version:** 3.1~70bde63c
- **Operating system:** Xubuntu 22.04

### Expected behavior

After extruding the final sketch (g006), no faces should have errors.

### Actual behavior

The northern face of the extrusion is marked in red as an error.

### Additional information

![image](https://github.com/solvespace/solvespace/assets/5593874/e3474e4e-7afd-4f96-b3c4-c20f0ebaea3e)

[extrusion-error.zip](https://github.com/solvespace/solvespace/files/12496421/extrusion-error.zip)

This happens no matter how I try to create a workplane on top of g005. The only thing that prevents the error is offsetting my workplane from the surface of g005 by some fractional value.

Issue #1291, opened 2022-09-06
Rounded corner cut (fillet) issue


phkahler:
### System information

- **SolveSpace version:** All
- **Operating system:** All

### Expected behavior

It should be possible to create a fillet by using a difference operation as shown:
[cube_cut.zip](https://github.com/solvespace/solvespace/files/9498557/cube_cut.zip)

![cube_cut](https://user-images.githubusercontent.com/14852918/188678388-8e4ad83b-2da4-4476-b911-27697e2dec74.png)

### Actual behavior

This file has curve control points P1 and P2 in group g004-sketch-fillet as well as the "corner" point of the fillet. When the corner point and P1 are coincident with the top cube edge (use point on line constraints) the NURBS fail.  Having just one of them on the cube edge is not enough to cause the problem.

The simplest work around for creating a fillet by cutting an arc is to place the corner point further away from the cube.

### Additional information
I made this sketch for easy experimentation with this issue. The curve-face tangency appears in several NURBS related issues, so this may be redundant but still a simple test case.

Since the tangent curve works with the corner point off-cube, and the corner point on-cube works if the curve is not tangent, this does not seem like a numerical issue (like finding the surface intersections) but more like a logic issue in determining which curves and surfaces to keep during the boolean operation. It is also notable that there is no failure if we switch from difference to an intersection boolean, even if the points are all colinear. Union also fails. It might also be a bug in merging the coincident surfaces. So somewhere in here:
https://github.com/solvespace/solvespace/blob/master/src/srf/boolean.cpp
https://github.com/solvespace/solvespace/blob/master/src/srf/merge.cpp

ruevs:
Good example! I'll add it to the tracking issue whem I am on a computer (next week). In addition I have an example involving only straight lines that passes with intersection and fails with difference and union. I created it two years ago when I implemented intersection, by have not gone back to it.

There seem to be at least five classes of "NURBS" failures that are emerging and there are more than enough examples in each group.

Off the top of my head (read edges as curves based on taste/case)
1. Zero length edges
2. Incorrectly split edges
3. Other bad intersections (numerical?)
4. Incorrectly selected edges
5. Incorrectly selected "patches" to keep.

I've only really ever looked at the code related to 1, 4, 5 and it always takes me a lot of time and thinking to even get into it... So I rarely start :-(

phkahler:
[...]

PR #1729, opened 2026-07-06
Fix NURBS boolean failures from unhandled edge-on-edge classification (fixes #1619, part of #1091)

### LINKED ISSUE #683: Invalid "The mesh has naked edges" error
### System information

SolveSpace version: 3.0~6894b0c9

Operating system: Windows 10

### Expected behavior

I should not see any errors

### Actual behavior
After extruding
image

After save - load
image

If remove last extruding layer  and create extruding again the error will be shown.
image

### Additional information
[wrong.zip](https://github.com/solvespace/solvespace/files/5112393/wrong.zip)

### LINKED ISSUE #728: Unexpected boolean operation fails in extrusion
### System information

SolveSpace version: master

Operating system: debian 9 amd64

### Expected behavior

Successful results

### Actual behavior

Boolean operation failure error, some red edges highlighted on the extrusion, and a spurious chord appears from intersecting geometry.

### Additional information

Open tapered-rafter-12-to-10-to-8-gusseted-truss.slvs in the attached zip to repro:

[gusseted-truss.zip](https://github.com/solvespace/solvespace/files/5315184/gusseted-truss.zip)

### LINKED ISSUE #1091: NURBS failure with countersunk holes

### LINKED ISSUE #1305: Surface experiments
3.1 70bde63c Version on Manjaro.

While drawing stuff my old friend surface problem showed up. This time I documented it a little for any debug use. 

Subtractive chamfer on corner drawn from the inside (0,0) worked but did not work from outside (inf,inf).  

errors
other side
When drawing the other side also from the inside what previously worked: 

other_side from innerplane

But partial is good:

other_side_partial

This is inconsistent as the with a little shaking of the point or going in or out of extrusion the red warning disappears. This fuzzy state goes away when the chord tolerance is raised to 0.01 from 0.1 and 10 segs instead 20.
Strange edges it got there too:

weird_curve

The first example lets me think of a failed inside outside detection for surface creation?

### LINKED ISSUE #1367: Edge_Fail - simple NURBS fail
### System information

- **SolveSpace version:** All
- **Operating system:** All

### Expected behavior

The attached sketch should not have naked edges.

### Actual behavior

The sketch fails depending on the location of points P1 - P4.

### Additional information

It appears that he real issue is misclassification of the edge(s) at the origin. Failure always show the short red edge at the origin in addition to a red rectangle where surface S3 protrudes from the first group. Moving P1 to the lleft can ffix it, as can a number of other movements ot P1-P4.

For discussion Lets call Surfaces S1 - S4 the surfaces connecting the origin to P1-P4. Also Edge normals (for in -surface edge normals) EN1-EN4 would be the in-surface normals of the main edge (pointing along each surface from the origin).  Also surface normals SN1-SN4 would be the outward surface normals - these should bee perpendicular to their ENx edge normal.

[Edge_Fail.zip](https://github.com/solvespace/solvespace/files/11412489/Edge_Fail.zip)

### LINKED ISSUE #1390: "failed to assemble polygon to trim nurbs surface in uv space" with a simple wedge cut-out
### System information

- **Operating system:** Gentoo Linux

### Expected behavior

There is no error if extrusion does not extend all the way through the other group:

image

### Actual behavior

image

### Additional information

[wedge-diff-nurbs.slvs.gz](https://github.com/solvespace/solvespace/files/12318087/wedge-diff-nurbs.slvs.gz)

### LINKED ISSUE #1398: unstable "failed to assemble polygon to trim nurbs surface in uv space"
### System information

- * **SolveSpace version:** 3.1~70bde63c
- * **Operating system:** Gentoo Linux

### Actual behavior

...depends on extrusion depth.

nurbs-2

### Additional information

[nurbs-2.slvs.gz](https://github.com/solvespace/solvespace/files/12435894/nurbs-2.slvs.gz)

### LINKED ISSUE #1619: NURBS failure when ClassifyEdge fails (was: Difference fail depending on compilation options (was: Difference fail with OpenMP))

BoykoNeov:
# Fix NURBS boolean failures from unhandled edge-on-edge classification (fixes #1619, part of #1091)

## Summary

Fixes #1619 ("failed to assemble polygon to trim nurbs surface", initially
suspected to be an OpenMP concurrency bug), and the unhandled knife-edge
classification case found while testing against the #1091 three-prisms
model.

`SShell::ClassifyEdge()` had two failure paths in the edge-on-edge case:

1. **Tangent at a shell edge** (#1619): when the edge is tangent to the shell
   at the shell's edge, it did `return false` without writing its
   `indir`/`outdir` outputs — and both call sites in
   `SSurface::MakeCopyTrimAgainst()` ignore the return value, so `KeepEdge()`
   was called with **uninitialized stack values**. Whether the edge was kept
   depended on whatever was on the stack, which is why the outcome flipped
   with build type, OpenMP, and sanitizers (see @iscgar's debug diff in the
   issue, and @ruevs' build-mode table). @jwesthues anticipated exactly this
   in https://github.com/solvespace/solvespace/issues/1091#issuecomment-904048397:
   "I suspect some paths now access uninitialized memory for
   `{in,out}dir_shell`."

2. **Knife edge** (#1091, `NURBS_ThreePrisms.slvs`): when regions of the
   shell join along a knife edge, an edge lying on that line intersects
   *four* (in general, any even number of) coincident shell edges. The
   `edge_inters == 2` block was skipped entirely, the code fell through to
   the ray-casting fallback, and the wrong classification broke the trim
   polygon (naked edges, "failed to assemble polygon").

The attached `Issue_OpenMP.slvs` hits path 1 because the subtracted wedge is
sketched exactly on the box's face and its hypotenuse plane touches the box's
bottom-front edge tangentially.

## The fixes

**`src/srf/raycast.cpp`, tangent case** — implement the missing
classification instead of returning false. When our edge is tangent to the
shell at the shell's edge, the two faces meeting at that edge lie on the same
side of our surface, so the regions on both sides of our edge get the same
class:

- **outside** the shell if the shell's edge is convex,
- **inside** if it's concave (reflex).

The vectors already computed in this function give this directly. Since
`inter_edge_n[i]` points away from face *i*'s material, the shell's edge is
convex iff face 1's material direction lies behind face 0's plane:

```cpp
Class c = (inter_edge_n[1].Dot(inter_surf_n[0]) > 0) ?
                Class::SURF_OUTSIDE : Class::SURF_INSIDE;
*indir  = c;
*outdir = c;
```

(Orientation convention cross-checked against the function's own
coincident-face branches and against `EdgeNormalsWithinSurface()`, whose
actual math treats `n×d` as the *outer* edge normal — the "inner normal"
comment at its line 457 is stale; note `pin = PointAt(muv.Minus(enuv))` and
the commented-out `*enout = enxyz`.)

**`src/srf/raycast.cpp`, knife-edge case** — handle `edge_inters >= 4`
(even). The faces along the knife edge divide the pencil of directions
perpendicular to the edge at p into sectors that are alternately inside and
outside the shell. Classify each side of our edge by the face *angularly
closest* to it: the side is outside the shell iff it lies on the outer side
of that face, and if it lies *along* the nearest face (within `DOTP_TOL`),
it's coincident, same/opposite by the surface normals — the same answers the
`edge_inters == 2` block produces, generalized to any number of faces. (This
is jwesthues' suggested experiment from #1091 made robust: rather than
"picking any pair and classifying as if n=2", the nearest face always bounds
the sector the queried direction lies in, so no arbitrary pairing is needed;
it also tolerates duplicate matches from split edge chains.)

**`src/srf/boolean.cpp`** — defensive hardening at both call sites:
initialize `indir_shell`/`outdir_shell` to a deterministic value and print a
debug warning if `ClassifyEdge()` ever returns false again, so any future
unhandled classification case fails loudly and reproducibly instead of
depending on uninitialized memory (per jwesthues' recommendation in #1091).

## Verification

All on Windows 11, MSVC 19.44 x64, `-G Ninja`, `ENABLE_OPENMP=ON`:

1. **Reproduced first** (unfixed master, Debug): `solvespace-cli export-mesh`
   on `Issue_OpenMP.slvs` prints `failed: I=3, avoid=2` and `failed to
   assemble polygon to trim nurbs surface in uv space` — same surface 3 as in
   the issue.
2. **Mechanism proven**: initializing the two variables with a "poison" value
   (`SURF_INSIDE`) made the failure 100% deterministic in every build config,
   with `ClassifyEdge()` returning false 3 times, all on surface 3.
3. **Geometry verified against ground truth**: the expected volume computed
   analytically from the sketch data is
   (59.3·80 − ½·59.3·14.667931)·59.3 = **255,529.4 mm³**; the fixed build
   exports a mesh of exactly 255,529.39 mm³ with the expected 6-quad-face
   topology (12 triangles). An unfixed *Release* build (the config that was
   "OK" by stack-garbage luck) produces the identical triangle set.
4. **Concave branch verified too, and covered by a testsuite case**
   (`test/group/boolean_tangent_edge/`): an L-section extrusion minus a
   triangular prism whose face plane passes tangentially through the L's
   concave vertical edge. Instrumentation confirms this one model exercises
   **both** new tangent branches at the same edge — the prism's edge against
   the L classifies concave → `SURF_INSIDE`, and the L's edge against the
   prism (where that edge is a convex wedge of the prism) classifies convex →
   `SURF_OUTSIDE`. The test checks that the mesh is watertight and
   non-self-intersecting and that its volume matches the analytic value
   (296,000 mm³) from the sketch. On unfixed code with the poison-init from
   (2), the same model fails deterministically — naked edges, wrong geometry
   (264,000 mm³), and the "failed to assemble polygon" error.
5. **Knife-edge case verified and covered by a testsuite case**
   (`test/group/boolean_knife_edge/`, fixture = ruevs' `NURBS_ThreePrisms`
   model from #1091): previously 12 × `bad, edge_inters=4`, surface 13
   failed to assemble, 21 triangles with 5 naked edges. Fixed: clean export,
   watertight, 24 triangles, volume exactly the analytic
   (½·1·1.6126052 + ½·1.5498346·1)·1.0 + ½·1·1·0.4983402 = **1.8303900 mm³**
   (independently cross-checked against a force-to-triangle-mesh export of
   the same model: identical volume). Identical triangle sets across Debug,
   Release, and repeated OpenMP runs. The test fails with naked edges when
   the knife-edge classification is reverted.
6. **No regressions**: full test suite passes in Debug and Release
   (now 259 cases / 907 checks); `NURBS_TwoPrisms.slvs` still exports
   cleanly, and repeated OpenMP runs of all repro models produce identical
   geometry. (I avoided `CHECK_SAVE`/render comparisons in both new tests on
   purpose: the Boolean-derived surface cache contains a few last-ulp values
   that could differ across compilers/FP-contraction settings, and
   `PrepareSavefile` does not normalize `Curve`/`CCtrl` lines.)

## Notes / limitations

- **Stacked on the companion race-fix PR** ("Fix an OpenMP use-after-free on
  into->curve in AddExactIntersectionCurve()"): stress-testing the
  three-prisms model exposed a pre-existing OpenMP use-after-free in
  `SSurface::AddExactIntersectionCurve()` that intermittently crashes
  Booleans on models with many identical exact intersection curves — and the
  new `boolean_knife_edge` test model is the heaviest known trigger (the
  test suite segfaults in roughly half of Debug OpenMP runs without that
  fix). The race is an independent, pre-existing bug on master, so it's a
  separate PR; this branch is based on it and it should merge first.
- The other #1091 sub-issues (tangent curve-surface intersections through a
  cylinder's silhouette, the "proximity" failures, trim-curve splitting) are
  different mechanisms and remain open; this PR only fixes the edge-on-edge
  classification cases.
- jwesthues noted in #1091 that the Booleans assume non-degenerate geometry
  without regions joining along knife edges; the new `edge_inters >= 4`
  branch handles the classification half of that assumption. If some other
  stage still trips on a knife-edge model, it will now at least do so
  deterministically thanks to the boolean.cpp hardening.

### DIFF
--- a/src/srf/raycast.cpp
+++ b/src/srf/raycast.cpp
@@ -522,6 +522,72 @@ bool SShell::ClassifyEdge(Class *indir, Class *outdir,
         return true;
     }
 
+    if(edge_inters >= 4 && edge_inters % 2 == 0) {
+        // More than two surfaces of our shell meet edge-on-edge along the
+        // given edge: regions of the shell join along a knife edge there
+        // (e.g. two prisms sharing a single edge, issue #1091). The faces
+        // along the knife edge divide the directions perpendicular to the
+        // edge at p into sectors that are alternately inside and outside
+        // the shell, so classify each side of our edge against the face
+        // that lies angularly closest to it: that face bounds the sector
+        // containing it, and its normal points out of that sector iff the
+        // sector is outside the shell.
+        Vector edge_d = (eb.Minus(ea)).WithMagnitude(1);
+        Vector en[2] = { edge_n_in, edge_n_out };
+        Vector ep[2];
+        double bestdot[2] = { VERY_NEGATIVE, VERY_NEGATIVE };
+        Vector bestn[2];
+        for(int i = 0; i < 2; i++) {
+            ep[i] = (en[i].Minus(edge_d.ScaledBy(en[i].Dot(edge_d))))
+                        .WithMagnitude(1);
+        }
+
+        for(SSurface &srf : surface) {
+            if(srf.LineEntirelyOutsideBbox(ea, eb, /*asSegment=*/true)) continue;
+
+            SEdgeList *sel = &(srf.edges);
+            SEdge *se;
+            for(se = sel->l.First(); se; se = sel->l.NextAfter(se)) {
+                if((ea.Equals(se->a) && eb.Equals(se->b)) ||
+                   (eb.Equals(se->a) && ea.Equals(se->b)) ||
+                    p.OnLineSegment(se->a, se->b))
+                {
+                    Point2d pm;
+                    srf.ClosestPointTo(p, &pm, /*mustConverge=*/false);
+                    Vector n = srf.NormalAt(pm);
+                    // A vector tangent to the intersecting surface at p,
+                    // pointing into that face, perpendicular to our edge.
+                    Vector f = ((se->b).Minus(se->a)).Cross(n);
+                    f = f.Minus(edge_d.ScaledBy(f.Dot(edge_d)));
+                    if(f.Magnitude() < LENGTH_EPS) continue;
+                    f = f.WithMagnitude(1);
+                    for(int i = 0; i < 2; i++) {
+                        double dot = ep[i].Dot(f);
+                        if(dot > bestdot[i]) {
+                            bestdot[i] = dot;
+                            bestn[i]   = n;
+                        }
+                    }
+                }
+            }
+        }
+
+        Class *dir[2] = { indir, outdir };
+        for(int i = 0; i < 2; i++) {
+            double dotn = en[i].DirectionCosineWith(bestn[i]);
+            if(bestdot[i] > 0 && fabs(dotn) < DOTP_TOL) {
+                // This side of our edge lies along the nearest face, so
+                // our surface is locally coincident with it there.
+                *dir[i] = (surf_n.Dot(bestn[i]) > 0) ? Class::SURF_COINC_SAME
+                                                     : Class::SURF_COINC_OPP;
+            } else {
+                *dir[i] = (dotn > 0) ? Class::SURF_OUTSIDE
+                                     : Class::SURF_INSIDE;
+            }
+        }
+        return true;
+    }
+
     if(edge_inters != 0) dbp("bad, edge_inters=%d", edge_inters);
 
     // Next, check for edge-on-surface. The ray-casting for edge-inside-shell
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -63,6 +63,7 @@ set(testsuite_SOURCES
     request/line_segment/test.cpp
     request/ttf_text/test.cpp
     request/workplane/test.cpp
+    group/boolean_knife_edge/test.cpp
     group/boolean_tangent_edge/test.cpp
     group/link/test.cpp
     group/translate_asy/test.cpp
--- a/test/group/boolean_knife_edge/normal.slvs
+++ b/test/group/boolean_knife_edge/normal.slvs
@@ -1,0 +1,2229 @@
+±²³SolveSpaceREVa
+
+
+Group.h.v=00000001
+Group.type=5000
+Group.name=#references
+Group.color=ff000000
+Group.skipFirst=0
+Group.predef.swapUV=0
+Group.predef.negateU=0
+Group.predef.negateV=0
+Group.visible=1
+Group.suppress=0
+Group.relaxConstraints=0
+Group.allowRedundant=0
+Group.allDimsReference=0
+Group.scale=1.00000000000000000000
+Group.remap={
+}
+AddGroup
+
+Group.h.v=00000002
+Group.type=5001
+Group.order=1
+Group.name=sketch-in-plane
+Group.activeWorkplane.v=80020000
+Group.color=ff000000
+Group.subtype=6000
+Group.skipFirst=0
+Group.predef.q.w=1.00000000000000000000
+Group.predef.origin.v=00010001
+Group.predef.swapUV=0
+Group.predef.negateU=0
+Group.predef.negateV=0
+Group.visible=1
+Group.suppress=0
+Group.relaxConstraints=0
+Group.allowRedundant=0
+Group.allDimsReference=0
+Group.scale=1.00000000000000000000
+Group.remap={
+}
+AddGroup
+
+Group.h.v=00000003
+Group.type=5100
+Group.order=2
+Group.name=extrude
+Group.opA.v=00000002
+Group.color=00646464
+Group.subtype=7000
+Group.skipFirst=0
+Group.predef.entityB.v=80020000
+Group.predef.swapUV=0
+Group.predef.negateU=0
+Group.predef.negateV=0
+Group.visible=1
+Group.suppress=0
+Group.relaxConstraints=0
+Group.allowRedundant=0
+Group.allDimsReference=0
+Group.scale=1.00000000000000000000
+Group.remap={
+    1 00040000 1002
+    2 00040001 1002
+    3 00040002 1002
+    4 00040000 1001
+    5 00040001 1001
+    6 00040002 1001
+    7 00040000 1004
+    8 00040001 1003
+    9 00040002 1003
+    10 00050000 1002
+    11 00050001 1002
+    12 00050002 1002
+    13 00050000 1001
+    14 00050001 1001
+    15 00050002 1001
+    16 00050000 1004
+    17 00050001 1003
+    18 00050002 1003
+    19 00060000 1002
+    20 00060001 1002
+    21 00060002 1002
+    22 00060000 1001
+    23 00060001 1001
+    24 00060002 1001
+    25 00060000 1004
+    26 00060001 1003
+    27 00060002 1003
+    28 00070000 1002
+    29 00070001 1002
+    30 00070002 1002
+    31 00070000 1001
+    32 00070001 1001
+    33 00070002 1001
+    34 00070000 1004
+    35 00070001 1003
+    36 00070002 1003
+    37 00080000 1002
+    38 00080001 1002
+    39 00080002 1002
+    40 00080000 1001
+    41 00080001 1001
+    42 00080002 1001
+    43 00080000 1004
+    44 00080001 1003
+    45 00080002 1003
+    46 00090000 1002
+    47 00090001 1002
+    48 00090002 1002
+    49 00090000 1001
+    50 00090001 1001
+    51 00090002 1001
+    52 00090000 1004
+    53 00090001 1003
+    54 00090002 1003
+    55 000a0000 1002
+    56 000a0001 1002
+    57 000a0002 1002
+    58 000a0000 1001
+    59 000a0001 1001
+    60 000a0002 1001
+    61 000a0000 1004
+    62 000a0001 1003
+    63 000a0002 1003
+    64 000b0000 1002
+    65 000b0001 1002
+    66 000b0002 1002
+    67 000b0000 1001
+    68 000b0001 1001
+    69 000b0002 1001
+    70 000b0000 1004
+    71 000b0001 1003
+    72 000b0002 1003
+    73 80020000 1002
+    74 80020000 1001
+    75 80020001 1002
+    76 80020002 1002
+    77 80020001 1001
+    78 80020002 1001
+    79 80020002 1003
+    80 00000000 1001
+    81 00000000 1002
+    82 000f0000 1002
+    83 000f0001 1002
+    84 000f0002 1002
+    85 000f0000 1001
+    86 000f0001 1001
+    87 000f0002 1001
+    88 000f0000 1004
+    89 000f0001 1003
+    90 000f0002 1003
+    91 00100000 1002
+    92 00100001 1002
+    93 00100002 1002
+    94 00100000 1001
+    95 00100001 1001
+    96 00100002 1001
+    97 00100000 1004
+    98 00100001 1003
+    99 00100002 1003
+}
+AddGroup
+
+Group.h.v=00000004
+Group.type=5001
+Group.order=3
+Group.name=sketch-in-plane
+Group.activeWorkplane.v=80040000
+Group.color=00646464
+Group.subtype=6001
+Group.skipFirst=0
+Group.predef.origin.v=80030042
+Group.predef.entityB.v=8003001c
+Group.predef.entityC.v=80030037
+Group.predef.swapUV=0
+Group.predef.negateU=0
+Group.predef.negateV=1
+Group.visible=1
+Group.suppress=0
+Group.relaxConstraints=0
+Group.allowRedundant=0
+Group.allDimsReference=0
+Group.scale=1.00000000000000000000
+Group.remap={
+}
+AddGroup
+
+Group.h.v=00000005
+Group.type=5100
+Group.order=4
+Group.name=extrude
+Group.opA.v=00000004
+Group.color=00646464
+Group.subtype=7000
+Group.skipFirst=0
+Group.predef.entityB.v=80040000
+Group.predef.swapUV=0
+Group.predef.negateU=0
+Group.predef.negateV=0
+Group.visible=1
+Group.suppress=0
+Group.relaxConstraints=0
+Group.allowRedundant=0
+Group.allDimsReference=0
+Group.scale=1.00000000000000000000
+Group.remap={
+    1 000c0000 1002
+    2 000c0001 1002
+    3 000c0002 1002
+    4 000c0000 1001
+    5 000c0001 1001
+    6 000c0002 1001
+    7 000c0000 1004
+    8 000c0001 1003
+    9 000c0002 1003
+    10 000d0000 1002
+    11 000d0001 1002
+    12 000d0002 1002
+    13 000d0000 1001
+    14 000d0001 1001
+    15 000d0002 1001
+    16 000d0000 1004
+    17 000d0001 1003
+    18 000d0002 1003
+    19 000e0000 1002
+    20 000e0001 1002
+    21 000e0002 1002
+    22 000e0003 1002
+    23 000e0020 1002
+    24 000e0000 1001
+    25 000e0001 1001
+    26 000e0002 1001
+    27 000e0003 1001
+    28 000e0020 1001
+    29 000e0001 1003
+    30 000e0002 1003
+    31 000e0003 1003
+    32 80040000 1002
+    33 80040000 1001
+    34 80040001 1002
+    35 80040002 1002
+    36 80040001 1001
+    37 80040002 1001
+    38 80040002 1003
+    39 00000000 1001
+    40 00000000 1002
+    41 000e0000 1004
+}
+AddGroup
+
+Param.h.v.=00010010
+AddParam
+
+Param.h.v.=00010011
+AddParam
+
+Param.h.v.=00010012
+AddParam
+
+Param.h.v.=00010020
+Param.val=1.00000000000000000000
+AddParam
+
+Param.h.v.=00010021
+AddParam
+
+Param.h.v.=00010022
+AddParam
+
+Param.h.v.=00010023
+AddParam
+
+Param.h.v.=00020010
+AddParam
+
+Param.h.v.=00020011
+AddParam
+
+Param.h.v.=00020012
+AddParam
+
+Param.h.v.=00020020
+Param.val=0.50000000000000000000
+AddParam
+
+Param.h.v.=00020021
+Param.val=0.50000000000000000000
+AddParam
+
+Param.h.v.=00020022
+Param.val=0.50000000000000000000
+AddParam
+
+Param.h.v.=00020023
+Param.val=0.50000000000000000000
+AddParam
+
+Param.h.v.=00030010
+AddParam
+
+Param.h.v.=00030011
+AddParam
+
+Param.h.v.=00030012
+AddParam
+
+Param.h.v.=00030020
+Param.val=0.50000000000000000000
+AddParam
+
+Param.h.v.=00030021
+Param.val=-0.50000000000000000000
+AddParam
+
+Param.h.v.=00030022
+Param.val=-0.50000000000000000000
+AddParam
+
+Param.h.v.=00030023
+Param.val=-0.50000000000000000000
+AddParam
+
+Param.h.v.=00040010
+AddParam
+
+Param.h.v.=00040011
+Param.val=-1.61260523905064823147
+AddParam
+
+Param.h.v.=00040013
+AddParam
+
+Param.h.v.=00040014
+AddParam
+
+Param.h.v.=00070010
+AddParam
+
+Param.h.v.=00070011
+AddParam
+
+Param.h.v.=00070013
+Param.val=-1.00000000000000000000
+AddParam
+
+Param.h.v.=00070014
+AddParam
+
+Param.h.v.=000a0010
+AddParam
+
+Param.h.v.=000a0011
+AddParam
+
+Param.h.v.=000a0013
+AddParam
+
+Param.h.v.=000a0014
+Param.val=1.00000000000000000000
+AddParam
+
+Param.h.v.=000b0010
+Param.val=1.54983460368207648372
+AddParam
+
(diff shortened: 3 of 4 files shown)


src/srf/raycast.cpp
```diff
@@ -506,9 +506,84 @@ bool SShell::ClassifyEdge(Class *indir, Class *outdir,
...
+            // behind face 0's plane.
+            Class c = (inter_edge_n[1].Dot(inter_surf_n[0]) > 0) ?
+                            Class::SURF_OUTSIDE : Class::SURF_INSIDE;
+            *indir  = c;
+            *outdir = c;
+        }
+        return true;
+    }
+
+    if(edge_inters >= 4 && edge_inters % 2 == 0) {
```
ruevs, line 525: @jwesthues @phkahler Please look at this section - I stepped through it with the debugger, I think I understand it, it definitely works; but I'm still not 100% sure it's "the right thing" to do.

I'll attach a test model (in addition to https://github.com/solvespace/solvespace/files/7030785/NURBS_ThreePrisms.zip).

src/srf/raycast.cpp
```diff
@@ -506,9 +506,84 @@ bool SShell::ClassifyEdge(Class *indir, Class *outdir,
...
-            return false;
+            // Our edge is tangent to the shell at the shell's edge: the
+            // two faces meeting at the shell's edge lie on the same side
+            // of our surface, so the regions on both sides of our edge
+            // get the same class. That's outside the shell if the shell's
+            // edge is convex, and inside if it's concave (reflex).
+            // inter_edge_n[i] points away from face i's material, so the
+            // shell's edge is convex iff face 1's material direction lies
+            // behind face 0's plane.
+            Class c = (inter_edge_n[1].Dot(inter_surf_n[0]) > 0) ?
```
ruevs, line 517: Very interesting! Was it really so simple?!

ruevs (commented):
When the [LLMs generate slop it is not worth looking at it](https://github.com/solvespace/solvespace/blob/master/CONTRIBUTING.md#llm-or-ai-contributions). But this is not slop!

This is in fact _very_ good. The code is coherent, the comments are coherent, the analysis is correct. Thanks @BoykoNeov for setting [Claude Fable 5](https://platform.claude.com/docs/en/about-claude/models/introducing-claude-fable-5-and-claude-mythos-5) onto this. This is the first time when I'm truly impressed by a programming problem solved by AI.

Speaking of problem - this actually fixes two of the most commonly reported NURBS bugs (See the comments in the individual commits for details):

- [Classifying edges tangent to the shell at a shell edge](https://github.com/solvespace/solvespace/pull/1729/changes/93659278a7c637908261e71121a62189b0d94778)
- [Classifying edges lying on a knife edge of the shell, where more than two
surfaces meet edge-on-edge.](https://github.com/solvespace/solvespace/pull/1729/changes/a1422a8fae0a4daa8c11ba8b5fb0d3ebc8b24482)

For the second one I am attaching a new test model that underlines the solution:
[1729_TestCase_SixPrismsShareingAnEdge.zip](https://github.com/user-attachments/files/29750075/1729_TestCase_SixPrismsShareingAnEdge.zip)
this is a simple 2mm³ triangular prism rotated 6 times along one of its edges. You can increase the number of repeats in the step-rotating group up to 25 - when the total volume becomes 50mm³ without anything breaking. Try the same with master before this PR or with 3.2.

## Here is a full list of everything that this fixes - I personally tested every model:

### Fixed in this PR:

- https://github.com/solvespace/solvespace/issues/683#issuecomment-683274604
  - https://github.com/solvespace/solvespace/files/5145107/wrong2.zip - works in both 3.2 (sometimes, based on stack junk?) and here.

- https://github.com/solvespace/solvespace/issues/683#issuecomment-730602943
  - https://github.com/solvespace/solvespace/files/5569373/wrong3.zip

- https://github.com/solvespace/solvespace/issues/683#issuecomment-731718242
  - https://github.com/solvespace/solvespace/files/5579199/wrong4.zip

- https://github.com/solvespace/solvespace/issues/728#issue-713235181
  - https://github.com/solvespace/solvespace/files/5315184/gusseted-truss.zip

- https://github.com/solvespace/solvespace/issues/1091#issuecomment-903661168
  - https://github.com/solvespace/solvespace/files/7030785/NURBS_ThreePrisms.zip

- https://github.com/solvespace/solvespace/issues/1091#issuecomment-945190823
  - https://github.com/solvespace/solvespace/files/7360794/corner-nurbs_slvs3x.zip

- https://github.com/solvespace/solvespace/issues/1305#issuecomment-1297213481
  - https://github.com/solvespace/solvespace/files/9902223/1305MinimalPrizmCamferNURBSBug.zip

- https://github.com/solvespace/solvespace/issues/1367#issue-1698644174
  - https://github.com/solvespace/solvespace/files/11412489/Edge_Fail.zip

- https://github.com/solvespace/solvespace/issues/1390#issue-1846087458
  - https://github.com/solvespace/solvespace/files/12318087/wedge-diff-nurbs.slvs.gz

- https://github.com/solvespace/solvespace/issues/1398#issue-1866321687
  - https://github.com/solvespace/solvespace/files/12435894/nurbs-2.slvs.gz

- https://github.com/solvespace/solvespace/pull/1410#issuecomment-1731480056
  - https://github.com/solvespace/solvespace/files/12700083/IntersectionReallyReallyReallyReallyNastyTest_DifferenceFailsBecauseItShould.zip - my  #favorite creation :-)

- https://github.com/solvespace/solvespace/issues/1619#issue-3328296115
  - https://github.com/user-attachments/files/21822164/Issue_OpenMP.zip

### Zip files with some models fixed and some not:

- https://github.com/solvespace/solvespace/issues/35#issuecomment-531602706
  - https://github.com/solvespace/solvespace/files/3614327/NURBSTests.zip
    - IntersectionReallyReallyReallyNastyTest_DifferenceFailsBecauseItShould.slvs - fails in 3.2, **fixed here**
    - IntersectionDifferenceFailsRandomly.slvs - still fails here (maybe this is "couldn't find an ear! fail"? See below)
    - all others work in both 3.2 and here.

- https://github.com/solvespace/solvespace/issues/1291#issuecomment-1288057464
  - https://github.com/solvespace/solvespace/files/9846083/CylinderTangentFromTheInside_AKA_Fillet_NURBSTests.zip
    - PrismChamferMinimalTestCase.slvs - **fixed here**
    - PrismFilletMinimalTestCase.slvs - different issue - still fails (tangent curved surfaces)
    - PrismFilletMinimalTestCase_OnlyOneFailingEdge.slvs - different issue - still fails (tangent curved surfaces)

- https://github.com/solvespace/solvespace/issues/1305#issuecomment-1294219759
  - https://github.com/solvespace/solvespace/files/9883625/Cluster_set.zip
    - cluster_repro.slvs - fails in 3.2, **fixed here**
    - Case_error_occured.slvs - works in both 3.2 and here
    - Case_solved.slvs - fails in both 3.2 and here


### Still failing - but this is a different issue ("couldn't find an ear! fail")

- https://github.com/solvespace/solvespace/issues/537#issuecomment-669068154
  - https://github.com/solvespace/solvespace/files/5027208/Prisms_EarFail.zip

- https://github.com/solvespace/solvespace/pull/594#issuecomment-625587508
  - https://github.com/solvespace/solvespace/files/4596617/CouldntFindAnEar_NURBSTests.zip

- https://github.com/solvespace/solvespace/issues/683#issue-683996514
  - https://github.com/solvespace/solvespace/files/5112393/wrong.zip

<details>
  <summary>Details</summary>

// When I first discovered the using of junk on the stack (but did not realize what it was): @ruevs ruevs on Nov 22, 2020
https://github.com/solvespace/solvespace/issues/683#issuecomment-731763446

  
// Dot products: @ruevs ruevs on Oct 25, 2022
https://github.com/solvespace/solvespace/issues/1291#issuecomment-1289998801
  
</details>

src/srf/surfinter.cpp
```diff
@@ -44,12 +44,17 @@ void SSurface::AddExactIntersectionCurve(SBezier *sb, SSurface *srfB,
...
+                sc.pts.Add(v);
+            }
+        }
     }// end omp critical
     if(existing) {
-        SCurvePt *v;
-        for(v = existing->pts.First(); v; v = existing->pts.NextAfter(v)) {
-            sc.pts.Add(v);
-        }
         if(backwards) sc.pts.Reverse();
```
phkahler, line 58: Should the next line that may reverse be moved inside the critical section too?

if(backwards) sc.pts.Reverse();

src/srf/raycast.cpp
phkahler: Regions of a single shell that meet along an edge must alternate inside/outside/inside/outside  otherwise that shell isn't valid. Once we bring a second shell into the picture things get complicated since the two (assuming only one new coincident edge) new surfaces can lie anywhere relative to the originals. I'm not sure I can reason about this very well, but the comment looks correct. I'm not sure how it's finding the closest one yet...

phkahler:
This is astonishing.  @BoykoNeov is it possible to see the conversation/prompts that lead to this? Had you studied the code and problems much prior to engaging the AI?  There are bits of spatial reasoning in there that I would not expect an LLM to understand because to me they are visual representations. Here I thought a mathematician dropped in to solve some of our bugs, but how much was you and how much was AI?

phkahler:
BTW I like the test that uses volume rather than an image. Images all need to be updated if we change the colors for example.

ruevs:
> This is astonishing. @BoykoNeov is it possible to see the conversation/prompts that lead to this? Had you studied the code and problems much prior to engaging the AI? There are bits of spatial reasoning in there that I would not expect an LLM to understand because to me they are visual representations. Here I thought a mathematician dropped in to solve some of our bugs, but how much was you and how much was AI?

This PR was prompted (pun intended) by a conversation about vibe coding @BoykoNeov and I had (in person) so I think I know the answer, but I'll let him answer :-)

<details>
In the meantime - I'll start planning for retirement and/or dealing with a Terminator/WALL-E/Blade Runner/The Matrix etc. scenarios ;-) 'been thinking about it for the last <a href="https://en.wikipedia.org/wiki/Attention_Is_All_You_Need">nine years</a>...
</details>

phkahler:
Can't keep hand off that "Rebase and Merge" button.....

phkahler:
Fixes #1398

I checked a LOT of sketches and saw no regressions.

phkahler:
AFAICT this doesn't fix the original issue in #683 in wrong.zip.

ruevs:
> AFAICT this doesn't fix the original issue in #683 in wrong.zip.

You are right - it does not fix it. I did not read carefully enough - one needs to delete an re-create the extrusion to reproduce the bug. I corrected my comment.

> Fixes #1398

Indeed! But **why** does it fix that?! Probably the junk on the stack. I'll have to debug...
EDIT: I debugged - it is the exact same problem. And I confirmed it by always returning a "wrong" classification in the `} else { // Our edge is tangent to the shell at the shell's edge:` - and it always fails. Good.

> Can't keep hand off that "Rebase and Merge" button.....

:-) Indeed. If/when we merge this a 3.3 release is immediately in order, in my opinion (after we fix the again [failing macOS build](https://github.com/solvespace/solvespace/actions/runs/28905261211/job/85750859358#step:3:13)).

But please wait a bit.

I want to see if @jwesthues will take a look at the "alternating knife edges" logic
image

BoykoNeov:
> This is astonishing. @BoykoNeov is it possible to see the conversation/prompts that lead to this? Had you studied the code and problems much prior to engaging the AI? There are bits of spatial reasoning in there that I would not expect an LLM to understand because to me they are visual representations. Here I thought a mathematician dropped in to solve some of our bugs, but how much was you and how much was AI?
 
 I am a very basic programmer and here - practically a complete layman - it was 100% LLMs work. As ruevs noted - it was just born out of a conversation about vibe coding (with which I am experimenting in the past month). Practically theese are all my meaningful prompts lol :
 
solvespace/solvespace#1619 - investigate the issue, replicate if needed, propose and test solutions.
commit this on a branch and draft a PR locally for now
investigate the open ends
add the concave model as a testsuite case
why was the ruevs case not fixed?
ok , work on 1091 now

Here is the trimmed transcript of the sessions, alas - the internal thinking is encrypted and Fable 5 shares relatively few (compared to Opus) of its thaughts verbally.
https://gist.github.com/BoykoNeov/17dd078385c634db338355a75acd1050

If you want to test what Fable can do on other issues, say the word, we have until 12th of July (or when I run out of tokens :) )

ruevs:
> https://gist.github.com/BoykoNeov/17dd078385c634db338355a75acd1050

Yep. The LLM is **_truly_** impressive.

A small example from the log:

```
**The case ruevs pointed at in #1291 — the missing `dotp[0] > TOL && dotp[1] < -TOL` branch — *is* what our fix fixes.** That comment describes exactly the tangent edge-on-edge path in `ClassifyEdge()` where the two dot products have opposite signs and the function used to fall through with `return false` and unwritten outputs. In the #1619 thread ruevs said "I fixed this case, but not [#1091] — it is harder," and then retracted: "My 'fix' is not good in the general case. More work needed." He never published his patch, so I can't say exactly what was wrong with it — but the most likely gap is the one our version handles explicitly: the classification isn't a constant in that branch. It depends on whether the shell's edge is convex or concave (reflex) — convex → both sides `SURF_OUTSIDE`, concave → both sides `SURF_INSIDE`. A fix that always picks one answer works on the common convex model (like the #1619 repro) and silently produces wrong geometry on a concave tangency — which is exactly why the new testsuite case exercises both branches at the same edge.
```

and this is exactly what I missed and exactly why my fix (that I never published) did not work in all cases...

ruevs:
Interestingly this does **not** fix https://github.com/solvespace/solvespace/issues/1452 and https://github.com/solvespace/solvespace/issues/1401

ruevs:
> If you want to test what Fable can do on other issues, say the word, we have until 12th of July (or when I run out of tokens :) )

@BoykoNeov If you want to burn more tokens on SolveSpace, then continuing from this branch (and context?) ask the model to work on https://github.com/solvespace/solvespace/issues/1291 next,  specifically concentrating on the `PrismFilletMinimalTestCase_OnlyOneFailingEdge.slvs` file from here https://github.com/solvespace/solvespace/files/9846083/CylinderTangentFromTheInside_AKA_Fillet_NURBSTests.zip

phkahler:
> ask the model to work on #1291 next

@ruevs and @BoykoNeov 

That is the exact one I was going to point to.

BoykoNeov:
@phkahler @ruevs ok, this evening I will point it to 1291 and we will see. I pointed it now to 1730 
https://github.com/solvespace/solvespace/pull/1730

BoykoNeov:
@phkahler  @ruevs 
  https://github.com/solvespace/solvespace/pull/1731

src/srf/surfinter.cpp
ruevs: As @phkahler pointed out in #1728 - no

phkahler:
@ruevs  I understand the logic now. You provided a picture where many edges are coincident in a single shell. If you go around the edge say 1/2 radius, you'll pass through all the faces that intersect that central edge and you'll alternate inside/outside/inside... (assuming Z depth is correct) Those faces normal vector is going to point CC or CCW depending which side is inside/outside.   So if that's shell A and we want to merge it with shell B, we need to classify every face of B that shares that central edge. For any face of B we just need to find the nearest of the radial faces of A and decide if the B face is on the inside or outside of that face from A. It might also be coincident, in which case we need to know if the normal is same or opposite.

It's a very weird special case, but also very common in CAD. We can of course treat any coincident edges the same way even if shell A only has 2 faces along the edge, those from B can still be classified relative to A with this method. Ummm does that mean it replaces all the existing coincident edge classification?

ruevs:
## This is merged - as part of merging the first nine commits in #1731, the first four of which are the ones from here.

Closing

@phkahler the ones from the eight linked issues (on the right) - that are fully fixed - will have to be closed by hand. I'm out of time for now - please do it. My comment above will be helpful - it lists which test models are fixed.

PR #1730, opened 2026-07-08
Fix missing faces when coplanar faces join edge-on-edge across multiple unions (fixes #1452)

### LINKED ISSUE #1452: Naked edges with only rectangles and extrusion

BoykoNeov:
Fixes #1452.

## The bug

Models that union several cuboids face-to-face, so that many coplanar faces join edge-on-edge across multiple union steps, could end up with an entire side face missing from the NURBS shell (visible as naked edges unless *Force NURBS to triangle mesh* is enabled). Reconstructing the model from the recipe in #1452 — cuboid A extruded down, B extruded up from the same plane, C on top of B, then D sketched on the shared side-face plane and extruded away from the stack — reproduces it reliably: it fails exactly when D's face overlaps the side face of C by more than a sliver, matching the reporter's observations ("if D touches just A and B everything is ok", "if D is just a little bit above C, it still works").

## Root cause

Two cooperating defects in `src/srf/boolean.cpp`:

1. **Phantom vertices from padded merged surfaces.** After each Boolean, `SShell::MergeCoincidentSurfaces()` merges coplanar faces and pads the merged surface's control-point bounds by a tolerance (for STEP export robustness). Exact intersection curves computed in a later Boolean get trimmed to those padded bounds, so they can end just *past* the real geometry (e.g. at z = 10.69 on an edge whose real vertex is at z = 10). `FindVertsOnCurve()` treated the endpoints of the shell's exact curves as vertices and split other curves at those phantom points — but the trims adjacent across the edge are not split there, which produces T-junctions (naked edges) and trim-edge copies with mismatched interior splits.

2. **Intersection edges duplicating the trim boundary.** When an intersection curve runs exactly along an edge of the surface's original trim polygon in the same direction (`SBspUv::Class::EDGE_PARALLEL`), `MakeCopyTrimAgainst()` kept both the intersection edge and the original boundary edge. The keep decision for both reduces to the same predicate (keep iff the region on the in-side of the shared line is kept), so the intersection edge is always redundant. Keeping it was only ever survivable because the two copies used to be split identically, letting `CullExtraneousEdges()` remove the duplicate; with the mismatched splits from (1), the overlapping duplicates survive the cull, `AssemblePolygon()` fails ("failed to assemble polygon to trim nurbs surface in uv space"), and the whole face disappears.

## The fix

* `FindVertsOnCurve()` now collects the shell's vertices from the start/finish points of its surfaces' trims rather than from raw curve endpoints. A curve endpoint the shell actually uses is such a trim endpoint anyway; unused tails past the last real vertex no longer contribute phantom split points.
* `EDGE_PARALLEL` intersection chains are discarded, since the original trim edge (classified identically by the loop above) already supplies whatever trim is needed along that line.

Each commit is independently safe: the first prevents the polygon-assembly failure layer, the second removes the phantom splits that trigger it (and the T-junction leaks).

## Validation

* New regression test `test/group/boolean_coplanar_union` with a model reconstructed from the issue's recipe; it fails on current master (`failed to assemble polygon…`, leaking mesh) and passes with this PR, checking both watertightness and the analytically expected volume (like the tests in #1729).
* The model family was reconstructed programmatically (parameterized cuboid stack) and swept over 2,304 configurations — extrusion directions, workplane-normal orientations, sketch windings, D's plane/height/depth/extent, including all the exact-coincidence heights. Zero failures with the fix; 576 failing configurations without it. The generator is available on request.
* Full test suite passes; also verified stacked on top of #1729 (no conflicts — that PR touches `raycast.cpp`, this one `boolean.cpp`; all 258 tests pass combined).

This addresses a case from the second group of #738 (coincident coplanar faces).

### DIFF
--- a/src/srf/boolean.cpp
+++ b/src/srf/boolean.cpp
@@ -41,26 +41,27 @@ static void FindVertsOnCurve(List<SInter> *l, const SCurve *curve, SShell *sh) {
     Vector amax, amin;
     curve->GetAxisAlignedBounding(&amax, &amin);
 
-    for(const auto &sc : sh->curve) {
-        if(!sc.isExact) continue;
-        
-        Vector cmax, cmin;
-        sc.GetAxisAlignedBounding(&cmax, &cmin);
+    // The vertices of the shell are the endpoints of its trims, not the
+    // endpoints of its curves: an exact intersection curve may extend past
+    // the real geometry on both sides (e.g. out to the padded bounds of a
+    // surface that got enlarged by SShell::MergeCoincidentSurfaces()), and
+    // splitting some other curve at such a phantom point--which is not a
+    // vertex of the trims adjacent across that other curve--produces
+    // T-junctions and naked edges in the triangulated shell (issue #1452).
+    for(const SSurface &ss : sh->surface) {
+        for(const STrimBy &stb : ss.trim) {
+            for(int i = 0; i < 2; i++) {
+                Vector pt = (i == 0) ? stb.start : stb.finish;
+                if(pt.OutsideAndNotOn(amax, amin)) continue;
 
-        if(Vector::BoundingBoxesDisjoint(amax, amin, cmax, cmin)) {
-            // They cannot possibly intersect, no curves to generate
-            continue;
-        }
-        
-        for(int i=0; i<2; i++) {
-            Vector pt = sc.exact.ctrl[ i==0 ? 0 : sc.exact.deg ];
-            double t;
-            curve->exact.ClosestPointTo(pt, &t, /*must converge=*/ false);
-            double d = pt.Minus(curve->exact.PointAt(t)).Magnitude();
-            if((t>LENGTH_EPS) && (t<(1.0-LENGTH_EPS)) && (d < LENGTH_EPS)) {
-                SInter inter;
-                inter.p = pt;
-                l->Add(&inter);
+                double t;
+                curve->exact.ClosestPointTo(pt, &t, /*mustConverge=*/false);
+                double d = pt.Minus(curve->exact.PointAt(t)).Magnitude();
+                if((t > LENGTH_EPS) && (t < (1.0 - LENGTH_EPS)) && (d < LENGTH_EPS)) {
+                    SInter inter = {};
+                    inter.p = pt;
+                    l->Add(&inter);
+                }
             }
         }
     }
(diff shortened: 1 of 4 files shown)


phkahler:
This is absolutely amazing.  @BoykoNeov was this done on top of the previous PR?  We will have to merge these in the correct order.

ruevs:
> was this done on top of the previous PR? We will have to merge these in the correct order.

It's on master. The two PRs are completely orthogonal. This turns out to be a different problem! I'll have to look at the code carefully. But I have no time now.

I tested it and it fixes the [model](https://github.com/solvespace/solvespace/files/14521863/bug-report-naked-edges-with-only-cuboids-extrude-and-unions.zip) from #1452:
image

phkahler:
@ruevs looks like you and Claude were both trying to rebase this? That was strange.

ruevs:
> @ruevs looks like you and Claude were both trying to rebase this? That was strange.

I did the rebasing with my work Git user by mistake. So I rebased it a second time - just to change the "commiter" name and email - Claude was not involved.

Issue #651, opened 2020-07-12
segfault on gtkmm filechooser set_current_name


xzcvczx:
### System information

SolveSpace version: 3.0-c4ca4be9 

Operating system: kiss linux, gtk+3.24.2  

### Expected behavior

run solvespace (gui) click "Sketch image from file"
expect dialog to open to select image

### Actual behavior

solvespace segfaults with no output in console

### Additional information

if you have gtk3 build with debug=yes you get the following message in console and it doesn't crash, however with debug=no it segfaults
```
Gtk-CRITICAL **: 23:56:40.947:
                 gtk_file_chooser_widget_set_current_name: assertion 'priv->action ==
                 GTK_FILE_CHOOSER_ACTION_SAVE || priv->action ==
                 GTK_FILE_CHOOSER_ACTION_CREATE_FOLDER' failed
```

based on the documentation on gtkmm solvespace is using incorrect functions, see #650  for fixes

PR #1733, opened 2026-07-10
Fix Gtk-CRITICAL assertion in file chooser when opening files

AnibalPinto:
Guard SetCurrentName() and SuggestFilename() calls against being invoked on open file dialogs, where GTK's gtk_file_chooser_widget_set_current_name asserts that the action is GTK_FILE_CHOOSER_ACTION_SAVE or GTK_FILE_CHOOSER_ACTION_CREATE_FOLDER.

Follow the same guard pattern already used in FilterChanged() and CheckForUntitledFile() in the same file.

### DIFF
--- a/src/platform/guigtk.cpp
+++ b/src/platform/guigtk.cpp
@@ -1407,7 +1407,9 @@ public:
     }
 
     void SetCurrentName(std::string name) override {
-        gtkChooser->set_current_name(name);
+        if(gtkChooser->get_action() != Gtk::FILE_CHOOSER_ACTION_OPEN) {
+            gtkChooser->set_current_name(name);
+        }
     }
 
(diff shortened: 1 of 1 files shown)


AnibalPinto:
This PR fixes #651

phkahler:
@AnibalPinto   Are you sure?  I tried to reproduce #651 and don't have any problem. I go to Sketch->Image, and I select a png and it just works.

BTW I'm running Gnome on Fedora 42 I think.

phkahler:
@AnibalPinto   I take that back.  I do get the debug message but it doesn't crash. This PR eliminates the message.

Thank you for the contribution!

AnibalPinto:
@phkahler Thanks for the **good first issue** and **Linux** tags :+1:

PR #1737, opened 2026-07-17
SVG export: add 'Z' close-path command for closed contours

### LINKED ISSUE #1368: SVG export: for closed paths, svg path definition should contain a final "Z"
### System information

- **SolveSpace version:** 3.1~70dbde63c
- **Operating system:** windows

### Expected behavior

When exporting a closed sketch to SVG, the svg path definition should have a final "Z" (even if the last point is equal to the first point)

Example:
`<path d='M5.000 5.000 L15.700,56.500 L54.000,49.300 L51.500,10.100 L5.000,5.000  Z' class='s1' /> 
`
### Actual behavior

`<path d='M5.000 5.000 L15.700,56.500 L54.000,49.300 L51.500,10.100 L5.000,5.000' class='s1' />
`

AnibalPinto:
Fixes #1368.

For closed paths (loops), the SVG path definition now ends with 'Z' to explicitly close the path, rather than relying on the last point being equal to the first point.

- Add virtual ClosePath() method to VectorFileWriter base class
- Override ClosePath() in SvgFileWriter to emit ' Z'
- Call ClosePath() in OutputLinesAndMesh when SBezierLoop::IsClosed()
- Add export/svg/closed_path_has_z test case


phkahler:
@AnibalPinto  It might be better to add a "IsClosed" flag for the FinishPath function and pass the flag to that.  Then in the SVG version of FinishPath, check the flag and output the Z as needed there.  I think this is maybe better than adding a new virtual function just for this. This is also something we don't need a test case for.

AnibalPinto:
@phkahler thanks for the notification. It changes more code but avoid more functions.

I tried to use clang to check format but or I am doing something wrong or format isn't used.

phkahler:
Thanks for this. Better file output is always good.

PR #1738, opened 2026-07-21
Datum point isn't a construction entity

### LINKED ISSUE #1064: Datum Point not shown if added in solid group
### System information

- **SolveSpace version:** `3.0~` https://github.com/solvespace/solvespace/commit/4ee869906540343522db76f1ed170e74f3dbed8c

### Expected behavior

*Datum Point* objects (points added via <kbd>P</kbd> shortcut) **added in any group** should be visible if set *"Show points"* + *"Hide construction entities"*.

### Actual behavior

1. Launch SolveSpace;
2. Draw simple rectangle `skecth-in-plane` group;
3. Extrude sketch
4. Add  *Datum Point* **in resulted solid group**;
5. Set *"Show points"* + *"Hide construction entities"*;
6. Added *Datum Point* hides.

In same time *Datum Points* added in initial `skecth-in-plane` or added in new `skecth-in-plane` or `sketc-in-3d` groups are shown as expected.

AnibalPinto:
Fixes #1064

Datum Point isn't a construction entity.

### DIFF
--- a/src/mouse.cpp
+++ b/src/mouse.cpp
@@ -824,7 +824,8 @@ hRequest GraphicsWindow::AddRequest(Request::Type type, bool rememberForUndo) {
     Request r = {};
     r.group = activeGroup;
     Group *g = SK.GetGroup(activeGroup);
-    if(g->type == Group::Type::DRAWING_3D || g->type == Group::Type::DRAWING_WORKPLANE) {
+    if(g->type == Group::Type::DRAWING_3D || g->type == Group::Type::DRAWING_WORKPLANE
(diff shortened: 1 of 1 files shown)


ruevs:
This works and is clean and simple.

In a workplane datum points are indeed not treated as construction and there is a separate button for hiding them, so making them behave the same way in 3d is probably OK.

phkahler:
@ruevs  I was going to say the solution might be to have "show points" override "hide construction" but if you're OK with this fix, so am I.

phkahler:
@AnibalPinto  Thank you for the contribution!

PR #1742, opened 2026-07-24
Web: Enable OpenMP for Emscripten

### LINKED ISSUE #1697: Enable OpenMP in the web version
I don't think we're using multithreading in the web version of Solvespace. Emscripten support for OpenMP is still experimental but we only use some of the more basic features, so this should be possible. This should not involve changing any application code, just build setting and such.

On desktop builds the OMP status is shown at the bottom of the configuration screen. Same should be true of a proper web build.

ruevs:
`cmake/Platform/Emscripten.cmake` is taken from Emsdk 6.0.4 (fe5be6afdff43ad58860d821fcc8572a23f92d19)

Unlike #1734 this works:
image

The OpenMP version is definitely faster.

### DIFF
--- a/.github/scripts/build-web.sh
+++ b/.github/scripts/build-web.sh
@@ -22,6 +22,6 @@ emcmake cmake .. \
   -DENABLE_CLI="OFF" \
   -DENABLE_TESTS="OFF" \
   -DENABLE_COVERAGE="OFF" \
-  -DENABLE_OPENMP="OFF" \
+  -DENABLE_OPENMP="ON" \
   -DENABLE_LTO="ON"
 cmake --build . --config "${BUILD_TYPE}" -j$(nproc) --target solvespace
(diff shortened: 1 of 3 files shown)


ruevs:
The CI works https://github.com/ruevs/solvespace/actions/runs/30106627790 so I'm merging this.

phkahler:
@ruevs   Brilliant!!

phkahler:
It uses 8 threads on my phone!

ruevs:
> It uses 8 threads on my phone!

Same here. But unfortunately in Firefox on Android (tried two different devices) it crashes as soon as I extrude something. Will have to investigate... :-(

ruevs:
> > It uses 8 threads on my phone!
> 
> Same here. But unfortunately in Firefox on Android (tried two different devices) it crashes as soon as I extrude something. Will have to investigate... :-(

Hmmm now it works... And I did not change anything....

phkahler:
I googled "emscripten openmp android crash" perhaps not that order. It suggested build options and something about critical sections. It was all outside my territory, but sounded like using some redundant build flags might help. Maybe openmp still has issues with critical sections or thread local storage. Those can result in intermittent or unpredictable bugs.

Even our first loop - triangulating surfaces in parallel uses a critical section around the line that adds a surface triangle list to the whole shell triangle list.

PR #1746, opened 2026-07-29
Fix #1743 cross vertex split and fix #1291 tangent crossing

### LINKED ISSUE #1291: Rounded corner cut (fillet) issue

### LINKED ISSUE #1743: Edge split failure
This revisits the model from #1291 Rounded Corner cut.

In the first attachment the 1291 fillet sketch is adjusted by aligning the straight lines of the fillet sketch with the pre-existing rectangular box. If the length of the fillet is equal or less than the box thickness, the previous fixes allow it to work without error. Here I have made the fillet extrusion two sided so that it extends past the top of the box. I have also made it longer so that even with it single sided it will extend past the other side of the box. In either case the boolean difference fails.

I believe the problem is that one of the edges of the cube fails to split. Actaully two of them for the same reason. A straight edge of the box intersects the sharp edge of the fillet. That fillet edge lies on one face of the box, but is cut by the adjacent face of the box which is 90 degrees to it and produces a clean intersection and a new vertex along that edge. The edge of the cube however is not split at that point. The cube edge is tangent to both faces of the fillet along that fillet edge, so no intersection is found and there are no other fillet surfaces passing though that point.

Image

[cube_cut_2.zip](https://github.com/user-attachments/files/30440798/cube_cut_2.zip)

It looks like SCurve::MakeCopySplitAgainst() is splitting one of the curves which creates a new vertex. But that vertex also needs to be added to the curve on the other shell which FindVertsOnCurve() would do if it were a preexisting vertex, but it is only found during the current boolean operation so does not exist already.

One solution would be to split curves from each shell agains curves from the other shell. That may produce redundant verticies when - like here - curves are split against a surface and an edge. redundant verticies are removed, but how reliably?

I think the core problem is that the unsplit curve lies in the plane of a surface from the other shell and needs to be intersected with that surfaces trim curves, or at least those curves that have coincident surfaces along the curve.

To offer another example I'll attach another old model that seems to suffer the same failure:

Image

[curve_curve.zip](https://github.com/user-attachments/files/30441042/curve_curve.zip)
This one has a box with a rounded corner along one edge of the top. The curved surface is tangent to the flat box top. A circle is drawn on the box top straddling the line between the flat and curved top and difference extruded down into the box. It appears that some of the curves in the plane of the box top fail to get split along other curves in the same plane. There may be other NURBS failures in addition to that, but I can't be certain.

Now I see that curve_curve can actually work depending on the exact placement of the circle center. I have it constrained to the median line of the first extrusion, but if it's off-axis I think it will be harder to make it work.

BTW I believe a large number of our remaining NURBS failures are similar to this cube_cut_2 issue failing to split curves. Perhaps as many as 20-40 percent of the remaining issues.

ruevs:
This PR combines https://github.com/BoykoNeov/solvespace/pull/5 with https://github.com/BoykoNeov/solvespace/pull/6 rebased on top of it. In addition I added an extra commit that adds [1291_1743_cube_cut_tangent_outside_still_fails_simplified.zip](https://github.com/user-attachments/files/30458058/1291_1743_cube_cut_tangent_outside_still_fails_simplified.zip) as a test case that covers both #1291 and #1743.

The original comments form Claude Opus 5:

# https://github.com/BoykoNeov/solvespace/pull/5

Fixes the tangent-outside case from #1291 — @ruevs' `1291_cube_cut_tangent_outside_still_fails.slvs`, which still failed after the nine commits of #1731 that are now in master.

This continues that work and sits directly on current master. It does **not** include the two commits from #1731 that you decided against (the `SKdNode::FindEdgeOn` relaxation and its tests) — this fix is independent of them, and I rebased it off them so there is nothing to re-litigate here.

### Both of the obvious explanations are wrong

I had two candidates going in, and I want to record that they're dead so nobody spends time on them:

1. **"The tangent intersection curve is never generated"** — the interior-profile-tangency limitation I documented in #1731. It does not apply. The profile cubic has `ctrl[2] = (30, −12.2621)` and `ctrl[3] = (30, 18.7884)`, both at `x = 30`, so the tangency is at the bezier's *endpoint*, and the exact-in-plane path from that PR fires correctly: `EXACT-IN-PLANE (30, 18.7884, 30) → (30, 18.7884, 0)`.
2. **"906cf739 discards it"** (*Don't keep Boolean intersection edges that lie along the trim boundary*) — ablating that commit gives byte-identical output. Not the culprit.

The curve is generated correctly and then **thrown away by misclassification.** It reaches `MakeCopyTrimAgainst` as three chains, all with `KEEP = 0` (`ins = COINC_OPP, outs = OUTSIDE` in one direction, reversed in the other). The cube's `x = +30` face is left with five edges, `AssemblePolygon` fails with `failed: I=5, avoid=4`, and the face disappears from the output.

### Root cause

In `SShell::ClassifyEdge()`'s edge-on-edge case (`src/srf/raycast.cpp`), the `edge_inters == 2` **mixed** branches — where one face's direction cosine is ~0 and the other's is not — assume that a face whose normal is parallel to ours is *coincident* with our surface. It may instead be merely **tangent** at the shell's edge and curving away immediately. First-order data at the point of contact cannot tell those apart.

This is the same class of defect as the both-tangent branch fixed in #1731, in the branches that PR left alone. I flagged it there as out of scope for lack of a model that reached it; @ruevs' file is that model.

### The fix

Extract the both-tangent branch's existing geometry probe as `ProbeTangentFace()` and apply it to the mixed branches too: probe the parallel-normal face a short distance into itself (scaled by that face's own control-net size) and see whether it stays in the tangent plane or leaves it. If it curves away, that side is classified `SURF_INSIDE` / `SURF_OUTSIDE` by the sign of `dev * surf_n·inter_surf_n` rather than being called coincident.

Faces that really are flat take the old path unchanged, so nothing that worked before changes answer.

### Regression test

`test/group/boolean_tangent_crossing/` — @ruevs' file verbatim, checking that the display mesh is watertight (no naked or self-intersecting edges) and that the volume matches. Fails with `(leaks) = true` without the fix. Volume rather than image checks, consistent with the other Boolean cases.

### Verification

- Full suite passes in Debug **and** Release: 263 cases / 931 checks (master is 262 / 925).
- The NURBS result matches the mesh-Boolean ground truth of the same model (`forceToMesh = 1`) **to the last digit**: 38 triangles / volume 204432.7149 / 0 naked edges, against 54 triangles / 204432.7149 / 0 naked edges.
- The 12-model sweep of every repro from #1291, #1619 and #1091 is unchanged, with every previously documented volume still exact.
- Since this branch does *not* carry the naked-edge checker change from #1731, the "0 naked edges" above was re-established with the **unpatched** checker.

### Not addressed

**@phkahler's two-sided-extrusion case.** There is no model attached to the issue for it yet. Flipping a working model's group subtype 7000 → 7001 does produce failures, but that is not his geometry — it doubles the tool length so it pokes out of the top of the cube, and the tangency still involves one tangent and one crossing face rather than being tangent to both fillet faces. Those failures are pre-existing and unchanged by this PR. I'd rather wait for the real model than claim a fix I can't demonstrate.

# https://github.com/BoykoNeov/solvespace/pull/6 

Fixes the `cube_cut_2` case from #1743. Two commits on top of master `790bf74c`.

## Root cause

@phkahler's hypothesis in #1743 is right, and this is the trace that proves it.

Where the two shells **touch** rather than pass through each other, a curve of one shell can cross a curve of the other at a point that is a vertex of neither input shell, and that lies on no surface which either curve crosses transversally. `SCurve::MakeCopySplitAgainst()` splits curves against *surfaces*, so at such a point it splits only whichever of the two curves does meet a transversal surface there, and leaves the other one whole.

The unsplit curve then contributes a single trim edge whose two halves classify differently. That cannot be represented: `MakeCopyTrimAgainst()` assembles the trim polygon from chains that `FindChainAvoiding()` builds, classifies each chain from one representative edge, and keeps or discards it whole. So the edge is lost along with the rest of its chain, `AssemblePolygon()` fails, and the face goes missing.

@ruevs' simplified model is the clearest example. A 2×2×2 triangular prism, minus a tool swept along a bezier profile from (1,0) to (0,1), z ∈ [−0.5, 0.5]. The tool's flat base face lies in the plane y=0, coincident with the prism's y=0 face, and the profile is tangent to y=0 at its (1,0) end. So the tool's corner edge x=1, y=0 runs *within* the prism's y=0 face, and crosses the prism's own bottom-front edge y=0, z=0 at **(1,0,0)**.

The tool's edge does get split there — it crosses the prism's z=0 face transversally. The prism's edge does not: the only tool surfaces it meets at x=1 are the coplanar base face and the tangent swept face, neither of which it crosses. An instrumented run then shows the whole 5-edge boundary of the prism's y=0 face arriving as one chain:

```
ORIGCHAIN A3 n=5 (0,0,0.5)->(2,0,2) ins=CSAME outs=CSAME KEEP=0
...
final: 2 edges
failed: I=3, avoid=3
```

`CSAME` is correct for the half of that boundary inside the coincident patch (x ≤ 1) and wrong for the half outside it — which is exactly the situation the missing split creates.

The asymmetry within the same model is worth noting, because it shows the rule at work: at the profile's *other* end the tool's third face crosses the prism's x=0 face transversally, so **that** prism edge did get split, at y=1. Only the touching end fails.

## The fix

`SShell::SplitCurvesAtCrossings()`, called from `MakeFromBoolean()` immediately after the two `CopyCurvesSplitAgainst()` calls: collect the interior `vertex == true` points of every source-A curve and every source-B curve, and insert each into the other operand's curves wherever it lies within one of their piecewise linear segments (axis-aligned bounding box prefilter, strict 0 < t < 1, within `LENGTH_EPS` of the chord) and — for exact curves — within `LENGTH_EPS` of the exact curve, since the chord of a curved segment is not the curve.

Interior points only, deliberately. A curve's endpoints are its own shell's vertices, which `FindVertsOnCurve()` already propagates from the trims, and an exact intersection curve's endpoints may lie outside the real geometry entirely — which is what #1452 was about. `test/group/boolean_coplanar_union` stays green.

It runs before `MakeIntersectionCurvesAgainst()`, and `RemoveShortSegments()` runs after it, so a point landing very close to an existing one cannot leave a degenerate segment behind (the `Equals` check already rejects anything within `LENGTH_EPS` of an existing point, so the shortest new segment is longer than that).

## Regression test

`test/group/boolean_cross_vertex/` — @phkahler's `cube_cut_2` from the issue, unmodified. Watertight, not self-intersecting, and a volume check.

The expected volume is the value the same model gives when built as a **mesh** Boolean instead of a NURBS one (`Group.forceToMesh`): 203485.224213202 against the NURBS result's 203485.224213201, which is agreement to 2.9e-15 relative. That is the real validation — the constant is not just "whatever this build printed".

No `CHECK_SAVE`/`CHECK_RENDER`, following the other Boolean cases here: the surface cache holds last-ulp floats that `PrepareSavefile` does not normalize.

The committed test against unmodified `boolean.cpp` fails `(inters) = true` with two `failed to assemble polygon to trim nurbs surface in uv space` messages.

## Verification

| model | master | this branch | mesh-Boolean truth |
|---|---|---|---|
| `cube_cut_2` | 28 tris / 171862.60 / **17 naked** | 38 / **202243.1316** / **0 naked** | 56 / 202243.1315 / 0 |
| ruevs' simplified | 22 / **10 naked** | 24 / **6 naked** | 48 / 0 naked |

- Full suite **263 cases / 931 checks**, Debug **and** Release (master baseline 262 / 925).
- The 12 repro models from #1619 / #1091 / #1291 all still exact, every documented volume unchanged to the last digit, Debug and Release identical.
- 5 repeated runs with OpenMP on give a single sorted triangle set for both models.
- `Vector::WithMagnitude(1) of zero vector!` appears three times while this test runs. It is the pre-existing triangulator diagnostic on tangent geometry, not something this patch introduces: on unmodified master `boolean_knife_edge` prints it 18 times, `boolean_tangent_spline` 6, `boolean_tangent_fillet` 3.

## What this does not fix

**ruevs' simplified model still has 6 naked edges on this branch alone.** It needs a second, independent fix — the tangent-crossing classification in `ClassifyEdge()` — and with that stacked on top it comes out clean at 28 triangles, 0 naked. The two defects are genuinely separate: `cube_cut_2` is fixed by this change alone, and the classification change alone leaves `cube_cut_2` at 17 naked edges.

**`curve_curve.slvs` is a third, different defect, and this does not touch it.** It is not the original-curve split at all. The failing curves are the two **intersection** curves on the plane y = 21.873392, which should each be split where they cross the box's x=0 face at z = ∓13.164015. One of them is, by `AllPointsIntersecting()` inside surfinter.cpp's own `MakeCopySplitAgainst()` call — leaving a spurious segment 1.9e-5 long. Its mirror image is not: its nearest piecewise linear point sits 2.5e-5 away from the point it needs. `FindVertsOnCurve()` is not involved anywhere on this model; instrumented, it yields zero candidates. So the gap is about 1e-5 wide and `LENGTH_EPS` is 1e-6 — propagating vertices cannot bridge it, and closing it means changing a tolerance, which I did not want to do speculatively.

So the two models @phkahler attached are **not** two instances of one cause. `cube_cut_2`'s two missing faces are (that is the "two of them for the same reason"); `curve_curve` is a separate accuracy problem in the intersection-curve split.

One more thing worth knowing about `curve_curve`: it is **OpenMP-nondeterministic on master and on every branch here**, cycling through four outcomes (88 tris/4 naked, 75/13, 71/15, 63/17) and deterministic only with `OMP_NUM_THREADS=1`. Single-run comparisons on that model will mislead whoever picks it up.

---

If this looks right, please fetch and fast-forward rather than using the merge button, so the history stays linear for upstream.

### DIFF
--- a/src/srf/boolean.cpp
+++ b/src/srf/boolean.cpp
@@ -203,6 +203,115 @@ SCurve SCurve::MakeCopySplitAgainst(SShell *agnstA, SShell *agnstB,
     return ret;
 }
 
+//-----------------------------------------------------------------------------
+// Insert p as a vertex of the piecewise linear curve sc, if it lies within one
+// of sc's segments (and on sc's exact curve, where we have one) and isn't
+// already one of its points. Returns true if we changed the curve.
+//-----------------------------------------------------------------------------
+static bool InsertVertexIntoCurve(SCurve *sc, Vector p) {
+    for(const SCurvePt &scpt : sc->pts) {
+        if(scpt.p.Equals(p)) return false;
+    }
+
+    int i;
+    for(i = 1; i < sc->pts.n; i++) {
+        Vector a = sc->pts[i-1].p,
+               d = (sc->pts[i].p).Minus(a);
+        double m = d.MagSquared();
+        if(m < LENGTH_EPS*LENGTH_EPS) continue;
+        double t = (p.Minus(a)).Dot(d)/m;
+        if((t <= 0.0) || (t >= 1.0)) continue;
+        if((p.Minus(a.Plus(d.ScaledBy(t)))).Magnitude() > LENGTH_EPS) continue;
+        break;
+    }
+    if(i >= sc->pts.n) return false;
+
+    // The pwl chord of a curved segment is not the curve, so make sure that
+    // the point really lies on the curve and not just near the chord.
+    if(sc->isExact) {
+        double t;
+        sc->exact.ClosestPointTo(p, &t, /*mustConverge=*/false);
+        if((p.Minus(sc->exact.PointAt(t))).Magnitude() > LENGTH_EPS) return false;
+    }
+
+    SCurvePt scpt = {};
+    scpt.p      = p;
+    scpt.vertex = true;
+
+    List<SCurvePt> pts = {};
+    pts.ReserveMore(sc->pts.n + 1);
+    for(int j = 0; j < sc->pts.n; j++) {
+        if(j == i) pts.Add(&scpt);
+        pts.Add(&(sc->pts[j]));
+    }
+    sc->pts.Clear();
+    sc->pts = pts;
+    return true;
+}
+
+//-----------------------------------------------------------------------------
+// A curve of one shell may cross a curve of the other shell at a point that is
+// a vertex of neither input shell, and that lies on no surface which either
+// curve crosses transversally; that happens where the two shells touch without
+// passing through each other, for example where an edge of one runs within a
+// coincident face of the other. MakeCopySplitAgainst() splits curves against
+// surfaces, so at such a point it splits only whichever of the two curves does
+// meet a transversal surface there, and leaves the other one whole. The
+// unsplit curve then contributes a single trim edge whose two halves classify
+// differently, which cannot be represented: the edge is kept or discarded
+// whole, and the trim polygon fails to assemble, so the face is lost.
+//
+// So propagate the splits that we just made. Any vertex that splitting created
+// within a curve from one operand, and that lies within a curve from the other
+// operand, becomes a vertex of that curve too.
+//-----------------------------------------------------------------------------
+void SShell::SplitCurvesAtCrossings() {
+    List<Vector> vert[2] = {};
+    for(SCurve &sc : curve) {
+        int i;
+        if(sc.source == SCurve::Source::A) {
+            i = 0;
+        } else if(sc.source == SCurve::Source::B) {
+            i = 1;
+        } else {
+            continue;
+        }
+        // Only the interior points that a split created; the endpoints of a
+        // curve are its shell's own vertices, and an exact curve's endpoints
+        // may lie outside the real geometry entirely (issue #1452).
+        for(int j = 1; j < sc.pts.n - 1; j++) {
+            if(sc.pts[j].vertex) vert[i].Add(&(sc.pts[j].p));
+        }
+    }
+
+    if(!vert[0].IsEmpty() || !vert[1].IsEmpty()) {
+        for(SCurve &sc : curve) {
+            int i;
+            if(sc.source == SCurve::Source::A) {
+                i = 1;
+            } else if(sc.source == SCurve::Source::B) {
+                i = 0;
+            } else {
+                continue;
+            }
+
+            Vector cmax = {VERY_NEGATIVE, VERY_NEGATIVE, VERY_NEGATIVE},
+                   cmin = {VERY_POSITIVE, VERY_POSITIVE, VERY_POSITIVE};
+            for(const SCurvePt &scpt : sc.pts) {
+                scpt.p.MakeMaxMin(&cmax, &cmin);
+            }
+
+            for(const Vector &p : vert[i]) {
+                if(p.OutsideAndNotOn(cmax, cmin)) continue;
+                InsertVertexIntoCurve(&sc, p);
+            }
+        }
+    }
+
+    vert[0].Clear();
+    vert[1].Clear();
+}
+
 void SShell::CopyCurvesSplitAgainst(bool opA, SShell *agnst, SShell *into) {
 #pragma omp parallel for
     for(int i=0; i<curve.n; i++) {
@@ -858,6 +967,11 @@ void SShell::MakeFromBoolean(SShell *a, SShell *b, SSurface::CombineAs type) {
     // shell.
     a->CopyCurvesSplitAgainst(/*opA=*/true,  b, this);
     b->CopyCurvesSplitAgainst(/*opA=*/false, a, this);
+
+    // Where the two shells touch without crossing, a curve of one shell can
+    // cross a curve of the other at a point that splitting against surfaces
+    // finds on only one of the two curves; carry those points across.
+    SplitCurvesAtCrossings();
 
     // Generate the intersection curves for each surface in A against all
     // the surfaces in B (which is all of the intersection curves).
--- a/src/srf/surface.h
+++ b/src/srf/surface.h
@@ -402,6 +402,7 @@ public:
     void MakeFromIntersectionOf(SShell *a, SShell *b);
     void MakeFromBoolean(SShell *a, SShell *b, SSurface::CombineAs type);
     void CopyCurvesSplitAgainst(bool opA, SShell *agnst, SShell *into);
+    void SplitCurvesAtCrossings();
     void CopySurfacesTrimAgainst(SShell *sha, SShell *shb, SShell *into, SSurface::CombineAs type);
     void MakeIntersectionCurvesAgainst(SShell *against, SShell *into);
     void MakeClassifyingBsps(SShell *useCurvesFrom);
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -66,6 +66,7 @@ set(testsuite_SOURCES
     request/workplane/test.cpp
     group/boolean_coplanar_union/test.cpp
     group/boolean_cross_vertex/test.cpp
+    group/boolean_cross_vertex_tangent_crossing/test.cpp
     group/boolean_knife_edge/test.cpp
     group/boolean_tangent_crossing/test.cpp
     group/boolean_tangent_edge/test.cpp
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -65,6 +65,7 @@ set(testsuite_SOURCES
     request/ttf_text/test.cpp
     request/workplane/test.cpp
     group/boolean_coplanar_union/test.cpp
+    group/boolean_cross_vertex/test.cpp
     group/boolean_knife_edge/test.cpp
     group/boolean_tangent_crossing/test.cpp
     group/boolean_tangent_edge/test.cpp
--- a/test/group/boolean_cross_vertex/normal.slvs
+++ b/test/group/boolean_cross_vertex/normal.slvs
@@ -1,0 +1,2130 @@
+±²³SolveSpaceREVa
+
+
+Group.h.v=00000001
+Group.type=5000
+Group.name=#references
+Group.color=ff000000
+Group.skipFirst=0
+Group.predef.swapUV=0
+Group.predef.negateU=0
+Group.predef.negateV=0
+Group.visible=1
+Group.suppress=0
+Group.relaxConstraints=0
+Group.allowRedundant=0
+Group.allDimsReference=0
+Group.scale=1.00000000000000000000
+Group.remap={
+}
+AddGroup
+
+Group.h.v=00000002
+Group.type=5001
+Group.order=1
+Group.name=sketch-square
+Group.activeWorkplane.v=80020000
+Group.color=ff000000
+Group.subtype=6000
+Group.skipFirst=0
+Group.predef.q.w=1.00000000000000000000
+Group.predef.origin.v=00010001
+Group.predef.swapUV=0
+Group.predef.negateU=0
+Group.predef.negateV=0
+Group.visible=1
+Group.suppress=0
+Group.relaxConstraints=0
+Group.allowRedundant=0
+Group.allDimsReference=0
+Group.scale=1.00000000000000000000
+Group.remap={
+}
+AddGroup
+
+Group.h.v=00000003
+Group.type=5100
+Group.order=2
+Group.name=extrude-cube
+Group.opA.v=00000002
+Group.color=00646464
+Group.subtype=7001
+Group.skipFirst=0
+Group.predef.entityB.v=80020000
+Group.predef.swapUV=0
+Group.predef.negateU=0
+Group.predef.negateV=0
+Group.visible=1
+Group.suppress=0
+Group.relaxConstraints=0
+Group.allowRedundant=0
+Group.allDimsReference=0
+Group.scale=1.00000000000000000000
+Group.remap={
+    1 00040000 1002
+    2 00040001 1002
+    3 00040002 1002
+    4 00040000 1001
+    5 00040001 1001
+    6 00040002 1001
+    7 00040000 1004
+    8 00040001 1003
+    9 00040002 1003
+    10 00050000 1002
+    11 00050001 1002
+    12 00050002 1002
+    13 00050000 1001
+    14 00050001 1001
+    15 00050002 1001
+    16 00050000 1004
+    17 00050001 1003
+    18 00050002 1003
+    19 00060000 1002
+    20 00060001 1002
+    21 00060002 1002
+    22 00060000 1001
+    23 00060001 1001
+    24 00060002 1001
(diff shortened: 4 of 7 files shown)


ruevs:
@phkahler one of the fixes here: 2ecb3417af4c91bb4e558d943d7ae7cd44013200 is right up your alley. You worked on that code and logic a lot in the past. I think the new `SplitCurvesAtCrossings` makes a lot of sense and I'll make a detailed review with comments here but I hope you will take a look as well.

What is more I think this commit may result in a performance regression similar to ab10e38d44b26de3ee64b60db0d154278d1ee8ac that you later fixed in 0f1ece2b8e76f52cc675c0c3349749eb2e0d9470 .

phkahler:
@ruevs  I had thoughts about doing this "shove the vertex into the other curve" fix but realized there is no simple way to find the other curve than search. This implementation looks slow but worth trying. It probably won't work when both shells have that kind of knife-edge and those intersect since *neither* curve will find the vertex.

I was leaning toward modifying SSurface::PointOnCurve()  but I see the culprit might be SSurface::PointIntersectingLine()   The later is spitting out a lot of "didn't converge (surface intersecting line)" and "parallel (surface intersecting line)" messages. I think because it's trying to intersect the straight edge with the edge of the curved surface.

Just thinking out loud, I propose that (those) function, instead of failing if the point is near the edge, test for curve-curve intersection with the curve of the surface edge taken from control points. If that fails, then let the function fail. If there is an intersection it would be valid and should be returned. This would allow the vertex to be found on the other trim without adding any overhead other than trying a new test only when that function fails. If should also find the vertex in the case of two shells meeting along edges like that.

Another observation - that function does not seem to limit the range of u,v coordinates within the surface while its iterating. Not sure if that might be a problem. A point might fly far away in some cases.

phkahler:
For now I only want the BoykoNeov#5  which is the first two commits here.

ruevs:
https://github.com/BoykoNeov/solvespace/pull/5 is now merged in master, so rebased leaving only https://github.com/BoykoNeov/solvespace/pull/6 "Split Boolean curves where they cross a curve of the other shell. Fixes the cube_cut_2 case from #1743"

src/srf/boolean.cpp
```diff
@@ -203,6 +203,115 @@ SCurve SCurve::MakeCopySplitAgainst(SShell *agnstA, SShell *agnstB,
...
 }
 
+//-----------------------------------------------------------------------------
+// Insert p as a vertex of the piecewise linear curve sc, if it lies within one
+// of sc's segments (and on sc's exact curve, where we have one) and isn't
+// already one of its points. Returns true if we changed the curve.
+//-----------------------------------------------------------------------------
+static bool InsertVertexIntoCurve(SCurve *sc, Vector p) {
+    for(const SCurvePt &scpt : sc->pts) {
+        if(scpt.p.Equals(p)) return false;
```
ruevs, line 213: The point is a vertex of the curve already - nothing to do.

src/srf/boolean.cpp
```diff
@@ -203,6 +203,115 @@ SCurve SCurve::MakeCopySplitAgainst(SShell *agnstA, SShell *agnstB,
...
+    }
+
+    int i;
+    for(i = 1; i < sc->pts.n; i++) {
+        Vector a = sc->pts[i-1].p,
+               d = (sc->pts[i].p).Minus(a);
+        double m = d.MagSquared();
+        if(m < LENGTH_EPS*LENGTH_EPS) continue;
```
ruevs, line 221: The current and next vertex are too close - skip.

src/srf/boolean.cpp
```diff
@@ -203,6 +203,115 @@ SCurve SCurve::MakeCopySplitAgainst(SShell *agnstA, SShell *agnstB,
...
+        double t = (p.Minus(a)).Dot(d)/m;
```
ruevs, line 222: `t` is the location closest to `p` between the current vertex and the next vertex expressed as a range [0,1]. Can be thought of as the location of `p` projected onto the line segment between the current vertex and the next vertex.

src/srf/boolean.cpp
```diff
@@ -203,6 +203,115 @@ SCurve SCurve::MakeCopySplitAgainst(SShell *agnstA, SShell *agnstB,
...
+        if((t <= 0.0) || (t >= 1.0)) continue;
+        if((p.Minus(a.Plus(d.ScaledBy(t)))).Magnitude() > LENGTH_EPS) continue;
```
ruevs, line 224: The `p` and it's projected location of the curve segment are not close enough - skip.

src/srf/boolean.cpp
```diff
@@ -203,6 +203,115 @@ SCurve SCurve::MakeCopySplitAgainst(SShell *agnstA, SShell *agnstB,
...
+        break;
+    }
+    if(i >= sc->pts.n) return false;
```
ruevs, line 227: At this point `i` is either the index of the vertex after which we should insert `p` or has reached the end of the list - in which case do not insert it.

src/srf/boolean.cpp
```diff
@@ -203,6 +203,115 @@ SCurve SCurve::MakeCopySplitAgainst(SShell *agnstA, SShell *agnstB,
...
+
+    // The pwl chord of a curved segment is not the curve, so make sure that
+    // the point really lies on the curve and not just near the chord.
+    if(sc->isExact) {
+        double t;
+        sc->exact.ClosestPointTo(p, &t, /*mustConverge=*/false);
+        if((p.Minus(sc->exact.PointAt(t))).Magnitude() > LENGTH_EPS) return false;
```
ruevs, line 234: `p` was close enough to the curve PWL line segment (the chord) but is not close enough to the exact curve - do not insert it.

src/srf/boolean.cpp
```diff
@@ -203,6 +203,115 @@ SCurve SCurve::MakeCopySplitAgainst(SShell *agnstA, SShell *agnstB,
...
+    }
+
+    SCurvePt scpt = {};
```
ruevs, line 237: Actually insert the point.

src/srf/boolean.cpp
```diff
@@ -203,6 +203,115 @@ SCurve SCurve::MakeCopySplitAgainst(SShell *agnstA, SShell *agnstB,
...
+            }
+
+            Vector cmax = {VERY_NEGATIVE, VERY_NEGATIVE, VERY_NEGATIVE},
+                   cmin = {VERY_POSITIVE, VERY_POSITIVE, VERY_POSITIVE};
+            for(const SCurvePt &scpt : sc.pts) {
+                scpt.p.MakeMaxMin(&cmax, &cmin);
+            }
+
+            for(const Vector &p : vert[i]) {
+                if(p.OutsideAndNotOn(cmax, cmin)) continue;
```
ruevs, line 305: The above few lines are essentially a bounding box check.

ruevs (commented):
@phkahler I reviewed this and it actually looks perfectly sensible to me. I still have a nagging feeling of "copy-paste" - in the sense that this logic can be combined with something else in this file, but I have not spent the time to confirm or reject the "nagging feeling".

phkahler:
@ruevs  When I read it I was worried about a performance regression too. But with this rebased and as PR here I can try it out tonight or tomorrow. Thank you for the rebase, it will make testing easy.

Last night I started working on a fix for what I believe to be the root of the problem [right here](https://github.com/solvespace/solvespace/blob/220c144344ee0509b5bc936cdc2dce2807ec189a/src/srf/raycast.cpp#L223)

When a PWL segment is tested for intersection against a surface, the code above that comment calls PointIntersectingLine() which bails out with debug messages if they become parallel. That function returns false, and this caller in raycast.cpp does nothing. It turns out if you print the u,v coordinates in there it's close to 0, 0.5   or 1, 0.5 with your test sketch. It is essentially another unhandled case.

So my plan is to use the nearest edge of the surface (based on U,V) and do a curve-curve intersection to see if we can find it. If not, we will still fail after having looked at the edge of the surface. This check will only be done where the tangency occurs so there will be no impact on performance except when we check for this case. Because this will be done at the lowest level of curve-surface intersection, it will make the vertex pass through all the regular checks - like is it inside the trim polygon of the surface.

I think there is another case where two fillet cuts intersect and neither edge will be split with the existing code. My proposed edge-edge intersection should find it.

Give me a day or two to test this PR and try my alternative.

phkahler:
This has been superseded by #1751.  Still an interesting PR, but the root cause of the issues is the failures to find certain intersections. This is more of a patch it after-the-fact.

I don't think it will be able to split the edges in #1755 because both shells have all surfaces tangent at one vertex. The issue in that is different, but the edges do get split by #1751 which is now on master.

phkahler:
Closing

Issue #1723, opened 2026-06-04
unstable redundant constraints


vthriller:
(huh, issue template missing? anyway...)

[bug-parallel-redundant-not-always.slvs.gz](https://github.com/user-attachments/files/28580355/bug-parallel-redundant-not-always.slvs.gz)

- Mark `r00b` and `r00e` as parallel. This is immediately marked as redundant.
- Do the same for the other pair of lines, `r00c` and `r010`. No immediate message about redundancy, it only shows when you try to drag any of the points.
  - It also shows when followed by immediate undo-redo cycle.
  - Also, while dragging, c014-eq-length is constantly popping in and out of redundant constraints list.

Reproducible with 3.1`~`70bde63c and 3.2`~`27b6a080.

ruevs:
Indeed. And if one moves the (already fully constrained apart from moving and rotating) sketch around a bit the two sets of line segments can switch places - in other words `r00c` and `r010` can react immediately and `r00b` and `r00e` only after moving.

I'm not an expert on how the constraint solver works but I suspect that numerically in one case the [rank](https://github.com/solvespace/solvespace/blob/master/src/system.cpp#L249) [test](https://github.com/solvespace/solvespace/blob/master/src/system.cpp#L505) of the matrix (somehow) works out OK - because the lines are already parallel - and in the other it does not...

Someone with better knowledge will have to say e.g. @Evil-Spirit, @jwesthues or perhaps @iscgar. The solver is a very stable piece of code that [few people have had to touch](https://github.com/solvespace/solvespace/commits/master/src/system.cpp) in the last 10 years - at least not in a way that actually changes how the core logic woks.

ruevs:
This could be related https://github.com/solvespace/solvespace/issues/21#issuecomment-258877279

Evil-Spirit:
> The solver is a very stable piece of code

Without a thousands tests - not. So bugs can happen, especially in rank test under specal conditions. We can solve specific problem, but without test coverage we can't guarantee this will not break something in antoher place (not introducing false positive redundancy)

Issue #1466, opened 2024-06-15
constrain symmetric causing "constraints are incompatible"


wolfv6:
### System information

- **SolveSpace version:** 3.1~70bde63c

- **Operating system:** Fedora 40

### Expected behavior

The assembly should have constrained without error. 

### Actual behavior

The error message in the Property Browser said:
the following constraints are incompatible
- same-orientation
- pt-on-line

### Additional information

The error is reproduced in the attached [box_on_foor.zip](https://github.com/user-attachments/files/15849354/box_on_foor.zip) > box_on_floor-asm_error.slvs
I think the error has something to do with the "constrain symmetric" in box.slvs

Constraining the other and of the box in the same way works fine in box_on_floor-asm.slvs

On https://solvespace.com/forum.pl?action=viewthread&parent=5754, Tom said it worked for him using 3.0, so maybe it's a new bug?

wolfv6:
Also getting same error constraining assembly that does not use "Constrain symmetric".

The error is reproduced in the attached [box_on_foor_3.zip](https://github.com/user-attachments/files/15854970/box_on_foor_3.zip) > box_on_floor-asm_error.slvs

Constraining a different point in the same way works fine in box_on_floor-asm.slvs

ruevs:
This is a duplicate to #246 #1105 #1354

It boils down to the solver having "convergence" problems with "large" objects. If you decrease the size of box.slvs to 127x40 and the size of floor.slvs to 152 it works fine.

There was a discussion on this topic recently on the forum https://solvespace.com/forum.pl?action=viewthread&parent=5740&tt=1718184803

wolfv6:
Thanks for the links.

The dimensions of the box and floor where already smaller than 127x40 and 152 x 152 (inches).
What units are you using?

ruevs:
Millimeters - internally everything is in millimeters :-)

wolfv6:
A simple work around is to use mm or .1-inch were inches are meant.
Should I close this issue?

ruevs:
Let it stay open. We can easily close them all when this is fixed.

Issue #1378, opened 2023-06-09
Arc length constraints not applying correctly until next invocation of solve


thekakkun:
### System information

- **SolveSpace version:** 3.1~70bde63c
- **Operating system:** Windows 10

### Expected behavior

Constraints regarding arc lengths apply upon solve.

### Actual behavior

There are cases in which constraints regarding arc lengths aren't solved when the values are set, but only when the solver is called again.

[arcs.zip](https://github.com/solvespace/solvespace/files/11709820/arcs.zip)

In the attached file, updating the ratio constraint between arcs to be 5:1 will not actually apply correctly. The constraint is solved, only once the solver is called again (by dragging, adding a new constraint, etc.)

### Additional information

I believe that the constraint formulas are based off the initial central angle of the arc, and different formulas are used depending on whether that angle is greater/less than 180 degrees. If that threshold is crossed during the solve step, the formula is not updated, so the solver ends up solving for the incorrect arc.

Issue #1354, opened 2023-03-27
Large Dimensions result in Incompatible Constraints


zegelin:
Trying to model large systems that have dimensions in metres causes the certain constraint types to fail with an "Incompatible Constraints" error. For example, two 4 metre long lines cannot be constrained perpendicular to each other, as shown in the screenshot below.

(I have View -> Dimension Units set to "Dimensions in Meters".)

![Capture](https://user-images.githubusercontent.com/19296634/227827807-d1402481-36ae-4b9a-9c57-3e258d1726f5.PNG)

Are there dimensional limits to the solver? E.g. keep everything smaller than X mm?
A work around is to model everything at a smaller scale and export with a scale factor, but it would be nice to have the model match the project scale.

### System information

- **SolveSpace version:** 3.1
- **Operating system:** Windows 10

### Expected behavior

Requests and their constraints spanning large distances should work.

### Actual behavior

Requests and their constraints spanning large distances fail with an "Incompatible Constraints" error.

### Additional information

[Incompatible Constraints Bug.zip](https://github.com/solvespace/solvespace/files/11074149/Incompatible.Constraints.Bug.zip)

ruevs:
That's an interesting bug in the solver. Thank you for the minimal reproduction model.

The interesting part is that if one deletes the constraint "c002-pt-on-line" and recreates it then it works irrespective of how long one makes the line segments. In other words if the perpendicular constraint is created first and the point on line after that it is OK, Otherwise it breaks.

Alternatively if **one** of the dimensions is decreased to 1851mm or less it works... and breaks again of it goes above.

This needs to be debugged. Useful code https://github.com/solvespace/solvespace/issues/1247#issuecomment-1147481539

ruevs:
#1105 for cross reference.

zwn:
Another way the problem can be recreated is by drawing 10m x 10m x 2m box as a part, importing it into an an empty file and constrain any point on origin.

> I rebased https://github.com/Evil-Spirit/solvespace-master/commits/solver-debug onto master here: https://github.com/ruevs/solvespace/tree/EvilSpirit_solver-debug it is interesting to watch the solve fail, and see the equations but no immediate help. I need to think/debug much more...
> 
> #226 is the original issue for which this code was created.

I tried the code on my failing case:

- [Debug / Equations][1]
- [Debug / Folded Equations][2]

But `Debug / Show Solve Way` is not showing any blue lines. I've seen them only once but I am not able to recreate it anymore.

[...]

PR #1748, opened 2026-07-29
Damp the Newton step in the solver when the full step makes things worse. Fixes #1247

### LINKED ISSUE #1247: SolveSpace fails to solve solvable constraints

ruevs:
# Based on https://github.com/BoykoNeov/solvespace/pull/3. Original PR text from Claude Opus 5:

Fixes #1247 — @tc7000's sketch where changing an angle dimension from 60° to 30° reports "didn't converge", while 60° → 40° is fine, and the same 30° works if you approach it in smaller steps.

### It is not a rank or redundancy problem

That was my first hypothesis, and it is wrong. `rankOk` is true at every failure, and `TestRank` / `CalculateRank` never fire on this model — the `DIDNT_CONVERGE` comes straight out of `System::NewtonSolve()`. Worth saying explicitly, because the symptoms (a threshold that depends on the *previous* value, order dependence) look exactly like a redundancy problem and it's a natural place to start digging.

### Root cause: an undamped Newton step lands on a critical point

After substitution this sketch reduces to `m = 2, n = 4`, effectively one-dimensional in the height `h` of the triangle, with

```
```

`f'(0) = 0` — the function is flat at the origin. Solving for 30° means finding `h = 57.735`, and from the 60° solution the full Newton step overshoots it and lands at `h ≈ 4.2`, right on top of that critical point. With the derivative there near zero, the next steps are 161 mm, then 349 mm, and the geometry runs out to ±1.9e6 mm, where the parameters are so large that steps collapse to ~1e-16 and the solver gives up with the residual still at `cos 30°`.

60° → 40° lands at `h = 50`, where the derivative is healthy, and converges in four iterations. That is the whole "threshold-shaped, order-dependent" story: whether the full step happens to clear the flat spot.

### The fix

A backtracking line search in `System::NewtonSolve()` (`src/system.cpp`): try `relax = 1, ½, ¼ … 1/128`, and accept the first that strictly reduces `‖F‖₂²`. **If none does, take the full step exactly as before** — so no step the old code would have taken is ever rejected, and a model that converges today cannot start failing because of this.

`IsReasonable()` failures now shrink the step instead of aborting the solve outright. (Note for anyone reading that function: it is misnamed upstream — it returns **true** when the value is *un*reasonable.)

### Regression test

`test/core/solver/angle_step_over_critical_point` — set 60°, then 30°, and check both solve and that the triangle really has the right height, not just that a result code came back OKAY. Fails at the 30° step without the fix.

The repro is path-dependent: the previous solution is the next solve's initial guess, so it cannot be reproduced by editing the `.slvs` and reloading. The second commit therefore teaches `test/debugtool.cpp` to load a sketch and then set a dimension and re-solve repeatedly (`solve`), and to sweep a range of start/target pairs (`sweep`). That is what produced the numbers above and it is useful for any solver issue of this shape.

### Verification

- A sweep of 138 (start, target) angle pairs: **all 32 previously-failing pairs now solve, with zero regressions** among the rest.
- Full suite passes in Debug **and** Release: 263 cases / 930 checks (master is 262 / 925).
- With `src/system.cpp` reverted, the new test fails at the 30° step.

### Concerns you may want to raise

- **The merit function mixes units.** `‖F‖₂²` sums dimensionless cosine residuals from angle constraints — which the `mult` factor can gain up to ~1000× near 0°/180° — together with mm-valued distance residuals. This is harmless for a decrease-only acceptance test with a full-step fallback (the worst case is that a good step is judged "not an improvement" and we fall back to exactly today's behaviour), but it is not a principled merit function, and a residual scaling pass would be a real improvement. I did not attempt one.
- **One visible behaviour change in the sweep:** 59° → 25° now lands on +25° where master lands on −25°. Both are exact, mirrored solutions of the same constraint system, and the baseline already flips arbitrarily depending on the starting angle — but it is a difference, so I'd rather point at it than have it found later.
- **1/128 and 8 tries are arbitrary.** They were enough for every pair in the sweep; there's no theory behind the specific bound.

This PR is **independent of my other three** and can be taken on its own. In particular, note that a failed solve on this branch still deletes the constraints that failed — that is a separate bug with its own PR, deliberately not mixed in here.


ruevs:
Keep in mind the discussion:

https://github.com/BoykoNeov/solvespace/pull/3#issuecomment-5118302555
https://github.com/BoykoNeov/solvespace/pull/3#issuecomment-5121902740

BoykoNeov:

@ruevs Your objection sent me back to measure, and it turned into a better fix than this one. It is on my fork as `fix-1247-trust-region` (BoykoNeov/solvespace#7). **Which of the two lands is your call** — you opened this PR and you have already tested it; I am not asking you to close anything. But the two are either/or rather than stackable, so the comparison is worth having in one place.

**You were right about the mechanism and, I think, wrong about the price.** You said you could not see how to get predictability "apart from internally *animating* each change in smaller steps and solving multiple times", and did not like it. The steps that need to be smaller are not the user's edit — they are the solver's Newton steps. Shrinking *those* costs almost nothing: `EvalJacobian()` and `SolveLeastSquares()` stay outside the retry loop, so a rejected trial step is one residual evaluation, with no Jacobian and no factorization. No intermediate solves, no regeneration, nothing to see on screen.

So instead of the line search in this PR, `NewtonSolve()` keeps a radius it trusts the linearization over: clip each step to it, double it when a step strictly reduces `‖F‖₂²`, quarter it when one does not, up to 12 tries. The radius starts at **half** the first Newton step. If nothing along the direction improves, take the full step exactly as the old code does.

## Measured against master, same 1° grid of start × target angle (6480 pairs, driven through the edit-a-dimension-and-re-solve path)

| | pairs solved | flips, on the 5662 master also solves | regressions |
|---|---|---|---|
| master | 5662 | **619** | — |
| this PR (line search) | 6480 | 547 | 0 |
| trust region | **6480** | **0** | **0** |

Scoped deliberately: **zero flips wherever master has an answer at all.** Across the whole grid there are still 417, and every one of them is inside the 818 pairs master cannot solve — there is no previous behaviour there to be surprising relative to. So it is a large improvement on the axis you named, not flip-freedom.

Two things I should not let you discover on your own:

**It is not a guarantee.** Animating the constraint value is continuation in the parameter, and under regularity that genuinely tracks a branch. A trust region limits step length in the *state* and tracks nothing — it exploits the same locality intuition without the theorem. 85° → 30° still crosses to the mirror: the offending step is a legitimate descent step that even a textbook ρ-ratio rule accepts. Landing on the nearer root is a policy preference, and no branch-blind descent method gets it for free.

**It gives up this PR's best safety argument.** The line search only engages when the full step increases the residual, so its set of trajectories is a strict superset of today's — that is a complete argument by itself, and it is why I would understand you preferring it. The trust region clips the first step unconditionally, so **every solve takes a different path**, and the defence is empirical: 0 regressions on the grid above, a 12-model NURBS boolean corpus whose output is byte-identical to master (triangle counts and volumes to every digit), suite iteration count 957 against master's 956 over the 263 pre-existing cases, and green in Debug and Release. Suite is 265 cases / 941 checks with two new regression cases; both fail on master's `system.cpp`.

Cost in the regime that actually matters, since `NewtonSolve()` runs once per mouse-move frame and the radius is a fresh local each call: **+1 Newton iteration per frame** (2.00 → 3.00 over 40 small monotone increments). Caveat — that is a dimension increment, so `sys.dragged` is empty and the reduced system is not identical to a real drag. Right regime, but a proxy. Choosing to halve the first step rather than divide it by 8 came out of that measurement; the headline result above is unchanged for every divisor ≥ 2, so it is not a tuned constant.

If that per-frame iteration bothers you, there is an obvious lever I chose not to pull, and you are better placed to judge it than I am. `sys.dragged` is non-empty exactly during an interactive drag, so the radius could start at the *full* Newton step while dragging and at half of it otherwise — parity with master in the regime where the branch preference matters least (a drag moves in tiny increments, so there is no big jump to overshoot), keeping the near-root landing for the case that actually produces it, a typed dimension edit. I left it out deliberately: it would ship an interactive code path that my harness cannot exercise even in principle — a dimension increment leaves `dragged` empty, which is precisely the branch such a conditional would *not* take — and it turns one constant into two UI-dependent regimes. If you think it is worth the frame, that is the change, and you can test it by dragging in the GUI in a way I cannot.

Finally, a retraction. On BoykoNeov/solvespace#3 I offered you a post-solve branch preference and said it would need "no double-solving". That was wrong on its own terms — picking a different mirror after the fact needs either a second solve or a reflection, and "the mirror" only exists for specially structured systems. It also turned out to be unnecessary: the step rule delivers the preference, so there is nothing left to select afterwards.

One correction to the record in the other direction, offered as data rather than as a rebuttal: your specific case is exactly as you describe — 59° → 25° flips here and not on master. But master flips on 619 of the 5662 pairs it solves, so the predictability is not a property it has today either; both solvers land where the iteration happens to take them, and the trust region is the first of the three that lands there on purpose.

jwesthues:
The damping is a textbook solution to nonconvergence in a Newton's method. I don't think there should be any significant speed penalty in cases where this isn't helping, since the inner loop stepping the relaxation factor will then exit after the first iteration. So this looks like a basically promising approach to me.

In the case where the minimum step still makes the error worse, this code takes the maximum step and keeps iterating. It justifies that as a fallback to the previous behavior so that "nothing that converges now can start failing", but I don't think that's obviously true--if we took a damped step on a prior outer iteration then we're now linearized about a different point and who knows what might happen.

So my intuition would probably have been to take the minimum step in that case, or just fail entirely. Their minimum step size seems reasonable to me, though it would be interesting to see the actual step sizes on a diverse set of problematic sketches.

Is there any benefit to still checking (the misnamed) `IsReasonable()` now that we check that the squared error decreased? That would deserve a comment explaining why NaN is handled correctly, but I think it is (since NaN < err is false; the tricky part is that NaN >= err is also false).

ruevs:
https://github.com/BoykoNeov/solvespace/pull/7 / https://github.com/solvespace/solvespace/pull/1749 is a better solution [in my opinion](https://github.com/BoykoNeov/solvespace/pull/7#issuecomment-5133145847), so I'm converting this to a draft for now.

BoykoNeov:

@jwesthues — a correction to my comment above, and one to a claim I made for this PR.

I wrote that the line search "only engages when the full step increases the residual, so its set of trajectories is a strict superset of today's — that is a complete argument by itself." **That is wrong.** You had already said why, in this thread, before I measured anything:

> I don't think that's obviously true--if we took a damped step on a prior outer iteration then we're now linearized about a different point and who knows what might happen.

You were right. On a breadth sweep I ran for #1749, this branch regresses `problem.slvs` from #1105 — constraint c4 set to 5°, master `OKAY` in 11 iterations, this branch `DIDNT_CONVERGE` — on a pristine, uninstrumented build, in both Debug and Release. (I built from my own `fix-1247-newton-damping`, and checked its `system.cpp` is byte-identical to this PR's head `c375ae2d` before making any claim about your code.) One damped step relinearizes about a different point and the rest of the trajectory is no longer master's. The safety argument has to be withdrawn from both PRs.

Also superseded from that comment: the cost figure "957 against 956" was measured on the test suite, which is 263 mostly-trivial solves. Over 5662 real solves it is **+5.6%** Jacobian+least-squares for the trust region (4.47 → 4.72 per solve). That is the number to use.

None of this makes this PR the worse of the two — on breadth it is clearly the better one. Two separate corpora, and it matters which is which, because the regression above is in the second one:

- **664 in-repo fixture cases** over 31 distinct sketches — this branch **0** regressions, 6 solves landing elsewhere, Jacobian cost **0.996×** master; the trust region **0** regressions, **80** landing elsewhere, **1.291×**. Both produce the same single win.
- **160 reporter-model cases** from #1105, #1378, #1466, #1723 — this is where `problem.slvs` lives. Like-for-like with #1354's least-squares fix applied to both sides, this branch has **1** regression against the trust region's **3**.

So "0 regressions" and "regresses `problem.slvs`" are both true and describe different corpora. The full write-up and the rest of the retraction are on #1749.

PR #1752, opened 2026-08-05
Export: warn instead of crashing on freehand/zigzag stipple

### LINKED ISSUE #1506: SolveSpace crashes when exporting 2D View with zigzag line style
### System information

- **Operating system:** Manjaro Linux 24.2.0 

### Expected behavior

It successfully exports a 2D view with a zigzag line style.
### Actual behavior

The program suddenly crashes.

### Additional information
Running it from the terminal shows the following error:
```
File ./src/exportvector.cpp, line 676, function MakeStipplePattern:
Assertion failed: false.
Message: Freehand and zigzag export not implemented.
```
Assuming it's really not implemented yet, shouldn't it show a warning popup instead of crashing?

AnibalPinto:
Fixes #1506.

Exporting a 2D view with a line style that uses the freehand or zigzag stipple pattern crashed with an assertion failure in MakeStipplePattern(), since those patterns have no vector-export equivalent. The on-screen renderer handles them via the OpenGL StippleAtlas, so the crash only occurred on vector export (SVG/PDF/EPS/DXF).

Replace the ssassert() with a dbp() warning and fall through to export the line as continuous, so a valid file is produced instead of aborting.

### DIFF
--- a/src/exportvector.cpp
+++ b/src/exportvector.cpp
@@ -674,7 +674,9 @@ static std::string MakeStipplePattern(StipplePattern pattern, double scale, char
 
         case StipplePattern::FREEHAND:
         case StipplePattern::ZIGZAG:
-            ssassert(false, "Freehand and zigzag export not implemented");
+            dbp("Freehand and zigzag export not implemented; "
+                "exporting as continuous line");
+            break;
     }
     std::replace(result.begin(), result.end(), '_', delimiter);
     return result;


phkahler:
I'll merge this, but I'm going to re-open #1506 because we should be able to export stipple patterns.

ruevs:
> I'll merge this, but I'm going to re-open #1506 because we should be able to export stipple patterns.

You decided to close it after all?

ruevs:
https://github.com/solvespace/solvespace/blame/master/src/exportvector.cpp#L676

Issue #1165, opened 2021-12-26
Add option to set custom fonts folder path


Symbian9:
### System information

- **SolveSpace version:** `3.x`, `4.x`
- **Operating system:** all supported

### Expected behavior

Add option to set custom fonts folder (instead of use system default fonts)

Few ways to solve this feature request:

1. as path input field in "Property Browser > configuration" tab (should be then stored);
1.1. there should be something like: `fonts folder:` `~/path/to/fonts-folder` `[apply]` (path should be clickable to input path as text)
1.2. if possible, add also `[change]` button (before `[apply]`) for interact as GTK folder choosing dialog instead of path input as text;
2. as CLI variable `-fonts-folder` (should be not saved in settings, just set per session).

It would be good if both ways may be implemented, but first one way is preferred here.

### Actual behavior

Actually SolveSpace uses system default fonts folder and if there are a lot of fonts installed on PC list is very huge and its hard to scroll it to need font each time.

PR #1757, opened 2026-08-21
Implement 1165 Add option to set custom fonts folder path

AnibalPinto:
Implementation for #1165 

I am not able to test outside Linux... testers welcomed.

Didn't implement:

>  1.2. if possible, add also [change] button (before [apply]) for interact as GTK folder choosing dialog instead of path input as text;

Not sure if is a good idea mixing UI code ... would like opinions before trying to implement it.

### DIFF
--- a/src/confscreen.cpp
+++ b/src/confscreen.cpp
@@ -214,6 +214,11 @@ void TextWindow::ScreenChangeAnimationSpeed(int link, uint32_t v) {
     SS.TW.edit.meaning = Edit::ANIMATION_SPEED;
 }
 
+void TextWindow::ScreenChangeFontsFolder(int link, uint32_t v) {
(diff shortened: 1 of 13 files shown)


phkahler:
@AnibalPinto  Do you actualy *use* solvespace? Are you an AI?  This PR doesn't even make sense. BTW if you're not actually using the program please stop submitting AI generated pull requests for it.

AnibalPinto:
@phkahler 

> @AnibalPinto Do you actualy _use_ solvespace? 

I did some tutorials about it and an original model some weeks ago that was  3d printed.

> Are you an AI?

Nop

> This PR doesn't even make sense. BTW if you're not actually using the program please stop submitting AI generated pull requests for it.

I followed the requirements from #1165 or at least what I understood from the issue.

I see advantages on being able to control the number font used in the project, not a high priority thing but it may be helpful.

Issue #1345, opened 2023-02-21
Crashing with upon application of symmetry constraints


IntegrativeMechanobiologyLaboratory:
### System information

- **SolveSpace version:** 
- 3.1~d6e97091
- **Operating system:** 
Ubuntu

### Expected behavior

Selected an edge of a rectangle, then clicked "Constraints" > "Symmetric". The rectangle should get the symmetry constraint. 

### Actual behavior

Software crashed.

### Additional information

**Error message on the terminal:**
SolveSpace!

Missing (absent) translation for 'A&ngle / Equal Angle'
Missing (absent) translation for 'E&qual Length / Radius'
connect failed: No such file or directory
Vector::WithMagnitude(2.3) of zero vector!
Vector::WithMagnitude(2.3) of zero vector!
Vector::WithMagnitude(2.3) of zero vector!
Vector::WithMagnitude(2.3) of zero vector!
Vector::WithMagnitude(2.3) of zero vector!
Vector::WithMagnitude(2.3) of zero vector!
File ./src/dsc.h, line 526, function FindById:
Assertion failed: t != nullptr.
Message: Cannot find handle.
Aborted (core dumped)

ruevs:
This is a regression, probably around 3d3d5c789d7035517bec6a4df6e7590f5f22e680 I need to bisect and/or debug it.

77maxikov:
https://github.com/solvespace/solvespace/blob/d6e970918f5ea77757a6d21da8b11f9c66882422/src/constraint.cpp#L583 should be exchanged with the next one

77maxikov:
@ruevs, should I make a PR for this?

PR #1759, opened 2026-08-22
Improved newton (BoykoNeov#4)

### LINKED ISSUE #246: Circle inexplicable cannot resize more than 16384
### System information

solvespace 2.3+repack1-2~bpo8+1

Operating system: (e.g. Debian testing) debian/jessie

### Expected behavior

In a sketch with a single circle. a single tangent and an intersecting radial, and no other constraints, all values of circle diameter should be feasible.

### Actual behavior

Circle cannot be dragged beyond ~16384.

Hitting space makes the solve error disappear, and then the error strikes again if you so much as touch and move any point.

However, if the sketch is redrawn to scale, with diameter = 1638, then it can be resized happily --- till it reaches ~16384. (or if you start beyond ~16384 diameter).

### Additional information

Attached repro bug-earth-resize-stops-at-16384.slvs and working.slvs
[bug-earth-resize-stops-at-16384.zip](https://github.com/solvespace/solvespace/files/965867/bug-earth-resize-stops-at-16384.zip)

### LINKED ISSUE #267: Wrong detection of redundant constraint
### System information

SolveSpace version: 2.3~

Operating system: Linux Mint Debian Edition 2 Betsy (using repos from Debian Jessie)

### Expected behavior

Program should solve sketch correctly. DOF should be equal to 0.

### Actual behavior

"SOLVE FAILED!" text appears in "Property Browser" window.

### Additional information

Problematic file attached: [bug-angle.slvs.zip](https://github.com/solvespace/solvespace/files/1165143/bug-angle.slvs.zip)

### LINKED ISSUE #1105: Cannot extend reference line in simple sketch

### LINKED ISSUE #1354: Large Dimensions result in Incompatible Constraints

### LINKED ISSUE #1466: constrain symmetric causing "constraints are incompatible"

phkahler:
This is BoykoNeov #4 applied to solvespace master.  This is the one we seem to agree is good.  FYI  I could not automatically rebase #1748 on this so it's not clear which shoud go first - assuming we even want that one.


phkahler:
Closes #246
Closes #267
Closes #1105
Closes #1354
Closes #1466

BoykoNeov:
I measured the five issues in the `Closes` list against this PR, since three of them had never been checked. Three are genuinely fixed, two are not shown to be — and one of the three reverses what I'd previously assumed.

## Method

Two trees differing in **nothing but** `src/system.cpp` plus the new test:

| tree | contents |
|---|---|
| base | `master` = `f801618a` |
| fix | `f801618a` + `21a74715` + `b3ebbd36` |

`src/system.cpp` in the fix tree verified byte-identical to this PR's head. Both trees were given the **same** `debugtool.cpp`, so the tooling can't bias the comparison, and both builds were confirmed to exit 0 before any result was believed. Debug, Ninja/MSVC, OpenMP on, GUI off.

The stepping path is what matters here: several of these only fail when a dimension is **edited and the sketch re-solved from the previous solution**, which is what the GUI does. Loading a file can't reproduce them.

## Results

| issue | on master `f801618a` | with this PR | verdict |
|---|---|---|---|
| **#1105** cannot extend reference line | stepping `1000 → 1200 → 2000 → 2500` **DIDNT_CONVERGE**, never recovers through 4000 | every step **OKAY** through 4000 | **fixed** |
| **#1354** large dimensions | **DIDNT_CONVERGE**, dof=4; bad: pt-on-line, pt-pt-distance, perpendicular | **OKAY** | **fixed** |
| **#1466** constraints are incompatible | group `g004-box` **DIDNT_CONVERGE**; bad: same-orientation, pt-on-line — exactly the two the reporter named | **OKAY**, dof=1 | **fixed** |
| **#246** circle won't resize past 16384 | **does not reproduce** by driving the diameter | unchanged | not demonstrated |
| **#267** wrong redundant detection | **does not reproduce** — loads OKAY, dof=0 | unchanged | not demonstrated |

Test suite: base **264 cases / 937 checks** → fix **265 / 944**, all green. The delta is exactly the seven checks in `test/constraint/large_dimensions/test.cpp`. **No regressions.**

## #1105 — this one reverses my own earlier read, and it has a trap in it

I had assumed #1105 was a *stepping* problem in the #1247 family rather than a conditioning one, and so would **not** be fixed here. That was based on your own 2021 comment about needing ever-smaller steps and getting stuck around 3629. Measured, it is fixed by this PR.

The trap, which is why it looks unfixed if you check it the obvious way: **a direct 1000 → 4000 jump succeeds even on master.** Only the gradual path fails. If you re-check this, step it.

## #1466 — checked twice, because it was the thinnest evidence

It's a linked assembly, so a bare OKAY could be a converged-but-wrong solution. Two things settle it:

1. The load path is already a full regenerate — `AfterNewFile()` → `GenerateAll(Generate::ALL)` — so the linked parts really are reloaded and rebuilt, not taken from cache.
2. The fix lands the failing file in the **same group, same dof, same result** (`g004-box` OKAY dof=1) as the reporter's own *working* sibling `box_on_floor-asm.slvs`, while disturbing neither that file nor `box.slvs` / `floor.slvs`.

## #246 and #267 — what I can and can't say

Neither reproduces on master, so closing them hides no live bug. But "doesn't reproduce" is not "fixed by this PR", and for #246 it isn't even the whole test:

- **#246** — I could not reproduce it **by driving the diameter constraint**, from 16069 up to 50000, every step OKAY. But #246 is reported as a **dragging** bug, and dragging is a different code path that a headless tool can't exercise. Claiming it either way needs a GUI drag test, which I haven't done.
- **#267** — the file loads and solves OKAY at dof=0. The report is vague enough ("SOLVE FAILED in the property browser") that this may simply not be what the reporter saw.

Both are from 2017. I'd rather say what was tested than round it up.

## One mechanical note on the `Closes` list

The keywords are in [your comment](https://github.com/solvespace/solvespace/pull/1759#issuecomment-5383092117), not in the PR description, and GitHub only creates the closing link from the **description or a commit message**. As the PR stands, `closingIssuesReferences` returns exactly one issue — **#1354** — and that comes from the `Fixes #1354` trailer in `21a74715`, not from the comment. So merging as-is would close #1354 only, and leave the other four open. Worth knowing either way, since it's easy to fix by moving the lines into the description if that's what you want.

If you do move them: **#1105, #1354 and #1466 are earned and now measured.** For #246 and #267 I'd close them as not-reproducible rather than as fixed here, or leave them.

Nothing above argues against merging.

---

ruevs:
I tested all models with the "solver" label and reported my results here https://github.com/BoykoNeov/solvespace/pull/4#issuecomment-5181097079 but did not compare with master - so some of them may have indeed already be fixed. https://github.com/solvespace/solvespace/issues/1345 is one such.

Issues linked in "development".

Edit: #1354 indeed

phkahler:
I was just rebasing those commits here so it'd be easy to apply. My list of fixes was based on @ruevs [comment here](https://github.com/BoykoNeov/solvespace/pull/4#issuecomment-5181097079) with an assumption on my part that he meant to type [1354](https://github.com/solvespace/solvespace/issues/1354) instead of 1345. I have no idea how to link issues that will be closed by the PR - the dropdown/search on the right side of github doesn't work. Adding the comments here apparently doesn't work either even though I've seen that work in the past.

We seemed to be in agreement that this is a good fix. For merging, my only blocking question is if we want the other one first. Claude rejected the later fix (was it number 7?) in a discussion I can't find. @ruevs had indicated 3 and 7 both fixed issue #1247 but seemed to not like 3, which the larger discussion had jwesthues and Claude both prefering 3 to 7. I agree about not taking 7 after reading the discussion.

So we're going to take this change. Do we want BoykoNeov 3 ( #1748 ) first or not?

ruevs:
> I was just rebasing those commits here so it'd be easy to apply. My list of fixes was based on @ruevs [comment here](https://github.com/BoykoNeov/solvespace/pull/4#issuecomment-5181097079) with an assumption on my part that he meant to type [1354](https://github.com/solvespace/solvespace/issues/1354) instead of 1345. I have no idea how to link issues that will be closed by the PR - the dropdown/search on the right side of github doesn't work. Adding the comments here apparently doesn't work either even though I've seen that work in the past.
> 
> We seemed to be in agreement that this is a good fix. For merging, my only blocking question is if we want the other one first. Claude rejected the later fix (was it number 7?) in a discussion I can't find. @ruevs had indicated 3 and 7 both fixed issue #1247 but seemed to not like 3, which the larger discussion had jwesthues and Claude both prefering 3 to 7. I agree about not taking 7 after reading the discussion.
> 
> So we're going to take this change. Do we want BoykoNeov 3 ( #1748 ) first or not?

This is orthogonal to [#7](https://github.com/BoykoNeov/solvespace/pull/7) / [#1749](https://github.com/solvespace/solvespace/pull/1749) and/or [#3](https://github.com/BoykoNeov/solvespace/pull/3) / [#1748](https://github.com/solvespace/solvespace/pull/1748). @jwesthues has valid objections to both of those.

But we can merge this. I'll merge it.

vthriller:
I was about to file another issue with redundant constraints, but this PR seems to fix it. @phkahler @ruevs are you still interested in sample files, and if so, what's the best place to put them (i.e. here, in a new issue, somewhere else)?

phkahler:
@vthriller  I prefer each sample file to be a separate github issue. That way when a fix comes along we can close every issue it fixes and only those that it fixes.

For example,  In my mind we had 4 main categories of NURBS issues, 3 of which recently got fixed. But there were actually multiple bugs for each category - no one fix resolved an entire category of problem sketches.

Same looks true for solver issues.

Thank you for asking.

Issue #1507, opened 2024-12-16
Trackpad gestures on Linux


polyfloyd:
I use Solvespace exclusively on my laptop which has a touchpad. The current controls are sufficient to model with, but the right trackpad button is does not feel nice to use.

It would be nice to also support zoom/pan gestures on the Linux/GTK based platform.

Related, gestures on Mac: #68

I would be happy to spend some time on this myself. However, I am wondering whether it would be better to first finish the port to GTK4 before working on this :thinking:

phkahler:
> I would be happy to spend some time on this myself. However, I am wondering whether it would be better to first finish the port to GTK4 before working on this 🤔

@polyfloyd  To be clear, I don't see myself leading this thing forever. But my vision for the project is GTK4 with single window (see the web version) on all platforms. I've tried the GTK4 port myself and it's outside my area of expertise so it'll be up to others to make it happen. Having better trackpad functionality would be great in the present, and might also carry over to GTK4. No need to wait IMHO.

Personally, I want the arrow keys to do rotations as well. AFAICT they're not handled the same as other keyboard inputs and that might also be platform dependent. Just something to consider when improving the laptop experience.

polyfloyd:
I'm a little busier these days, but maybe I will have some time to look into this still.

@phkahler can your GTK4 attempt be found online so I can perhaps start from there?

phkahler:
> I'm a little busier these days, but maybe I will have some time to look into this still.
> 
> [@phkahler](https://github.com/phkahler) can your GTK4 attempt be found online so I can perhaps start from there?

Not very useful but you can start with my [GTK4 branch](https://github.com/phkahler/solvespace/tree/gtk4).  I would only take the first 2 commits though, which just do basic changes - bumping the C++ version used (GTK4 requires newer than current solvespace) and change the includes, make a new .cpp file.

I would ignore the later commits because I view the attempt as a complete failure. A better approach might be to make stub functions out of everything in that main file, and then start filling it in. Window creation would be the first thing to make work, followed by I suppose getting menus to populate.

Solvespace has an abstraction layer between it and the GUI toolkit, so there are functions that go both ways. It has it's own "event" objects that get passed to solvespace for example - populated from whatever the chosen toolkit provides. It also has functions to build menus, which need to call whatever the toolkit requires for that - and object wrappers I think, so there is a solvespace menu class but it'll have to wrap a GTK4 object of some kind.

[...]

Issue #68, opened 2016-10-11
Support trackpad gestures


d235j:
macOS has scroll and zoom gestures which are essential for usability on a trackpad (as right-dragging is tricky and there's no center button). This should be possible on Windows and Linux as well, at least on computers that have multitouch trackpads.

Wishlist: support new Apple/Mac pressure-sensitive Force Touch trackpad.

whitequark:
OS X docs: https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/EventOverview/HandlingTouchEvents/HandlingTouchEvents.html
Windows docs: https://blogs.msdn.microsoft.com/jennifer/2010/12/14/multitouch-part-2-support-for-gestures-in-windows-7/
GTK docs: lol

The Windows API is more or less the same as OS X one.

tpimh:
I think this is the GTK+ doc that you are missing: https://developer.gnome.org/gtk3/stable/GtkGesture.html

whitequark:
@tpimh It is a really bad doc.

ss0git:
Since my issue was closed I'll copy my suggestion here. It is more specific than what is proposed in this support ticket.

UX improvement - Switch to using Shift + "Two finger drag" (no click) on trackpads to Orbit around

Proposed change to switch from Shift + "Two finger click and drag" to Shift + "Two finger drag" (no click) on trackpads to Orbit around.

### System information

SolveSpace version: 2.3

Operating system: OS X Sierra

### Expected behavior

Orbiting around using Shift + "Two fingers drag on trackpad" (no click) should work better. Pushing down to click is very cumbersome. You want to spin around the model like spinning an earth globe in school, effortless.

### Actual behavior

Shift + "Two fingers click and drag on trackpad" orbits around but not pushing down hard enough to click causes the app to Zoom.

To better see the change i made this table:

Current behaviour

_ | Pan | Zoom | Orbit
------------ | ------------- | ------------- | -------------
Flick fingers | _ | X  | _
Push down + Flick fingers | X | _  | X
Shift | _ | _  | X

Proposed behaviour

_ | Pan | Zoom | Orbit
------------ | ------------- | ------------- | -------------
Flick fingers | _ | X  | X
Push down + Flick fingers | X | _  | _
Shift | _ | _  | X

### Possible incompatibility with 

"Two fingers drag on trackpad" (mainly up and down) are used for Zooming (scrolling) now which is fine, but it might lead to a possible incompatibility.

[...]

PR #1763, opened 2026-08-29
Add Android front end

RainbowC0:
shot


ruevs:
Thank you @RainbowC0 

This is very interesting. Porting to Android has been discussed many times during the years but (I guess) there was never an Android developer interested in implementing it.

I have a few questions:

- Is this a human implementation or was it prompted/vibe coded?
- Is using Java necessary? I know that it (and Kotlin?) is the main language on android, but I thought the SDK supported C++ only/native applications?
- If C++ only is possible is there an advantage to using some Java?

(Even though I've never developed for Android and have last written in Java in 2001) I'll go through the code later and add some specific comments (if I have any).

phkahler:
@RainbowC0  First off thanks! This look interesting. There is currently AFAICT no comparable CAD option available on Android (meaning native, local, 3D).  But you dropped a review bomb on us so my initial response is probably going to be scattered.

This is going to take some time, but for starters is the image above the result of doing the tutorial on Android, or just showing that it can open a file?  I ask because I don't see evidence of incorporating touchscreen gestures in the PR, so I'm not sure how you could even make that model without a mouse.

There are some invasive changes like  #ifdef __ANDROID__  in export.cpp and mouse.cpp

**The GZIP stuff:**
There are calls to LoadStringFromGzip replaced with LoadString scattered around but ALSO an #ifdef around that function in resource.cpp.
1) Why not use gzip?
2) If it can't be used, replace the function LoadStringFromGzip with an android specific function that strips the ".gz" in just one place. Or relocate the additon of ".gz" into the function and have that one line skipped for android.  The goal here is to minimize platform specific #defines outside the /platform folder.  I'm not sure there are *any* such defines today but I'm probably wrong.
3) The following change is completely unnecessary and should not have been mixed in with a big PR like this.
```
    if(SS.showToolbar) {
        if(ToolbarMouseDown((int)mx, (int)my)) return;
```
changed to:
```
    if(SS.showToolbar && ToolbarMouseDown((int)mx, (int)my)) {
        return;
    }
```

**In solvespace.cpp**
Why this change, which will likely affect *something*
```
                GW.window->SetTitle(saveFile.raw.substr(saveFile.raw.find_last_of("/\\") + 1));

                GW.window->SetTitle(saveFile.FileName());
```
**Changes in toolbar.cpp**
Is this change needed, or fixing a latent bug?
```
--    return withinToolbar;

++    return withinToolbar || toolbarPressed != Command::NONE;
```

Renaming hitCommand to commandHit was completely unnecessary.
Why the new distinction between toolbarHovered and toolbarPressed?
Why was ToolbarMouseup() function added? It's called at the end of MouseLeftUp() what is that all about?

**Dependencies**
Why not zlib? pixman? cairo?
By removing cairo I think you removed the ability to output .pdf files

My guess is that these are non-trivail to build for Android?

@ruevs should we push this to a branch? It's certainly not ready to be in master and I'd rather it get worked on on a branch rather than in a single PR.

RainbowC0:
@ruevs :
- Most of this is human implementation.
- Java is necessary, though it's okay to use pure C++ for Android development but that's mainly used in game development without system components. To access components in pure C++, we have to call java functions via JNI reflection, thus complicating the work.

@phkahler :
- LoadStringFromGzip is only used to load unifonts and threejs from resources. Due to the [behaviour](https://raw.githubusercontent.com/SimonMarquis/TIL/main/docs/android.md#1) of AAPT (automatically unziping .gz files when packaging), I have to remove the .gz extension and use LoadFromString.
- It's okay to redefine LoadStringFromGzip as LoadString, but all calls to the function use a literal .gz extension, so if it's okay to remove the .gz extension from LoadStringFromGzip and define a macro to append the extension so that it compat the parameters for LoadString and the concating operation could be optimized when compiling as well?
- Android platform mainly uses touchscreen instead of mouse for input. If touching up, all hover states should be cleared.
- General behaviour of a gui button is to get triggered when mouse up and the target button not changed since mouse down rather than only mouse down. toolbarPressed is just for recording the down command. Especially on Android, if mouse down triggers it, a message box will appear instantly and intercept the following mouse events, making the received mouse events not completed.
- I use SAF to access files on android to avoid storage permissions needed, but the returned path is a url with content:// scheme, and the file path is all in percent-encoding. I've modified the Path::Filename for android to use %2F as separator to get the file name.
- I searched the project and found that cairo and pixman are only used for cli, and zlib is for LoadStringFromGzip. And this project generates pdf manually in bytes (in exportvector.cpp:845-)
```cpp
fprintf(f,
"%%PDF-1.1\r\n"
"%%%c%c%c%c\r\n",
        0xe2, 0xe3, 0xcf, 0xd3);

    xref[1] = (uint32_t)ftell(f);
    fprintf(f,
"1 0 obj\r\n"
"  << /Type /Catalog\r\n"
"     /Outlines 2 0 R\r\n"
"     /Pages 3 0 R\r\n"
"  >>\r\n"
"endobj\r\n");

    xref[2] = (uint32_t)ftell(f);
    fprintf(f,
"2 0 obj\r\n"
"  << /Type /Outlines\r\n"
"     /Count 0\r\n"
"  >>\r\n"
"endobj\r\n");
```

This is still bugly when I test with mouse input and I have to fix severe bugs for merge.

RainbowC0:
@phkahler With the help of vibe coding, there are some 3d parametic cad apps on Android today and you can have a comparison:
- [Digit 3D](https://digit3d.top/)
- [Materializer](https://github.com/materializr-cad/materializr)

RainbowC0:
This is the built apks for tests:
- [app-arm64_v8a.zip](https://github.com/user-attachments/files/31681920/app-arm64_v8a.zip)
- [app-x86_64.zip](https://github.com/user-attachments/files/31681900/app-x86_64.zip)

And how about add a CI/CD workflow to build the apks?

My following plains:
- [short-term] Publish the Android port on [F-Droid](https://f-droid.org) if no severe bugs.
- [long-term] More accessible touchscreen operations to select, move, scale and orbit. Android devices often recognize three-pointer sliding to capture screen, and the operation is also mapped to middle-button operation on the current version. This is also necessary on any other devices with touchscreen like Surface Go, Matebook and my Lenovo Yoga 14c. There's also news that Macbook Pro would support touchscreen this year. But till now solvespace has not support trackpad well enough (#1507, #68)
- [long-term] Single window on both phone and tablet screens. I prefer a foldable bottom window for the text window on portrait phone screen and a right side window on landscape or large screen.
- [long-term] Fully asynchronous callbacks support. An Android's app process has a main ui looper to process input events and ui callbacks. If any events spans over 5s for foreground or 20s for background on main thread, Application No Response (ANR) would occur and cause crashes. I temporarily take over the main looper when waiting for result to achieve synchronous callbacks, however which is dangerous.

PR #1765, opened 2026-09-01
Support load multiple fonts from .ttc

RainbowC0:


phkahler:
@RainbowC0  What was this in response to?  Doesn't this have potential to dramatically expand the list of fonts in the text window?  I suppose that's the point though, if some are missing.

RainbowC0:
@phkahler I have a system font NotoSansCJK.ttc which contains  chinese characters for SC, TW, JP, KR, but before this change I can only get the JP variation, which displays different with SC characters like 门, 令. TTC font file usually contains a set of font variations which shares some common glyphs, so it's necessary to load all variations.

RainbowC0:
> @RainbowC0 What was this in response to? Doesn't this have potential to dramatically expand the list of fonts in the text window? I suppose that's the point though, if some are missing.

I got 220 fonts on the text window before this change, and 246 after it on my win11. Most system font files on Windows (also others) are in ttf format, which contains only one font face (fontFace->num_faces==1), so it's not the point.

Anything else to consider?

PR #1771, opened 2026-09-07
Windows: ship solvespace-cli in the release assets

### LINKED ISSUE #1770: Windows releases don't ship solvespace-cli.exe even though CI builds it
### System information

- **SolveSpace version:** 3.2, and master @ 952c11c0
- **Operating system:** Windows 11 24H2

### Expected behavior

A Windows user can download `solvespace-cli` from the releases page, as on macOS (it is bundled inside `SolveSpace.app` in `SolveSpace.dmg`, and `sign-macos.sh` signs it) and on Linux (distribution packages and the snap both provide it). `README.md` documents `build/bin/solvespace-cli.exe` as a Windows build product, and the CLI is the only way to reach `export-view`, `export-wireframe`, `export-mesh`, `export-surfaces` and `thumbnail` without the GUI.

### Actual behavior

No official Windows download contains a CLI binary. The v3.2 assets are:

```
solvespace_x86.exe   solvespace_single_core_x86.exe
solvespace_x64.exe   solvespace_single_core_x64.exe
SolveSpace.dmg       solvespace-3.2.tar.xz
```

All four executables are the GUI. `src/platform/entrygui.cpp` treats a trailing argument as a file to open, so passing CLI options to them gives `Couldn't read file from '--version'` (#903).

WinGet and Scoop both pull `solvespace_x86.exe` straight from this release, so the CLI is missing from those channels too. Building from source is currently the only way for a Windows user to get it.

### Additional information

The binary is already built by the release jobs and then discarded. From the *Build Release Windows (OpenMP) x86* job of the most recent CD run on master:

```
solvespace-cli.vcxproj -> D:\a\solvespace\solvespace\build\bin\RelWithDebInfo\solvespace-cli.exe
```

`.github/scripts/build-windows.sh` renames only `solvespace.exe`, and each Windows job in `.github/workflows/cd.yml` uploads only that single path, so `solvespace-cli.exe` is left behind in the build directory.

It is statically linked (`/MT`, in-tree cairo/libpng/zlib) exactly like the GUI executable, so shipping it needs no extra runtime files — I ran one from a CI artifact on a clean Windows 11 machine and exported an SVG with it successfully.

I have a small patch for `build-windows.sh` and `cd.yml` and will open a PR.

Foadsf:
Fixes #1770.

The Windows release jobs already build `solvespace-cli.exe` — it shows up in every CD run:

```
solvespace-cli.vcxproj -> D:\a\solvespace\solvespace\build\bin\RelWithDebInfo\solvespace-cli.exe
```

but `build-windows.sh` renames only `solvespace.exe`, and each job uploads only that one path, so the CLI is discarded at the end of the build. macOS ships it inside the app bundle and Linux distributions package it; Windows is the only platform where it cannot be obtained without compiling.

This renames `solvespace-cli.exe` with the same suffix as the GUI binary it was built alongside, uploads it in the same artifact, and adds the four files to the release assets. No build configuration changes — the binary is already produced by the existing `ENABLE_CLI` default, and it is statically linked (`/MT`, in-tree cairo/libpng/zlib) like the GUI executable, so it needs no extra runtime files.

Verified by running the modified workflow end to end on a fork. All four Windows release jobs passed, and the `windows_x86` artifact contained both binaries:

```
solvespace_x86.exe        8,740,864
solvespace-cli_x86.exe    7,242,752
```

Running that `solvespace-cli_x86.exe` on a clean Windows 11 machine, with nothing else installed:

```
> solvespace-cli_x86.exe export-view --output "%.svg" --view top demo.slvs
Written 'demo.svg'.
```

Happy to trim this to just the two OpenMP builds if eight Windows assets is more than you want on the release page.

### DIFF
--- a/.github/scripts/build-windows.sh
+++ b/.github/scripts/build-windows.sh
@@ -43,13 +43,17 @@ cmake --build . --config "${BUILD_TYPE}" -t test_solvespace -- -maxcpucount
 if [ "$3" = "x64" ]; then
 	if [ "$2" != "openmp" ]; then
 		mv bin/$BUILD_TYPE/solvespace.exe bin/$BUILD_TYPE/solvespace_single_core_x64.exe
+		mv bin/$BUILD_TYPE/solvespace-cli.exe bin/$BUILD_TYPE/solvespace-cli_single_core_x64.exe
 	else
 		mv bin/$BUILD_TYPE/solvespace.exe bin/$BUILD_TYPE/solvespace_x64.exe
+		mv bin/$BUILD_TYPE/solvespace-cli.exe bin/$BUILD_TYPE/solvespace-cli_x64.exe
 	fi
 else
 	if [ "$2" != "openmp" ]; then
 		mv bin/$BUILD_TYPE/solvespace.exe bin/$BUILD_TYPE/solvespace_single_core_x86.exe
+		mv bin/$BUILD_TYPE/solvespace-cli.exe bin/$BUILD_TYPE/solvespace-cli_single_core_x86.exe
     else
         mv bin/$BUILD_TYPE/solvespace.exe bin/$BUILD_TYPE/solvespace_x86.exe
+        mv bin/$BUILD_TYPE/solvespace-cli.exe bin/$BUILD_TYPE/solvespace-cli_x86.exe
 	fi
 fi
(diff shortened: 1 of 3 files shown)


ruevs:
Consider my comment on the issue here https://github.com/solvespace/solvespace/issues/1770#issuecomment-5567862450 before deciding whether to merge this.

Issue #1722, opened 2026-06-02
Allow linking against system mimalloc


werdahias:
I am in the process in updating the debian package, and the only remaining issue is mimalloc. As Debian tries to avoid embedded code copies, when forcing to build against the packaged `libmimalloc-dev` the build fails:

```
[ 48%] Linking C executable ../bin/CDemo
cd /build/reproducible-path/solvespace-3.2+ds/obj-x86_64-linux-gnu/exposed && /usr/bin/cmake -E cmake_link_script CMakeFiles/CDemo.dir/link.txt --verbose=1
/usr/bin/x86_64-linux-gnu-ld.bfd: ../bin/libslvs.so.3.2: undefined reference to `mi_heap_new'
/usr/bin/x86_64-linux-gnu-ld.bfd: ../bin/libslvs.so.3.2: undefined reference to `mi_heap_zalloc'
/usr/bin/x86_64-linux-gnu-ld.bfd: ../bin/libslvs.so.3.2: undefined reference to `mi_heap_destroy'
collect2: error: ld returned 1 exit status
/usr/bin/cc -g -O2 -Werror=implicit-function-declaration -ffile-prefix-map=/build/reproducible-path/solvespace-3.2+ds=. -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -Wdate-time -D_FORTIFY_SOURCE=2 -ffile-prefix-map=/build/reproducible-path/solvespace-3.2+ds=. -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Werror=switch -O3 -DNDEBUG -Wl,--as-needed -Wl,-z,relro -Wl,--dependency-file=CMakeFiles/CDemo.dir/link.d CMakeFiles/CDemo.dir/CDemo.c.o -o ../bin/CDemo  -Wl,-rpath,/build/reproducible-path/solvespace-3.2+ds/obj-x86_64-linux-gnu/bin ../bin/libslvs.so.3.2
make[3]: *** [exposed/CMakeFiles/CDemo.dir/build.make:105: bin/CDemo] Error 1
make[3]: Leaving directory '/build/reproducible-path/solvespace-3.2+ds/obj-x86_64-linux-gnu'
make[2]: *** [CMakeFiles/Makefile2:332: exposed/CMakeFiles/CDemo.dir/all] Error 2
make[2]: *** Waiting for unfinished jobs....
[...]

Issue #1036, opened 2021-05-12
Windows error in solvespace 3.0 for windows 10.


orisvaldo19:
### System information

- **SolveSpace version:** 
- **Operating system:** 

### Expected behavior

The program open.

### Actual behavior

The error message to open solvespace im my windows 10 as is in attached image. The vc_redist.x86.exe is installed.

### Additional information

![Solvespace_error](https://user-images.githubusercontent.com/84055840/117909285-57aee500-b2b0-11eb-887d-196a7cb09fbb.jpg)

kk6mrp:
Your computer may not support a high enough OpenGL version. Can you try the attached version and report if this solves your issue?
[solvespace.zip](https://github.com/solvespace/solvespace/files/6463481/solvespace.zip)

Symbian9:
> Can you try the attached version and report if this solves your issue?

@kk6mrp, Please, describe what changes included in your build.

orisvaldo19:
Really. I don´t have a graphic card in this PC (is a reserve machine cause
my principal is with my son, a student and gamer), but the version that you
send me solves. Its open.
Thanks!

Orisvaldo Silva.

Em qua., 12 de mai. de 2021 às 13:44, app4soft ***@***.***>
escreveu:

phkahler:
Maybe we need to update the web site to indicate an OpenGL / graphics requirement?

Another option is to make a release build for XP or whatever. That will eventually die but it's still possible apparently :-)

ruevs:
> Another option is to make a release build for XP or whatever.

I'll make a Windows XP compatible build of 3.0 with the Open GL1 renderer. Will ping you when I get to it (matter of of sitting on a machine which very rarely gets used these days) and you can decide how "official" you want to make it.

Adding the Open GL requirement on the download page is a good idea.

orisvaldo19:
Cool, thanks!

Em qui., 13 de mai. de 2021 às 06:07, ruevs ***@***.***>
escreveu:

Symbian9:
> I'll make a Windows XP compatible build of SolveSpace 3.0 with the OpenGL1.x renderer.

@ruevs, It would be cool! What about OpenGL1.x-copmpatible build for Mac OS X? See: https://github.com/solvespace/solvespace/issues/102#issuecomment-261479046

> Adding the OpenGL requirement on the download page is a good idea.

[...]

PR #1772, opened 2026-09-09
an option to replace mimalloc with std::pmr::monotonic_buffer_resource

vthriller:
I don't even expect you to merge this in (especially given [this](https://github.com/solvespace/solvespace/pull/1665#issuecomment-3706281524)), just saw #1722 and decided to throw in an idea.


ruevs:
> I don't even expect you to merge this in (especially given [this](https://github.com/solvespace/solvespace/pull/1665#issuecomment-3706281524)), just saw #1722 and decided to throw in an idea.

And [this](https://github.com/solvespace/solvespace/blob/master/CONTRIBUTING.md#contributing-code)

Your expectation is correct ;-)

What is more, before `mimalloc` we had a "perfectly good" hand written allocator, which still works and I'm reverting to it when building for [Windows XP](https://github.com/ruevs/solvespace/releases/) (and [2000](https://github.com/solvespace/solvespace/issues/1036#issuecomment-3405830198)). See here https://github.com/ruevs/solvespace/commits/Win2k/

